PR #16, opened 2020-01-29
Add JupyterHub user data directories and integrate with Magpie for authentication

davidcaron:
## Todo
- [x] fix deploy scripts, using the tutorial-notebooks folder 
- [x] change the login message

Depends on https://github.com/Ouranosinc/jupyterhub/pull/1


davidcaron:
@tlvu Currently, I set permissions for the data directory by setting 1000:1000 as the owner. This works for the workflow-e2e tests docker image, but this will likely break for other images when the user running the docker image is not UID 1000.

I tried monkeypatching the `post_start_exec` method in the dockerspawner plugin, to get the UID by running exec inside the container after it started, but that didn't work... I'm not sure why, maybe it has to do with something I didn't understand about monkeypatching a twisted generator.

I was thinking we could, in the same `pre_spawn_hook` I used here, use `chmod 777` instead of `chown 1000:1000`. What do you think? The user id inside the container has no meaning outside it, so we might as well allow everyone to write to these files. And a user user still doesn't have access to another user's files inside the container.

tlvu:
> @tlvu Currently, I set permissions for the data directory by setting 1000:1000 as the owner. This works for the workflow-e2e tests docker image, but this will likely break for other images when the user running the docker image is not UID 1000.

For now, we only run our jupyter image.

We can easily test other official Jupyter images by uncommenting this section in `env.local`: https://github.com/bird-house/birdhouse-deploy/blob/ccf8c22d4179b227e7393b0746db10bcd4c16a97/birdhouse/env.local.example#L94-L105


> 
> I tried monkeypatching the `post_start_exec` method in the dockerspawner plugin, to get the UID by running exec inside the container after it started, but that didn't work... I'm not sure why, maybe it has to do with something I didn't understand about monkeypatching a twisted generator.

Does the PR include this monkeypatching or not?  Sorry if irrelevant question, I have not read the diff of this PR yet.

> 
> I was thinking we could, in the same `pre_spawn_hook` I used here, use `chmod 777` instead of `chown 1000:1000`. What do you think? The user id inside the container has no meaning outside it, so we might as well allow everyone to write to these files. And a user still doesn't have access to another user's files when they run their notebooks.

That's not a bad idea.  But that also means on the disk, it'll be 777, anyone having ssh to the server can delete/modify every users data. 

This is one of the advantages when using data volume instead of bind mounting a folder from the host, the data is less exposed so we can easier do things like that, unfortunately the data volume route do not work for us :(

davidcaron:
> Does the PR include this monkeypatching or not?

No, it doesn't... I might explore this a bit more.

> > I was thinking we could, in the same `pre_spawn_hook` I used here, use `chmod 777` instead of `chown 1000:1000`. What do you think? The user id inside the container has no meaning outside it, so we might as well allow everyone to write to these files. And a user still doesn't have access to another user's files when they run their notebooks.

> That's not a bad idea. But that also means on the disk, it'll be 777, anyone having ssh to the server can delete/modify every users data.

Well... we *do* have a [post_stop_hook](https://jupyterhub.readthedocs.io/en/stable/api/spawner.html#jupyterhub.spawner.Spawner.post_stop_hook) we can call that sets back the permissions when the container stops... This would at least reduce the damage, but I'll keep looking for a solution.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -1,21 +1,49 @@
 import os
 import logging
+import subprocess
 
 c.JupyterHub.bind_url = 'http://:8000/jupyter'
 c.JupyterHub.hub_ip = 'jupyterhub'
 
+c.JupyterHub.authenticator_class = 'jupyterhub_magpie_authenticator.MagpieAuthenticator'
+c.MagpieAuthenticator.magpie_url = "http://magpie:2001"
```
tlvu: Nice, using internal docker network.  Do you know why historically we do not use that more often?  We tend to use the public ip `PAVICS_FQDN` which depends on external routing.  However, just to be safe, have to add a `links` section in `docker-compose.yml` of jupyterhub to magpie.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -1,21 +1,49 @@
...
+
+# JupyterHub requires a single-user instance of the Notebook server, so we
+# default to using the `start-singleuser.sh` script included in the
+# jupyter/docker-stacks *-notebook images as the Docker run command when
+# spawning containers.  Optionally, you can override the Docker run command
+# using the DOCKER_SPAWN_CMD environment variable.
+
+# spawn_cmd = os.environ.get('DOCKER_SPAWN_CMD', "jupyterhub-singleuser")
+
+# c.DockerSpawner.extra_create_kwargs.update({ 'command': 'jupyterhub-singleuser' })
```
tlvu: Did something break if you do not use this singleuser command?

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -1,21 +1,49 @@
...
+
 c.DockerSpawner.image = os.environ['DOCKER_NOTEBOOK_IMAGE']
 c.DockerSpawner.use_internal_ip = True
 c.DockerSpawner.network_name = os.environ['DOCKER_NETWORK_NAME']
+
 notebook_dir = '/notebook_dir'
+host_user_data_dir = os.path.join(os.environ['JUPYTERHUB_USER_DATA_DIR'], "{username}")
```
tlvu: this `{username}` is a built-in variable or template expansion variable?

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -1,21 +1,49 @@
...
+
 c.DockerSpawner.notebook_dir = notebook_dir
-c.DockerSpawner.volumes = { os.environ['DOCKER_JUPYTERHUB_USER_PERSISTENCE_VOLUME']: notebook_dir }
+c.DockerSpawner.volumes = { host_user_data_dir: notebook_dir }
+
+def create_dir_hook(spawner):
```
tlvu: Here's the hook, interesting.

So it's able to create a new dir on the root mount even if the root mount is, I assume, writable by root only?

birdhouse/deployment/trigger-deploy-notebook
```diff
@@ -83,7 +77,7 @@ docker run --rm \
   -u root \
   -v $TMP_SCRIPT:/deploy-notebook:ro \
   -v $TMPDIR/$TUTORIAL_NOTEBOOKS_DIR:/$TUTORIAL_NOTEBOOKS_DIR:ro \
-  -v $DOCKER_JUPYTERHUB_USER_PERSISTENCE_VOLUME:$NOTEBOOK_DIR_MNT:rw \
+  -v "${JUPYTERHUB_USER_DATA_DIR}":$NOTEBOOK_DIR_MNT:rw \
```
tlvu: From memory, this script do not read `env.local` so do not know what is `JUPYTERHUB_USER_DATA_DIR`.

birdhouse/deployment/trigger-deploy-notebook
```diff
@@ -12,12 +12,6 @@
 #
 # Logs to /var/log/PAVICS/notebookdeploy.log, re-use existing logrotate.
 
-
-if [ -z "$DOCKER_JUPYTERHUB_USER_PERSISTENCE_VOLUME" ]; then
-    # dupe with pavics-compose.sh
-    DOCKER_JUPYTERHUB_USER_PERSISTENCE_VOLUME=jupyterhub_user_persistence
```
tlvu: Since this script do not read `env.local`, I had to set the default value here.  It's a trade-off to make it independent of `env.local` or rather the location of `env.local`.  From memory this script is dropped directly into /etc/cron.hourly as a standalone script, not as a cronjob calling script with ability to pass the path to `env.local` as arguments.

birdhouse/docker-compose.yml
```diff
@@ -330,7 +336,7 @@ services:
     restart: always
 
   jupyterhub:
-    image: pavics/jupyterhub:1.0-20190501
+    image: pavics/jupyterhub:1.0-20200130
```
tlvu: `1.0.0-date` ?

davidcaron:
```bash
$ for image in pavics/workflow-tests:191122 jupyter/scipy-notebook jupyter/r-notebook jupyter/tensorflow-notebook jupyter/datascience-notebook jupyter/pyspark-notebook jupyter/all-spark-notebook; do docker run --rm $image id -a; done 
uid=1000(jenkins) gid=1000(jenkins) groups=1000(jenkins) 
uid=1000(jovyan) gid=100(users) groups=100(users) 
uid=1000(jovyan) gid=100(users) groups=100(users) 
uid=1000(jovyan) gid=100(users) groups=100(users) 
uid=1000(jovyan) gid=100(users) groups=100(users) 
uid=1000(jovyan) gid=100(users) groups=100(users) 
uid=1000(jovyan) gid=100(users) groups=100(users)
```

So... I guess calling `chown -R 1000` should be good enough?

birdhouse/env.local.example
```diff
@@ -16,7 +16,7 @@ export TOMCAT_NCWMS_PASSWORD=ncwmspass
 export SUPPORT_EMAIL=helpdesk@example.com
 export CMIP5_THREDDS_ROOT=birdhouse/CMIP5/CCCMA
 export JUPYTERHUB_ADMIN_USERS="{'admin'}"  # python set syntax
-export JUPYTERHUB_USERS_PASS="admin:admin public:public"  # space separated 'username:passwd' format
+export JUPYTERHUB_USER_DATA_DIR=/data/jupyterhub_user_data
```
tlvu: This should be an optional var instead of mandatory vars.  It's not like a password were user is much encouraged to change the default value.

Optional var do not break the deployment if `env.local` do not contain it (much easier for the auto deployment).  New vars introduced should be optional with sensible default.

Mandatory is reserved when we really want user to take action, like to set a new password, else deployment won't work.

birdhouse/pavics-compose.sh
```diff
@@ -125,19 +125,11 @@ if [[ $1 == "up" ]]; then
...
-  # this is shared for all users, no multi users isolation
-  # see open issue https://github.com/moby/moby/issues/32582 [feature] Allow
-  # mounting sub-directories of named volumes
-  # for multi user isolation, we should leverage sub-directories of named
-  # volumes
-  docker run --rm --name set_permissions \
-    -v ${DOCKER_JUPYTERHUB_USER_PERSISTENCE_VOLUME}:/notebook_dir bash \
-    bash -c "chown 1000:1000 /notebook_dir"
+  # Create the path to contain user notebooks if it doesn't exist
+  mkdir -p "${JUPYTERHUB_USER_DATA_DIR}"
```
tlvu: Ah !  The base user data is created here !  This will likely not work since the base `/data` is most likely owned by `root` while the user running this `pavics-compose.sh` script must not be `root` else we run into weirdo problems (Dominique Lenoir found out about this the hard way previously).

You'll probably need the `docker run --rm --name set_permissions` trick below to create this base dir.

birdhouse/scripts/backup-datavolume.sh
```diff
@@ -1,15 +1,15 @@
 #!/bin/sh -x
-# Backup DATA_VOL_NAME to BACKUP_OUT_DIR/DATA_VOL_NAME.tgz
+# Backup DATA_VOL_DIR to BACKUP_OUT_DIR/DATA_VOL_DIR.tgz
 
-DATA_VOL_NAME="$1"; shift
+DATA_VOL_DIR="$1"; shift
 BACKUP_OUT_DIR="$1"; shift
 
 docker run --rm \
-  --name backup_data_vol_$DATA_VOL_NAME \
```
tlvu: Don't change this script.  This script is to backup a data volume, now we use bind mount.  We will simply not use this script and use something else for bind mount.

Keep this script as-is for other data volume backup in the future.

birdhouse/scripts/backup-juputerhub-notebooks.sh
```diff
@@ -5,12 +5,7 @@ if [ -z "$BACKUP_OUT_DIR" ]; then
     BACKUP_OUT_DIR=/tmp
 fi
 
-if [ -z "$DOCKER_JUPYTERHUB_USER_PERSISTENCE_VOLUME" ]; then
-    # dupe with pavics-compose.sh
-    DOCKER_JUPYTERHUB_USER_PERSISTENCE_VOLUME=jupyterhub_user_persistence
-fi
-
-./backup-datavolume.sh "$DOCKER_JUPYTERHUB_USER_PERSISTENCE_VOLUME" "$BACKUP_OUT_DIR"
```
tlvu: Don't call `backup-datavolume.sh`, adapt this script for bind mount.

Realistically, since we are under `/data` on the host, I assume there is already something that backup the entire `/data` so this script is probably obsoleted.

birdhouse/deployment/deploy.sh
```diff
@@ -164,15 +164,6 @@ cd $COMPOSE_DIR
 # immediately cache the new notebook image for faster startup by JupyterHub
 docker pull $DOCKER_NOTEBOOK_IMAGE
 
-# deploy new README.ipynb to Jupyter instance
-docker run --rm --name deploy_README_ipynb \
-    -v $DOCKER_JUPYTERHUB_USER_PERSISTENCE_VOLUME:/notebook_dir \
-    -v $REPO_ROOT/docs/source/notebooks:/nb:ro \
-    -u root \
-    bash \
-    bash -c "cp -v /nb/README.ipynb /notebook_dir/.; \
```
tlvu: Missing deployment of `README.ipynb` into each and every user data dir.

A read-only bind mount of this file into the user data dir is better than a copy.  Should use same trick for the `tutorials-notebook` dir as well.

tlvu:
> ```shell
> $ for image in pavics/workflow-tests:191122 jupyter/scipy-notebook jupyter/r-notebook jupyter/tensorflow-notebook jupyter/datascience-notebook jupyter/pyspark-notebook jupyter/all-spark-notebook; do docker run --rm $image id -a; done 
> uid=1000(jenkins) gid=1000(jenkins) groups=1000(jenkins) 
> uid=1000(jovyan) gid=100(users) groups=100(users) 
> uid=1000(jovyan) gid=100(users) groups=100(users) 
> uid=1000(jovyan) gid=100(users) groups=100(users) 
> uid=1000(jovyan) gid=100(users) groups=100(users) 
> uid=1000(jovyan) gid=100(users) groups=100(users) 
> uid=1000(jovyan) gid=100(users) groups=100(users)
> ```
> 
> So... I guess calling `chown -R 1000` should be good enough?

Oh wow what a luck, both Jenkins ssh agent image and Jupyter images are using `1000` as uid !  Yes `chown -R 1000` will do it.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
davidcaron: I think they should be used more... is the `links` section necessary? The network created by docker-compose isn't enough?

birdhouse/config/jupyterhub/jupyterhub_config.py.template
davidcaron: I think it's used by default, when the `default_url` is '/lab'. I'll remove this commented out code.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
davidcaron: It's replaced by JupyterHub's spawner https://jupyterhub.readthedocs.io/en/stable/api/spawner.html#jupyterhub.spawner.Spawner.notebook_dir

birdhouse/config/jupyterhub/jupyterhub_config.py.template
davidcaron: The hook runs in the main JupyterHub container, which is running as root. So yes, the main JUPYTERHUB_USER_DATA_DIR should be writable by root only.

birdhouse/docker-compose.yml
davidcaron: yes

birdhouse/config/jupyterhub/jupyterhub_config.py.template
tlvu: `links` is not necessary but might prevent problems later.  Ex: docker will start magpie before jupyterhub, if someone want to extract dependencies between the containers, `links` will explicitly show that dependency/connection.

birdhouse/deployment/trigger-deploy-notebook
davidcaron: I realized that... not sure what we should do here? This script doesn't have any knowledge of the pavics repo... previously it was using the docker volume so it was accessible globally.

Maybe the pavics repo path should be passed when running `install-deploy-notebook`? And check the current folder when this variable is not given?

davidcaron: Would appending the variable in /env/environment be a solution?

birdhouse/pavics-compose.sh
davidcaron: I... think it's not even necessary to create it here. It's always mounted by docker, which will create it as root if it doesn't exist. Do you think of a use case where this folder is not mounted by docker?

birdhouse/deployment/trigger-deploy-notebook
davidcaron: Please check https://github.com/bird-house/birdhouse-deploy/pull/16/commits/a34b2af59a19f233eb142cf5f58f6c1503ebdcdc, I used envsubst when installing the cron script in /etc/hourly

birdhouse/scripts/backup-datavolume.sh
davidcaron: Done

birdhouse/scripts/backup-juputerhub-notebooks.sh
davidcaron: See https://github.com/bird-house/birdhouse-deploy/pull/16/commits/c55725635f18f54a24f0743b31f4546f908a0ba5

davidcaron:
Here is what the tree looks like for the user inside jupyterhub:
image

tutorial-notebooks and README.ipynb are readonly

birdhouse/pavics-compose.sh
tlvu: Agreed, probably not needed to be created since docker will create itself on demand.

birdhouse/env.local.example
```diff
@@ -104,6 +103,11 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
 #                                   }
 #"
 
+# The parent folder where all the user notebooks will be stored.
+# For example, a user named "bob" will have his data in $JUPYTERHUB_USER_DATA_DIR/bob
+# and this folder will be mounted when he logs into JupyterHub.
+export JUPYTERHUB_USER_DATA_DIR=/data/jupyterhub_user_data
```
tlvu: This line should be commented out and a default value added to `./pavics-compose.sh`.  The whole point is to make the existence of optional var optional in `env.local`.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -1,21 +1,59 @@
...
 notebook_dir = '/notebook_dir'
+jupyterhub_data_dir = os.environ['JUPYTERHUB_USER_DATA_DIR']
+host_user_data_dir = join(jupyterhub_data_dir, "{username}")
+
 c.DockerSpawner.notebook_dir = notebook_dir
-c.DockerSpawner.volumes = { os.environ['DOCKER_JUPYTERHUB_USER_PERSISTENCE_VOLUME']: notebook_dir }
+
+c.DockerSpawner.volumes = {host_user_data_dir: notebook_dir}
+
+host_tutorial_notebooks_dir = join(jupyterhub_data_dir, "tutorial-notebooks")
```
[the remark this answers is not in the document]
davidcaron: I know 😄 I'll try creating one just to see...

tlvu:
> The only issue is that tutorials-notebooks and README.ipynb are copied on the drive, because they are mounted in a "nested" way inside docker... I think I will create a "workspace" folder that is writable, and everything else will be readonly.

Or we can bind mount 2 mounts onto the Jupyterhub.  1 for all the user data, 1 for all the read-only stuff.  Then backup of the user data do not uselessly also backup all the read-only stuff.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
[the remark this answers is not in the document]
davidcaron: The user was no workspace, and can't write anywhere...

birdhouse/deployment/trigger-deploy-notebook
tlvu: Nice, very clever user of envsubst.

birdhouse/env.local.example
davidcaron: Done

davidcaron:
> Or we can bind mount 2 mounts onto the Jupyterhub. 1 for all the user data, 1 for all the read-only stuff. Then backup of the user data do not uselessly also backup all the read-only stuff.

My only problem is how should we present it to the user so that it's intuitive where to create notebooks, and what is persisted.

Here is what I have now:

image

Only workspace is writable, and the top folder is not. The only problem with this configuration is that when the user tries to create a notebook in the top folder, they get a confusing 403 not authorized error...

tlvu:
> Only workspace is writable, and the top folder is not. The only problem with this configuration is that when the user tries to create a notebook in the top folder, they get a confusing 403 not authorized error...

How about calling it `writable-workspace` so it kinda hint that elsewhere is not writable?

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -13,7 +13,7 @@ c.MagpieAuthenticator.magpie_url = "http://magpie:2001"
 c.JupyterHub.cookie_secret_file = '/persist/jupyterhub_cookie_secret'
 c.JupyterHub.db_url = '/persist/jupyterhub.sqlite'
 
-c.JupyterHub.template_paths = ['/template_paths']
+c.JupyterHub.template_paths = ['custom_templates']
```
tlvu: Lost the `/` in front?  Do you really need to rename?

davidcaron: Yes, good catch. I thought template_paths was the name of the configuration parameter, containing a list of paths. it made more sense to me to call it like a folder that contains templates, but it's not required...

tlvu:
Have you tried `vagrant provision`?  I am getting this weird error:
```
    lvupavics: /usr/local/bin/docker-compose: line 1: Not: command not found
    lvupavics: /usr/local/bin/docker-compose: line 1: Not: command not found
    lvupavics: /usr/local/bin/docker-compose: line 1: Not: command not found
```
and `pavics-compose.sh up -d` is not run at all.

birdhouse/config/jupyterhub/custom_templates/login.html
```diff
@@ -1,8 +1,21 @@
-{% extends "templates/login.html" %}
-{% set announcement_login = '
+{% extends "templates/login.html" %} {% set announcement_login = '
 <p><strong>Public login:</strong> public/public</p>
-<p>Given this public nature, anyone can tamper with your notebooks so please <strong>export your valuable notebooks elsewhere</strong> if you want to preverve them.</p>
-<p>Your <strong>notebooks will persist</strong> accross reboots <strong>only if</strong> saved under <strong>home (/notebook_dir)</strong>.</p>
-<p>Please <strong>be considerate</strong> with the amount of <strong>disk space usage</strong> on this Jupyter instance.</p>
-<p>This Jupyter instance can restart every day.  <strong>Long running processes will be killed without notice.</strong></p>
+<p>
+  Given this public nature, anyone can tamper with your notebooks so please
```
tlvu: I think we can remove this "public nature" paragraph since your PR effectively make each user workspace private to him only.

davidcaron: So maybe add a note on how to ask for an account? If so, could you write something like 'contact Ouranos for an account'?

davidcaron: ah, there is a SUPPORT_EMAIL variable in env.local

davidcaron:
> 
> 
> Have you tried `vagrant provision`? I am getting this weird error:
> 
> ```
>     lvupavics: /usr/local/bin/docker-compose: line 1: Not: command not found
>     lvupavics: /usr/local/bin/docker-compose: line 1: Not: command not found
>     lvupavics: /usr/local/bin/docker-compose: line 1: Not: command not found
> ```
> 
> and `pavics-compose.sh up -d` is not run at all.

related? https://github.com/docker/compose/issues/6268

tlvu:
Yeah, docker-compose release problem, their latest tag is not available for download !

Previous version:
```sh
$ curl --silent --include https://github.com/docker/compose/releases/download/1.25.1/docker-compose-Linux-x86_64 | head -1
HTTP/1.1 302 Found
```

Latest version:
```sh
$ curl --silent --include https://github.com/docker/compose/releases/download/1.25.2/docker-compose-Linux-x86_64 | head -1
HTTP/1.1 404 Not Found
```

birdhouse/config/jupyterhub/jupyterhub_config.py.template
tlvu: It's okay, keep the rename.

tlvu:
> Or we can bind mount 2 mounts onto the Jupyterhub. 1 for all the user data, 1 for all the read-only stuff. Then backup of the user data do not uselessly also backup all the read-only stuff.

By the way, I am okay with the current `writetable-workspace` solution.  It's not ideal that user data and read-only data is mixed up but it's not the end of the world, we try something else later if this becomes a real problem.

birdhouse/config/jupyterhub/custom_templates/login.html
tlvu: Oh wait, actually we still need to keep that public nature paragraph for the public/public demo user !  

But we can add "contact us for account" like you said if user want a personalized private account with private workspace.  We can use the private workspace as a selling point for user to make the effort to contact us.

birdhouse/deployment/install-deploy-notebook
```diff
@@ -29,9 +29,10 @@ if [ ! -e "$REPO_ROOT/birdhouse/deployment/trigger-deploy-notebook" ]; then
     exit 2
 fi
 
+. "$REPO_ROOT/birdhouse/env.local"
```
tlvu: Humm, since we can not expect `JUPYTERHUB_USER_DATA_DIR` to always be in `env.local` since it is optional, `install-deploy-notebook` can't use it in `envsubst`.

tlvu: I think it's fine this is not configurable, then we can hardcode it.  All the other bind mount locations are not configurable either.

Oh right, move it to `common.env`, then `./pavics-compose.sh` have access and this `install-deploy-notebook` can source it and use in `envsubst` again.

davidcaron: see https://github.com/bird-house/birdhouse-deploy/pull/16/commits/75b7926ac79e366a38c2a038701e2ce55faef1aa

tlvu: Yeah, or old style hardcode like I did.

tlvu:
The login banner is gone https://lvupavicsdev-lvu.pagekite.me/jupyter/hub/login

tlvu:
> The login banner is gone https://lvupavicsdev-lvu.pagekite.me/jupyter/hub/login

Nevermind it's there now, not sure what happended.

birdhouse/deployment/deploy.sh
```diff
@@ -166,7 +166,7 @@ docker pull $DOCKER_NOTEBOOK_IMAGE
 
 # deploy new README.ipynb to Jupyter instance
 docker run --rm --name deploy_README_ipynb \
-    -v $DOCKER_JUPYTERHUB_USER_PERSISTENCE_VOLUME:/notebook_dir \
+    -v "$JUPYTERHUB_USER_DATA_DIR":/notebook_dir \
```
tlvu: Same for this one, `JUPYTERHUB_USER_DATA_DIR` do not exist in `env.local`.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -1,21 +1,60 @@
...
+
 c.DockerSpawner.notebook_dir = notebook_dir
-c.DockerSpawner.volumes = { os.environ['DOCKER_JUPYTERHUB_USER_PERSISTENCE_VOLUME']: notebook_dir }
+
+c.DockerSpawner.volumes = {host_user_data_dir: container_workspace_dir}
+
+host_tutorial_notebooks_dir = join(jupyterhub_data_dir, "tutorial-notebooks")
+if os.path.exists(host_tutorial_notebooks_dir):
+    c.DockerSpawner.volumes[host_tutorial_notebooks_dir] = {
+        "bind": join(notebook_dir, "tutorial-notebooks"),
```
tlvu: Will have to test tomorrow, bind mount might refresh only if container is restarted.  Typical user do not know how to restart their jupyter server.

davidcaron: Good catch, see https://github.com/bird-house/birdhouse-deploy/pull/16/commits/6ddaddc74d384299e45b0dc8d50a63e59b3cc0d5

davidcaron:
User settings like theme and keymaps are now persisted between reboots. The $HOME environment variable controls where jupyterlab keeps its settings.

tlvu:
> User settings like theme and keymaps are now persisted between reboots. The $HOME environment variable controls where jupyterlab keeps its settings.

Nice, didn't know we can change theme and stuff !

tlvu:
I'll start committing a few stuff as well, remember to always pull before doing a commit.

davidcaron:
I just found out about the `options_form` property of the spawner. We could ask the user if they want to see tutorials, if so show only read-only tutorials, if not, show only their workspace.

tlvu:
> I just found out about the `options_form` property of the spawner. We could ask the user if they want to see tutorials, if so show only read-only tutorials, if not, show only their workspace.

I think it's safer to always show the tutorials and README so when these are updated, user will be able to see them.

tlvu:
@davidcaron This is looking pretty good.  Thanks for the PR.

I added a few commits and fixes for you to review: https://github.com/bird-house/birdhouse-deploy/compare/c408051...4a9f764

birdhouse/deployment/deploy.sh
```diff
@@ -137,10 +137,12 @@ set -x
 # stop all to force reload any changed config that are volume-mount into the containers
 ./pavics-compose.sh stop
 
-# this container is not managed by docker-compose, have to handle it manually
+# user containers are not managed by docker-compose, have to handle them manually
 # rm and not just stop to force spawning newer image
-docker stop jupyter-public
-docker rm jupyter-public
+for jupyter_cont in `docker ps --format '{{.Names}}' | grep jupyter-`; do
+    docker stop $jupyter_cont
```
davidcaron: I think this is handled by the dockerspawner plugin. Everytime I restarted the jupyterhub image, the user containers got stopped and removed also. 

I'm not sure, but maybe manually stopping and removing them this way could interfere with post-stop hooks (although we don't have any right now).

davidcaron:
I looked at your commits and I agree with your changes.

I've got nothing to add, so in my opinion, this would be ready to merge.

Just remember to create the public/public user in magpie before merging.

tlvu:
@huard @tlogan2000 @moulab88 

Summarizing this PR from DavidC for your review:

* **JupyterHub authentication is integrated with Magpie**
  * Adding new user and password management is done via Magpie
  * User public/public needs to exist on Magpie for the demo account to work
  * **Since user "public" exists on Magpie, any "secure" folder on Thredds that needs authentication will be visible to that user !** Probably okay for now since everything on Thredds is public but we have to keep that in mind.  We will need that fine grained permission capabilities soon.

* **Each user have their own private workspace and Jupyter server**
  * Can only see and modify/delete their own files
  * Can only restart their own Jupyter server so can not kill other people running jobs
  * Of course this **do not apply to the demo public/public user account**

* **Login banner changed with `SUPPORT_EMAIL` from `env.local` for users to request private account/workspace**
  * Review wording please
  * Is this a good email to use?  For Ouranos it is pavics@ouranos.ca

* **New persistence on disk: `/data/jupyterhub_user_data`**
    * To include in backup

My public test server https://lvupavicsdev-lvu.pagekite.me with default Magpie admin account if you guys need to create more accounts to test the multi user support .  It has public/public, pub/pub and admin users for now.

davidcaron:
>       * **Since user "public" exists on Magpie, any "secure" folder on Thredds that needs authentication will be visible to that user !** Probably okay for now since everything on Thredds is public but we have to keep that in mind.  We will need that fine grained permission capabilities soon.

Depending on how that folder is protected in magpie, you could remove the public user from the "users" groups. Make it only part of the "anonymous" group. That might work.

birdhouse/deployment/deploy.sh
tlvu: Oh !  This is a behavior change that was not there in the previous JupyterHub version we used.  I didn't notice that, thanks for bringing it up.  That's why I had to manually kill the jupyter server of the public user previously.

Not sure I like this new behavior.  With the old behavior, I can deploy a new jupyterhub (`./pavics-compose.sh up -d`) without killing jobs running in the user jupyter server.  And when the JupyterHub is back, user can resume working as if nothing happened.

`deploy.sh` runs during night time only for prod server so it was okay that it kills user jupyter server.

tlvu: You are right, `./pavics-compose.sh stop jupyterhub` kills all the user jupyter servers !  This sucks !  Wonder why this behavior change is needed !

Can not do live update of JupyterHub during the day anymore.  I used to live update JupyterHub to publish new Jupyter image, then let Travis know so he restarts his user jupyter server at his convenient time to not kill his running jobs.

tlvu:
> Depending on how that folder is protected in magpie, you could remove the public user from the "users" groups. Make it only part of the "anonymous" group. That might work.

This is a great idea.  Is there a way to create that public user in Magpie via code (like the `permissions.cfg` file for services)?

davidcaron:
> Is there a way to create that public user in Magpie via code (like the `permissions.cfg` file for services)?

@fmigneault ?

An alternative would be to change the 'anonymous' user to 'public' (this should be possible in the config) and either provide password access to the anonymous user (now sure if possible) or pass the information to the jupyter-magpie login that the public user should always be able to connect...

tlvu:
> This is a great idea. Is there a way to create that public user in Magpie via code (like the `permissions.cfg` file for services)?

I tried this, but the user did not get created:
```sh
$ git diff
diff --git a/birdhouse/config/magpie/permissions.cfg.template b/birdhouse/config/magpie/permissions.cfg.template
index a6e23e3..76b8c1c 100644
--- a/birdhouse/config/magpie/permissions.cfg.template
+++ b/birdhouse/config/magpie/permissions.cfg.template
@@ -30,3 +30,7 @@ permissions:
     permission: read
     user: ${CATALOG_USERNAME}
     action: create
+
+  - user: public
+    group: anonymous
+    action: create
```

Also, is there a way to set the password?

davidcaron:
I think the cleanest solution would be for the jupyterhub-magpie connector to query magpie for the name of the anonymous user, and if this username was entered, let it in with any password.

tlvu:
> I think the cleanest solution would be for the jupyterhub-magpie connector to query magpie for the name of the anonymous user, and if this username was entered, let it in with any password.

Don't do this code change for now.  I was just looking for existing code that allow me to create the user automatically.

The real ideal long term scenario is user self sign up so we know who is doing what on the system.  Currently we do not know "who" is the "public" user, not good for security and stats.

fmigneault:
@tlvu 
permissions.cfg will only create the user if it needs to be created for a specific permission on a resource or service.
The config doesn't handle defining only a user/group by itself without them.

fmigneault:
@tlvu 
As @davidcaron said, on way could be to defined `MAGPIE_ANONYMOUS_USER=public`.

Otherwise, maybe call this script: 
https://github.com/Ouranosinc/Magpie/blob/9f31c9543f00048441675705cfb3432c63ba6991/magpie/helpers/create_users.py

tlvu:
> Otherwise, maybe call this script:
> https://github.com/Ouranosinc/Magpie/blob/9f31c9543f00048441675705cfb3432c63ba6991/magpie/helpers/create_users.py

The script hardcode the username same as the user email and a random password, not what I need.

I'll just add a manual steps to the README for now.

tlvu:
Auto-deploy is not enough to handle the changes in this PR.  The complete data migration and deployment steps are documented here https://github.com/bird-house/birdhouse-deploy/pull/17#issuecomment-581623219.

Went live on production Boreas at 
```
deploy finished START_TIME=2020-02-03T15:02:17-0500                                                                                               
deploy finished   END_TIME=2020-02-03T15:06:14-0500
```

PR #17, opened 2020-02-03
jupyter data migration: touch new location else jupyterhub won't bind mount them

tlvu:
See PR https://github.com/bird-house/birdhouse-deploy/pull/16
or commit
https://github.com/bird-house/birdhouse-deploy/commit/53576cc9d36642c50e4a649ca58fc8339559fd4a

See the `if os.path.exists` in the `jupyterhub_config.py`:
https://github.com/bird-house/birdhouse-deploy/blob/53576cc9d36642c50e4a649ca58fc8339559fd4a/birdhouse/config/jupyterhub/jupyterhub_config.py.template#L36-L48

### DIFF
--- a/birdhouse/scripts/migrate-jupyterhub-user-persistence
+++ b/birdhouse/scripts/migrate-jupyterhub-user-persistence
@@ -15,5 +15,7 @@ set -x
 $BASE_CMD bash -c "mkdir -p /data/jupyterhub_user_data/public \
     && cp -r /jpuser/* /data/jupyterhub_user_data/public/ \
     && rm -rf /data/jupyterhub_user_data/public/tutorial-notebooks \
+    && touch /data/jupyterhub_user_data/tutorial-notebooks \
     && rm /data/jupyterhub_user_data/public/README.ipynb \
+    && touch /data/jupyterhub_user_data/README.ipynb \
     && chown -R 1000:1000 /data/jupyterhub_user_data/public"


tlvu:
@davidcaron Sorry very tiny fall off from the previous PR.  Just need an approval since merging policy in this repo is very strict.

davidcaron:
My reasoning was that if the file or folder doesn't exist, we shouldn't try to mount it... but I understand that the files can come afterwards, when the `trigger-deploy-notebook` script runs.

Although for the `README.ipynb`, I think if the file is updated, the user's notebook maybe will not see the new file, but still the old one... not sure about that. The file id changes, and docker can't notice it.

tlvu:
Oh what you did is good.  I just had to make sure the data migration to the new location works fine.  No worries.

tlvu:
By the way, just FYI, the `README.ipynb` was deployed by `deploy.sh` after `./pavics-compose.sh up -d` so it was "too late".

`trigger-deploy-notebook` runs every hour so it'll just fill up the folder `tutorial-notebooks` later.  Or you can simply run it manually.

So if you had to repeat this migration on CRIM side, here are the steps for your reference:

* Update `env.local` if you want a different location for `JUPYTERHUB_USER_DATA_DIR`

* Deploy the new notebook deploy script: `birdhouse/deployment/install-deploy-notebook $PWD`

* Migrate the data: `birdhouse/scripts/migrate-jupyterhub-user-persistence`

* Trigger deployment: `birdhouse/deployment/deploy.sh birdhouse`

tlvu:
Tagged `1.8.0`.  Bumping the minor version since the changes to JupyterHub is quite significant.

PR #21, opened 2020-02-06
Increase JupyterHub security

tlvu:
Read each commit comments for more infos.

### DIFF
--- a/birdhouse/config/jupyterhub/jupyterhub_config.py.template
+++ b/birdhouse/config/jupyterhub/jupyterhub_config.py.template
@@ -60,6 +60,11 @@ def create_dir_hook(spawner):
 
     subprocess.call(["chown", "-R", "1000:1000", user_dir])
 
+    if username == 'public':
+        # Restrict resources for the public user
+        spawner.cpu_limit = 0.25  # 25% of 1 CPU, seems not honored by DockerSpawner
+        spawner.mem_limit = "500M"  # 500 MB
+
 c.Spawner.pre_spawn_hook = create_dir_hook
 
 ## Disable per-user configuration of single-user servers.
--- a/birdhouse/config/jupyterhub/jupyterhub_config.py.template
+++ b/birdhouse/config/jupyterhub/jupyterhub_config.py.template
@@ -4,6 +4,10 @@ import logging
 import subprocess
 
 c.JupyterHub.bind_url = 'http://:8000/jupyter'
+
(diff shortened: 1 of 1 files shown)


tlvu:
@davidcaron Please leave comments if you have any.  I just had to merge this PR quickly so I can go-live on our prod instance.

tlvu:
Went live on prod Boreas:
```
deploy finished START_TIME=2020-02-06T15:25:35-0500
deploy finished   END_TIME=2020-02-06T15:28:29-0500
```

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -62,6 +62,9 @@ def create_dir_hook(spawner):
 
 c.Spawner.pre_spawn_hook = create_dir_hook
 
+## Disable per-user configuration of single-user servers.
+c.Spawner.disable_user_config = True
```
davidcaron: Not sure, but I think this has no effets when using dockerspawner?

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -67,4 +67,7 @@ c.JupyterHub.log_level = logging.DEBUG
 
 c.Spawner.debug = True
 
+## Extra arguments to be passed to the single-user server.
+c.Spawner.args = ["--NotebookApp.terminals_enabled=False"]
```
davidcaron: The terminal is disabled, but the user can still launch bash commands from a notebook... If the terminal is not used, you're right that it should be disabled.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
tlvu: Apparently user can influence how the jupyter user server starts with stuff in HOME/.jupyter, and we persist HOME now ... so I'd rather be on the cautious side.  But agree by using dockerspawner, the risk is already reduced or this setting might have no effect.  Was following the doc in the commit message.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
tlvu: You're right, this just make it less obvious to get to the terminal and we have no use-case for the terminal yet.

tlvu:
Tagged `1.8.1`.

Issue #8, opened 2020-01-29
Monitor all components in Canarie node, both public and internal url


tlvu:
Migrated from old PAVICS https://github.com/Ouranosinc/PAVICS/issues/172.  Lots of discussion, summarized below.

The motivation was the need for some quick dashboard for the working state of all the components, not to get more stats.

Right now we bypassing twitcher, which is not real life, it's not what real users will experience.

This is ultra cheap to add and provide very fast and up-to-date (every minute) result. It's like an always on sanity check that can quickly help debugging any connectivity issues between the components.

We do not intend to scale the canarie-api monitoring to become a full blown monitoring app.

PR #22, opened 2020-02-07
Optionally monitor all components behind Twitcher using canarie api

tlvu:
Fixes https://github.com/bird-house/birdhouse-deploy/issues/8

The motivation was the need for some quick dashboard for the working state of all the components, not to get more stats.

Right now we bypassing Twitcher, which is not real life, it's not what real users will experience.

This is ultra cheap to add and provide very fast and up-to-date (every minute) result. It's like an always on sanity check that can quickly help debugging any connectivity issues between the components.

It is optional because it assumes all components are publicly accessible.  Might not be the case for everyone.  We can also override the override :D

All components in config/canarie-api/docker_configuration.py.template that do not have public (behind Twitcher) monitoring are added.

Also added Hummingbird and ncWMS2 public monitoring.

@tlogan2000 This will catch accidental Thredds public url breakage like last time and will leverage the existing monitoring on https://pavics.ouranos.ca/canarie/node/service/stats by @moulab88.

@davidcaron @dbyrns This is optional so if the CRIM do not want to enable it, it's fine.

New node monitoring page:

Screenshot_2020-02-07 Ouranos - Node Service

### DIFF
--- a/birdhouse/common.env
+++ b/birdhouse/common.env
@@ -1,4 +1,10 @@
 # All env this common.env can be overridden by env.local.
 
+# Jupyter single-user server image
 export DOCKER_NOTEBOOK_IMAGE="pavics/workflow-tests:200120"
+
+# Folder on the host to persist Jupyter user data (noteboooks, HOME settings)
 export JUPYTERHUB_USER_DATA_DIR="/data/jupyterhub_user_data"
+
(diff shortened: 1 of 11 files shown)


tlvu:
@tlogan2000 Jupyter will restart tonight, please notify your users.

tlvu:
Tagged `1.8.2`.

PR #24, opened 2020-02-12
deploy: do not delete user Jupyter server to avoid breaking user's running jobs

tlvu:
Each user will have to restart their Jupyter server themselves to get
update.

The original intend of the code was to ensure an up-to-date Jupyter
environment.  But it might be too disruptive for the users.

### DIFF
--- a/birdhouse/deployment/deploy.sh
+++ b/birdhouse/deployment/deploy.sh
@@ -137,13 +137,6 @@ set -x
 # stop all to force reload any changed config that are volume-mount into the containers
 ./pavics-compose.sh stop
 
-# user containers are not managed by docker-compose, have to handle them manually
-# rm and not just stop to force spawning newer image
-for jupyter_cont in `docker ps --format '{{.Names}}' | grep jupyter-`; do
-    docker stop $jupyter_cont
-    docker rm $jupyter_cont
-done
-
 # override git ssh command because this repo is private and need proper credentials
 #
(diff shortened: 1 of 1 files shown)


tlvu:
> My feeling is that this is a good idea... tutorial notebooks should always be updated inside the containers.

Just to be clear, this is not the auto deploy of the tutorial notebooks.  This is the auto deploy of the entire platform.

So even just a README update will trigger the auto deploy which will shutdown the user Jupyter server.  Basically we shutdown a lot more frequently than necessary.

Before, with only the public demo account, this was actually a good thing, we really do not want long running processes in the public demo account.

Now we have private accounts, we want to encourage production usage of Jupyter, breaking user's long running jobs is more of a concern.

> 
> I think short term for necessary updates you can notify the users that the containers will be restarted at a certain time (most people will not care if it's during the night), and it might be worth looking into ways to send notifitations to the users inside the jupyterlab environment. There might be plugins for that...

Yes, I was thinking along the same thing.  I'll just notify users and let them decide when the best moment for them to restart their server themselves.

Thanks for the review.

tlvu:
> I suppose the worry would be that some users will probably not think of restarting? Or even know how?

Exact.

tlvu:
Thinking about this, even if we do not shutdown the user Jupyter server anymore, during auto-deploy, all the services will be restarted so the running jobs could be impacted anyways if they uses PAVICS services.

PR #25, opened 2020-02-13
catalog: fix pavicsearch broken due to typo in config

tlvu:
The `thredds_host` should be the exact prefix of each document url found
in Solr, otherwise it is removed from the search result.

This explains why pavicsearch was returning nothing.

This will fix the `catalog_search.ipynb` notebook that keeps failing on Jenkins.

The typo was introduced in PR
https://github.com/bird-house/birdhouse-deploy/pull/5, commit
https://github.com/bird-house/birdhouse-deploy/commit/83c839178fff170dbcb4c4e0586e67d19b9cfbc5

Edit:
* add "This will fix the `catalog_search.ipynb` notebook that keeps failing on Jenkins."


tlvu:
Tagged `1.8.3`.

Issue #27, opened 2020-02-17
The autodeploy mechanism should also be able to autodeploy itself


tlvu:
Right now, everything can be "autodeployed" except the autodeploy mechanism itself and its environment (cron, git).

This is annoying and breaks the autodeploy "workflow".

User have additional manual steps to install the autodeploy directly on the host and manually deploy any updates.

We also had to deal multiple versions of git, see https://github.com/bird-house/birdhouse-deploy/blob/a45981a9bfb8a2b538a7e880dbf1ca69990291b6/birdhouse/deployment/deploy.sh#L16-L21

All the above annoyances can technically be solved if we run this autodeploy cron job inside a docker as well.  Enabling/disabling autodeploy can be configured in `env.local`, centralizing all configurations in the same place.

Having a scheduler part of PAVICS will also help share clean up jobs that's useful for all PAVICS deployments, ex: clean up of the `wps_outputs` folders so problem like this do not happen again https://github.com/bird-house/birdhouse-deploy/pull/4#issuecomment-587163363

PR #28, opened 2020-02-24
Autodeploy the autodeploy phase 1: tutorial notebooks only and not enabled by default

tlvu:
We do not need to manually deploy the autodeploy mechanism (manually run all the `install-*` stcripts under https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse/deployment) because the cronjob is also a container so all scripts are always up-to-date.

This is part 1, only the tutorial notebooks auto deployment is migrated, serving as proof-of-concept.  The install script is still left behind.  Will remove all install scripts and add full documentation once all autodeployment are migrated in another PR.

Part of issue https://github.com/bird-house/birdhouse-deploy/issues/27.

It's not immediately obvious but 3 separate containers are being spawned from inside each other, accessing the host docker engine.

* The cronjob runs inside the `scheduler` container, as seen in this PR.

* The `trigger-deploy-notebook` script (that was previously deployed at `/etc/cron.hourly/PAVICS-deploy-notebooks`) runs inside its own container `notebookdeploy`, as seen in this PR.

* The `trigger-deploy-notebook` also spawn another container `deploy_tutorial_notebooks`, not seen in this PR since it's existing code 
https://github.com/bird-house/birdhouse-deploy/blob/3ae6c14fc5888155fc37655287d3e6c3c29e2e4e/birdhouse/deployment/trigger-deploy-notebook#L88-L95

* Logging still goes to the same `/var/log/PAVICS/notebookdeploy.log` on the docker host as before.

### DIFF
--- a/birdhouse/components/scheduler/config.json
+++ b/birdhouse/components/scheduler/config.json
@@ -10,6 +10,6 @@
     "comment":"Auto-deploy tutorial notebooks",
     "schedule":"@every 15m",
     "command":"${COMPOSE_DIR}/deployment/trigger-deploy-notebook",
-    "dockerargs":"--rm --name notebookdeploy --volume /var/run/docker.sock:/var/run/docker.sock:ro --volume /var/log/PAVICS:/var/log/PAVICS:rw --volume ${COMPOSE_DIR}:${COMPOSE_DIR}:rw --volume /data:/data:rw --env COMPOSE_DIR=${COMPOSE_DIR}",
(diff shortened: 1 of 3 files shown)


tlvu:
@dbyrns @davidcaron @moulab88 Any comments?  This PR is ready to be merged.  The subset of feature is actually fully functional.  I will continue down this road so if there's is something you do not like, please let me know.

davidcaron:
I didn't have time to check this and I'm outside for the rest of the week. I only read the description and I think it's a good idea.

PR #33, opened 2020-03-12
Thredds: New "Datasets" top level for NCML files

tlvu:
http://lvupavics-lvu.pagekite.me/twitcher/ows/proxy/thredds/catalog/datasets/catalog.html (only gridded_obs/nrcan.ncml works on my dev server).

Add a new top-level "Datasets" at the same level as the existing "Birdhouse".

The content of the new top-level comes from `/data/ncml` from the host.  For comparison content of existing "Birdhouse" was coming from `/data/datasets`.

Given it's a new top level, I was able to remove WCS and WMS links, now thinking more about this ... I am not sure anymore if I should do this, suggestions?

We should also document somewhere this new "Datasets" top level is for NCML files and regular `.nc` files are still under "Birdhouse".  Let me know where to modify the doc.

### DIFF
--- a/birdhouse/config/thredds/catalog.xml.template
+++ b/birdhouse/config/thredds/catalog.xml.template
@@ -13,14 +13,6 @@
         <service name="wms" serviceType="WMS" base="/twitcher/ows/proxy/thredds/wms/" />
     </service>
 
-    <service name="ncmlservice" serviceType="Compound" base="" >
-        <service name="http" serviceType="HTTPServer" base="/twitcher/ows/proxy/thredds/fileServer/" />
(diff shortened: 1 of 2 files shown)


huard:
I think we should keep WMS and WCS if they actually work (I'm not sure). 

Let's not advertise this capability too loudly for now as it's still experimental, but yes, eventually we'll need docs for this.

tlvu:
Went live on prod Boreas:
```
triggerdeploy finished START_TIME=2020-03-17T05:07:01-0400
triggerdeploy finished   END_TIME=2020-03-17T05:08:18-0400
```

tlvu:
Tagged `1.8.6`.

PR #35, opened 2020-03-20
Jupyter: make configurable public demo user name, passwd, resource limit, login banner

tlvu:
For security reasons, the public demo username and password are not hardcoded anymore.

Compromising of one PAVICS deployment should not compromise all other PAVICS deployments if each deployment use a different password.

The password is set when the public demo user is created in Magpie, see the `birdhouse/README.md` update.

The login banner do not display the public demo password anymore.  If one really want to display the password, can use the top or bottom section of the login banner that is customizable via `env.local`.

Login banner is updated with more notices, please review wording.

Resource limits (only memory limit seems to work with the `DockerSpawner`) is also customizable.

All changes to `env.local` are live after a `./pavics-compose.sh up -d`.

Test server: https://lvupavics-lvu.pagekite.me/jupyter/ (ask me privately for the password :D)

### DIFF
--- a/birdhouse/README.md
+++ b/birdhouse/README.md
@@ -60,7 +60,7 @@ postgres instance. See [`scripts/create-wps-pgsql-databases.sh`](scripts/create-
 * Click "Save".
 
 
-### Create `public` user in Magpie for JupyterHub login
+### Create public demo user in Magpie for JupyterHub login
 
 * Go to
   https://<PAVICS_FQDN>/magpie/ui/login, login with the `admin` user,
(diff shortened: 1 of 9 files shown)


tlvu:
> where do we now configure the public user password

It's when we create that public demo user in Magpie, see README update.

tlvu:
Tagged `1.8.8`.

tlvu:
Auto-deployed on prod Boreas:
```
triggerdeploy finished START_TIME=2020-03-21T05:07:01-0400
triggerdeploy finished   END_TIME=2020-03-21T05:08:10-0400
```

PR #36, opened 2020-03-31
Allow components to be from other extra repos and add Emu to Canarie monitoring

tlvu:
Allow to have components that are in a completely separate repo (ex: thunderbird from @nikola-rados) to hook into our PAVICS stack without any code change.  Useful for new birds to test compatibility/integration with PAVICS stack early on in their development cycle.

The architecture was already ready for it, just missing template expansion to be performed on those extra repos.

Just realized Canarie monitoring can be "hooked" in as well, after my work in PR https://github.com/bird-house/birdhouse-deploy/pull/22 so I enable it for Emu for demo.

Tested with both Emu and Thunderbird in our PAVICS stack on my dev server: https://lvupavics-lvu.pagekite.me/canarie/node/service/status
Screenshot_2020-03-30 Ouranos - Node Service

### DIFF
--- a/birdhouse/optional-components/emu/.gitignore
+++ b/birdhouse/optional-components/emu/.gitignore
@@ -1,3 +1,4 @@
 wps.cfg
 emu-magpie-permission.cfg
 emu-magpie-provider.cfg
+emu_canarie_api_monitoring.py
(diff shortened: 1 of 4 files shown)


tlvu:
Relate to https://github.com/pacificclimate/thunderbird/pull/10

PR #39, opened 2020-04-03
Autodeploy the autodeploy phase 2: everything operational but a few compatibility issues remain

tlvu:
Part of https://github.com/bird-house/birdhouse-deploy/issues/27

Activating the `./components/scheduler` will do everything.  All configurations are centralized in the `env.local` file.

One missing feature is piece-wise choice of platform or notebook autodeploy only, like with the old manual `install-*` stcripts under https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse/deployment.  Right now it's all or nothing.  I can work on this if you guys think it's needed.

Remaining compatibility issues with Medus (Vagrant box works fine):
* Notebook autodeploy do not work. It looks like using the `bash` docker image, I am unable to wget any httpS address.  This same `docker run` command works fine on my Vagrant box as well.  So there's something on Medus.

```
$ docker run --rm --name debug_wget_httpS -u root bash bash -c "wget https://google.com -O -"
Connecting to google.com (172.217.13.206:443)
wget: error getting response: Connection reset by peer
```

* All the containers are being recreated when `./pavics-compose.sh` runs inside the container (first migration to the new autodeploy mechanism).  To investigate but I suspect this might be due to older version of `docker` and `docker-compose` on Medus.

* This one looks like due to older kernel on Medus:
```
sysctl: error: 'net.ipv4.tcp_tw_reuse' is an unknown key
sh: 0: unknown operand
```

* All the files updated by `git pull` are now owned by `root` (the user inside the container).  I'll have to undo this ownership change, somehow.  This one is super weird, I should have got it on my Vagrant box.  Probably Vagrant did some magic to always ensure files under `/vagrant` is always owned by the user even if changed by user `root`.

* Documentation: update README and list relevant configuration variables in `env.local` for this new `./component/scheduler`.

Migrating to this new mechanism requires manual deletion of all the artifacts created by the old install scripts: `sudo rm /etc/cron.d/PAVICS-deploy /etc/cron.hourly/PAVICS-deploy-notebooks /etc/logrotate.d/PAVICS-deploy /usr/local/sbin/triggerdeploy.sh`.  Both can not co-exist at the same time.

Maximum backward-compatibility has been kept with the old existing install scripts style:
* Still log to the same existing log files under `/var/log/PAVICS`.
* Old single ssh deploy key is still compatible, but the new mechanism allows for different ssh deploy keys for each extra repos (again, public repos should use https clone path to avoid dealing with ssh deploy keys in the first place)
* Old install scripts are kept

Features missing in old existing install scripts or how this improves on the old install scripts:
* Autodeploy of the autodeploy itself !  This is the biggest win.  Previously, if `triggerdeploy.sh` or `PAVICS-deploy-notebooks` script changes, they have to be deployed manually.  It's very annoying.  Now they are volume-mount in so are fresh on each run.
* `env.local` now drive absolutely everything, source control that file and we've got a true DevOPS pipeline.
* Configurable platform and notebook autodeploy frequency.  Previously, this means manually editing the generated cron file, less ideal.
* Do not need any support on the local host other than `docker` and `docker-compose`.  cron/logrotate/git/ssh versions are all locked-down in the docker images used by the autodeploy.  Recall previously we had to deal with git version too old on some hosts.
* Each cron job run in its own docker image meaning the runtime environment is traceable and reproducible.
* The newly introduced scheduler component is made extensible so other jobs can added into it as well (ex: backup), via `env.local`, which should source control, meaning all surrounding maintenance related tasks can also be traceable and reproducible.

This is a rather large PR.  For a less technical overview, start with the diff of README.md, env.local.example, common.env.  If a change looks funny to you, read the commit description that introduce that change, the reasoning should be there.


tlvu:
@moulab88 Found problem when testing on Medus: the notebook autodeploy do not work (it works fine on my Vagrant box).

It looks like using the `bash` docker image, I am unable to wget any httpS address.  This same `docker run` command works fine on my Vagrant box as well.  So there's something on Medus.

```
$ docker run --rm --name debug_wget_httpS -u root bash bash -c "wget https://google.com -O -"
Connecting to google.com (172.217.13.206:443)
wget: error getting response: Connection reset by peer
```

Other errors noticed on Medus that are not on my Vagrant box:

* All the containers are being recreated when `./pavics-compose.sh` runs inside the container (first migration to the new autodeploy mechanism).  To investigate but I suspect this might be due to older version of `docker` and `docker-compose` on Medus.

* This one looks like due to older kernel on Medus:
```
sysctl: error: 'net.ipv4.tcp_tw_reuse' is an unknown key
sh: 0: unknown operand
```

* All the files updated by `git pull` are now owned by `root` (the user inside the container).  I'll have to undo this ownership change, somehow.  This one is super weird, I should have got it on my Vagrant box.  Probably Vagrant did some magic to always ensure files under `/vagrant` is always owned by the user even if changed by user `root`.

moulab88:
$ docker run --rm --name debug_wget_httpS bash bash -c "apk add wget; wget https://google.com -O -"
 
# sysctl -w net.ipv4.tcp_tw_reuse=0
net.ipv4.tcp_tw_reuse = 0

tlvu:
@moulab88 thanks for quick investigation.

> $ docker run --rm --name debug_wget_httpS bash bash -c "apk add wget; wget https://google.com -O -"

It's the same identical `bash` docker image, meaning `wget` is already there and working, why install it again?!  This work-around is fishy for me.  I can buy this work-around if we don't have a choice, but let me know if you can think of something else.

> # sysctl -w net.ipv4.tcp_tw_reuse=0
> 
> net.ipv4.tcp_tw_reuse = 0

Not sure I understand your answer.

moulab88:
oh sorry for the big font. I want just to show you the command that I usually use to modify kernel parameters:
image

tlvu:
> oh sorry for the big font. I want just to show you the command that I usually use to modify kernel parameters:
> image

@moulab88, it's the following lines in `pavics-compose.sh` that cause the error `sysctl: error: 'net.ipv4.tcp_tw_reuse' is an unknown key`:

https://github.com/bird-house/birdhouse-deploy/blob/748caf9bc5b46e7b36d8b594fe59332acd5b1592/birdhouse/pavics-compose.sh#L81-L86

I think I'll build a Vagrant using Centos7 image to try reproducing this problem on my side.

birdhouse/deployment/deploy.sh
```diff
@@ -50,12 +53,10 @@
 #   are re-read.  docker-compose is not aware of any changes outside of the
 #   docker-compose.yml file.
 
-
-GREEN=$(tput setaf 2)
-YELLOW=$(tput setaf 3)
-RED=$(tput setaf 1)
-NORMAL=$(tput sgr0)
-
```
fmigneault, line 58: Why remove this?
Makes it easier to spot important items in the console that often spit out too much text to be able to read all.

moulab88:
image

birdhouse/common.env
```diff
@@ -19,3 +19,16 @@ export JUPYTER_LOGIN_BANNER_BOTTOM_SECTION=""
...
+# Note when overriding this variable in env.local, do not use HOME environment
+# var, use its fully resolved value.  This default value is suitable only for
+# backward-compatibility when autodeploy do not run in its own container.
+export AUTODEPLOY_DEPLOY_KEY_ROOT_DIR="$HOME/.ssh"
+
+# Daily at 5:07 AM
+export AUTODEPLOY_PLATFORM_FREQUENCY="7 5 * * *"
+
+# Hourly
+export AUTODEPLOY_NOTEBOOK_FREQUENCY="@hourly"
```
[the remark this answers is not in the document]
fmigneault: Ok. Saw the reference to the function that parses this in env.local.example
Maybe add a link here too to help guide what is supported here.

tlvu:
> image

It's when `pavics-compose.sh` runs inside a container on Medus that it does not work (the new autodeploy mechanism in this PR is fully containerized).  The same script + container works fine on my Ubuntu Vagrant box.  That's why I am planning to make a Centos7 Vagrant box to reproduce locally.

birdhouse/pavics-compose.sh
```diff
@@ -90,7 +94,7 @@ if [ -z "$PAVICS_FQDN_PUBLIC" ]; then
   export PAVICS_FQDN_PUBLIC="$PAVICS_FQDN"
 fi
 
-if [ "$ALLOW_UNSECURE_HTTP" = "True" ]; then
+if [ x"$ALLOW_UNSECURE_HTTP" = x"True" ]; then
```
fmigneault, line 97: I think the 'x' in this situation only makes sense if the double quotes are not present to avoid the syntax error on missing variable expansion. With the double quotes it expands as `"" = "True"` which should behave correctly.

birdhouse/pavics-compose.sh
```diff
@@ -111,15 +115,23 @@ if [ -z "$VERIFY_SSL" ]; then
...
-  while IFS= read -r -d $'\0' FILE
+find ./config ./components ./optional-components $EXTRA_CONF_DIRS -name '*.template' |
+  while read FILE
   do
     DEST=${FILE%.template}
     cat ${FILE} | envsubst "$VARS" | envsubst "$OPTIONAL_VARS" > ${DEST}
   done
 
-if [[ $1 == "up" ]]; then
+if [ x"$1" = x"up" ]; then
```
fmigneault, line 134: same here for 'x'

birdhouse/pavics-compose.sh
```diff
@@ -144,16 +156,16 @@ ERR=$?
...
+  [ ${ERR} -gt 0 ] && { echo "Error occurred with docker-compose, not running post-compose"; exit $?; }
   post-compose $*
 fi
 
 
-while [[ $# -gt 0 ]]
+while [ $# -gt 0 ]
 do
-  if [[ $1 == "up" ]]; then
+  if [ x"$1" = x"up" ]; then
```
fmigneault, line 168: and here

birdhouse/README.md
```diff
@@ -79,6 +89,10 @@ postgres instance. See [`scripts/create-wps-pgsql-databases.sh`](scripts/create-
 
 ## Mostly automated unattended continuous deployment
 
+***NOTE***: this section about automated unattended continuous deployment is
+superseded by the new `./components/scheduler` that can be entirely
+enabled/disabled via the `env.local` file.
```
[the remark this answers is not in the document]
tlvu: The list of variables to configure the autodeploy feature is right in the diff hunks above, in the same file !  But sure I can add a reference to the section above instead.

birdhouse/common.env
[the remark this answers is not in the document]
tlvu: This is the `common.env` file, a committed file containing default values, not meant for any customizations.  All customizations are in `env.local` and the reference to the parsing function is already there.

birdhouse/deployment/deploy.sh
tlvu: See why console coloring are removed in this commit message https://github.com/bird-house/birdhouse-deploy/pull/39/commits/a2dad6214b8a4786f4f9b25e0176eaad5e181ecf

birdhouse/pavics-compose.sh
tlvu: I've read it also helps to prevent code injection in shell script.  The 'x' in front basically change the command if someone put the command there instead of the expected value.  Agreed this is a bit on a paranoid side, but having the 'x' there does not harm the functionality.

tlvu:
@fmigneault @dbyrns @moulab88 I'd like to merge this PR quickly.

It's a new optional feature not enabled by default (I've removed the README update that advertize it).  All other changes are backward-compatible.  The old-style install scripts are still there.

So even if this new feature still have problems, it does not break anything.  The remaining problems look like compatibility problems rather than broken feature (it's fully functional on my Vagrant box).

I'll make another PR to fix the remaining problems.

So waiting for your approval unless you see something absolutely blocking.

tlvu:
> Maybe make an issue to keep track of it so it doesn't get lost.

@fmigneault I added a note in this PR description, in the "remaining issue" section.  It's the stuff that the next PR will need to address.

tlvu:
Tagged `1.8.10`.

Auto deployed to prod:
```
triggerdeploy finished START_TIME=2020-04-10T05:07:01-0400
triggerdeploy finished   END_TIME=2020-04-10T05:08:23-0400
```

PR #41, opened 2020-04-09
Add generic_bird optional component

tlvu:
Could be used as work-around for issue https://github.com/bird-house/finch/issues/121 so we have a Finch dedicated for async request only.

This component is generic and configurable enough it could even accommodate any WPS.  It has its own Postgres instance so we can use it later to experiment with different version of Postgres.

Test server: https://lvupavics-lvu.pagekite.me/twitcher/ows/proxy/finchasync, https://lvupavics-lvu.pagekite.me/canarie/node/service/status

See the `README.md` update for more info.

Note the docker image, service name and port is customizable (my test server do not use the default service name).

It starts out as a Finch2 but now its rather a generic skeleton to plugin any birds we need. I wish I had this when I was testing out the new Thunderbird from PCIC.

At some point in the past, we had the idea of 2 simultaneous FlyingPigeon because the new one is so different we wanted to keep the old one around for a short while. This PR could have helped.

So do not view this PR as Finch 2 anymore, view this PR as a mechanism to quickly deploy another bird, any bird, without having to change code.

This opens up many options, testing is just one of the possible usage.


tlvu:
> I'm uncomfortable with this. I suppose it's ok as a temporary measure, but I don't like the idea of having multiple versions of birds without a clear purpose and framework.

It's an optional generic component, if this looks confusing, we don't need to enable it.

> 
> We now have a secondary server for thredds and finch. Should we uniformize how we deal with this ? Ie thredds-test, finch-test, etc.

It starts out as a Finch 2 but now its rather a generic skeleton to plugin any birds we need.  I wish I had this when I was testing out the new Thunderbird from PCIC.

At some point in the past, we had the idea of 2 simultaneous FlyingPigeon because the new one is so different we wanted to keep the old one around for a short while.  This PR could have helped.

So do not view this PR as Finch 2 anymore, view this PR as a mechanism to quickly deploy another bird, any bird, without having to change code.

> Should test servers inherit their config from their parent ?

No, this is a standalone bird.

> How do these test servers differ from our staging environment ?

Good question, same reason why we have Test Thredds: the staging and prod environment is not exactly identical, not same data on Thredds.

This opens up many options, testing is just one of the possible usage.

tlvu:
Auto deployed to prod:
```
triggerdeploy finished START_TIME=2020-04-16T05:07:01-0400
triggerdeploy finished   END_TIME=2020-04-16T05:08:25-0400
```

PR #42, opened 2020-04-22
vagrant: add centos7 and LetsEncrypt SSL cert support, fix scheduler autodeploy remaining issues

tlvu:
Fixes https://github.com/bird-house/birdhouse-deploy/issues/27.

Centos7 support added to Vagrant to reproduce problems found on Medus in PR https://github.com/bird-house/birdhouse-deploy/pull/39 (commit https://github.com/bird-house/birdhouse-deploy/commit/6036dbd5ff072544d902e7b84b5eff361b00f78b):

Problem 1: wget httpS url not working in bash docker image breaking the notebook autodeploy when running under the new scheduler autodeploy: **not reproducible**

Problem 2: all containers are destroyed and recreated when alternating between manually running `./pavics-compose.sh up -d` locally and when the same command is executed automatically by the scheduler autodeploy inside its own container: **not reproducible**

Problem 3: `sysctl: error: 'net.ipv4.tcp_tw_reuse' is an unknown key` on `./pavics-compose.sh up -d` when executed automatically by the scheduler autodeploy inside its own container: **reproduced** but seems **harmless** so **not fixing** it.

Problem 4: current user lose write permission to birdhouse-deploy checkout and other checkout in `AUTODEPLOY_EXTRA_REPOS` when using scheduler autodeploy: **fixed**

Problem 5: no documentation for the new scheduler autodeploy: **fixed**

Another autodeploy fix found while working on this PR: notebook autodeploy broken when `/data/jupyterhub_user_data/tutorial-notebooks` dir do not pre-exist.  Regression from this commit https://github.com/bird-house/birdhouse-deploy/pull/16/commits/6ddaddc74d384299e45b0dc8d50a63e59b3cc0d5 (PR https://github.com/bird-house/birdhouse-deploy/pull/16): before that commit the entire dir was copied, not just the content, so the dir was created automatically.

Centos7 Vagrant box experience is not completely automated as Ubuntu box, even when using the same vagrant-disksize Vagrant plugin as Ubuntu box.  Manual disk resize instruction is provided.  Candidate for automation later if we destroy and recreate Centos7 box very often.  Hopefully the problem is not there for Centos8 so we can forget about this annoyance.

Automatic generation of SSL certificate from LetsEncrypt is also added for both Ubuntu and Centos Vagrant box.  Can be used outside of Vagrant so Medus and Boreas can also benefit next time, if needed.  Later docker image of `certbot` is used so should already be using ACMEv2 protocol (ACMEv1 is being deprecated).

Pagekite is also preserved for both boxes for when exposing port 80 and 443 directly on the internet is not possible but PAVICS still need a real SSL certificate.

Test server: https://lvupavicsmaster.ouranos.ca (Centos7, on internet with LetsEncrypt SSL cert).

Jenkins run only have known errors: http://jenkins.ouranos.ca/job/ouranos-staging/job/lvupavicsmaster.ouranos.ca/4/console

2020-04-22-070604_1299x1131_scrot

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -36,6 +36,7 @@ Vagrant.configure("2") do |config|
 
   config.vm.provision :shell, path: "birdhouse/vagrant-utils/provision.sh", env: {"VM_HOSTNAME" => settings['hostname'],
                                                                                   "VM_DOMAIN" => settings['domain'],
+                                                                                  "LETSENCRYPT_EMAIL" => settings['letsencrypt_email'],
                                                                                   "KITENAME" => settings.fetch('kitename',''),
                                                                                   "KITESECRET" => settings.fetch('kitesecret',''),
                                                                                   "KITESUBDOMAIN" => settings.fetch('kitesubdomain',''),
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,7 +4,7 @@ settings=YAML.load_file(config_file)
 
 Vagrant.configure("2") do |config|
   config.vagrant.plugins = ["vagrant-disksize", "vagrant-vbguest"]
-  config.vm.box = "ubuntu/bionic64"
+  config.vm.box = settings.fetch('box', "ubuntu/bionic64")
   config.vm.define settings['hostname']
   config.vm.hostname = settings['hostname']
   # thin provisioning, won't take 100G upfront
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -14,6 +14,10 @@ Vagrant.configure("2") do |config|
   # 2019-07-03 with VirtualBox 6.0.8
   config.vbguest.auto_update = false
 
+  # https://blog.centos.org/2018/01/updated-centos-vagrant-images-available-v1801-01/
+  # Fix /vagrant shared folders (together with vagrant-vbguest) for Centos 7.
+  config.vm.synced_folder ".", "/vagrant", type: "virtualbox"
+
   # bridge networking to get real DNS name on local network, PAVICS does not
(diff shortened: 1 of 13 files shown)


tlvu:
> We also need to verify that the renew will work after 1 june 2020:
> https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430

Interesting, will keep this in mind.  I use the latest certbot docker image, I did not have to register any account, everything just works.  I would assume this latest certbot client uses ACMEv2 already.

tlvu:
> > We also need to verify that the renew will work after 1 june 2020:
> > https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430
> 
> Interesting, will keep this in mind. I use the latest certbot docker image, I did not have to register any account, everything just works. I would assume this latest certbot client uses ACMEv2 already.

Found this in the help of the certbot client, looks like it's ACMEv2 by default already:
```
  --server SERVER       ACME Directory Resource URI. (default:                
                        https://acme-v02.api.letsencrypt.org/directory) 
```

tlvu:
Tagged `1.9.0`.  Bumping the minor since having the autodeploy self-update itself is quite milestone.

Migrating to this new mechanism requires manual deletion of all the artifacts created by the old install scripts: `sudo rm /etc/cron.d/PAVICS-deploy /etc/cron.hourly/PAVICS-deploy-notebooks /etc/logrotate.d/PAVICS-deploy /usr/local/sbin/triggerdeploy.sh`.  Both can not co-exist at the same time.

For more info on this new mechanism, see https://github.com/bird-house/birdhouse-deploy/pull/39.

PR #54, opened 2020-06-17
flyingpigeon: set maxsingleinputsize = 2097152000.0

tlvu:
Requested by @nilshempelmann for a soilmoisture data set.

This is a 2G file limit !  I wonder if it is possible to give a URL path and the WPS service will fetch the file itself so we do not have to set such a big file limit.

### DIFF
--- a/birdhouse/config/flyingpigeon/wps.cfg.template
+++ b/birdhouse/config/flyingpigeon/wps.cfg.template
@@ -1,6 +1,7 @@
 [server]
 outputurl = https://${PAVICS_FQDN_PUBLIC}/wpsoutputs
 outputpath = /data/wpsoutputs
+maxsingleinputsize = 2097152000.0
 
 [logging]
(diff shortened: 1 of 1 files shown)


tlvu:
@nilshempelmann new FP settings deployed to https://lvupavicsdev.ouranos.ca, can you test it works as intended before we go to prod with this.

nilshempelmann:
@tlvu 
> @nilshempelmann new FP settings deployed to https://lvupavicsdev.ouranos.ca, can you test it works as intended before we go to prod with this.

`Invalid username or password` gues I need an valid login. can you send it per mail?

nilshempelmann:
'ProcessSucceeded'

tlvu:
@nilshempelmann Thanks Nils for the test, this will be auto deployed during the night by our automated system.  I can force the deployment now if you have an urgent need for it.

Issue #12, opened 2020-01-29
Need to add some monitoring and notification to the automated deployment system


tlvu:
Migrated from old PAVICS https://github.com/Ouranosinc/PAVICS/issues/140

Automated deployment was triggered but not performed on `boreas` because of
```
++ git status -u --porcelain
+ '[' '!' -z '?? birdhouse/old_docker-compose.override.yml_18062019' ']'
+ echo 'ERROR: unclean repo'
ERROR: unclean repo
+ exit 1
```

We need some system to monitor the logs and send notification if there are any errors.  This log file error monitoring and notification can be generalized to watch any systems later so each system is not forced to reimplement monitoring and notification.

This problem has triggered this issue https://github.com/Ouranosinc/PAVICS/issues/176

huard:
Managed to get Fluentd to parse NGinx logs. 

1. Build docker image with the prometheus plugin
```
FROM fluent/fluentd:edge

# Use root account to use apk
USER root

# below RUN includes plugin as examples elasticsearch is not required
# you may customize including plugins as you wish
RUN apk add --no-cache --update --virtual .build-deps \
        sudo build-base ruby-dev \
 && sudo gem install fluent-plugin-prometheus \
 && sudo gem sources --clear-all \
 && apk del .build-deps \
 && rm -rf /tmp/* /var/tmp/* /usr/lib/ruby/gems/*/cache/*.gem

# COPY fluent.conf /fluentd/etc/
COPY entrypoint.sh /bin/

USER fluent
```

2. Configure `fluent.conf`:
```
<source>
    @type tail
    path /var/log/nginx/access_file.log.*
    # pos_file /var/log/td-agent/nginx-access-file.pos  # Store tail position across restarts
    # follow_inodes true  # Without this parameter, file rotation causes log duplication.
    refresh_interval 2
    <parse>
      @type regexp
      expression /^(?<remote>[^ ]*) (?<host>[^ ]*) (?<user>[^ ]*) \[(?<time>[^\]]*)\] \"(?<method>\w+)(?:\s+(?<path>[^\"]*?)(?:\s+\S*)?)?\" (?<status_code>[^ ]*) (?<size>[^ ]*)(?:\s"(?<referer>[^\"]*)") "(?<agent>[^\"]*)" (?<urt>[^ ]*)$/
        time_format %Y-%m-%dT%H:%M:%S%:z
        keep_time_key true
        types size:integer,urt:float
    </parse>
    tag nginx
[...]

PR #56, opened 2020-06-20
Monitoring for host and each docker container

tlvu:
Screenshot_2020-06-19 Docker and system monitoring - Grafana

For host, using Node-exporter to collect metrics:
* uptime
* number of container
* used disk space
* used memory, available memory, used swap memory
* load
* cpu usage
* in and out network traffic 
* disk I/O

For each container, using cAdvisor to collect metrics:
* in and out network traffic
* cpu usage
* memory and swap memory usage
* disk usage

Useful visualisation features:
* zoom in one graph and all other graph update to match the same "time range" so we can correlate event
* view each graph independently for more details
* mouse over each data point will show value at that moment

Prometheus is used as the time series DB and Grafana is used as the visualization dashboard.

Node-exporter, cAdvisor and Prometheus are exposed so another Prometheus on the network can also scrape those same metrics and perform other analysis if required.

The whole monitoring stack is a separate component so user is not forced to enable it if there is already another monitoring system in place.  Enabling this monitoring stack is done via `env.local` file, like all other components.

The Grafana dashboard is taken from https://grafana.com/grafana/dashboards/893 with many fixes (see commits) since most of the metric names have changed over time.  Still it was much quicker to hit the ground running than learning the Prometheus query language and Grafana visualization options from scratch.  Not counting there are lots of metrics exposed, had to filter out which one are relevant to graph.  So starting from a broken dashboard was still a big win.  Grafana has a big collection of existing but probably un-maintained dashboards we can leverage.

So this is a first draft for monitoring.  Many things I am not sure or will need tweaking or is missing:
* Probably have to add more metrics or remove some that might be irrelevant, with time we will see.
* Probably will have to tweak the scrape interval and the retention time, to keep the disk storage requirement reasonable, again we'll see with time.
* Missing alerting.  With all the pretty graph, we are not going to look at them all day, we need some kind of alerting mechanism.

Test system: http://lvupavicsmaster.ouranos.ca:3001/d/pf6xQMWGz/docker-and-system-monitoring?orgId=1&refresh=5m, user: admin, passwd: the default passwd

Part of issue https://github.com/bird-house/birdhouse-deploy/issues/12


tlvu:
http://medus.ouranos.ca:3001/d/pf6xQMWGz/docker-and-system-monitoring?orgId=1&from=now-24h&to=now&refresh=5m (cAdvisor not working on Medus for now, so all stats related to containers are missing).

tlvu:
Medus now working properly after 2 days uptime: http://medus.ouranos.ca:3001/d/pf6xQMWGz/docker-and-system-monitoring?orgId=1&refresh=5m (on Medus had to perform full yum update to get new kernel and new docker engine for cAdvisor to work properly).

Tagged `1.10.0`.  Bumping the minor as this is a new component added.

tlvu:
Autodeployed to prod:
```
triggerdeploy finished START_TIME=2020-07-03T05:07:02+0000
triggerdeploy finished   END_TIME=2020-07-03T05:10:35+0000
```

Issue #62, opened 2020-07-30
Monitoring settings in env.local is required even if monitoring component is not enabled


tlvu:
Given https://github.com/bird-house/birdhouse-deploy/blob/d048bd5e6102efcb2fc9837c2bc2ceba68060825/birdhouse/pavics-compose.sh#L28-L29

User are forced to set those vars even if the Monitoring component is not enabled.

Further more, the following data volumes are also created even when the Monitoring component is not enabled.
https://github.com/bird-house/birdhouse-deploy/blob/d048bd5e6102efcb2fc9837c2bc2ceba68060825/birdhouse/pavics-compose.sh#L151-L153

We need a way for each component to inject their various requirements without hardcoding them in `pavics-compose.sh` script or anywhere else than inside their own component folder.

Found out by @nikola-rados.

PR #64, opened 2020-08-05
Improved plugable component architecture

tlvu:
Before this PR, components needing default values, needing template variable substitution, needing to execute commands pre and post `docker-compose up` are hardcoding their needs directly to the "core" system, basically "leaking" their requirements out even when they are not activated (fixes https://github.com/bird-house/birdhouse-deploy/issues/62).

This PR provides true plugable architecture for the components so they can provide all their needs without having to modify the code of the "core" system.

All the components (monitoring, generic_bird, emu, testthredds) are modified to leverage the new plugable architecture, with additional customizations given it is cleaner/easier to have default configuration values.

Given this PR both changes the architecture and modify many components at the same time, it is best to read each commit separately to easier understand which code change belongs to which "goal".

Deployed here https://lvupavicsmaster.ouranos.ca with all the impacted components activated to test the change:
* Canarie: https://lvupavicsmaster.ouranos.ca/canarie/node/service/status
* Generic bird (using Finch): https://lvupavicsmaster.ouranos.ca/twitcher/ows/proxy/generic_bird?service=WPS&version=1.0.0&request=GetCapabilities
* Emu: https://lvupavicsmaster.ouranos.ca/twitcher/ows/proxy/emu?service=WPS&version=1.0.0&request=GetCapabilities
* Test Thredds: https://lvupavicsmaster.ouranos.ca/testthredds/catalog.html
* Prometheus: http://lvupavicsmaster.ouranos.ca:9090/alerts
* AlertManager: http://lvupavicsmaster.ouranos.ca:9093/
* Grafana dashboard: http://lvupavicsmaster.ouranos.ca:3001/d/pf6xQMWGz/docker-and-system-monitoring?orgId=1&refresh=5m


tlvu:
> I see more and more image name moving from the compose file to an env, what is the rational for this? (My fear is for the component tracking system with auto PR that will have a hard time to update the image version)

The reason for image moving from the `docker-compose.yml` into the `default.env` file is for multiple components to  be able to refer to the same image (ex: thredds and testthredds, finch and generic_bird).

I also found an unexpected benefit to this, having the image in the `default.env` file means it can be overridden by `env.local` which in turns means we can quickly deploy new image without even committing code (just override the value in `default.env` in `env.local`).  Useful for emergency fix not having to wait for PR approval or on test machine to test new build without having to modify the code.  New Jupyter image is deployed to our staging Medus even before the PR is merged via this override mechanism.

For the future auto image update detection, maybe we can move all images into the `default.env` file and the tool will lookup for a list of image variable to scan for update, something like `IMAGES_TO_UPDATE='$FINCH_IMAGE $THREDDS_IMAGE ...` and so on.

> Also if you are not in a rush before your vacation, I would like that Mathieu take a look at this PR too.

I wanted to add @MatProv as reviewer but he was not in the list of contributor for this repo so I sent him an invite last night already.

birdhouse/pavics-compose.sh
```diff
@@ -66,6 +66,18 @@ cd $(dirname $(readlink -f $0 || realpath $0))
...
+  COMPONENT_DEFAULT_ENV="$adir/default.env"
+  if [ -f "$COMPONENT_DEFAULT_ENV" ]; then
+    echo "reading '$COMPONENT_DEFAULT_ENV'"
+    . "$COMPONENT_DEFAULT_ENV"
+  fi
+done
+
+# Re-read env.local to make sure it can override ALL defaults from all
+# components.
+[ -f env.local ] && . ./env.local
```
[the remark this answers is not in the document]
tlvu: Same priority as before with the `docker-compose-extra.yml` fragment https://github.com/bird-house/birdhouse-deploy/blob/cfaf768bcd79140afd609585b4b0c4ebc7ac8915/birdhouse/env.local.example#L55-L57

I can elaborate a little bit more.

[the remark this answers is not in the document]
tlvu: done be1581a

dbyrns:
> I wanted to add @MatProv as reviewer but he was not in the list of contributor for this repo so I sent him an invite last night already.

Thank you for that.

birdhouse/components/monitoring/default.env
```diff
@@ -0,0 +1,28 @@
...
+export ALERTMANAGER_EXTRA_RECEIVERS=""
+
+
+
+# add vars only needed to be substituted in templates
+
+VARS="
+  $VARS
+  \$ALERTMANAGER_ADMIN_EMAIL_RECEIVER
+  \$SMTP_SERVER
```
MatProv, line 19: +1 on this one
I was wondering why these variable needed to be defined even if monitoring wasn't enabled.
This offers more flexibility, great.

birdhouse/pavics-compose.sh
```diff
@@ -25,8 +25,6 @@ VARS='
   $POSTGRES_PAVICS_PASSWORD
   $POSTGRES_MAGPIE_USERNAME
   $POSTGRES_MAGPIE_PASSWORD
-  $ALERTMANAGER_ADMIN_EMAIL_RECEIVER
-  $SMTP_SERVER
```
MatProv, line 29: +1 on this one
I was wondering why these variable needed to be defined even if monitoring wasn't enabled.
This offers more flexibility, great.

birdhouse/config/thredds/entrypointwrapper
```diff
@@ -20,8 +20,12 @@ fi
...
 WEBAPPS_ROOT="/usr/local/tomcat/webapps"
 EXISTING_CONTEXT_ROOT="thredds"
-WANTED_CONTEXT_ROOT="twitcher/ows/proxy/thredds"
-WANTED_CONTEXT_ROOT_WARFILE_NAME="twitcher#ows#proxy#thredds"
+if [ -z "$WANTED_CONTEXT_ROOT" ]; then
+    WANTED_CONTEXT_ROOT="twitcher/ows/proxy/thredds"
+fi
+if [ -z "$WANTED_CONTEXT_ROOT_WARFILE_NAME" ]; then
+    WANTED_CONTEXT_ROOT_WARFILE_NAME="twitcher#ows#proxy#thredds"
+fi
```
MatProv, line 28: Not sure of both variable names starting with `WANTED_`. 
Would only `CONTEXT_ROOT` do the job instead?

birdhouse/env.local.example
```diff
@@ -52,9 +52,19 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
...
+#   * Last docker-compose volume mount to same destination win over the
+#     previous docker-compose volume mount.
+#
+#   * Last default.env can change the values of all previous default.env.
+#
+#   * Last pre/post docker-compose-up script can potentially undo actions
+#     from previous scripts.
+#
+# Suggested to keep the private-config-repo last in the list to be able to
+# override anything.
```
MatProv, line 67: Thanks for specifying. 
It is worth mentioning since with the pre/post hooks and custom configs, precedence can become tricky at a certain point.

MatProv:
> The reason for image moving from the `docker-compose.yml` into the `default.env` file is for multiple components to be able to refer to the same image (ex: thredds and testthredds, finch and generic_bird).
> 
> I also found an unexpected benefit to this, having the image in the `default.env` file means it can be overridden by `env.local` which in turns means we can quickly deploy new image without even committing code

Great benefits. Of course we need to manually update the `env.local` to contain the updated versions specified in `default.env`. Simple solution is to comment image versions overrides in `env.local` to use the ones from `default.env`. @dbyrns Looking at our multiple projects using `compose.sh` scripts, I don't see reason why this practice might not be good. Especially for templates overrides, which sometimes require to have the image version to be passed at other places, that version declaration syntax gets pretty useful.

I went through this PR and I agree with the changes, especially the one related to extra config not having to be defined if the extra is not used. 

@tlvu Please re-send me the invitation since it was a time limited offer

birdhouse/config/thredds/entrypointwrapper
tlvu: Existing variable name, I did not want to change the existing code too much.  There is `EXISTING_CONTEXT_ROOT` and this `WANTED_CONTEXT_ROOT`.  Does this naming really bother you a lot?

birdhouse/env.local.example
tlvu: Agreed.  The variable names in each component `default.env` file should not collide else one will override the other one.  Probably not an issue for components in this repo but if we pull in components from other repos, this might become an issue.

birdhouse/config/thredds/entrypointwrapper
MatProv: Not at all. Seemed strange when I first saw it but I would probably not even notice if I haven't reviewed this part.

tlvu:
Tagged `1.11.0` as this is an architecture design change that allows much more freedom and increased capabilities for the components.

tlvu:
The automated deployment is broken due to the rename of `common.env` to `default.env`.

Error in logs:
```
+ . ./default.env
/tmp/latestdeploy.sh: .: line 124: can't open './default.env': No such file or directory
```

Easy and one time only work-around is to manually perform a `git pull` in this `birdhouse-deploy` checkout, followed by `./deployment/deploy.sh . ./env.local`, assuming current path is `<birdhouse-deploy checkout>/birdhouse`.

PR #72, opened 2020-10-09
Sync Raven testdata to Thredds for Raven tutorial notebooks

tlvu:
Leveraging the cron daemon of the scheduler component, sync Raven testdata to Thredds for Raven tutorial notebooks.

Activation of the pre-configured cronjob is via `env.local` as usual for infra-as-code.

New generic `deploy-data` script can clone any number of git repos, sync any number of folders in the git repo to any number of local folders, with ability to cherry-pick just the few files needed (Raven testdata has many types of files, we only need to sync `.nc` files to Thredds, to avoid polluting Thredds storage `/data/datasets/testdata/raven`).

Limitation of the first version of this `deploy-data` script:
* Do not handle re-organizing file layout, this is a pure sync only with very limited rsync filtering for now (tutorial notebooks deploy from multiple repos, need re-organizing the file layout)

So the script has room to grow.  I see it as a generic solution to the repeated problem "take files from various git repos and deploy them somewhere automatically".  If we need to deploy another repo, juste write a new config file, stop writing boilerplate code again.

Minor unrelated change in this PR:
* README update to reference the new birdhouse-deploy-ouranos.
* Make sourcing the various pre-configured cronjob backward-compat with older version of the repo where those cronjob did not exist yet.

### DIFF
--- a/birdhouse/README.md
+++ b/birdhouse/README.md
@@ -25,7 +25,10 @@ for your organization.  For an example of possible override, see how the [emu
 service](optional-components/emu/docker-compose-extra.yml)
 ([README](optional-components/README.md)) can be optionally added to the
 deployment via the [override
-mechanism](https://docs.docker.com/compose/extends/).
+mechanism](https://docs.docker.com/compose/extends/).  Ouranos specific
+override can be found in this
+[birdhouse-deploy-ouranos](https://github.com/bird-house/birdhouse-deploy-ouranos)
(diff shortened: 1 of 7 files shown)


huard (approved):
May want a second review from someone more competent in bash.

MatProv (approved):
LGTM

tlogan2000 (approved):

PR #102, opened 2020-11-10
bump magpie and twitcher to 3.2.1

MatProv:
* bump `magpie` to 3.2.0
* bump `twitcher` to magpie-3.2.0
* let the anonymous user browse Thredds


MatProv:
@tlvu @fmigneault This PR replaces [bump_twitcher_to_magpie-3.1.0 -> master](https://github.com/bird-house/birdhouse-deploy/pull/89), with a cleaner diff.

fmigneault (approved):

tlvu:
Thanks for the cleaner diff !  Much appreciated.

@fmigneault will move part of https://github.com/bird-house/birdhouse-deploy/pull/99#issuecomment-725027773 here, then we can test everything together.

And let me do this merge myself once I also test an upgrade on my perso server then our staging server.  The pipeline only test fresh new install, but in production it's actually an upgrade with existing Magpie DB.  This is a fairly big change so I'd like to be more cautious.

tlvu:
@MatProv can you also enable @fmigneault new `./optional/components/secure-thredds` (in this PR) so we can test this new revoke permission feature?

MatProv:
> can you also enable new `./optional/components/secure-thredds` (in this PR) so we can test this new revoke permission feature?

@tlvu It's now enabled and will be applied in future instance creation. Will need to be remove as soon as we don't need it, in test instances.

tlvu:
I deployed this PR on my dev server with `all-public-access` and `secure-thredds` components both activated and my secure folder https://lvupavicsdev.ouranos.ca/twitcher/ows/proxy/thredds/catalog/birdhouse/testdata/secure/catalog.html is still accessible publicly.

In the logs I see
```
[2020-11-13 22:17:00,930] INFO       [MainThread][magpie.register] Processing permissions from configuration (4/4).                                   
[2020-11-13 22:17:00,930] INFO       [MainThread][magpie.register] Found 3 permissions to evaluate from configuration.                                
[2020-11-13 22:17:00,930] WARNING    [MainThread][magpie.register] Unknown permission [browse-deny-recursive] [permission #0], skipping...            
[2020-11-13 22:17:00,930] WARNING    [MainThread][magpie.register] Unknown permission [read-deny-recursive] [permission #1], skipping...              
[2020-11-13 22:17:00,931] WARNING    [MainThread][magpie.register] Unknown permission [write-deny-recursive] [permission #2], skipping...  
```

Looks like the config is not applied properly.

fmigneault:
@tlvu  FYI
I found the problem, working on a fix that should come up soon.

tlvu:
@MatProv is it possible for the DACCS-iac pipeline to save docker logs of all the containers to the job before killing the server?  Then to save those files to the Jenkins job you can do something like https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/blob/5d5a9aa2251386378406efb5b414b3aa6db0b37e/Jenkinsfile#L80-L81

MatProv:
> @MatProv is it possible for the DACCS-iac pipeline to save docker logs of all the containers to the job before killing the server? Then to save those files to the Jenkins job you can do something like https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/blob/5d5a9aa2251386378406efb5b414b3aa6db0b37e/Jenkinsfile#L80-L81

@tlvu Yes absolutely, there might be a way to forward the instance's logs to jenkins artifacts. I will look at this soon, thanks.

MatProv:
@tlvu FYI docker-compose logs are now stored as jenkins artifacts.
You can also access it using `date | netcat IP_MACHINE 9875 > docker-compose.log`.

fmigneault:
@MatProv @tlvu 
Magpie `3.2.1` is available with fix of permission parsing. 
I would suggest also adding `type: directory` field to THREDDS entries to make sure there is not any ambiguity when the resource must be created by Magpie.

tlvu:
> I would suggest also adding `type: directory` field to THREDDS entries to make sure there is not any ambiguity when the resource must be created by Magpie.

@fmigneault Thanks for the fix.  Since you know it best, can you add it?

Since CRIM owns Twitcher/Magpie, I prefer to let you guys deal with any config changes required for the update.

I can help final testing once you guys confirm the PR is good to go.  You guys should make sure to test both fresh deployment and upgrade existing deployment.

fmigneault:
@tlvu @MatProv 
8f8eb2a should do the trick

MatProv:
run tests

tlvu:
Crawler didn't work https://host-140-4.rdext.crim.ca/wpsoutputs/catalog/f749898e-29e0-11eb-b5a4-0242ac130014.xml, regression in Magpie or infra issue?
```
<?xml version="1.0" encoding="UTF-8"?>
<wps:ExecuteResponse xmlns:wps="http://www.opengis.net/wps/1.0.0" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wps/1.0.0 ../wpsExecute_response.xsd" service="WPS" version="1.0.0" xml:lang="en-US" serviceInstance="http://localhost/wps?request=GetCapabilities&amp;amp;service=WPS" statusLocation="https://host-140-4.rdext.crim.ca/wpsoutputs/catalog/f749898e-29e0-11eb-b5a4-0242ac130014.xml">
    <wps:Process wps:processVersion="0.1">
        <ows:Identifier>pavicrawler</ows:Identifier>
        <ows:Title>PAVICS Crawler</ows:Title>
        <ows:Abstract>Crawl thredds server and write metadata to SOLR database.</ows:Abstract>
	</wps:Process>
    <wps:Status creationTime="2020-11-18T20:59:44Z">
        <wps:ProcessFailed>
            <wps:ExceptionReport>
                    <ows:Exception exceptionCode="NoApplicableCode" locator="None">
                            <ows:ExceptionText>Process error: method=wps_pavicrawler.py._handler, line=146, msg=Traceback (most recent call last):
  File &#34;/usr/local/lib/python2.7/dist-packages/pavics_datacatalog-0.6.11-py2.7.egg/pavics_datacatalog/wps_processes/wps_pavicrawler.py&#34;, line 144, in _handler
    headers=headers, verify=self.verify)
  File &#34;/usr/local/lib/python2.7/dist-packages/pavics/catalog.py&#34;, line 476, in pavicrawler
    headers=headers, verify=verify)
  File &#34;/usr/local/lib/python2.7/dist-packages/pavics/catalog.py&#34;, line 280, in thredds_crawler
    verify=verify):
  File &#34;/usr/local/lib/python2.7/dist-packages/threddsclient/client.py&#34;, line 33, in crawl
    for ds in crawl(ref.url, skip, depth - 1, **kwargs):
  File &#34;/usr/local/lib/python2.7/dist-packages/threddsclient/client.py&#34;, line 33, in crawl
    for ds in crawl(ref.url, skip, depth - 1, **kwargs):
  File &#34;/usr/local/lib/python2.7/dist-packages/threddsclient/client.py&#34;, line 33, in crawl
    for ds in crawl(ref.url, skip, depth - 1, **kwargs):
  File &#34;/usr/local/lib/python2.7/dist-packages/threddsclient/client.py&#34;, line 28, in crawl
    cat = read_url(url, skip, **kwargs)
  File &#34;/usr/local/lib/python2.7/dist-packages/threddsclient/client.py&#34;, line 52, in read_url
    return read_xml(req.text, url)
  File &#34;/usr/local/lib/python2.7/dist-packages/threddsclient/client.py&#34;, line 73, in read_xml
    raise ValueError(&#34;Does not appear to be a Thredds catalog&#34;)
ValueError: Does not appear to be a Thredds catalog
</ows:ExceptionText>
                    </ows:Exception>
            </wps:ExceptionReport>
        </wps:ProcessFailed>
	</wps:Status>
</wps:ExecuteResponse>
```

tlvu:
@fmigneault The various icons are not showing up.  See https://host-140-4.rdext.crim.ca/thredds/ vs working prod https://pavics.ouranos.ca/thredds/.  Pretty sure it's permissions issues.

I'd suggest to not hardcode those .gif filename but allow them to be configured via a config file so if upstream Thredds change them, we do not have to modify the code.

MatProv:
As discussed with @tlvu , the issue is probably related to the crawler not able to access thredds, since permissions introduced in latest commits.

tlvu:
@fmigneault by the way, can you add browse and read recursive on `/birdhouse/testdata/secure` for user `authtest`.  That user can not access those resources anymore since group `anonymous` is locked out from that resource.

tlvu:
@fmigneault I remembered we have discussed that there will be a `thredds-config.yml` of some sort that will list which url is data vs metadata.  Should that new config file be added in this PR to match the new magpie?

tlvu:
@fmigneault @MatProv Investigation about the crawler failure:

The crawler seems to use this magpie user to perform crawling https://github.com/bird-house/birdhouse-deploy/blob/8be8eeac211d3f5c2de620781db8832fdb8f9093/birdhouse/config/catalog/catalog.cfg.template#L15-L16

That user is granted full read permission here (is that config still up-to-date with the new magpie?):
https://github.com/bird-house/birdhouse-deploy/blob/8f8eb2a3d614e1cd9d143d9f4c43b9066e24d69e/birdhouse/config/magpie/permissions.cfg.template#L29-L32

The user password is defined here:
https://github.com/bird-house/birdhouse-deploy/blob/8be8eeac211d3f5c2de620781db8832fdb8f9093/birdhouse/env.local.example#L22-L23

So my question are

* Where/how is this user created on a fresh instance?

* Is that user being created with the expected password?

* Now that the new magpie requires minimum 12 password length, should we update that sample password value?

* I think, I am not sure, the crawler seems to be using this code https://github.com/Ouranosinc/PAVICS-DataCatalog/blob/0.6.11/pavics_datacatalog/magpie_utils.py to interface with magpie, is this code still good with the new magpie?

tlvu:
Unrelated to the crawler issue but related to magpie in general, there is this hardcoded `anonymous` user being granted read access on the api service:

https://github.com/bird-house/birdhouse-deploy/blob/8f8eb2a3d614e1cd9d143d9f4c43b9066e24d69e/birdhouse/config/magpie/permissions.cfg.template#L18-L22

* I do not see where that `anonymous` user is being used.

* What's the password for that `anonymous` user?

* What is the api service?  Sorry magpie noob question.

fmigneault:
> Unrelated to the crawler issue but related to magpie in general, there is this hardcoded `anonymous` user being granted read access on the api service:
> 
> https://github.com/bird-house/birdhouse-deploy/blob/8f8eb2a3d614e1cd9d143d9f4c43b9066e24d69e/birdhouse/config/magpie/permissions.cfg.template#L18-L22
> 
>     * I do not see where that `anonymous` user is being used.
> 
>     * What's the password for that `anonymous` user?
> 
>     * What is the api service?  Sorry magpie noob question.

`anonymous` is the special (default) value for "public"
https://github.com/Ouranosinc/Magpie/blob/master/magpie/constants.py#L96

There is no actual password related to it as it represents "no user connected".

I don't think there is any reason to have that service (not sure where it comes from or what actual service it refers to). It is not present in the providers.cfg.

fmigneault:
>So my question are
>
>    Where/how is this user created on a fresh instance?
>
>    Is that user being created with the expected password?
>
>    Now that the new magpie requires minimum 12 password length, should we update that sample password value?
>
>    I think, I am not sure, the crawler seems to be using this code https://github.com/Ouranosinc/PAVICS-DataCatalog/blob/0.6.11/pavics_datacatalog/magpie_utils.py to interface with magpie, is this code still good with the new magpie?

I believe the user would be created "inplace" during the creation of the corresponding `${CATALOG_THREDDS_SERVICE}`. The `CATALOG_PASSWORD` will not be used though. 
It would be better to have an entry like here to ensure proper creation with desired username/password: 

https://github.com/Ouranosinc/Magpie/blob/3d3d899d8b0c00f64d854323fb892631e6b146f3/config/config.yml#L23-L27

It is a good idea to update the password for something else to match expectation.

The code will need the `browse` permission instead of `read` if its intention is only to check "visible" access to the directory/file. Read is to fetch the actual data.

fmigneault:
> I remembered we have discussed that there will be a `thredds-config.yml` of some sort that will list which url is data vs metadata. Should that new config file be added in this PR to match the new magpie?

It is integrated directly within the providers.cfg file, for services typed `thredds`. 
You can add it explicitly if the desired behavior doesn't match the default: 
https://github.com/Ouranosinc/Magpie/blob/master/docs/services.rst#servicethredds

fmigneault:
@tlvu 
> The various icons are not showing up. See https://host-140-4.rdext.crim.ca/thredds/ vs working prod https://pavics.ouranos.ca/thredds/. Pretty sure it's permissions issues.

I can confirm: 

image

Maybe adding a config option with default `ignore_paths: [ "*.gif" ]` would be reasonable?
I don't think there are any use of gif for thredds other than for display purposes.

tlvu:
> `anonymous` is the special (default) value for "public"
> https://github.com/Ouranosinc/Magpie/blob/master/magpie/constants.py#L96
> 
> There is no actual password related to it as it represents "no user connected".
> 
> I don't think there is any reason to have that service (not sure where it comes from or what actual service it refers to). It is not present in the providers.cfg.

@fmigneault are you saying we can delete that block?

tlvu:
> It would be better to have an entry like here to ensure proper creation with desired username/password:
> 
> https://github.com/Ouranosinc/Magpie/blob/3d3d899d8b0c00f64d854323fb892631e6b146f3/config/config.yml#L23-L27
> 

@fmigneault Agreed, better to directly provision any mandatory "system" user.  Since this `${CATALOG_USERNAME}` user should have read access everywhere on Thredds, should that user belongs to the `admin` group?  Does belonging to the `admin` group automatically override any deny access?

Can you make this config change to auto create the `${CATALOG_USERNAME}` with the proper `${CATALOG_PASSWORD}`.

> It is a good idea to update the password for something else to match expectation.
> 

At the same time, you can update the sample password to something more suitable as well.

> The code will need the `browse` permission instead of `read` if its intention is only to check "visible" access to the directory/file. Read is to fetch the actual data.

I think the crawler will need dods access so `read` access as well, in addition to `browse`.

tlvu:
> Maybe adding a config option with default `ignore_paths: [ "*.gif" ]` would be reasonable?
> I don't think there are any use of gif for thredds other than for display purposes.

@fmigneault Given you've updated the thredds section in `providers.cfg`, I think making this configurable at the same location is not too much work?

I'd rather not assume anything so prefer to not hardcode in the code.  We have exposed `.txt` for licences on our Thredds.  Some other org might expose other `.gif` or other image formats for other reasons.

Speaking of this. does that means our `.txt` file would have been blocked like all those `.gif`?

tlvu:
> Speaking of this. does that means our `.txt` file would have been blocked like all those `.gif`?

Sample licence `.txt` file currently on our Thredds https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/fileServer/birdhouse/ecmwf/CAMS_ECMWF_License.txt

fmigneault:
@tlvu 
> are you saying we can delete that block?

Yes

> Since this ${CATALOG_USERNAME} user should have read access everywhere on Thredds, should that user belongs to the admin group? Does belonging to the admin group automatically override any deny access?

It depends on what `CATALOG_USERNAME` user does. If it is supposed to dynamically update permissions on the service for other users, then it *has* to be a member of `admins` (to be allowed corresponding POST/DELETE API requests).

Otherwise, if it roles is only to check/fetch "any" resource under the service, I would restrict it to specific service permissions without `admin` group. Adding any user to `admins` indeed grant full access and bypasses all specific permissions checks. 

> Given you've updated the thredds section in providers.cfg, I think making this configurable at the same location is not too much work? 
> Speaking of this. does that means our .txt file would have been blocked like all those .gif?

Yes & Yes
I'm planning to add such field configurable at the same place via providers definitions.
I will set default `ignore_paths: []` and override should use `ignore_paths: [ "*.gif", "*License.txt" ]` or similar.


# self-reminders

- [ ] change to auto create the ${CATALOG_USERNAME} with the proper ${CATALOG_PASSWORD}
- [ ] update the sample password to something more suitable as well. 
- [ ] `browse` + `read` permissions for `CATALOG_USERNAME`
- [ ] update thredds section in providers.cfg with overrides configs

tlvu:
@fmigneault 

> It depends on what `CATALOG_USERNAME` user does. If it is supposed to dynamically update permissions on the service for other users, then it _has_ to be a member of `admins` (to be allowed corresponding POST/DELETE API requests).
> 

No need to dynamically update permissions on the service for other users.

> Otherwise, if it roles is only to check/fetch "any" resource under the service.

Exact, only browse and read absolutely everything on Thredds for crawler.

Basically each time I deny something for other groups, I do not want to have to go back to this user to manually reverse that decision.  This is the key part.

tlvu:
@fmigneault 

Would this also have worked for the `.gif` and `.txt` file issue?  If we actually want to have access control on them?  Meaning some `.gif` and `.txt` under some folder is private?

```yaml
providers:
  LocalThredds:
    file_patterns:
      # note: make sure to employ quotes and double escapes to avoid parsing YAML error
      - ".*\\.nc"
      - ".*\\.gif"
      - ".*\\.txt"
    metadata_type:
      prefixes:
        - null  # note: special YAML value evaluated as `no-prefix`, use quotes if literal value is needed
        - "catalog\\.\\w+"  # note: special case for `THREDDS` top-level directory (root) accessed for `BROWSE`
        - "\\*\\.gif"
        - "\\*\\.txt"
        - ncml
        - (...)
    data_type:
      prefixes:
        - "\\*\\.gif"
        - "\\*\\.txt"
        - fileServer
        - (...)
```

fmigneault:
> Would this also have worked for the `.gif` and `.txt` file issue? If we actually want to have access control on them? Meaning some `.gif` and `.txt` under some folder is private?
> 
> ```yaml
> providers:
>   LocalThredds:
>     file_patterns:
>       # note: make sure to employ quotes and double escapes to avoid parsing YAML error
>       - ".*\\.nc"
>       - ".*\\.gif"
>       - ".*\\.txt"
>     metadata_type:
>       prefixes:
>         - null  # note: special YAML value evaluated as `no-prefix`, use quotes if literal value is needed
>         - "catalog\\.\\w+"  # note: special case for `THREDDS` top-level directory (root) accessed for `BROWSE`
>         - "\\*\\.gif"
>         - "\\*\\.txt"
>         - ncml
>         - (...)
>     data_type:
>       prefixes:
>         - "\\*\\.gif"
>         - "\\*\\.txt"
>         - fileServer
>         - (...)
> ```

We will have to try it out. It could work, but due to the prefix handling, I am not 100% sure. Maybe will require top-level resources named like the .gif and .txt files.

fmigneault:
@tlvu 
I did some testing and something we didn't consider occurred with above custom config.
When doing the request on Twitcher, it doesn't know about that config since it is loaded via providers.cfg on Magpie side.
I will need to add that parsing/loading in `MagpieAdapter` for Twitcher.
The providers config will also need to be mounted into the Twitcher container.

tlvu:
@fmigneault By the way, does this Twitcher/Magpie upgrade will also include https://github.com/bird-house/twitcher/pull/98 ?

fmigneault:
@tlvu 

>  By the way, does this Twitcher/Magpie upgrade will also include [bird-house/twitcher#98](https://github.com/bird-house/twitcher/pull/98) ?

We can include it while we are at it.

tlvu:
> > By the way, does this Twitcher/Magpie upgrade will also include [bird-house/twitcher#98](https://github.com/bird-house/twitcher/pull/98) ?
> 
> We can include it while we are at it.

@fmigneault Oh it's not already in this PR?  Then would it be possible to just have the Twitcher fix separately so it can be quickly merged?

fmigneault:
> Oh it's not already in this PR? Then would it be possible to just have the Twitcher fix separately so it can be quickly merged?

@tlvu merged https://github.com/Ouranosinc/Magpie/pull/377

but I'm not planing to build another tag until https://github.com/Ouranosinc/Magpie/pull/373 is complete
(which is extremely close to be done, but travis-ci is making it longer as I must run everything locally beforehand)

tlvu:
> > Oh it's not already in this PR? Then would it be possible to just have the Twitcher fix separately so it can be quickly merged?
> 
> @tlvu merged [Ouranosinc/Magpie#377](https://github.com/Ouranosinc/Magpie/pull/377)
> 
> but I'm not planing to build another tag until [Ouranosinc/Magpie#373](https://github.com/Ouranosinc/Magpie/pull/373) is complete
> (which is extremely close to be done, but travis-ci is making it longer as I must run everything locally beforehand)

Thanks @fmigneault 

It's not 100% clear to me the inter dependencies between the 2 Magpie PR above and the Twitcher performance fix.

I simply prefer to have only the Twitcher performance fix alone without all the new features in Magpie.  Is that possible so the Twitcher performance fix can go-live quickly?

fmigneault:
@tlvu 
Sure. 
Can be done, but tests with THREDDS will still fail until that PR is merged, so it will need to be bumped again very soon.

tlvu:
> Sure. **edit** pushed 3.4.0, should be building docker soon enough.
> Tests with THREDDS will still fail until next PR is merged, so it will need to be bumped again very soon.

@fmigneault Why?  We are only bumping Twitcher, not Magpie so no Magpie behavior change so why would Thredds tests fail?

tlvu:
> Why? We are only bumping Twitcher, not Magpie so no Magpie behavior change so why would Thredds tests fail?

@fmigneault We currently have also a very old twitcher `pavics/twitcher:magpie-1.7.3`.  You mentioned in the meeting to alter the config of twitcher to get the performance fix, would that work with our current old twitcher?

If that is simple, can you then just submit a separate PR so we get the Twitcher perf fix in prod without having to wait for this big Magpie upgrade?

fmigneault:
@tlvu 

What I mean is that running the old Magpie, THREDDS will remain fully open. The PAVICS-end2end test that is added to validate `secure` directory is protected will not work. 

If what you desire is only to backport Twitcher 0.5.4, to match Magpie `1.7.x`, there is no actual change in code, so no new version needed either.

What I propose is to run `gunicorn` directly. The only thing that changes in that new Twitcher is the "default" WSGI server runner. 
So this can be achieved with modified `twitcher.ini` that we override anyway such that it says 
```
[server:main]
use = egg:gunicorn#main
host=localhost
port=8001
```
instead of 
```
[server:main]
use = egg:waitress#main
bind = 0.0.0.0:8000
```

tlvu:
For the record here, tested @fmigneault proposed backport of Twitcher perf fix, no improvements at all, see comment https://github.com/bird-house/twitcher/issues/97#issuecomment-742568089

MatProv:
> @MatProv can you also enable @fmigneault new `./optional/components/secure-thredds` (in this PR) so we can test this new revoke permission feature?

@tlvu 
Adding `./optional/components/secure-thredds` is what prevents the server from correctly booting, potentially due to missing rights for some THREDDS path, leading to OWS exception when triggering the pavics crawler `[ValueError: Does not appear to be a Thredds catalog](https://host-140-46.rdext.crim.ca/wpsoutputs/catalog/7ae7cbb4-3d39-11eb-83e6-0242ac130016.xml)`.

However, running the test environment without the `secure-thredds` component makes the build pass. 

Would it be possible that `ValueError: Does not appear to be a Thredds catalog` is due to an inaccessible path?

tlvu:
> Adding `./optional/components/secure-thredds` is what prevents the server from correctly booting, potentially due to missing rights for some THREDDS path, leading to OWS exception when triggering the pavics crawler [ValueError: Does not appear to be a Thredds catalog](https://host-140-46.rdext.crim.ca/wpsoutputs/catalog/7ae7cbb4-3d39-11eb-83e6-0242ac130016.xml).
> 
> However, running the test environment without the `secure-thredds` component makes the build pass.
> 
> Would it be possible that `ValueError: Does not appear to be a Thredds catalog` is due to an inaccessible path?

@fmigneault Could you take a look since you wrote `./optional-components/secure-thredds`?

tlvu:
@fmigneault @MatProv 

Found out `./optional-components/secure-thredds` was merged without the pipeline being triggered (PR https://github.com/bird-house/birdhouse-deploy/pull/99), aka it was never tested.

We should be much more strict to not merge if pipeline do not pass in the future.

MatProv:
> Found out `./optional-components/secure-thredds` was merged without the pipeline being triggered (PR #99), aka it was never tested.
> 
> We should be much more strict to not merge if pipeline do not pass in the future.

@tlvu This is due to the fact that PR commits done by Francis weren't picked up by the pipeline automatically, therefore not updating build status and not underlining potential issues. 

New PR commits are now supposed to trigger the pipeline (known issue).

fmigneault:
@tlvu it wasn't merged in master, but in this branch because it needs the new configs of #99 to do its job.
 What is the problem with that?

tlvu:
> it wasn't merged in master, but in this branch becaus e it needs the new configs of #99 to do its job.
> What is the problem with that?

@fmigneault I never said anything about #99 breaking `master`.  

@MatProv found out #99 breaks this branch so you should take a look.  And maybe get used to use the pipeline for testing because #99 was not tested when it was merged.

fmigneault:
@tlvu The "breaking merge" is intended, and it was known that it would break because Magpie priority permission is not completed. There is no plan to merge this new version of Magpie (before added "breaking merge"), as it doesn't bring anything "new" by itself.

tlvu:
@fmigneault I can close this PR (since replaced by PR https://github.com/bird-house/birdhouse-deploy/pull/107) and delete this branch as well?

tlvu:
wait what??? merge into master???  Aren't you supposed to just close the PR?

tlvu:
@fmigneault Can you make a new PR to undo this accidental merge?

tlvu:
@fmigneault you are merging everything, well the other PR https://github.com/bird-house/birdhouse-deploy/pull/107 already include this one so there was no need to merge this one anyways.

PR #107, opened 2020-11-25
Magpie thredds

fmigneault:
- add custom config to allow display of UI catalog elements of thredds service protected behind magpie/twitcher
- bump versions to 3.x to get custom config stored directly in db

PR #114, opened 2021-01-12
Feature/dac 300 weaver compose

### LINKED ISSUE #10: Free disk space problem due to docker images logging
Migrated from old PAVICS https://github.com/Ouranosinc/PAVICS/issues/152, discussion summarized below:

Containers that generate a lot of logging, for example Twitcher/Magpie on every request, cause disk to fill up rapidly. Following config (or similar) should be added to containers in compose.

```
    logging:
      driver: "json-file"
      options:
        max-size: "100m"
```

stdout/stderr logging in docker is by default placed within each container folder:
`du -h /var/lib/docker/containers//.log | sort -h`

We do have a few containers logging like crazy here, not clear if those logs are valuable. Maybe a variable max-size depending on the container expectation/logging needs, ping @moulab88 

```
# This is on Boreas
# du -sh */*.log
96.0K   01f8c3a7c20d234b023d73d798e56c20fa4a54cf9fffa12ff9f71b9cf9c74e18/01f8c3a7c20d234b023d73d798e56c20fa4a54cf9fffa12ff9f71b9cf9c74e18-json.log
16.0K   06b14f1163225843605288e5a8f203612f4dd3ae3761ff27713d6122333a030c/06b14f1163225843605288e5a8f203612f4dd3ae3761ff27713d6122333a030c-json.log
8.0K    0e1cc179a4114d7d10fde79c77cde728e6e37a6006e70b1f21120cbb673cf2c5/0e1cc179a4114d7d10fde79c77cde728e6e37a6006e70b1f21120cbb673cf2c5-json.log
* 155.0M  194af2ac8810d44c463c8da5b5942e48e3bf5485ccbeb47b21c03b8468780057/194af2ac8810d44c463c8da5b5942e48e3bf5485ccbeb47b21c03b8468780057-json.log
2.0M    1d2d23dc5d0dfb1049efdc9efa6e45386e2eea7e12227717e8ddca0935634f22/1d2d23dc5d0dfb1049efdc9efa6e45386e2eea7e12227717e8ddca0935634f22-json.log
32.0K   31814345a0d379063062b2015695ad5f6fb623c5f2c48619a85f8f0823398642/31814345a0d379063062b2015695ad5f6fb623c5f2c48619a85f8f0823398642-json.log
344.0K  353b6f48121b9d6ce6b42a1ed588455e91002225a9ad0e5626628db0f5e17865/353b6f48121b9d6ce6b42a1ed588455e91002225a9ad0e5626628db0f5e17865-json.log
436.0K  3d1a8e0c535ff110399b872d5cd759546c6f2c9356ee0f068841931ce785605f/3d1a8e0c535ff110399b872d5cd759546c6f2c9356ee0f068841931ce785605f-json.log
*** 19.4G   3f2f4c0457ccfd600fb48d245e323f58894ad791427af8387d95400766161747/3f2f4c0457ccfd600fb48d245e323f58894ad791427af8387d95400766161747-json.log
56.0K   3f92224708a09870c0dfa079ebae00461128880ed28ed23b3de9964d4c0a1b6f/3f92224708a09870c0dfa079ebae00461128880ed28ed23b3de9964d4c0a1b6f-json.log
40.0K   49312c398b724442922a8233458eb43f4832e66c1703193d20fe81ce6bd3551c/49312c398b724442922a8233458eb43f4832e66c1703193d20fe81ce6bd3551c-json.log
80.0K   54461d996891c145b912f5ee308ce625cc5367f59b319b42ea4fee603e58762d/54461d996891c145b912f5ee308ce625cc5367f59b319b42ea4fee603e58762d-json.log
4.0K    655a6a7b5978675c5d6c5382b9ee9e42d3e78432916545218f33f51ea7f91820/655a6a7b5978675c5d6c5382b9ee9e42d3e78432916545218f33f51ea7f91820-json.log
44.0K   65e0ac0df5c4da26eaf856b1677ef176ad31d1797697260c633bd8f0e79c1868/65e0ac0df5c4da26eaf856b1677ef176ad31d1797697260c633bd8f0e79c1868-json.log
18.7M   8084e627be8cc40cd30a5e03489c4b29e5190ec101429723038e45835d687f59/8084e627be8cc40cd30a5e03489c4b29e5190ec101429723038e45835d687f59-json.log
6.6M    97d2e7553bc6ee5fb3db183f11702c58d823544f6970d26e9e532f8cc1ade23d/97d2e7553bc6ee5fb3db183f11702c58d823544f6970d26e9e532f8cc1ade23d-json.log
76.0K   99fda25d27c8647732e3e3157d91b506f315ae2f591d8a53327271c7dce9bf2b/99fda25d27c8647732e3e3157d91b506f315ae2f591d8a53327271c7dce9bf2b-json.log
72.0K   a1e4f3b79fdae839f2e8f2367f11945872a14e8fc64393741d00b93335508a2f/a1e4f3b79fdae839f2e8f2367f11945872a14e8fc64393741d00b93335508a2f-json.log
216.0K  a5d35f52bd480a21c1768a90e616456eda307ec7ea05ce37ec63c34dff174850/a5d35f52bd480a21c1768a90e616456eda307ec7ea05ce37ec63c34dff174850-json.log
116.0K  c389703d90ba267d471f7f86f9f03c0ed4470a85fcaec64efc772478e888224f/c389703d90ba267d471f7f86f9f03c0ed4470a85fcaec64efc772478e888224f-json.log
** 551.4M  d237df0d3bb21c72dacd4098f8c2ac04737aac9357740638a06e11242098c077/d237df0d3bb21c72dacd4098f8c2ac04737aac9357740638a06e11242098c077-json.log
180.0K  dfa70f3379cda36dc1c76604b1f5f048e1232d66bb4c8e11ba568d8ea39b0e59/dfa70f3379cda36dc1c76604b1f5f048e1232d66bb4c8e11ba568d8ea39b0e59-json.log
48.0K   e3e758b108a81b06f7d5ed00124cc93066e470858aabff56840558ce22583328/e3e758b108a81b06f7d5ed00124cc93066e470858aabff56840558ce22583328-json.log
24.0K   e630ee7c268dad10d4e9c9d38cffa859595d1b4dee3620439cc2263084bfc815/e630ee7c268dad10d4e9c9d38cffa859595d1b4dee3620439cc2263084bfc815-json.log
1.1M    f36d34e11cd7732f277c2c1031705c745389cdc1a1bc9c4a5c78bb50e1b6c050/f36d34e11cd7732f277c2c1031705c745389cdc1a1bc9c4a5c78bb50e1b6c050-json.log
160.0K  f8dfbe7e3b15c710f6de5d650107be5c8e3eccb9a5772f4754ca165001581e04/f8dfbe7e3b15c710f6de5d650107be5c8e3eccb9a5772f4754ca165001581e04-json.log
```

fmigneault:
Live instance (at CRIM with daccs-iac): 
https://host-140-36.rdext.crim.ca/twitcher/ows/proxy/weaver
(OpenAPI doc endpoint for info: https://host-140-36.rdext.crim.ca/weaver/api)

Example of `flyingpigeon`'s WPS-1 `subset_bbox` process converted to WPS-REST via Weaver: 
https://host-140-36.rdext.crim.ca/weaver/providers/flyingpigeon/processes/subset_bbox

# Dependencies

- #152 (docker-compose v3.4 & logging)
- #153 (optional components - db ports & wps-healthchecks)
- #154 (env var `DATA_PERSIST_ROOT` and server-related docs) 

note: until they are merged, changes of this PR will continue to contain all of them
preview of simplified modifications: [changes](https://github.com/bird-house/birdhouse-deploy/compare/tmp-everything-except-weaver...feature/DAC-300-weaver-compose)
all above PRs are combined in branch [tmp-everything-except-weaver](https://github.com/bird-house/birdhouse-deploy/tree/tmp-everything-except-weaver) to generate the this diff changes

# Changes

- Adds [Weaver](https://github.com/crim-ca/weaver) to the stack using optional-component. 
- Adds worker image & docker-proxy to allow Weaver booting a docker Application Package as sibling container.
- Adds Magpie permissions for Weaver.
- Adds all WPS birds of the stack as known providers of Weaver.
- Bump docker-compose to 3.4 to resolve network issue. 
- Add `x-logging` config to reuse definition across compose. (fixes #10)
- Add generic metadata variables that can be shared and displayed in responses of multiple services.
- Add `DATA_PERSIST_ROOT` to allow changing default `/data` where persistence dirs are generated.

# Extra 

- Add optional-component `wps-healthchecks` (developed while testing Weaver integration, but not directly needed by it any more).
- Add optional-component `database-external-ports`, to allow exposing port for dev instances, but that should definitely not be open by default.

Resolves https://www.crim.ca/jira/browse/DAC-25
Resolves https://www.crim.ca/jira/browse/DAC-300
Resolves https://github.com/crim-ca/weaver/issues/13

daccs-iac config : https://www.crim.ca/stash/projects/DACCS/repos/daccs-configs/pull-requests/1


birdhouse/optional-components/testthredds/docker-compose-extra.yml
```diff
@@ -45,3 +45,6 @@ services:
 volumes:
   testthredds_persistence: {}
   testwps_outputs: {}
+  testdatasets:
+    driver_opts:
+      device: /data/testdatasets
```
dbyrns, line 50: Why not using `${DATA_PERSIST_ROOT}` rather than `/data`?
And what's the point of moving this volume as a named volume? I don't see it used anywhere else.

birdhouse/optional-components/weaver/conf.extra-service.d/weaver.conf.template
```diff
@@ -0,0 +1,22 @@
...
+
+    location /${WEAVER_MANAGER_NAME} {
+        proxy_pass https://${PAVICS_FQDN}${TWITCHER_PROTECTED_PATH}/${WEAVER_MANAGER_NAME};
+        proxy_set_header Host $host;
+        proxy_buffering off;
+        include /etc/nginx/conf.d/cors.include;
+    }
+
+    # location where process job outputs will be accessible
+    location ^~ ${WEAVER_WPS_OUTPUTS_PATH}/ {
```
dbyrns, line 20: Could we mount weaver under WPS_OUTPUTS_PATH like other static birds so that all outputs are together?

birdhouse/optional-components/weaver/config/weaver/data_sources.json.template
```diff
@@ -0,0 +1,22 @@
...
+    "localhost": {
+        "netloc":"localhost",
+        "ades":"http://localhost:4001",
+        "default":"true"
+    },
+    "public": {
+        "netloc":"${PAVICS_FQDN}",
+        "ades":"https://${PAVICS_FQDN}${TWITCHER_PROTECTED_PATH}/${WEAVER_MANAGER_NAME}"
+    },
+    "crim": {
```
dbyrns, line 11: I don't think we should add this location.

birdhouse/optional-components/weaver/config/weaver/weaver.ini.template
```diff
@@ -0,0 +1,156 @@
...
+pyramid.default_locale_name = en
+
+# mongodb
+mongodb.host = mongodb
+mongodb.port = 27017
+mongodb.db_name = weaver
+
+# --- Weaver Configuration ---
+weaver.configuration = ${WEAVER_CONFIG}
+weaver.url = https://${PAVICS_FQDN}/weaver
```
dbyrns, line 27: /weaver is hardcoded but nginx is configured with `${WEAVER_CONFIG}` and `${WEAVER_MANAGER_NAME}`

birdhouse/optional-components/weaver/config/weaver/wps_processes.yml.template
```diff
@@ -0,0 +1,27 @@
+# Details:
+#   https://github.com/crim-ca/weaver/blob/master/config/wps_processes.yml.example
+#
+# For each of the below WPS service providers, Weaver will iterate over listed WPS processes returned by GetCapabilities
+# and will register them as individual WPS-REST processes. Each process will be referenced
+
+# using 'providers' instead of 'processes' will register services directly without fetching each process (local copy)
+# this will make resulting deployed processes dynamic according to what those providers actually offer at any given time
+providers:
```
dbyrns, line 9: Does weaver have a "service user" with some privileges that would allow him to access the following routes if they are not public or do we need to validate that every provider is public?

birdhouse/optional-components/weaver/default.env
```diff
@@ -0,0 +1,46 @@
...
+# real names of the weaver/worker services
+# these values will generate "<server-fqdn>/<name>" URI to access them behind secured proxy
+export WEAVER_MANAGER_NAME=weaver
+export WEAVER_WORKER_NAME=worker
+export WEAVER_WPS_NAME=${WEAVER_MANAGER_NAME}-wps
+# WPS-1/2 path under Weaver's root endpoint
+export WEAVER_WPS_PATH=/wps
+# data persistence of wps outputs (directory path)
+# for the moment, this is hardcoded in 'birdhouse/docker-compose.yml'
+export WEAVER_WPS_OUTPUTS_DIR="${DATA_PERSIST_ROOT}/wps_outputs"
```
dbyrns, line 34: Following my previous comment on `WEAVER_WPS_OUTPUTS_DIR` : So you choose to dump result under the same directory but allow it to be different. Other birds dump their results into wps_outputs/<bird_name>. To respect this convention, should we set weaver output dir to `${DATA_PERSIST_ROOT}/wps_outputs/weaver` ?

dbyrns (commented):
I got some questions, but overall this is a nice PR.

fmigneault:
Jenkins-bot failing due to https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/issues/58

birdhouse/optional-components/testthredds/docker-compose-extra.yml
fmigneault: I experimented with named volumes and found out that docker-compose doesn't handle them the same way and mappings in services. 
For example, it will not create the missing directory, only attempt to mount an existing one and fail if missing. 

In this case, I will move it within the `volumes` of `testthredds` with the variable since a path location is specified. 
Named volumes are useful now only when `{}` is used to let docker manage it automatically (but volatile if removed).

birdhouse/optional-components/weaver/default.env
fmigneault: Yes. We can change it to follow this same convention.

birdhouse/optional-components/testthredds/docker-compose-extra.yml
```diff
@@ -22,9 +22,9 @@ services:
       WANTED_CONTEXT_ROOT_WARFILE_NAME: $TESTTHREDDS_WARFILE_NAME
     volumes:
       - testthredds_persistence:/usr/local/tomcat/content/thredds
-      - /data/testdatasets:/pavics-testdata:ro
-      - /data/datasets:/pavics-data:ro
-      - /data/ncml:/pavics-ncml:ro
+      - testdatasets:/pavics-testdata:ro
```
MatProv, line 25: Why not under `${DATA_PERSIST_ROOT}/testdatasets`?

birdhouse/optional-components/weaver/conf.extra-service.d/weaver.conf.template
```diff
@@ -0,0 +1,22 @@
+
+    # NOTE:
+    #   Redirect to internal network of twitcher with Weaver root endpoint and alias allows to set
+    #   the same 'magpie' permissions on the 'weaver' service defined by "WEAVER_MANAGER_NAME".
+    #   This allows verification of the same service user/group permissions references regardless
+    #   whether the *shortcut* Weaver endpoint, the alias or the explicit 'twitcher' proxy route is used.
+    location /${WEAVER_CONFIG} {
+        # redirect to secured EMS/ADES
+        return 302 https://${PAVICS_FQDN}${TWITCHER_PROTECTED_PATH}/${WEAVER_MANAGER_NAME};
```
MatProv, line 9: We do not need to specify port here and assume that it is on :80?

MatProv (approved):

birdhouse/optional-components/weaver/config/weaver/weaver.ini.template
fmigneault: Good point. I also removed the `/${WEAVER_CONFIG}` location as that will be weird once some kind of hybrid EMS/ADES config gets introduced.

birdhouse/optional-components/weaver/config/weaver/wps_processes.yml.template
fmigneault: For the moment, all providers are anonymous-read recursively since the processes are accessible anyway via their respective WPS-1 endpoints. Technically, the POST of jobs could also be anonymous-accessible since it is possible to run the process execution via the WPS-1 GET requests of each provider real location.

Local processes (eg: deployed docker) do actually require read/write permissions to be added. For example, https://132.217.140.4/twitcher/ows/proxy/weaver/processes/jsonarray2netcdf fails if you are not logged in as Magpie admin.

dbyrns: Ok, but my point is only for static providers. To register their processes weaver will need to acces them while booting. What if they are not public? Maybe add a note in the file saying that weaver fetches processes anonymously so they must be public.

birdhouse/optional-components/weaver/conf.extra-service.d/weaver.conf.template
fmigneault: It is https route, do default should be 443

birdhouse/optional-components/weaver/default.env
```diff
@@ -0,0 +1,47 @@
...
+# real names of the weaver/worker services
+# these values will generate "<server-fqdn>/<name>" URI to access them behind secured proxy
+export WEAVER_MANAGER_NAME=weaver
+export WEAVER_WORKER_NAME=worker
+export WEAVER_WPS_NAME=${WEAVER_MANAGER_NAME}-wps
+# WPS-1/2 path under Weaver's root endpoint
+export WEAVER_WPS_PATH=/wps
+# data persistence of wps outputs (directory path)
+# for the moment, this is hardcoded in 'birdhouse/docker-compose.yml'
+export WEAVER_WPS_OUTPUTS_DIR="${DATA_PERSIST_ROOT}/wps_outputs/weaver"
```
dbyrns, line 34: I'm not sure this would do the trick.
First, just to be clear `WEAVER_WPS_OUTPUTS_DIR` refers to a path used internally in weaver, while `DATA_PERSIST_ROOT` refers to a path of the host. Nothing prevent weaver from using the same root for data but I think it can cause useless confusion.

Ok, main course now ;) 
wps_outputs compose volume is mapped on the host like this : `wps_outputs:/pavics-data/wps_outputs`
Weaver map the wps_outputs volume like this : `wps_outputs:${WEAVER_WPS_OUTPUTS_DIR}`
So the weaver directory `${DATA_PERSIST_ROOT}/wps_outputs/weaver` will be mapped to the host directory `/pavics-data/wps_outputs` and weaver will write outputs directly under wps_outputs on the host.

We want weaver to write in the host directory `/pavics-data/wps_outputs/weaver`.
So we should have something like that : 

1. `export WEAVER_MAPPED_WPS_OUTPUTS_DIR="/data/wpsoutputs"`
2. Weaver mapping  `wps_outputs:${WEAVER_MAPPED_WPS_OUTPUTS_DIR}`
3. export WEAVER_WPS_OUTPUTS_DIR="${WEAVER_MAPPED_WPS_OUTPUTS_DIR}/weaver"

dbyrns (changes requested):
I think the volume mapping needs to be reviewed.

birdhouse/config/postgres/credentials.env.template
```diff
@@ -1,3 +1,3 @@
 POSTGRES_USER=${POSTGRES_PAVICS_USERNAME}
 POSTGRES_PASSWORD=${POSTGRES_PAVICS_PASSWORD}
-POSTGRES_DB=pavics
+POSTGRES_DB=postgres-pavics
```
tlvu: Should merge with `master`.  I already fixed this problem here 56274f3e5fbdd9b4eb71bfd3acd3039dc965de7d.  This commit is not required anymore.

birdhouse/config/magpie/permissions.cfg.template
```diff
@@ -18,7 +18,7 @@ permissions:
   - service: api
     resource: /api
     permission: read
-    user: anonymous
+    group: anonymous
```
tlvu, line 21: What is this api service?  I have a vague memory in the other magpie upgrade PR we are going to remove it?

birdhouse/docker-compose.yml
```diff
@@ -212,10 +234,9 @@ services:
       - "8983:8983"
       - "48983:9001"
     volumes:
-      - /data/solr:/data
-    volumes_from:
-      - mongodb
```
tlvu, line 217: I personally never understood why Solr and ncwms2 needs volume_from mongodb.  But was afraid to change since I do not know Solr and ncwms2 well.  So you confident this do not break anything?

birdhouse/docker-compose.yml
```diff
@@ -266,14 +287,18 @@ services:
   mongodb:
     image: mongo:3.4.0
     container_name: mongodb
-    # Mongodb crash with permission denied errors if the command is not overriden like this
+    volumes:
+      - ${DATA_PERSIST_ROOT}/mongodb_persist:/data/db
```
tlvu, line 291: Good catch !!!  Just curious, without persist, I have not yet found something broken or lost.  Did you actually got some problems without persist?  The bigger question is, what is mongodb used for? !

birdhouse/optional-components/README.md
```diff
@@ -129,3 +148,44 @@ How to enable in `env.local` (a copy from
 * Add `./optional-components/all-public-access` to `EXTRA_CONF_DIRS`.
 
 The anonymous user will now have all the permissions described in [`./optional-components/all-public-access/all-public-access-magpie-permission.cfg`](all-public-access/all-public-access-magpie-permission.cfg).
+
+
+## Weaver
```
tlvu, line 153: I think Weaver will become pretty much an essential component for PAVICS stack?  In this case, better put in in `birdhouse/components`.

There are `components/` and `optional-components/` because as previously discussed when I first introduced this pluggable components mechanism, I was planning to break up the root `docker-compose.yml` into many components as well, leaving only the proxy in the root `docker-compose.yml`.  So all those previously in the root `docker-compose.yml` will go into `components/`.

`optional-components/` are for nice to have, test helper, optional components, like `testthredds`, `emu`, `all-public-access` and so on.

tlvu (changes requested):
Just starting review.  Not done yet.  1 change request: move Weaver to birdhouse/components/ instead.

birdhouse/optional-components/weaver/default.env
fmigneault: I find there is a design problem in the stack itself if all data directories are accessible from any bird. 
Really, there should only be a mounting of `"/pavics-data/wps_outputs/weaver:/data/wps_outputs"` and weaver shouldn't worry about creating a sub-directory named `weaver` for itself. It should only dump in the specified location. 
I don't know of any way to combine named-volume with extra path. This can be done only with string mapping and variables.

fmigneault:
> 
> 
> Just starting review. Not done yet. 1 change request: move Weaver to birdhouse/components/ instead.

I thought the whole point was to make it optional? 
@dbyrns can you confirm?

I refactored everything as optional for this...

birdhouse/config/magpie/permissions.cfg.template
fmigneault: `ServiceAPI` is the service that converts `/some/sub/route` into resources: 
```
some:
  sub: 
    route:
```
with permissions applicable on each path element.

It is definitely not being removed. It is the 2nd most used after `ServiceWPS`. Both will probably be combined to form a "super" `ServiceADES` (https://github.com/Ouranosinc/Magpie/issues/360) that combines all WPS-1/REST process interfaces.

birdhouse/docker-compose.yml
fmigneault: For the moment, I think most usage were for temporary data that did not matter much (if even used at all?). 
With Weaver though, deployed providers, processes and jobs execution are stored in mongo, which should be persisted.

birdhouse/docker-compose.yml
dbyrns: I'm not really worried about ncwms2 as : 
1) we are now using thredds only
2) I don't see any reason at all why ncwms would use mongo

For Solr though, it can use mongo as backend, but could this be done using only a volumes_from? I don't know, but as Long said we never got confident enough to remove it. 

Booting is a thing, does the indexation occurs and the search works? If we now have tests to prove that, then I think we could get rid of that for good.

birdhouse/docker-compose.yml
dbyrns: Mongo were used in solr, ncwms and phoenix. Phoenix stores jobs and results in it and maybe solr used to index that by the volume_from?

birdhouse/optional-components/README.md
dbyrns: We tried to be nice here and don't push weaver too far until it is fully integrated in the stack (with birdy for example). If you don't mind to have that component meanwhile, we can put it in components.

dbyrns: But just to be clear the mecanism for components and optional-components is the same in that they are both pluggable?

tlvu: > But just to be clear the mecanism for components and optional-components is the same in that they are both pluggable?

Exact, both are pluggable and not enabled by default.  Stuff in `components/` just "look" more essential and have better README documentation (auto gen table of content).

birdhouse/optional-components/weaver/config/weaver/wps_processes.yml.template
fmigneault: Yes they will fail if they are not publicly accessible.

birdhouse/optional-components/weaver/default.env
fmigneault: I think that volumes should be ok now. Please see notes I added in 7aa56f0. This is an extra difficulty that must be considered in the case of Weaver on top of the `/data/wpsoutputs/<bird>` format.

birdhouse/docker-compose.yml
fmigneault: I will add the `/data/solr` one back in case, but the `volumes_from` didn't do anything since `mongo` did not have any mapped volumes to start with, unless overridden in some private config somewhere.

birdhouse/docker-compose.yml
fmigneault: As other comment. Original `mongodb` service did not even have any mapped volumes. So wherever they were writing, it was not persisted unless overridden in private config. I don't feel like any of those service should write directly into the `mongodb` managed directory either.

birdhouse/docker-compose.yml
dbyrns: If the following test https://github.com/Ouranosinc/pavics-sdi/blob/e50335e76d0018cf813bec1c963bff576b881345/docs/source/notebooks/catalog_search.ipynb (thank you @MatProv to provided me the link) succeed, we will be ok (not ok right now). The fastest way to know if volumes_from is the culprit, would be to revert the change and test again. Once we found what is absolutely required, we should add a note so that we know in the future what cannot be removed.

Please also note that we are not using the stock mongodb but the one build by birdhouse buildout mecanism. Their is a recipe for solr and I looked at it trying to figured out what could link mogo and solr without any success.

birdhouse/optional-components/README.md
fmigneault: Moved.

birdhouse/components/README.rst
```diff
@@ -56,12 +56,12 @@ Given the unattended nature, there is no UI.  Logs are used to keep trace.
 How to Enable the Component
 ---------------------------
 
-- Edit ``env.local`` (a copy of env.local.example_)
+- Edit ``env.local`` (a copy of `env.local.example`_)
```
tlvu, line 59: Not sure what's the difference surrounding `env.local.example` with backticks changes the rendering display.  Here's the original version from `master` https://github.com/bird-house/birdhouse-deploy/tree/e10a2ef9bd77498f1af22267cad5d181d0da65ef/birdhouse/components I see no difference.

birdhouse/config/magpie/permissions.cfg.template
tlvu: Oh ok thanks.  I mixed up I guess.

birdhouse/docker-compose.yml
tlvu: I was not there when `mongodb` was added so not sure why there is no persistence but I totally agree we should add the persistence now that Weaver is using it.

birdhouse/components/weaver/config/weaver/weaver.ini.template
```diff
@@ -0,0 +1,157 @@
...
+weaver.wps_output_dir = ${WEAVER_WPS_OUTPUTS_DIR}
+weaver.wps_output_url = https://${PAVICS_FQDN_PUBLIC}${WEAVER_WPS_OUTPUTS_PATH}
+weaver.wps_output_path =
+weaver.wps_workdir = ${WEAVER_WPS_WORKDIR}
+
+# --- Weaver WPS metadata ---
+weaver.wps_metadata_identification_title=Weaver
+weaver.wps_metadata_identification_abstract=${SERVER_SYNOPSIS}
+weaver.wps_metadata_identification_keywords=Weaver,WPS,OGC,${WEAVER_CONFIG}
+weaver.wps_metadata_identification_accessconstraints=https://${PAVICS_FQDN}/magpie
```
tlvu, line 54: `PAVICS_FQDN_PUBLIC` ?

birdhouse/components/weaver/config/weaver/weaver.ini.template
```diff
@@ -0,0 +1,157 @@
...
+#USE_CELERYCONFIG = True
+BROKER_URL = mongodb://mongodb:27017/celery
+
+###
+# wsgi server configuration
+###
+
+[server:main]
+use = egg:gunicorn#main
+bind = 0.0.0.0:4001
```
tlvu, line 109: I've seen this port 4001 a few times.  Should this port be move to `weaver/default.env` so the same value is enforced everywhere?

birdhouse/components/weaver/config/weaver/data_sources.json.template
```diff
@@ -0,0 +1,18 @@
+{
```
tlvu, line 1: So weaver can handle .yml config file.  This seems to be the only .json config file.  Any plan later to make this .json a .yml as well so we can add comment to the config file for inplace clarification?

birdhouse/components/weaver/config/weaver/wps_processes.yml.template
```diff
@@ -0,0 +1,37 @@
...
+#   referenced WPS DescribeProcess/Execute response to provide a live response.
+
+# Note:  
+#   All service providers and underlying processes must be publicly accessible for GetCapabilities and DescribeProcess
+#   at boot time in order to allow Weaver to retrieve the information. The application will skip unresolvable references
+#   including unauthorized and forbidden access.
+
+# processes:    # static local snapshot
+providers:      # dynamic remote reference
+  - name: catalog
```
tlvu, line 20: Oh, so all the providers are already listed here.  So why need them to have anon read access in magpie again?  Sorry this might be a dumb question since I am new to weaver.

tlvu: If there is a new bird, can it hook itself in to weaver like it can hook itself in to magpie right now?

birdhouse/components/weaver/default.env
```diff
@@ -0,0 +1,59 @@
...
+  $WEAVER_WPS_OUTPUTS_DIR
+  $WEAVER_WPS_OUTPUTS_PATH
+  $WEAVER_WPS_PATH
+  $WEAVER_WPS_WORKDIR
+  $WEAVER_MANAGER_LOG_LEVEL
+  $WEAVER_WORKER_LOG_LEVEL
+'
+# extend the original 'VARS' from 'birdhouse/pavics-compose.sh' to employ them for template substitution
+# adding them to 'VARS', they will also be validated in case of override of 'default.env' using 'env.local'
+VARS="$VARS $EXTRA_VARS"
```
tlvu, line 21: You sure you want to add to `VARS` and not `OPTIONAL_VARS`?  I have not checked but if all these have default values, `OPTIONAL_VARS` is much more appropriate.  See https://github.com/bird-house/birdhouse-deploy/blob/e10a2ef9bd77498f1af22267cad5d181d0da65ef/birdhouse/components/monitoring/default.env for example

birdhouse/components/weaver/docker-compose-extra.yml
```diff
@@ -0,0 +1,124 @@
...
+      - raven
+      - worker
+    links:
+      - mongodb
+    volumes:
+      - ./components/weaver/config/weaver/weaver.ini:/opt/local/src/weaver/config/weaver.ini
+      - ./components/weaver/config/weaver/data_sources.json:/opt/local/src/weaver/config/data_sources.json
+      - ./components/weaver/config/weaver/request_options.yml:/opt/local/src/weaver/config/request_options.yml
+      - ./components/weaver/config/weaver/wps_processes.yml:/opt/local/src/weaver/config/wps_processes.yml
+      - /usr/bin/docker:/usr/bin/docker   # FIXME: see above docker-sibling note
```
tlvu, line 55: Not sure I like this.  If you need the `docker` cli, can you directly install docker in the weaver image and only volume mount the `docker.sock` to access the daemon from the host?

birdhouse/components/weaver/docker-compose-extra.yml
```diff
@@ -0,0 +1,124 @@
...
+      - ~/.docker/config.json:/root/.docker/config.json   # FIXME: see above docker-sibling note
```
tlvu, line 56: Whooo, what do you need special from the user's docker config?  Is it documented somewhere?  Not sure I like this either.

birdhouse/components/weaver/docker-compose-extra.yml
```diff
@@ -0,0 +1,124 @@
...
+  driver: "json-file"
+  options:
+    max-size: "200k"
+    max-file: "10"
+
+services:
+
+  proxy:
+    volumes:
+      - ./components/weaver/conf.extra-service.d:/etc/nginx/conf.extra-service.d/weaver
```
tlvu, line 14: specify `ro` for all volumes that do not need write

birdhouse/default.env
```diff
@@ -1,6 +1,23 @@
 # All env in this default.env can be overridden by env.local.
 # All env in this default.env must not depend on any env in env.local.
 
+# General metadata details about the instance
+# Details a shared across various containers wherever applicable
```
tlvu, line 5: Details "are" shared I assume?  Typo here.

birdhouse/default.env
```diff
@@ -1,6 +1,23 @@
...
+export SERVER_INSTITUTION="Ouranos"
+export CONTACT_NAME="support"
+export CONTACT_POSITION="support"
+export CONTACT_PHONE=""
+export CONTACT_FAX=""
+export CONTACT_DELIVERY_POINT=""
+export CONTACT_CITY="Montréal"
+export CONTACT_STATE_OR_PROVINCE="Québec"
+export CONTACT_COUNTRY="Canada"
+export CONTACT_POSTAL_CODE="H3A 1B9"
```
tlvu, line 16: LOL is this Ouranos's postal code?  Maybe put something bogus here, like the bogus support email.

birdhouse/optional-components/wps-healthchecks/docker-compose-extra.yml
```diff
@@ -0,0 +1,80 @@
...
+services:
+  catalog:
+    healthcheck:
+      test:
+        [
+          "CMD",
+          "curl",
+          "--insecure",
+          "--fail",
+          "https://localhost${TWITCHER_PROTECTED_PATH}/catalog?service=WPS&request=GetCapabilities",
```
tlvu, line 11: Humm, why go through twitcher?  This will force anon access.  I assume a health check for each of the bird standalone.  If you want to check if anon access is enabled, I am not sure this is the right thing to do.

tlvu: Humm, does `localhost` even works?  `localhost` means the container running the bird, that container do not know about twitcher.

birdhouse/pavics-compose.sh
```diff
@@ -25,11 +24,17 @@ VARS='
   $POSTGRES_PAVICS_PASSWORD
   $POSTGRES_MAGPIE_USERNAME
   $POSTGRES_MAGPIE_PASSWORD
+  $SERVER_SYNOPSIS
```
tlvu, line 27: The default values of all these in `default.env` should then be empty if they are to be in `VARS` list.

Recall settings in `VARS` list are the settings we want to force users to set a value because no default value can be appropriate.  So if the default value in `default.env` is not empty, then it defeats the check that user has set a personalized value.

tlvu (changes requested):
Finished first pass of my review.  Still need to deploy it and play with it.  Will probably have more feedbacks once I actually try to use it.

Big component, great work plugging into the stack.  I think you got the hang of how the pluggable architecture work.  As one of my comment, does weaver allow other birds to plug themselves into weaver config?

birdhouse/components/README.rst
fmigneault: There's no difference technically, but parsers detect it more easily for rendering in IDEs.

birdhouse/components/weaver/config/weaver/weaver.ini.template
fmigneault: We could do that, but all birds seem to have hardcoded ports all over the place. 
The only case were the value matters (possible conflict) is for exposed ports, which should be fixed by not exposing them at all.

birdhouse/components/weaver/config/weaver/data_sources.json.template
fmigneault: Yes. Easy to fix. Just need `yaml.safe_load` instead of `json.load` as it handles both YAML/JSON.

birdhouse/components/weaver/config/weaver/wps_processes.yml.template
fmigneault: Before registering the providers, weaver tries to do a GetCapabilities to validate the endpoint.

birdhouse/components/weaver/default.env
fmigneault: All variables are mandatory in the sense that if they are overridden by an empty string, compose won't work at all. 
So although they are *optional* in terms of "no need any override", they must be validated. 
My understanding of `OPTIONAL_VARS` was that empty value will still pass. 
What I wrong? If so, I don't understand why there are 2 variables.

birdhouse/default.env
fmigneault: Yes it is. Will replace.

birdhouse/optional-components/wps-healthchecks/docker-compose-extra.yml
fmigneault: I used the same definition that was defined on `thredds`. 
I haven't looked into it that much. You're probably right.

birdhouse/docker-compose.yml
dbyrns: Like I said, probably by Phoenix. For Solr I don't see how, but there is a chance too.

birdhouse/components/weaver/config/weaver/data_sources.json.template
fmigneault: https://github.com/crim-ca/weaver/pull/206

birdhouse/components/weaver/docker-compose-extra.yml
fmigneault: https://github.com/crim-ca/weaver/pull/208

birdhouse/components/weaver/docker-compose-extra.yml
fmigneault: It is needed for access some dockers. If not provided, remote private repositories cannot be accessed. 
We can make it a separate extra component. Not a 'hard must' and not needed if every reference dockers is public. 

Feature to provide auth via API is not implemented in Weaver (at least for now), and I'm not sure if CWL even supports it. 
That feature was pushed aside as not a priority.

dbyrns: Hum, I agree with @tlvu, Weaver only needs access to the Application Packages repositories. Those are not the same as the PAVICS components repositories. I would change that line with `./components/weaver/config/docker/config.json:/root/.docker/config.json`

birdhouse/optional-components/wps-healthchecks/docker-compose-extra.yml
fmigneault: fixed in fed17402df0fb43492715e018870a5abac92dc4b
tested each bird individually with `curl` / `python requests` according to what was available in containers

birdhouse/components/weaver/config/weaver/wps_processes.yml.template
```diff
@@ -0,0 +1,37 @@
+# Details:
+#   https://github.com/crim-ca/weaver/blob/master/config/wps_processes.yml.example
+#
+# Using 'processes' section:
+#   For each of the below WPS service providers, Weaver would iterate over listed WPS processes returned by their
+#   GetCapabilities response and would register them as individual WPS-REST processes. Each process will be defined
+#   as a local snapshot of the process at that point in time. They would be static an not consider process changes.
```
dbyrns, line 7: Little typo around `"an not consider"`

birdhouse/components/weaver/default.env
```diff
@@ -11,37 +13,47 @@ EXTRA_VARS='
...
+#   See important notes about volume directory paths in 'docker-compose-extra.yml'
 # WPS-1/2 path under Weaver's root endpoint
 export WEAVER_WPS_PATH=/wps
-# data persistence of wps outputs (directory path)
-# for the moment, this is hardcoded in 'birdhouse/docker-compose.yml'
-export WEAVER_WPS_OUTPUTS_DIR="${DATA_PERSIST_ROOT}/wps_outputs/weaver"
 # route URI where the resulting process data will be remotely accessible
-# NOTE: '/wpsoutputs/<bird>' is the directory location format employd everywhere
+# NOTE:
+#   '/wpsoutputs/<bird>' is the directory location format employd everywhere
```
dbyrns, line 49: Typo `"employd"`

birdhouse/components/weaver/docker-compose-extra.yml
```diff
@@ -0,0 +1,129 @@
...
+    image: pavics/weaver:${WEAVER_VERSION}-manager
+    container_name: ${WEAVER_MANAGER_NAME}
+    ports:
+      - "4001:4001"
+    environment:
+      HOSTNAME: ${PAVICS_FQDN}
+      FORWARDED_ALLOW_IPS: "*"
+    #env_file:
+    #  - ./config/mongodb/credentials.env
+    # ensure other referenced services are created beforehand to avoid booting timing issues
```
dbyrns, line 34: We had problems in the past by trusting too much `depends_on`, flyingpigeon in particular can take serious time to boot and be ready. Compose only garantee that the container is started, but not necessarily ready. 
Given the previous comments, the providers are checked at Weaver boot time, so it may ends up with some of these not available without special care.

birdhouse/optional-components/wps-healthchecks/docker-compose-extra.yml
```diff
@@ -5,13 +5,13 @@ services:
...
         [
           "CMD",
-          "curl",
-          "--insecure",
-          "--fail",
-          "https://localhost${TWITCHER_PROTECTED_PATH}/catalog?service=WPS&request=GetCapabilities",
+          "python",
+          "-c",
+          "import requests; \
+           assert requests.get('http://localhost?service=WPS&request=GetCapabilities').status_code == 200",
```
dbyrns, line 11: Was using twitcher path served on port 80, but now are you sure the catalog is available at port 80? I guess that if the healthcheck is ok it's fine.

dbyrns (approved):
We are there! Tests are Ok and changes LGTM.

fmigneault:
@dbyrns 
> Tests are Ok

have to be careful here because the checkmark refers to default test run that doesn't have weaver enabled
here is a branch with specific test for weaver added: 
http://daccs-jenkins.crim.ca/job/PAVICS-e2e-workflow-tests/job/weaver-config  (via https://github.com/Ouranosinc/pavics-sdi/pull/203)

birdhouse/optional-components/wps-healthchecks/docker-compose-extra.yml
fmigneault: Yes, I updated the healthchecks by doing each of the adjusted request manually within each corresponding container. 
The command is executed from within the container, so twitcher path causes http not found.

birdhouse/components/weaver/docker-compose-extra.yml
fmigneault: Could use this solution instead, but the boot of the full stack will wait until the longest bird to finish init : 
https://github.com/jwilder/dockerize#waiting-for-other-dependencies

tlvu:
I propose we focus on getting Magpie PR (https://github.com/bird-house/birdhouse-deploy/pull/107) merged to `master` first before this one, so I will focus my review effort on the Magpie PR.

birdhouse/components/weaver/conf.extra-service.d/weaver.conf
```diff
@@ -0,0 +1,25 @@
+
```
tlvu, line 1: Huh, why commit a `weaver.conf` if there is also a `weaver.conf.template` ?!

birdhouse/components/weaver/config/magpie/config.yml
```diff
@@ -0,0 +1,106 @@
+## see reference example for details:
```
tlvu, line 1: Same here, why commit `config.yml` when there is `config.yml.template`?

birdhouse/components/weaver/config/weaver/data_sources.json
```diff
@@ -0,0 +1,18 @@
+{
```
tlvu, line 1: Again, `data_sources.json` vs `data_sources.json.template`.

birdhouse/components/weaver/config/weaver/weaver.ini
```diff
@@ -0,0 +1,157 @@
+# Weaver settings configuration
```
tlvu, line 1: In fact you've probably committed all the "generated" files from their "template" origin.

tlvu (changes requested):
Do not commit generated config file.

fmigneault:
> Do not commit generated config file.

@tlvu 

Any idea why this might be happening? I purposely added this file for that reason: 
https://github.com/bird-house/birdhouse-deploy/blob/683ffc3390249a0d2439adf6dd5253c226e693ab/birdhouse/components/weaver/.gitignore

Relative links not working?

fmigneault:
@tlvu  
Except the WMS tests unrelated to Weaver, everything seem to work as before. 
Since Weaver is optional, it won't impact instances that don't add it in the config. 
I plan to merge this soon if you find it's ready too. It has been open long enough.

birdhouse/components/weaver/docker-compose-extra.yml
```diff
@@ -0,0 +1,178 @@
...
+          -wait 'http://flyingpigeon:8093/wps?service=WPS&request=GetCapabilities'
+          -wait 'http://hummingbird:8097/wps?service=WPS&request=GetCapabilities'
+          -wait 'http://malleefowl:8091/wps?service=WPS&request=GetCapabilities'
+          -wait 'http://raven:9099/wps?service=WPS&request=GetCapabilities'
+          -timeout 30s
+        && pserve '/opt/local/src/weaver/config/weaver.ini'"
+    ]
+
+  # Image 'worker' is the job runner of Weaver, possibly spawning Docker applications according to process definition
+  worker:
```
tlvu, line 103: Could this be named `weaver-worker` and not just plain `worker` to be clearer?  We might eventually have a `dask-worker` container in the stack.

birdhouse/components/weaver/docker-compose-extra.yml
```diff
@@ -0,0 +1,178 @@
...
+    restart: always
+    logging: *default-logging
+
+  # Image 'docker-proxy' is the connector between docker image 'worker' jobs and their sibling docker executions
+  docker-proxy:
+    # references
+    #   https://github.com/Tecnativa/docker-socket-proxy
+    #   https://stackoverflow.com/a/47293798/5936364
+    #   https://docs.docker.com/engine/api/
+    image: tecnativa/docker-socket-proxy
```
tlvu, line 146: no version pin?

birdhouse/docker-compose.yml
```diff
@@ -92,6 +92,9 @@ services:
       - ${SSL_CERTIFICATE}:/opt/birdhouse/etc/nginx/cert.pem
     links:
       - mongodb
+    networks:
+      - default
+      - mongodb
```
tlvu, line 97: Phoenix also uses mongodb?  I though nobody was using mongodb before.  Note I do not know much about Phoenix.

tlvu (changes requested):
This PR has changed since the last time I looked at it.  Found a few probably minor issues.  I agree we can merge once the issues are addressed since this is a component so it will not forcefully activate so no much harm.

I have another question that can also be for a different PR, let's not bloat this PR.

If another external bird is plugged into the stack, ex: all the new birds from Nik (PCIC), how will they add themselves to the hardcoded list of known birds to weaver?

birdhouse/components/weaver/Dockerfile
```diff
@@ -0,0 +1,12 @@
+ARG WEAVER_VERSION=latest
```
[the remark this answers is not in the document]
fmigneault: This is related to https://github.com/bird-house/birdhouse-deploy/pull/114#discussion_r596930805 posted by @dbyrns 

Because Weaver needs the other WPS birds to be *ready* to register them as providers, they must be started first. The problem with docker-compose `depends_on` is that it only ensure the other bird images are *started*, but not necessarily *ready* to receive requests. For instance, `flyingpigeon` is particularly long to boot because of the number of processes it has. 

The `Dockerfile` placed here combines the original Weaver docker (from its repo) with `korbin/dockerize`. 
The added utility allows to ping the birds (`GetCapabilities` requests) until they return 200. Only once those are received, `Weaver` will boot and retrieve the responses from the bird that are now *ready for real*.

The reason I placed this here is that this is very specific to this stack. 
Weaver itself doesn't need this tool and I don't believe it should be up to the source repo to handle external use cases like this. The extra docker-compose is defined such that it will try to build this combined image because it doesn't find it remotely.

fmigneault:
> If another external bird is plugged into the stack, ex: all the new birds from Nik (PCIC), how will they add themselves to the hardcoded list of known birds to weaver?

The `wps_processes.yml` would have to be overridden with whichever extra birds that must be registered.
The file currently defines only known birds of the stack. 
Probably the startup command would also need an update to make sure those birds are started before registration.

dbyrns:
> If another external bird is plugged into the stack, ex: all the new birds from Nik (PCIC), how will they add themselves to the hardcoded list of known birds to weaver?

Could we use a pluggable `post-docker-compose-up` script to POST a deploy request to weaver?

tlvu:
> > If another external bird is plugged into the stack, ex: all the new birds from Nik (PCIC), how will they add themselves to the hardcoded list of known birds to weaver?
> 
> Could we use a pluggable `post-docker-compose-up` script to POST a deploy request to weaver?
> The script will make sure that weaver and whatever bird it is are up and ready and then post the request.

Or can weaver support the same "multi config file load" like for magpie?  Currently magpie config can be extended by external component by volume mounting more config files under predefined folders inside the magpie container, like so https://github.com/bird-house/birdhouse-deploy/blob/7dded91a39c4557a3ea4a51d79b56b238a40d2b0/birdhouse/optional-components/generic_bird/docker-compose-extra.yml#L30-L33

tlvu:
> > If another external bird is plugged into the stack, ex: all the new birds from Nik (PCIC), how will they add themselves to the hardcoded list of known birds to weaver?
> 
> Could we use a pluggable `post-docker-compose-up` script to POST a deploy request to weaver?
> The script will make sure that weaver and whatever bird it is are up and ready and then post the request.

A POST request (in a `post-docker-compose-up` script) could also work, a sample equivalent pre script is here https://github.com/bird-house/birdhouse-deploy/blob/7dded91a39c4557a3ea4a51d79b56b238a40d2b0/birdhouse/components/monitoring/pre-docker-compose-up

I guess yours to choose which is simpler or provide more flexibility.

birdhouse/components/weaver/config/magpie/config.yml.template
```diff
@@ -0,0 +1,139 @@
...
+    permission: getcapabilities
+    group: anonymous
+    action: create
+  - service: ${WEAVER_WPS_NAME}
+    permission: describeprocess
+    group: anonymous
+    action: create
+
+## TESTING
+  # Grant write access to execute hummingbird's ncdump process (submit job)
```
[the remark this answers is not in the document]
fmigneault: Yes that could be something added dynamically. 
Something equivalent to `echo "<perms>" >> 'birdhouse/components/weaver/config/magpie/config.yml.template'`. 

When does this python config get executed? I'm not sure I understand its execution workflow. 
The config needs to be tweaked before booting Magpie. 
I would rather not have to get into fetching admin user/pwd to send requests afterwards to created the permissions.

birdhouse/components/weaver/Dockerfile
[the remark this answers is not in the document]
tlvu: Ah ok so I am buying "the reason I placed this here is that this is very specific to this stack.".   Can you just pin the version of the dockerize image so this build is reproducible?

birdhouse/components/weaver/docker-compose-extra.yml
```diff
@@ -0,0 +1,178 @@
...
+  weaver:
+    container_name: ${WEAVER_MANAGER_NAME}
+    # NOTE: Override the usual Weaver-Manager image by one that is built inline with extra dependency to 'dockerize'
+    #   The 'dockerize' CLI utility allows to wait until each WPS GetCapabilities responds correctly, such that once
+    #   Weaver is started, they will all be able to request them and register the providers correctly.
+    #   This is needed because simply adding 'depends_on' only ensures that containers are 'running', but startup of
+    #   WPS applications themselves are not necessarily completed. Successful HTTP responses ensure they are 'ready'.
+    #image: pavics/weaver:${WEAVER_VERSION}-manager
+    image: pavics/weaver:${WEAVER_VERSION}-wait-wps
+    build:
```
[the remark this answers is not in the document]
fmigneault: see other comment

birdhouse/components/weaver/config/magpie/config.yml.template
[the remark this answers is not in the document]
tlvu: > When does this python config get executed? I'm not sure I understand its execution workflow.
> The config needs to be tweaked before booting Magpie.

This is the template expansion feature.  I am pretty sure the actual config file that is given to magpie is just `config.yml` (without the `.template` suffix) so this instantiated `config.yml` will have all the template variable already expanded.

In that `JUPYTERHUB_CONFIG_OVERRIDE` example, it's value would be set in `env.local` and the instanciated `jupyterhub_config.py` file will have the value set in `env.local`.

birdhouse/docker-compose.yml
fmigneault: I'm not sure how it uses it, but it originally has a link to it. If it actually employs it behind the scene, it requires the network now.

I'm not sure if Phoenix is even relevant anymore in this stack.
It works with auth tokens that Twitcher we employ does not completely support nor validate with Magpie permissions. Most probably WPS operations done with it will fail unless the process is 100% public.

fmigneault:
> > Could we use a pluggable `post-docker-compose-up` script to POST a deploy request to weaver?
> > The script will make sure that weaver and whatever bird it is are up and ready and then post the request.
> 
> Or can weaver support the same "multi config file load" like for magpie? Currently magpie config can be extended by external component by volume mounting more config files under predefined folders inside the magpie container, like so 

@dbyrns  @tlvu 

Yes. That script would actually make everything easier to understand since there won't be any extra build step to plug the `dockerize` wait-operation.
 
For extending with more birds though, it would still require manual updates of some config somewhere to specify what to register. 

I would rather avoid going the multi-config way. 
It is easier to prepare the config in advance than handle their combinations after the fact in the code. 

The `post-docker-compose-up` seems like the most suitable option.

birdhouse/components/weaver/config/magpie/config.yml.template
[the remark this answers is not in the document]
fmigneault: Oh ok. I think I get what you meant. 
To confirm: 
Having everything under `## TESTING` defined literally in a variable `WEAVER_MAGPIE_TEST_CONFIG` within `env.local` and have `${WEAVER_MAGPIE_TEST_CONFIG}` in birdhouse/components/weaver/config/magpie/config.yml.template ?

birdhouse/components/weaver/config/weaver/data_sources.json.template
fmigneault: Yes

birdhouse/components/weaver/config/magpie/config.yml.template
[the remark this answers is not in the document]
fmigneault: I find extending the config with an optional template variable easier than extending a templated code snippet that extends the config dynamically.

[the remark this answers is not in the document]
tlvu: > I find extending the config with an optional template variable easier than extending a templated code snippet that extends the config dynamically.

Looks like we might be saying the same thing but differently.  Can you re-explain "your way" or implement it if not too complex?

birdhouse/docker-compose.yml
tlvu: I have a vague memory PCIC might use Phoenix.  Anyways, better be safe than sorry.  That network addition might not be needed, but it probably does no harm so good thing you add it.

birdhouse/components/weaver/config/magpie/config.yml.template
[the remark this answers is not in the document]
fmigneault: I think I will add a `test-weaver` optional-component similar to `testthredds`. 
That will be the easiest definition that follows current methodologies to extend features.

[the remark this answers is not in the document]
tlvu: Not bad this `test-weaver` optional-component idea.  Is that what you originally had in mind in comment https://github.com/bird-house/birdhouse-deploy/pull/114#discussion_r700568551?  Doesn't look like the same idea to me but I'll buy this `test-weaver`.

Looks like Magpie "multi config file load" is quite useful for customizing it's behavior as we just see with `test-weaver`.  Would you consider to implement the same for Weaver?

tlvu:
> Yes. That script would actually make everything easier to understand since there won't be any extra build step to plug the `dockerize` wait-operation.

FYI @fmigneault  `docker-compose up` returns when all the containers successfully started, not that they are ready so I don't think waiting for it solves your `dockerize` problem.

dbyrns:
> FYI @fmigneault `docker-compose up` returns when all the containers successfully started, not that they are ready so I don't think waiting for it solves your `dockerize` problem.

It "could" if the `post-docker-compose-up` script explicitely waits until all WPS are ready. This script may not be popular tough if it blocks until all WPS are ready.

tlvu:
> > FYI @fmigneault `docker-compose up` returns when all the containers successfully started, not that they are ready so I don't think waiting for it solves your `dockerize` problem.
> 
> It "could" if the `post-docker-compose-up` script explicitely waits until all WPS are ready. This script may not be popular tough if it blocks until all WPS are ready.

Ah true, then "wait for all WPS are ready" could be implemented for `weaver/post-docker-compose-up` only.  Each `post-docker-compose-up` can do whatever is needed for their specific component.

fmigneault:
@tlvu @dbyrns 
See https://github.com/bird-house/birdhouse-deploy/pull/192 for the script.

dbyrns:
> Ah true, then "wait for all WPS are ready" could be implemented for weaver/post-docker-compose-up only.

Exactly what I had in mind. Sorry if that was not obvious.

dbyrns:
> But once it is activated, we know it can start the first time but we have no idea how it integrate itself into the upgrade process (autodeploy) because that scenario is not tested by the `daccs-iac`.

@tlvu just so you know, indeed this process is not tested by the IAC, but it is tested by our staging environment that will be upgraded as soon as the PR is merged. So if Hirondelle goes well after the merge we will know everything is OK. I thought that we all agreed on that procedure. A PR is accepted based on the IAC tests. An organization deploys on prod using their fork after they check their own staging with master. An organization is not required to validate the upgrade on other organizations staging but can support them in case of documented issues.

tlvu:
> > But once it is activated, we know it can start the first time but we have no idea how it integrate itself into the upgrade process (autodeploy) because that scenario is not tested by the `daccs-iac`.
> 
> @tlvu just so you know, indeed this process is not tested by the IAC, but it is tested by our staging environment that will be upgraded as soon as the PR is merged. So if Hirondelle goes well after the merge we will know everything is OK. I thought that we all agreed on that procedure. A PR is accepted based on the IAC tests. An organization deploys on prod using their fork after they check their own staging with master. An organization is not required to validate the upgrade on other organizations staging but can support them in case of documented issues.

@dbyrns Just to be sure, there are 2 aspects to an upgrade during the autodeploy: 1) the code, 2) the exiting data that might need migration.

So I understand for number 2, each org with each different production data will be in a better position to test their own data.  

However for number 1, that code can be tested by the feature owner, weather it can run successfully inside the autodeploy process, at least without any errors for the most basic case where no data migration is required.

Both the autodeploy and weaver are optional components so it's not a blocking issue so I accepted this PR.  But it could still introduce a regression for autodeploy and I think we all agree regression is bad.

So that's why I mentioned in comment https://github.com/bird-house/birdhouse-deploy/pull/114#issuecomment-953263588 "Just indicate in the changelog clearly that autodeploy is not tested, in case another org (PCIC) wants to activate this."  This is to prevent another org that already has autodeploy currently activated to be careful with activating weaver at the same time.  And my comment has been ignored.

dbyrns:
For the changelog, you are right, a message could have been added, surely an honest mistake.
To test the autodeploy though, I still think that the staging is the right way to validate that, even for the code part. Let's talk about that in our next meeting.

birdhouse/docker-compose.yml
tlvu: @fmigneault I just double check again before going live on prod with this PR (without Weaver activated) and I notice this.

The `mongodb` network is added to `phoenix` container but not to the `mongodb` container !  This looks really odd.  Is that accidental or intentional?

fmigneault: Accidental. Must have happened during the many refactors into separate PRs.
I'll open another PR to add it.

birdhouse/components/weaver/config/canarie-api/weaver_config.py.template
```diff
@@ -0,0 +1,9 @@
+SERVICES['node']['monitoring'].update({
```
tlvu, line 1: @fmigneault another problem detected after go-live in prod yesterday:

This `weaver_config.py` should be added to `.gitignore`, else autodeploy is broken because it would detect an unclean checkout.

fmigneault: Do you want a separate PR for this fix only (with version bump and all?).

PR #118, opened 2021-01-25
Expose and restructure the documentation 

Zeitsperre:
The documentation can be viewed inline when browsing on Github or on ReadTheDocs https://birdhouse-deploy.readthedocs.io.

Convert all `.md` files to `.rst` files so they can be used on ReadTheDocs.

Due to the hybrid goal (view doc both while browsing Github and on ReadTheDocs), some compromise had to be made:

* all the relative links to other files in this repo are only working while browsing Github, not on RtD

* a "download" link is added to all those relative links for RtD, but annoyingly those "download" link are also visible while browsing Github but not working

The current state is not perfect.  We'll try to address the "relative links" vs "download links" work-around to have hybrid doc both when browsing Github and on ReadTheDocs another time.


docs/source/index.rst
```diff
@@ -20,6 +20,8 @@ Contents:
 
    installation
    deployment
+   birdhouse/components/README
```
tlvu, line 23: Since adding the components one, might as well add the optional-components one :)

birdhouse/optional-components/README.rst
```diff
@@ -0,0 +1,140 @@
+Optional components
```
tlvu, line 1: Not deleting the old optional-components/README.md ?

README.rst
```diff
@@ -15,6 +15,6 @@ PAVICS
 Power Analytics and Visualization for Climate Science - Powered by Birdhouse and other ESGF software
 
 How to deploy the entire PAVICS platform, see the
-`README <birdhouse/README.md>`_ and the various extra components 
+`README <birdhouse/README.rst>`_ and the various extra components
 `README for extra core components <birdhouse/components/README.rst>`_\ ,
 `README for optional components <birdhouse/optional-components/README.md>`_.
```
tlvu, line 20: optional-components/README.rst now too :)

birdhouse/README.rst
```diff
@@ -23,22 +19,18 @@ Requirements:
   Install latest docker-ce and docker-compose for the chosen distro (not the
   version from the distro).
 
-To run ``docker-compose`` for PAVICS, the `\ ``pavics-compose.sh`` <pavics-compose.sh>`_ wrapper script must be used.
-This script will source the ``env.local`` file, apply the appropriate variable substitutions on all the configuration files ".template", and run ``docker-compose`` with all the command line arguments given to ``pavics-compose.sh``. See `\ ``env.local.example`` <env.local.example>`_ for more details on what can go into the ``env.local`` file.
+To run ``docker-compose`` for PAVICS, the `pavics-compose.sh <birdhouse/pavics-compose.sh>`_ wrapper script must be used.
```
tlvu, line 22: This is odd, browsing on github, the link is also broken as well with the addition of `birdhouse` in front of `pavics-compose.sh`.  I imagine this is to fix the RtD build?

Before you spend more time on this, is it realistic to have the same kind of relative link that would work both on github and in RtD?  Maybe we want to test this feature first.

Note on github, relative links are relative to the current file, not to the root of the checkout.

birdhouse/README.rst
```diff
@@ -150,11 +138,11 @@ https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests with pre-configured
 Jenkins at https://github.com/Ouranosinc/jenkins-config.
 
 For that test suite to pass, run the script
-`\ ``bootstrap-instance-for-testsuite`` <scripts/bootstrap-instance-for-testsuite>`_
+`bootstrap-instance-for-testsuite <birdhouse/scripts/bootstrap-instance-for-testsuite>`_
 to prepare your new instance.  Further documentation inside the script.
 
 Optional component
-`all-public-access <optional-components#give-public-access-to-all-resources-for-testing-purposes>`_
+`all-public-access <birdhouse/optional-components#give-public-access-to-all-resources-for-testing-purposes>`_
```
tlvu, line 145: This format was with the old `.md`, wonder if samething with new `.rst`.

birdhouse/optional-components/README.rst
```diff
@@ -0,0 +1,140 @@
...
+
+So that the url ``https://<PAVICS_FQDN>/canarie/node/service/stats`` also return
+what the end user really see (a component might work but is not accessible to
+the end user).
+
+This assume all the WPS services are public.  If not the case, make a copy of
+this config and adjust accordingly.
+
+How to enable this config in ``env.local`` (a copy from
+`env.local.example <birdhouse/env.local.example>`_):
```
tlvu, line 15: Same problem with broken relative link when browsing on github.

tlvu (changes requested):
Github browsing still broken with relative links.

You should be previewing the changes directly on Github and not in your IDE I think, so you can catch these errors faster.

tlvu:
The links in RtD is also broken. ex: https://birdhouse-deploy--118.org.readthedocs.build/en/118/birdhouse/pavics-compose.sh.

birdhouse/optional-components/README.rst
Zeitsperre: I need to find a better way for linking to non-rst files. I feel like I'm close.

birdhouse/README.rst
Zeitsperre: Would it make more sense to provide the GitHub links instead?

birdhouse/README.rst
Zeitsperre: I'm testing things with symlinks and relpaths, maybe this approach isn't working.

birdhouse/optional-components/README.rst
Zeitsperre: This has now been addressed.

docs/source/index.rst
```diff
@@ -20,6 +20,9 @@ Contents:
...
+   birdhouse/docker/geoserver/README
+   birdhouse/optional-components/README
```
tlvu, line 25: This `optional-components/README` should be right below `components/README`.

birdhouse/README.rst
tlvu: > Would it make more sense to provide the GitHub links instead?

I think probably should get relative links to work on GH first, then find a way to make it work on RtD.

It was working before on GH so let's not break that.  Then RtD is the added bonus.

I think we might have to copy all the files not `.rst` we reference to on RtD as well, else for sure the links won't work on RtD.  Either that or RtD need a automatic relative link rewrite to make it absolute link pointing to GH.

So basically in the code it's relative links so it works on GH browsing.  But during the doc build on RtD, those relative links are transformed to absolute link so no files need to be copied to RtD.

Hope such a feature exists.

Zeitsperre: I think it does. It's been a LONG time since I've done some really in-depth documentation structuring. I'm gradually getting there.

Zeitsperre:
Went in a different direction and decided to make the GitHub links work by default, with the option of downloading a given file (they're copied to RtD as downloadable files). If we want, I can go in and replace all download links with the URL to these files on GitHub, but because they aren't markdown or `rst`'s, getting them to display as raw content is only viable with the `:download:` approach.

README.rst
```diff
@@ -14,7 +14,8 @@ PAVICS
 
 Power Analytics and Visualization for Climate Science - Powered by Birdhouse and other ESGF software
 
-How to deploy the entire PAVICS platform, see the
-`README <birdhouse/README.rst>`_ and the various extra components
-`README for extra core components <birdhouse/components/README.rst>`_\ ,
-`README for optional components <birdhouse/optional-components/README.md>`_.
+For GitHub navigation, see the following README pages:
```
tlvu, line 17: Why not just keep the original text "How to deploy the entire PAVICS platform, see the following README pages:" ... cuz it reads weird if user is actually on Github.

birdhouse/README.rst
```diff
@@ -19,17 +19,17 @@ Requirements:
   Install latest docker-ce and docker-compose for the chosen distro (not the
   version from the distro).
 
-To run ``docker-compose`` for PAVICS, the `pavics-compose.sh <birdhouse/pavics-compose.sh>`_ wrapper script must be used.
+To run ``docker-compose`` for PAVICS, the `pavics-compose.sh <pavics-compose.sh>`_ (:download:`download </birdhouse/pavics-compose.sh>`) wrapper script must be used.
```
tlvu, line 22: Those `:download:download </birdhouse/pavics-compose.sh>` are actually showing up when browsing on GH.

tlvu (changes requested):
Also from here https://birdhouse-deploy--118.org.readthedocs.build/en/118/, all the README links also do not resolve.

tlvu:
Oh true, those "download" links works on RtD https://birdhouse-deploy--118.org.readthedocs.build/en/118/deployment.html

birdhouse/README.rst
tlvu: Hum probably we have to live with seeing those download instruction when browsing GH.

tlvu: Instead of using the :download role, maybe the :doc role?  https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#the-doc-role (I have no idea what I am saying, I never used it, just found by googling).

Zeitsperre: The `:doc:` approach only works for linking to rendered pages. This doesn't work for things that need to be presented in their raw formats (shell scripts, YAML, JSON, etc.).

The alternative is to simply provide the GitHub address to the file and perform a link-check on build. The downside is that if you add or delete a file being pointed to, the docs will need to be pushed to master before you know they work (similar to the issues with adding new PyWPS services that need to be tested online). This could also be done?

Zeitsperre:
So I had a thought last night: GitHub navigation requires Markdown-like references (``` `LINKNAME <relpath/to/non-rst-file>`_ ```), while RtD/rST needs urls (``` `LINKNAME <URLrelpath/to/non-rst-file>`_ ```). If there were an easy way to template the URL to these relpaths when building the dcos? Maybe Jinja2 could help with that? Maybe this is a bit much.

tlvu:
> What I need to know is do we set links to the CFG/YAML/JSON files on GitHub or do we copy them to RtD and serve them from there?

I would prefer copy to RtD and serve them from there so it is snapshotted.  If we link to GH from RtD, I wonder how do we lock down the version (ie if RtD was built by `changeset1`, how do we link to the files on GH at `changeset1` and not `master` that is constantly changing).

By the way, I think we can accept having the `:download` role showing up when browsing GH for now.  We can revisit later.  I feel bad you spend too much time on this already.

So if you can copy the files to RtD so the relative links work both on RtD and GH and we can remove the annoying `:download` role, that would be best.  If that seems like another rabbit hole, we can stop here.

Zeitsperre:
> I would prefer copy to RtD and serve them from there so it is snapshotted. If we link to GH from RtD, I wonder how do we lock down the version (ie if RtD was built by `changeset1`, how do we link to the files on GH at `changeset1` and not `master` that is constantly changing).

Short answer: It would be a huge pain to do this. We'd need to add commit hashes to everything, which would be feasible with Jinja, but a nightmare to maintain. Maybe there's a better way, but I don't know it.

> By the way, I think we can accept having the `:download` role showing up when browsing GH for now. We can revisit later. I feel bad you spend too much time on this already.

It's fine, I needed the practice. I'm just glad it's nearly done.

> So if you can copy the files to RtD so the relative links work both on RtD and GH and we can remove the annoying `:download` role, that would be best. If that seems like another rabbit hole, we can stop here.

The only way to do that would be to convert all the non-rst/md files to HTML source files using a different converter (vanilla Sphinx doesn't support this out of the box). It's worth looking into but not today.

README.rst
Zeitsperre: I mention GutHub here because on RtD, clicking these links won't work, last I checked.

birdhouse/components/README.rst
```diff
@@ -113,24 +112,27 @@ Comparison between the old and new autodeploy mechanism
...
+* Old single ssh deploy key is still compatible, but the new mechanism allows for different ssh deploy keys for each extra
+  repos (again, public repos should use https clone path to avoid dealing with ssh deploy keys in the first place).
 * Old install scripts are kept and can still deploy the old way.
 
 Features missing in old install scripts or how the new mechanism improves on the old install scripts:
 
-* Autodeploy of the autodeploy itself !  This is the biggest win.  Previously, if ``triggerdeploy.sh`` or ``PAVICS-deploy-notebooks`` script changes, they have to be deployed manually.  It's very annoying.  Now they are volume-mount in so are fresh on each run.
-* ``env.local`` now drive absolutely everything, source control that file and we've got a true DevOPS pipeline.
+* Autodeploy of the autodeploy itself !  This is the biggest win.  Previously, if ``triggerdeploy.sh`` (:download:`download <../deployment/triggerdeploy.sh>`)
+  or ``PAVICS-deploy-notebooks`` (TODO: @tlvu Where is this file?) script changes, they have to be deployed manually.  It's very annoying.  Now they are volume-mount in so are fresh on each run.
```
Zeitsperre, line 122: @tlvu

tlvu: It's not a file in the source tree, it's the deployed location of the script (/etc/cron.hourly/PAVICS-deploy-notebooks).  Probably change it to ``/etc/cron.hourly/PAVICS-deploy-notebooks`` and leave it without download link.

docs/source/vagrant_variables.yml.example
```diff
@@ -0,0 +1 @@
+../../vagrant_variables.yml.example
```
tlvu, line 1: @Zeitsperre What was this symlink for again?

tlvu (approved):
I don't like the download link work-around but this will be for another day.

tlvu:
@Zeitsperre Check db4d5aab1bf24a8d2cd9972b2e0a638d35533e5f I found a way to directly copy some files to the html output but it only copies to the root, we are unable to specify any destination hierarchy.

Pretty much the only feature missing to have relative link working both while browsing Github and on RtD !!!

docs/source/vagrant_variables.yml.example
Zeitsperre: Ah, it was to address the file access problem. I was exploring symlinking into the docs folder all the files I needed, but looks like you took a different approach. Looking at it all now.

tlvu: My other "approach" only works for files in `birdhouse/` folder only, all other files still need the "download links" trick.  So no nothing new, no change of direction.

Zeitsperre: It would be interesting if we could use hard links both in  `/birdhouse`  and under `/docs/source/`, but that might cause issues in other ways.

tlvu: > It would be interesting if we could use hard links both in `/birdhouse` and under `/docs/source/`, but that might cause issues in other ways.

Not sure how this is going to solve our relative links on Github not working on RtD problem.   What we really need is either

* a way to copy files into the html output folder at any location we want (the option I found in db4d5aab1bf24a8d2cd9972b2e0a638d35533e5f only copy to the root of the html output folder)

* a way to instruct RtD/Sphinx to rewrite all relative links to absolute link to Github at the same changeset of the build on RtD (ex: a relative link like `../env.local.example` should be replace with `https://github.com/bird-house/birdhouse-deploy/blob/COMMIT_HASH_OF_RTD_BUILD/birdhouse/env.local.example`

PR #122, opened 2021-02-03
proxy: increase timeout since raven can take lots of time to process requests

tlvu:
Fix this timeout error in some Raven notebooks:
```
HTTPError: 504 Server Error: Gateway Time-out for url:
https://pavics.ouranos.ca/twitcher/ows/proxy/raven/wps
```

This is just a work-around since something is very wrong on our
production host Boreas (a physical host with 128G ram and 48 logical cpu).

During the test, Boreas was having this "load average: 6.35, 5.90,
4.33".  For its hardware specs, it is basically idle.

The increased timeout was not needed for my test VM (10G ram, 2 cpu),
medus.ouranos.ca (physical host with 16G ram, 16 logical cpu) and
hirondelle.crim.ca (VM with 32G ram, 8 cpu).

Should fix https://github.com/Ouranosinc/raven/issues/362 and fix https://github.com/Ouranosinc/raven/issues/357.

Ping @moulab88 to take a look at Boreas.  Just a wild guess, is it due for a reboot?

Ping @richardarsenault can you retry the 2 broken Raven notebooks?  I've already deployed this to prod.

### DIFF
--- a/birdhouse/README.rst
+++ b/birdhouse/README.rst
@@ -6,20 +6,15 @@ Docker instructions
 
 Requirements:
 
+* Centos 7 or Ubuntu Bionic (18.04), other distros untested.
 
-* 
-  Centos 7 or Ubuntu Bionic (18.04), other distros untested.
-
-* 
-  Hostname of Docker host must exist on the network.  Must use bridge
+* Hostname of Docker host must exist on the network.  Must use bridge
   networking if Docker host is a Virtual Machine.
 
-* 
-  User running ``pavics-compose.sh`` below must not be ``root`` but a regular user
+* User running ``pavics-compose.sh`` below must not be ``root`` but a regular user
   belonging to the ``docker`` group.
 
-* 
-  Install latest docker-ce and docker-compose for the chosen distro (not the
+* Install latest docker-ce and docker-compose for the chosen distro (not the
   version from the distro).
 
 To run ``docker-compose`` for PAVICS, the `pavics-compose.sh <pavics-compose.sh>`_ (:download:`download </birdhouse/pavics-compose.sh>`) wrapper script must be used.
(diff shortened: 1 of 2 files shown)


huard (approved):
I think we should document the fact any process running longer than this limit will fail. 
Also what happens on the PyWPS side, is the process killed or it becomes a zombie ?

tlvu:
> I think we should document the fact any process running longer than this limit will fail.
> Also what happens on the PyWPS side, is the process killed or it becomes a zombie ?

It just continue and return later ... when it's too late.  So no zombie or accumulated queue.

richardarsenault:
@tlvu I confirm that the 2 notebooks (Ouranosinc/raven#362 and Ouranosinc/raven#357) now work without time-outing!

moulab88:
The machine is up since 13 days, no need a reboot. I will analyze incoming traffic with more details.

tlvu:
> The machine is up since 13 days, no need a reboot. I will analyze incoming traffic with more details.

@moulab88 Just to be clear, the traffic is not blocked.  There is simply something that appear to slow down Raven responses from the client point of view.  There are quite a few possible reasons here:

* The request takes time to reach Raven (maybe OS firewall, Nginx, or Twitcher buffering the request)
* Raven itself is slower (slow disc or network access to get the data? There is plenty of ram and cpu so probably not a problem with ram/cpu)
* The response from Raven takes time to return to caller (again maybe OS firewall, Nginx, or Twitcher buffering the response)

Ping me if you need anything.

moulab88:
The OS firewall is disable. I will just check if there are dropped packets on the connections.

moulab88:
We have better latency on local disks / or /home than direct attached RAID5 direct disk which have better throughput:
image

To increase containers IO performance maybe it is better to move all docker directories to the local disks.

tlvu:
> We have better latency on local disks / or /home than direct attached RAID5 direct disk which have better throughput:
> 
> To increase containers IO performance maybe it is better to move all docker directories to the local disks.

Sensible suggestion.  Migration will take time, we have lots of images and data.

Also there might not be enough space.  `/` have 38G free, for sure not enough, `/home` have 114G free, probably not enough and for sure not enough future proof.

Given complicated migration, we probably should find a way to test the idea without doing the full migration.

moulab88:
Forget to mention  /var (265 GB Free), this space was reserved for this need.

tlvu:
We should also be using `overlay2` as the storage driver for Docker, see https://docs.docker.com/storage/storagedriver/overlayfs-driver/

Currently, it is not the case:

```
$ docker info
 Storage Driver: overlay
  Backing Filesystem: xfs
  Supports d_type: true
 Docker Root Dir: /pvcs1/var-lib/docker
WARNING: the overlay storage-driver is deprecated, and will be removed in a future release.
```

tlvu:
```
$ sudo du -sh /pvcs1/var-lib/docker/
478G    /pvcs1/var-lib/docker/
```

There is clean up to do and we'll need an even bigger partition !

moulab88:
oh!! it will be a good candidate to a new SSD/nVME disk/partition.

tlogan2000:
With Raven is there an issue `Progress=True` or `async` like we have seen for finch?

tlvu:
> With Raven is there an issue `Progress=True` or `async` like we have seen for finch?

This is an excellent question.  There should be Raven notebooks using async mode, since this issue exists https://github.com/Ouranosinc/raven/issues/353, but I have not seen the same "queue not cleared" problem with Raven !!!  This is very odd, why Finch have the problem and not Raven !

huard:
I don't know if we should read too much into that. Maybe Raven has not been exercised as intensively as Finch. 
One difference I see is that Finch uses sentry. Maybe it has an unhealthy relationship with PyWPS.

tlvu:
@moulab88 Performance seems to have improved on prod Boreas, did you change something?  See test results in https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/61

moulab88:
I did nothing on my side.

PR #125, opened 2021-02-16
cronjob: attempt LetsEncrypt SSL renewal weekly instead of daily

tlvu:
To avoid cutting connection daily since the proxy is restarted during renewal attempt.


tlogan2000 (approved):
Perfect thanks Long

tlvu:
@tlogan2000 Sorry, just to be clear, this do not warranty no proxy restart during the week.  If autodeploy detect there are new changes, it will still deploy the changes and the proxy is still restarted at the end of the autodeploy process.

Should still expect once or twice restart during the week.  At least it won't be everyday.

tlogan2000:
Ok sounds good

PR #126, opened 2021-02-19
Configurable Jupyterhub README

nikola-rados:
While the [`README.ipynb`](https://github.com/bird-house/birdhouse-deploy/blob/master/docs/source/notebooks/README.ipynb) provided by `birdhouse-deploy` is good, it does not quite fit our needs at PCIC. This PR allows users to configure their own `README` for Jupyterhub.

## Changes
- Adds `JUPYERHUB_README` as configuration option in the appropriate spots

PR #130, opened 2021-03-09
Custom Jupyter user images

ChaamC:
Adds CRIM's nlp and eo images to the available list of images in JupyterHub

The base image (pavics-jupyter-base) wasn't added to the list, because I assume the users will always be using the other more specialized images, but let me know if you think I should add it too.

We were already able to add/override Jupyter images but this PR makes it more integrated: those image will also be pulled in advanced so startup is much faster for big images since these images will already be cached.

Backward incompatible changes:
* `DOCKER_NOTEBOOK_IMAGE` renamed to `DOCKER_NOTEBOOK_IMAGES` and is now a space separated list of images.  Any existing override in `env.local` using the old name will have to switch to the new name.

Edit:
* [Long] note new behavior change that the image caching is better
* [Long] note backward incompatible variable name change


birdhouse/env.local.example
```diff
@@ -206,7 +206,9 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
 # allow jupyterhub user selection of which notebook image to run
 # see https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html
 #export ENABLE_JUPYTERHUB_MULTI_NOTEBOOKS="
-#c.DockerSpawner.image_whitelist = {'pavics': os.environ['DOCKER_NOTEBOOK_IMAGE'],
+#c.DockerSpawner.image_whitelist = {'pavics': os.environ['PAVICS_E2E_NOTEBOOK_IMAGE'],
+#                                   'eo-crim': os.environ['EO_NOTEBOOK_IMAGE'],
+#                                   'nlp-crim': os.environ['NLP_NOTEBOOK_IMAGE'],
```
MatProv, line 211: For test infra on a real jupyter instance, you'll need to change the test `env.local` in DACCS/daccs-configs repo, in order to see these images.

MatProv:
run tests

dbyrns (approved):
LGTM

birdhouse/deployment/deploy.sh
```diff
@@ -178,8 +178,10 @@ set -x
 # restart everything, only changed containers will be destroyed and recreated
 ./pavics-compose.sh up -d
 
-# immediately cache the new notebook image for faster startup by JupyterHub
-docker pull $DOCKER_NOTEBOOK_IMAGE
+# immediately cache the new notebook images for faster startup by JupyterHub
+docker pull $PAVICS_E2E_NOTEBOOK_IMAGE
+docker pull $EO_NOTEBOOK_IMAGE
+docker pull $NLP_NOTEBOOK_IMAGE
```
tlvu, line 184: I am not fond of adding all these images directly here.  I think we have to think bigger, that PAVICS will be deployed on many different organizations.  We do not want to force all the orgs to always download all these gigantic images.

The old `DOCKER_NOTEBOOK_IMAGE` var name was meant to be generic so if another org wants, they can simply override the default value in `default.env` and have their own flavor of Jupyter env.

So I would suggest we move these extra jupyter env into an `optional-components/` (https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse/optional-components).  There are plenty of existing examples you can based upon.  One new thing is this `docker pull` part and you can use the `post-docker-compose-up` script https://github.com/bird-house/birdhouse-deploy/blob/3718a320780bc08cc18632d3c79bb196b2749633/birdhouse/pavics-compose.sh#L193 to do what you need.

Here an example of a `pre-docker-compose-up` script that is equivalent to this `post-docker-compose-up` script: https://github.com/bird-house/birdhouse-deploy/blob/3718a320780bc08cc18632d3c79bb196b2749633/birdhouse/components/monitoring/pre-docker-compose-up#L3-L5

tlvu (changes requested):
Your implementation is good and it works.

But I would rather we leverage the "component" mechanism so the PAVICS stack remains generic for everyone.  Much like the previous PR https://github.com/bird-house/birdhouse-deploy/pull/126 from PCIC that started making the README file of Jupyter env configurable so they can provide their own file.

Otherwise, great work!

dbyrns:
@tlvu I agree with you that it's not a good idea to hardcode new image flavors as they come. However, I don't see additional flavors as an optional component. I think the better approach would be to allow setting multiple images using the DOCKER_NOTEBOOK_IMAGE(**S**) environment variable. This way we could only set one image to keep the current behavior or add more if flavors are available.

dbyrns:
It's true that we would need a way to advertise images. 
Maybe the env.local.exemple could simply list available images, so that orgs can keep only the needed ones?

tlvu:
> It's true that we would need a way to advertise images.
> Maybe the env.local.exemple could simply list available images, so that orgs can keep only the needed ones?

Yup, alongside the other existing public Jupyter images.

MatProv:
> It's true that we would need a way to advertise images.
> Maybe the env.local.exemple could simply list available images, so that orgs can keep only the needed ones?

Yes, nice and easy, I agree

ChaamC:
So, I tried the new solution suggested. Let me know if it corresponds to what you guys had in mind, or if you have a cleaner way to write certain parts (especially the split[0], split[1], ... in the env.local)!

birdhouse/default.env
```diff
@@ -1,8 +1,8 @@
 # All env in this default.env can be overridden by env.local.
 # All env in this default.env must not depend on any env in env.local.
 
-# Jupyter single-user server image
-export DOCKER_NOTEBOOK_IMAGE="pavics/workflow-tests:210216"
+# Jupyter single-user server images, can be overriden in env.local to have a list of multiple images
```
tlvu, line 4: Add "space separated list of multiple images".  Just to be extra clearer.

birdhouse/deployment/deploy.sh
```diff
@@ -178,8 +178,11 @@ set -x
 # restart everything, only changed containers will be destroyed and recreated
 ./pavics-compose.sh up -d
 
-# immediately cache the new notebook image for faster startup by JupyterHub
-docker pull $DOCKER_NOTEBOOK_IMAGE
+# immediately cache the new notebook images for faster startup by JupyterHub
+for IMAGE in ${DOCKER_NOTEBOOK_IMAGES}
+do
+  docker pull $IMAGE
```
tlvu, line 184: Ah true, forgot docker pull can not accept multiple images.

tlvu: Oh wait, this same docker pull was already in the `pavics-compose.sh` script which is called earlier already.  I think we can simply delete this.

tlvu (approved):
Looks good now.  Minor changes requested only.

tlvu (approved):
+1

dbyrns (approved):
Nice job! Generic and simple, I love that.

tlvu:
@ChaamC I'll let you do the merge.  I'd like the merge commit to contain the summary of the changes so I've updated the PR description with what is relevant.  Please copy the same PR description into the merge commit description.

ChaamC:
@MatProv 
Was waiting for the approval of this PR first, to be sure the new code added to the env.local files would be valid in `daccs-config` :P
I will go ahead and do this now ;)

MatProv:
> Was waiting for the approval of this PR first

@ChaamC Yes I understand your point. Only thing is that failed build status will remain until `daccs-config`'s updated. I'm pretty sure everything will be fine as well, it's only to enforce e2e test usage, to avoid surprises. Other than that great PR, it will be interesting to use custom images on the platform!

ChaamC:
run tests

ChaamC:
run tests

tlvu:
@ChaamC merge with `master` which has latest Finch that will fix recent (during my vacation) notebooks failure, see PR https://github.com/bird-house/birdhouse-deploy/pull/132 with the last commit triggering the pipeline passing (https://daccs-jenkins.crim.ca/job/DACCS-iac/237/).

ChaamC:
run tests

ChaamC:
@tlvu We think we have another problem as it doesn't even reach the tests. Maybe the daccs-iac timeout that is not long enough with the additional delay required for pulling more images. We're checking that out!

tlvu:
> @tlvu We think we have another problem as it doesn't even reach the tests. Maybe the daccs-iac timeout that is not long enough with the additional delay required for pulling more images. We're checking that out!

I thought you guys are caching all the docker images so all docker pull are basically no-op.

ChaamC:
@tlvu Yes there is caching, but here, since we had just added the images to the config, I guess they were not already in the cache and required a complete pulling.

ChaamC:
run tests

PR #131, opened 2021-03-15
bootstrap-testdata: new testdata for pavics-sdi regridding notebook

tlvu:
Need those new testdata on hosts other than `pavics.ouranos.ca` to run Jenkins end-to-end test suite against those hosts.

For notebook https://github.com/Ouranosinc/pavics-sdi/blob/bbe3061dceb98688e9fd58836fbc50992ee31568/docs/source/notebooks/regridding.ipynb

See PR https://github.com/Ouranosinc/pavics-sdi/pull/201


tlvu:
@huard 2 of those new files are outside of the `testdata/` folder, are we okay with that?

In the past, we already have files outside of the `testdata/` folder.  Do we need more enforcement going forward or it's okay to use files outside of the `testdata/` folder as long as there are noting suitable in the `testdata/` folder?

huard:
Good question. I don't feel like I'm able to give you a good answer though. I suggest we hot-fix this for now but come back to this question during our next meeting.

huard (approved):

tlvu:
Also for https://github.com/bird-house/birdhouse-deploy/pull/132

PR #135, opened 2021-03-20
doc LetsEncrypt SSL renew job should be sourced after SSL_CERTIFICATE in env.local

tlvu:
Found accidentally when manually harmonizing `env.local` produced by
Vagrant with the template `env.local.example`.

Vagrant automatic provisioning does the right thing to activate
`renew_letsencrypt_ssl_cert_extra_job.env` after the redefined
SSL_CERTIFICATE but if someone like me, manually re-order this, it
will break.

Fix the follow error, SSL_CERTIFICATE was set to /home/vagrant/certkey.pem.
```
+ '[' -s /tmp/tmp_certbotwrapper_ssl_cert.pem ]
+ diff /home/vagrant/certkey.pem /tmp/tmp_certbotwrapper_ssl_cert.pem
diff: can't stat '/home/vagrant/certkey.pem': No such file or directory
+ '[' 0 -eq 0 ]
+ cp -v /tmp/tmp_certbotwrapper_ssl_cert.pem /home/vagrant/certkey.pem
cp: can't create '/home/vagrant/certkey.pem': No such file or directory
```

### DIFF
--- a/birdhouse/components/scheduler/renew_letsencrypt_ssl_cert_extra_job.env
+++ b/birdhouse/components/scheduler/renew_letsencrypt_ssl_cert_extra_job.env
@@ -2,6 +2,9 @@
 # Configuration vars, set in env.local before sourcing this file.
 # This job assume the "scheduler" component is enabled.
 #
+# This job will write to the value of SSL_CERTIFICATE in env.local so make sure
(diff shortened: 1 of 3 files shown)


nikola-rados (approved):
Always appreciate extra context in the documentations :+1:

MatProv (approved):

PR #136, opened 2021-03-24
Upgrade geoserver to latest upstream kartoza/geoserver:2.18.2

tlvu:
Completely removed our custom build.  Upgrade will be much easier next time.

What is cool with this new upstream version, from deployment perspective:

* many plugins are pre-downloaded, we just have to enable them, see https://github.com/kartoza/docker-geoserver/blob/553ed2982685f366ddcbac3d3e1626cb493cf84b/scripts/setup.sh#L13-L41
  * full plugin list we can enable https://github.com/kartoza/docker-geoserver/blob/553ed2982685f366ddcbac3d3e1626cb493cf84b/build_data/stable_plugins.txt and https://github.com/kartoza/docker-geoserver/blob/553ed2982685f366ddcbac3d3e1626cb493cf84b/build_data/community_plugins.txt

* admin password can be set via config, no need for manual step post deployment anymore, sweet !!!

What might be different from the previous version:

* csw-plugin not found in stable_plugins.txt or community_plugins.txt so it is missing in this config.  Full disclosure, I did not search if it can be downloaded manually or it has been renamed or merged into another plugin or it has been deprecated/discontinued.

* Jai and Jai_ImageIO might be different from previous version.  The previous version (https://github.com/bird-house/birdhouse-deploy/tree/c0ffb413a3dff70bbe2c98c38690d6e919f11386/birdhouse/docker/geoserver/resources) we added them manually and there is a "native" component.  The new GeoServer seems to have Jai enabled by default (probably comes with Java 11) but without the "native" component, see https://docs.geoserver.org/stable/en/user/configuration/image_processing/index.html.   I have no idea what is the performance impact of this change.

Running on my dev VM https://lvupavicsmaster.ouranos.ca/geoserver/.  Contact me privately for the admin password if you want to test it out.  It's completely empty as I do not know how to load data into it.

I will also need your help to test it as I have no idea how it is being used.


tlvu:
https://daccs-jenkins.crim.ca/job/DACCS-iac/254/console
```
[CHECK] Waiting for canarie-api [https://host-140-91.rdext.crim.ca/canarie]...
.............................................................................................................................................................................................................................
[WARNING] Timeout of [1200] seconds reached. Exiting, but leaving the instance running.
```
@MatProv is the timeout too short if a new docker image needs to be downloaded?

huard (approved):
How do we connect the database from the previous GeoServer to this one ? Can we migrate the database ? 
Running the nb test suite will hit GeoServer.

MatProv:
> @MatProv is the timeout too short if a new docker image needs to be downloaded?

@tlvu 1200 seconds is long enough for new images too be downloaded, if any.
In this case, error occurs because of the new geoserver env variables added in this PR that are not yet defined in our test config.

huard:
CSW: https://docs.geoserver.org/latest/en/user/community/csw-iso/installing.html

@tlogan2000  As far as I know, we're not using directly using the CSW service at the moment. It could maybe be used by other users working with GIS client, but I'm not aware of it. My suggestion would be to open a new issue for this and prioritize it if we get user complaints.

tlvu:
> > @MatProv is the timeout too short if a new docker image needs to be downloaded?
> 
> @tlvu 1200 seconds is long enough for new images too be downloaded, if any.
> In this case, error occurs because of the new geoserver env variables added in this PR that are not yet defined in our test config.

@MatProv 
Ah yes indeed my bad, it was late last night, sorry!  I even wrote it down clearly in the commit message of 187e0f30171da09dee71eee01c4c7a492d330233 that the change is backward incompat and `env.local` will need a manual update.

Can you add the new config to `env.local`?  There no harm to have the new config without this change since unused config will just be ignored.

MatProv:
run tests

tlvu:
> How do we connect the database from the previous GeoServer to this one ? Can we migrate the database ?

I was not able to test data migration from previous version since my dev Geoserver is empty !  Will need help with how to load data and some data to load.

> Running the nb test suite will hit GeoServer.

Geoserver is one of the few components not tested by our notebooks on Jenkins currently, see https://github.com/Ouranosinc/pavics-sdi/issues/183

birdhouse/docker-compose.yml
```diff
@@ -101,16 +101,21 @@ services:
...
   geoserver:
-    image: pavics/geoserver:2.9.3
+    image: ${GEOSERVER_IMAGE}
     container_name: geoserver
     ports:
       - "8087:8080"
     environment:
       # for custom entrypoint
       PAVICS_FQDN_PUBLIC: $PAVICS_FQDN_PUBLIC
+      STABLE_EXTENSIONS: ${GEOSERVER_STABLE_EXTENSIONS}
```
MatProv, line 111: Maybe use a constant style of passing env variables. Eg: all with `${}` or not.

huard:
@tlvu Got it. There are a few tests in Raven and RavenPy that connect to GeoServer, so we would give those tests another URL to use. I suggest we run those tests manually for now rather than trying to solve #183. 

@Zeitsperre Could you discuss with Long 
1. how you upload data currently on GeoServer
2. new possibilities with the REST API to upload data

birdhouse/docker-compose.yml
tlvu: Done in 0c74cf3c12d1edf172563fb95e0e03e248502af2, thanks.

MatProv:
> Can you add the new config to `env.local`? There no harm to have the new config without this change since unused config will just be ignored.

@tlvu Yes, now passing.

tlvu:
> CSW: https://docs.geoserver.org/latest/en/user/community/csw-iso/installing.html
> 
> @tlogan2000 As far as I know, we're not using directly using the CSW service at the moment. It could maybe be used by other users working with GIS client, but I'm not aware of it. My suggestion would be to open a new issue for this and prioritize it if we get user complaints.

I opened a PR upstream: https://github.com/kartoza/docker-geoserver/pull/230, if accepted CSW will be available in their next release.

tlvu:
@huard, @Zeitsperre 

I just added this to the PR description, something else to pay attention when testing this new version.

* No more manual install of various NetCDF system libraries (zlib, hdf5, archive), see our previous custom build https://github.com/bird-house/birdhouse-deploy/blob/c0ffb413a3dff70bbe2c98c38690d6e919f11386/birdhouse/docker/geoserver/Dockerfile#L26-L35
  * Since we can enable `netcdf-plugin` on the fly so I am guessing those system libraries are not needed anymore but I do not know the actual real impact of this change.

tlvu:
Ping @f-PLT, you might be interested in this PR.

tlvu:
> I opened a PR upstream: [kartoza/docker-geoserver#230](https://github.com/kartoza/docker-geoserver/pull/230), if accepted CSW will be available in their next release.

Even better, upstream owner clarified that `csw-plugin` was even already enabled by default, so that's why it was not in `stable_plugins.txt`.  I was following the README, which was not up-to-date with the code !

Zeitsperre:
> @Zeitsperre Could you discuss with Long
> 
>     1. how you upload data currently on GeoServer
In a word; manually.

Geopackage-stored datasets are converted to Shapefile or another OGC format (GML usually), placed on the data store via scp, and loaded into the GeoServer using the admin frontend. For data with strange/non-standard CRSes, their PROJ4 definitions need to be added to the listing of available transformations (see: https://docs.geoserver.org/latest/en/user/configuration/crshandling/customcrs.html) and the GeoServer needs to be restarted. After that, the dataset is loaded as a `datastore` and `layers` are published from the store. This can involve tweaks (setting domain, specifying character encoding, etc.). It's a pretty labour-intensive process.

>     2. new possibilities with the REST API to upload data

I stumbled upon a project that shows promise for being a Python REST API interface for GeoServer (https://github.com/gicait/geoserver-rest) and have been opening issues and suggesting changes to make it a bit more usable. It's not production-ready, but I would say it's usable for automating a few steps of the process (loading, publishing, styling, etc.). We are in dire need of a standardization for our many datasets and if something like GeoPackage support is available at the outset (via `geopkg-plugin`?), we could merge most of our vector data into a handful of files (I'm unsure if raster support is fixed for GeoPackage+GeoServer; the maintainers told me it was in the plans a year or so ago).

For immediate needs, I can try loading some bunk data into @tlvu's instance and seeing whether things work as expected. We really don't use most of the capabilities, so I don't envision us needing most of the plugins we had enabled (we have never had netcdf data on our GeoServer, to my knowledge).

tlvu:
Quick recap discussion with @Zeitsperre today.

We will need to load data at least twice, once in currently empty new GeoServer and again on old version of GeoServer then upgrade to the new one to test upgrade scenario.  Might need to load data more than twice if we screw up during our testing.

So given loading data is a manual and tedious process, @Zeitsperre will try to test the water with the automation tools he found.  It might only automate haft the steps but it's half the steps he won't have to perform manually repeatedly.

It's only to have a quick and dirty prototyping of the automated process, not to productize it.  But will give us an idea what we need, how we want, what is missing to productize it later.

f-PLT:
It's supposed to be possible to migrate the Geoserver configurations (`/data/geoserver`), but 2.9 => 2.19 didn't work in my case. I did a small local test with shapefile data and did get it working up to 2.17.2, in incremental steps, but I couldn't get passed 2.17.2, trying to move on to 2.18 and above, Geoserver either had internal errors or 404's when accessing it. I have yet to investigate, but I figured this information could be useful nonetheless. I did not test PostGIS version updates along with this however.

As for using the Geoserver API to automate data loading, geoserver-rest is a good start. I know of a few others, but I'll have to look at my notes. I started playing with this in January and also with custom API requests. Louis-David from CRIM has also used the API for the Canadian Climate Data Portal. Unless I'm mistaken, we will be exploring this kind of automation for one of our new components, but I don't have an ETA for it yet.

Zeitsperre:
@f-PLT

Thanks for the heads-up on the migration experiments. I still have yet to do my end of the tests, but it's interesting to see that the database can be migrated to a version as new as 2.17. For Postgres, I don't believe Ouranos has any data in an SQL DB, so I wouldn't worry about PostGIS (unless CRIM does). The feature I'm most curious about is GeoPackage performance, as that would clean up our /data drive quite a lot. I'll probably create a file and can share that with you if you'd like.

I've been contributing some tooling changes to `geoserver-rest` and was going to start fixing up a few clear bugs as well as the docs and tests for it as well. The maintainer is solo on the development, and it looks like the capabilities listed are based on his present needs, so we can probably extend the features we want; I'm sure the maintainer will be happy to integrate most contributions.

birdhouse/pavics-compose.sh
```diff
@@ -146,6 +147,11 @@ if [ x"$1" = x"up" ]; then
   docker volume create jupyterhub_data_persistence  # jupyterhub db and cookie secret
   docker volume create thredds_persistence  # logs, cache
 
+  if [ ! -d "$GEOSERVER_DATA_DIR" ]; then
```
dbyrns, line 150: So this directory cannot be created on the host before launching geoserver, else the permissions will not be good?
And given that we give the ownership to 1000:10001 from the container, we could prevent access directly from the host... Not saying it's a bad thing, maybe just for backup purposes it could be more complicated.

dbyrns (approved):
Very clean update. Hopefully the migration will not cause too much issues.

birdhouse/pavics-compose.sh
tlvu: The root cause is the new geoserver runs as user with id 1000:10001 so the data folder needs to have the same ownership, else geoserver can not write to it.  And given the geoserver container do not run as root, it can not switch the ownership itself.

Now, when docker comes up, if the dir do not already exist, it will create it with ownership 0:0.

So this code basically ensure the dir creation is not done by docker but before hand with proper ownership.

Hopefully do not affect backup since backup processes probably run as root to have read-only on everything (the other stuff under `/data/` mostly owned by root).  And I hope backup is a read-only and not write so should not accidentally create a new file with wrong ownership.

MatProv (approved):

huard:
Maybe I'm wrong, but this sounds more complicated than it should be. 

What about doing a manual backup of the old database, spinning the new server on medus, restoring from the backup, and running the jenkins with the new geoserver. If the nb pass, all good and we do the same for prod. If it fails, then I agree we need a more elaborate migration procedure.

https://docs.geoserver.org/latest/en/user/community/backuprestore/usagegui.html

tlvu:
> Maybe I'm wrong, but this sounds more complicated than it should be.
> 
> What about doing a manual backup of the old database, spinning the new server on medus, restoring from the backup, and running the jenkins with the new geoserver. If the nb pass, all good and we do the same for prod. If it fails, then I agree we need a more elaborate migration procedure.
> 
> https://docs.geoserver.org/latest/en/user/community/backuprestore/usagegui.html

@Zeitsperre @f-PLT have you tried in the past this manual method of data migration on upgrade?

Could be our way out if a transparent automated data migration do not work.

Zeitsperre:
@tlvu I haven't tried it yet (frankly had no idea that that functionality was available), but will be sure to check things out. This database likely relies on the locations of source files being exactly the same, which is less than ideal for our current data on the GeoServer prod server.

If we can merge our data into a few files, this would make it the process much easier, as we probably don't have to re-publish/re-configure every layer individually.

huard:
@Zeitsperre That's not my interpretation of how the backup works.

f-PLT:
Unless the documentation is not up to date, the `Backup and Restore` community module simply saves the configuration files only, not the data. (https://docs.geoserver.org/latest/en/user/community/backuprestore/usagegui.html#performing-a-full-backup-via-ui)

I haven't tried this module, but I did manually copy the configuration files to make my own backups. Looks like a useful tool, but I would not expect it to handle path changes for the data.

At best, it could be used to back up current configurations before trying to update Geoserver in case the update fails and caused breaking changes to the configurations.

For a test on a secondary instance, at least a partial copy of the `/data/`  folder would need to be copied along with the backup file so the new instance can find some of the layers. The missing layers shouldn't cause Geoserver to crash, in my experience it just says it can't find it when I try to access a missing layer.

Zeitsperre:
@f-PLT The configurations are the difficult-to-automate part, so even that would be a win. If we can also automate the backup process somehow, the rest is a cinch. I sense that a good approach for now might be:
- Identify the datasets/workspaces we want to migrate over.
- Merge the layers into GeoPackages named for the respective workspace.
  - Save the layer styling information (is this supported/desired? Affects QGIS and ArcGIS users.)
- Copy them over to `/data/workspace_name/` and import them (manually, for now).
- Generate a backup file and zip the data file structure.
- Rebuild on another instance from the backup and zip.

Thoughts?

f-PLT:
@Zeitsperre

Overall, I think it's a good approach. If you already have the Backup and Restore community module set up, it should be quick.

If not, or if there are problems setting it up, it could be simpler to copy the `/data/geoserver/` directory (on-host directory) while only cherry-picking the datasets/workspaces in the `/data/geoserver/data/` folder. If there are also datasets outside of this folder, it's going to be more work to sift through though.

As for geopackages, if the styles are saved in SLD format, it should be possible, at the very least, to import them in Geoserver. I don't know if Geoserver automatically reads internally saved styles for geopackages.

huard:
@tlvu @Zeitsperre 
Are you making progress on this ? 

To keep it simple, could we "not support the upgrade process"? That is, we spin up a new GeoServer, upload data manually, leave the previous server alone, and deprecate it in a few months. This way, we don't have to "productize" the upgrade for other users. 

I really need this done, so I'm willing to take shortcuts.

huard:
@tlvu Yes. 

@Zeitsperre @tlogan2000 
The plan would be (let me know if it makes sense): 
1. Install a recent version of geoserver in geoservernew
2. Test functionality across NB and co with geoservernew
3. Announce version upgrade for a fixed date. The recommendation would be to migrate services to use geoservernew before said date.
4. At said date, decommission old geoserver, and create redirect so that both geoserver and geoservernew point to new geoserver
5. Two months later, kill geoservernew.

tlvu:
> Sync'ing all the geoserver data from prod to my machine to first test "restore" data to current existing version before attempting upgrade.

Restore from prod data seems to work !!! https://lvupavics.ouranos.ca/geoserver/web/wicket/bookmarkable/org.geoserver.web.demo.MapPreviewPage?0  @Zeitsperre can you take a quick look that everything is there?

Tomorrow I'll try to upgrade on my VM.

huard:
Looks good here !

@Zeitsperre @tlogan2000 I think this switch will be a good occasion to complete the clean-up (remove test files), review the naming of layers, etc.

tlvu:
Bad news for today.  I've tried the upgrade and the new geoserver do not even start with the existing data !

So I've followed some false leads what I thought was the error in the logs today.

Given they are all false leads, I suspect these errors were possibly in the previous version as well so they are not the real root cause.  Will have to restore to the old version and compare the logs so I can eliminate "existing not harmful errors".

tlvu:
I've opened an issue https://github.com/kartoza/docker-geoserver/issues/250 for the upstream GeoServer docker build, hopefully they can share some leads.

@f-PLT you've said you had problem upgrading GeoServer yourself, did you get the same errors I had in the issue above or something else?

I am now stuck at the upgrade path.  I think we should seriously consider finding a way to rebuild/re-import our data from scratch on the new version, @Zeitsperre will need your help for this.

f-PLT:
In my case, it was really all about the user:group permissions. Once that was done, everything worked well.

What I did was : 
- Start a compose with geoserver 2.9.4
- Load some data into it.
- `docker-compose down`, then `chown` on the data folder (GEOSERVER_DATA_DIR)
- Change the docker-compose.yml for geoserver 2.19.0 with .env file and variables
- `up` the compose again

However, today was the first time I was able to go past 2.18.2 and migrate to 2.19. Wanted to make sure I had the right configurations, so tried without the data_dir mounted, and it still didn't work, still got an error in the logs saying `Context [/geoserver] startup failed due to previous errors`

Turns out I think something was wrong with my geoserver image, as simply pulling a new copy solved everything on my end. Did a new test, as I described above and everything worked well. One thing I spotted; before pulling the new image, my geoserver's tomcat's version was at 10, while now it's at version 9.

If this ends up being not helpful, maybe trying to migrate to 2.18.2 first?

tlvu:
> One thing I spotted; before pulling the new image, my geoserver's tomcat's version was at 10, while now it's at version 9.

Looks like you hit this problem https://github.com/kartoza/docker-geoserver/issues/234#issuecomment-811561207

So looks like something in our data is breaking the new Geoserver docker image !  Very strange.

tlvu:
I got Geoserver 2.19.0 upgraded properly with our existing data !!! https://lvupavics.ouranos.ca/geoserver/web/wicket/bookmarkable/org.geoserver.web.demo.MapPreviewPage?1&filter=false

I was following the advice here https://github.com/kartoza/docker-geoserver/issues/250#issuecomment-845417632 to gradually delete stuff in our existing data to see which one is causing Geoserver to fail at starting up.  So I wipe everything gradually: workspaces, stores, layers, styles until I got to the extra admin users.  It is all the extra admin users that broke the start-up !  Who knows !

So @huard can you check if this version of Geoserver can do what you wanted to do for your project?

@Zeitsperre can you smoke test the new version is still behaving like the old version for the features that we care about?

huard:
Bien joué !

So I'm able to upload my layers using 
```
curl -v <user:pass> -XPUT -H "Content-type: application/zip", --data-binary @decamillenial_flood_CC.gpkg  https://lvupavics.ouranos.ca/geoserver/rest/workspaces/public/datastores/decamillenial_flood_CC/decamillenial_flood_CC.gpkg
```
But geoserver still truncates the database column names to 10 characters when I try to open the WFS served by geoserver from QGIS. 
So I'll have to dig deeper into this, but at least GeoServer recognizes gpkg flies and can server them, so that's a big step forward.

tlvu:
Problem found by @Zeitsperre with this new Geoserver https://github.com/Ouranosinc/pavics-sdi/issues/220

huard:
Ok. So let's put this on hold until we get solutions in. On my end, it's not a slam dunk either.

Zeitsperre:
@tlvu Everything on the Raven side that was blocking this has now been addressed. I think we can move forward with the upgrade!

tlvu:
> @tlvu Everything on the Raven side that was blocking this has now been addressed. I think we can move forward with the upgrade!

@Zeitsperre cool, let's deploy new Raven first, then this one can follow.

PR #143, opened 2021-04-08
magpie 3.x + gunicorn bind

fmigneault:
~should be tested after #151 for iterative updates~
Update directly to latest Magpie (3.8 -> ... -> 3.12)

Updates magpie 3.12 (see [CHANGES](https://github.com/Ouranosinc/Magpie/blob/master/CHANGES.rst#3120-2021-05-11) for details)

Changes: 
- Applies relevant modifications to configuration following `gunicorn` update in Magpie
  (docker image uses `pserve` CLI instead of directly calling `gunicorn`, which requires a small adjustment of `host`/`bind` value in the [magpie.ini](https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/config/magpie/magpie.ini.template) config)
- Provides improved webhook configuration and implementations (for eventual [cowbird](https://github.com/Ouranosinc/cowbird))
- Providers headers `WWW-Authenticate` and `Location-When-Unauthenticated` when HTTP Unauthorized [401] is returned, which can help browsers/clients/users find the authentication endpoint to obtain needed access to the targeted resource.
- Includes some accumulated bugfixes
- use a shared `MAGPIE_VERSION` variable from `default.env` to align Magpie API, Magpie CLI, and Twitcher Magpie Adapter references
- Replace `authtest` user permissions in `optional-components/secure-thredds/secure-access-magpie-permission.cfg` to instead use new `thredds-secure` group, which `authtest` user will be a member of (resolves user-creation vs permission creation timing issues to make tests pass)

### DIFF
--- a/birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
+++ b/birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
@@ -36,9 +36,12 @@ permissions:
     group: anonymous
     action: create
 
-  # following permissions provide access to 'secure' directory to any user that is member of 'thredds-secure' group
-  # this access will take priority over 'anonymous' group above that gets denied public access to 'secure' directory
-  # users that must be granted access to the data under 'secure' should be added to the group and will inherit access
+  # Following permissions provide access to 'secure' directory nested under 'thredds' service to any user that is member
+  # of 'thredds-secure-authtest-group'. This permission will take priority over 'anonymous' group above that gets denied
+  # public access to 'secure' directory. Users that must be granted access to data under 'secure' should be added
+  # to the group to inherit permissions.
+  # The group is created dynamically if it doesn't already exist. User 'authtest' from 'create-magpie-authtest-user'
+  # script when bootstrapping the instance for testing purpose.
 
   - service: thredds
     resource: /birdhouse/testdata/secure
@@ -47,7 +50,7 @@ permissions:
       name: browse
       access: allow
       scope: recursive
-    group: thredds-secure
+    group: thredds-secure-authtest-group
     action: create
 
   - service: thredds
@@ -57,7 +60,7 @@ permissions:
       name: read
       access: allow
       scope: recursive
-    group: thredds-secure
+    group: thredds-secure-authtest-group
     action: create
 
   - service: thredds
@@ -67,5 +70,5 @@ permissions:
       name: write
       access: allow
       scope: recursive
-    group: thredds-secure
+    group: thredds-secure-authtest-group
     action: create
--- a/birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
+++ b/birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
@@ -1,9 +1,3 @@
-users:
-  - username: authtest
-    password: authtest1234
-    email: authtest@example.com
-    group: anonymous
-
 permissions:
   # note:
   #   following permissions can be combined with others such as 'optional-components/all-public-access'
@@ -42,7 +36,10 @@ permissions:
     group: anonymous
     action: create
 
-  # preserve access for test-suite user
+  # following permissions provide access to 'secure' directory to any user that is member of 'thredds-secure' group
+  # this access will take priority over 'anonymous' group above that gets denied public access to 'secure' directory
+  # users that must be granted access to the data under 'secure' should be added to the group and will inherit access
+
   - service: thredds
     resource: /birdhouse/testdata/secure
     type: directory
@@ -50,7 +47,7 @@ permissions:
       name: browse
       access: allow
       scope: recursive
-    user: authtest
+    group: thredds-secure
     action: create
 
   - service: thredds
@@ -60,7 +57,7 @@ permissions:
       name: read
       access: allow
       scope: recursive
-    user: authtest
+    group: thredds-secure
     action: create
 
   - service: thredds
@@ -70,5 +67,5 @@ permissions:
       name: write
       access: allow
       scope: recursive
-    user: authtest
+    group: thredds-secure
     action: create
--- a/birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
+++ b/birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
@@ -1,3 +1,9 @@
+users:
+  - username: authtest
+    password: authtest1234
+    email: authtest@example.com
+    group: anonymous
+
 permissions:
   # note:
   #   following permissions can be combined with others such as 'optional-components/all-public-access'
--- a/birdhouse/scripts/create-magpie-users
+++ b/birdhouse/scripts/create-magpie-users
@@ -69,6 +69,9 @@ THIS_FILE="`realpath "$0"`"
 THIS_DIR="`dirname "$THIS_FILE"`"
 COMPOSE_DIR="`dirname "$THIS_DIR"`"
 
+# must retrieve MAGPIE_VERSION
+. "${COMPOSE_DIR}/default.env"
+
 if [ -f "$COMPOSE_DIR/env.local" ]; then
     # Get PAVICS_FQDN, MAGPIE_ADMIN_PASSWORD, MAGPIE_ADMIN_USERNAME
     . $COMPOSE_DIR/env.local
@@ -98,7 +101,7 @@ if [ -z "$MAGPIE_CLI_CONF" ]; then
 fi
 
 if [ -z "$MAGPIE_CLI_IMAGE" ]; then
-    MAGPIE_CLI_IMAGE="pavics/magpie:3.12.0"
+    MAGPIE_CLI_IMAGE="pavics/magpie:${MAGPIE_VERSION}"
 fi
 
 # End configurable config via env var.
--- a/birdhouse/scripts/create-magpie-users
+++ b/birdhouse/scripts/create-magpie-users
@@ -65,16 +65,21 @@
 # bogus03      bvNWVWCQi8M6     409 : User name matches an already existing user name.
 #
 
+RED=$(tput setaf 1)
+NORMAL=$(tput sgr0)
+
 THIS_FILE="`realpath "$0"`"
 THIS_DIR="`dirname "$THIS_FILE"`"
 COMPOSE_DIR="`dirname "$THIS_DIR"`"
 
-# must retrieve MAGPIE_VERSION
-. "${COMPOSE_DIR}/default.env"
+# must retrieve MAGPIE_VERSION from 'default.env' or 'env.local'
+if [ -f "$COMPOSE_DIR/default.env" ]; then
+    . "${COMPOSE_DIR}/default.env"
+fi
 
 if [ -f "$COMPOSE_DIR/env.local" ]; then
     # Get PAVICS_FQDN, MAGPIE_ADMIN_PASSWORD, MAGPIE_ADMIN_USERNAME
-    . $COMPOSE_DIR/env.local
+    . "${COMPOSE_DIR}/env.local"
 fi
 
 ##############################################################################
@@ -101,6 +106,11 @@ if [ -z "$MAGPIE_CLI_CONF" ]; then
 fi
 
 if [ -z "$MAGPIE_CLI_IMAGE" ]; then
+    # MAGPIE_VERSION must be provided by 'default.env', 'env.local' or directly
+    if [ -z "${MAGPIE_VERSION}" ]; then
+        echo "${RED}Error${NORMAL}: Required MAGPIE_VERSION is undefined or empty."
+        exit 1
+    fi
     MAGPIE_CLI_IMAGE="pavics/magpie:${MAGPIE_VERSION}"
 fi
 
--- a/birdhouse/scripts/create-magpie-users
+++ b/birdhouse/scripts/create-magpie-users
@@ -98,7 +98,7 @@ if [ -z "$MAGPIE_CLI_CONF" ]; then
 fi
 
 if [ -z "$MAGPIE_CLI_IMAGE" ]; then
-    MAGPIE_CLI_IMAGE="pavics/magpie:3.8.0"
+    MAGPIE_CLI_IMAGE="pavics/magpie:3.12.0"
 fi
 
 # End configurable config via env var.
--- a/birdhouse/README.rst
+++ b/birdhouse/README.rst
@@ -140,13 +140,14 @@ For that test suite to pass, run the script
 `scripts/bootstrap-instance-for-testsuite <scripts/bootstrap-instance-for-testsuite>`_ (:download:`download </birdhouse/scripts/bootstrap-instance-for-testsuite>`)
 to prepare your new instance.  Further documentation inside the script.
 
-Optional component
+Optional components
 `all-public-access <./optional-components#give-public-access-to-all-resources-for-testing-purposes>`_
-also need to be enabled in ``env.local``.
+and `secure-thredds <./optional-components/#control-secured-access-to-resources-example>`_
+also need to be enabled in ``env.local`` using ``EXTRA_CONF_DIRS`` variable.
 
 ESGF login is also needed for
 https://github.com/Ouranosinc/pavics-sdi/blob/master/docs/source/notebooks/esgf-dap.ipynb
-part of test suite.  ESGF credentails can be given to Jenkins via
+part of test suite.  ESGF credentials can be given to Jenkins via
 https://github.com/Ouranosinc/jenkins-config/blob/aafaf6c33ea60faede2a32850604c07c901189e8/env.local.example#L11-L13
 
 The canarie monitoring link
--- a/birdhouse/config/magpie/magpie.ini.template
+++ b/birdhouse/config/magpie/magpie.ini.template
@@ -64,7 +64,7 @@ prefix = /magpie
 
 [server:main]
 use = egg:gunicorn#main
-host = localhost
+host = 0.0.0.0
 port=2001
 timeout=10
 workers=3
--- a/birdhouse/default.env
+++ b/birdhouse/default.env
@@ -13,6 +13,9 @@ export FINCH_IMAGE="birdhouse/finch:version-0.7.4"
 
 export THREDDS_IMAGE="unidata/thredds-docker:4.6.15"
 
+# Tag version that will be used to update Magpie API, Magpie CLI, and matching Twitcher with Magpie Adapter
+export MAGPIE_VERSION=3.12.0
+
 # Root directory under which all data persistence should be nested under
 export DATA_PERSIST_ROOT="/data"
 
--- a/birdhouse/docker-compose.yml
+++ b/birdhouse/docker-compose.yml
@@ -310,7 +310,7 @@ services:
     logging: *default-logging
 
   magpie:
-    image: pavics/magpie:3.10.0
+    image: pavics/magpie:3.12.0
     container_name: magpie
     ports:
       - "2001:2001"
@@ -338,7 +338,7 @@ services:
     logging: *default-logging
 
   twitcher:
-    image: pavics/twitcher:magpie-3.10.0
+    image: pavics/twitcher:magpie-3.12.0
     container_name: twitcher
     ports:
       - "8000:8000"
--- a/birdhouse/docker-compose.yml
+++ b/birdhouse/docker-compose.yml
@@ -310,7 +310,7 @@ services:
     logging: *default-logging
 
   magpie:
-    image: pavics/magpie:3.8.0
+    image: pavics/magpie:3.10.0
     container_name: magpie
     ports:
       - "2001:2001"
@@ -338,7 +338,7 @@ services:
     logging: *default-logging
 
   twitcher:
-    image: pavics/twitcher:magpie-3.8.0
+    image: pavics/twitcher:magpie-3.10.0
     container_name: twitcher
     ports:
       - "8000:8000"
--- a/birdhouse/docker-compose.yml
+++ b/birdhouse/docker-compose.yml
@@ -310,7 +310,7 @@ services:
     logging: *default-logging
 
   magpie:
-    image: pavics/magpie:3.12.0
+    image: pavics/magpie:${MAGPIE_VERSION}
     container_name: magpie
     ports:
       - "2001:2001"
@@ -338,7 +338,7 @@ services:
     logging: *default-logging
 
   twitcher:
-    image: pavics/twitcher:magpie-3.12.0
+    image: pavics/twitcher:magpie-${MAGPIE_VERSION}
     container_name: twitcher
     ports:
       - "8000:8000"
--- a/birdhouse/scripts/create-magpie-authtest-user
+++ b/birdhouse/scripts/create-magpie-authtest-user
@@ -10,7 +10,7 @@ THIS_DIR="`dirname "$THIS_FILE"`"
 
 TMP_CONFIG_FILE="/tmp/create-magpie-authtest-user.yml"
 
-# Test user is added to 'thredds-secure' group to obtain access to 'secure' directory under 'thredds' service.
+# Test user is added to 'thredds-secure-authtest-group' to obtain access to 'secure' directory under 'thredds' service.
 # Those permissions are defined in 'optional-components/secure-thredds/secure-access-magpie-permission.cfg'
 # which should also be included in 'EXTRA_CONF_DIRS' of your custom 'env.local' file.
 # This user is also automatically added to 'anonymous' group without the need of explicit membership here.
@@ -19,7 +19,7 @@ users:
    - username: authtest
      password: authtest1234
      email: authtest@example.com
-     group: thredds-secure
+     group: thredds-secure-authtest-group
 __OEF__
 
 set -x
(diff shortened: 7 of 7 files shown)


MatProv:
@fmigneault Maybe also update the `MAGPIE_CLI_IMAGE` in `scripts/create-magpie-users`, for consistency.

MatProv (approved):
lgtm, only look at the comment I left before merging

birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
```diff
@@ -1,3 +1,9 @@
+users:
+  - username: authtest
+    password: authtest1234
+    email: authtest@example.com
+    group: anonymous
```
tlvu, line 5: We already create this same test user here https://github.com/bird-house/birdhouse-deploy/blob/edd8ab40f783b6b724b88c1abd5c2ac1d721f8dc/birdhouse/scripts/create-magpie-authtest-user#L13-L19, so why repeat it here again?

Also the component is called `secure-thredds`, so in this spirit, adding a publicly known user and password is not so "secure".

Granted the same publicly known user and password is in the `create-magpie-authtest-user` but at least the script says test user in there and do not imply anything "secure".

birdhouse/scripts/create-magpie-users
```diff
@@ -98,7 +98,7 @@ if [ -z "$MAGPIE_CLI_CONF" ]; then
 fi
 
 if [ -z "$MAGPIE_CLI_IMAGE" ]; then
-    MAGPIE_CLI_IMAGE="pavics/magpie:3.8.0"
+    MAGPIE_CLI_IMAGE="pavics/magpie:3.12.0"
```
tlvu, line 101: I would suggest we move the magpie image into `default.env` and make `create-magpie-users` source `default.env`, just like it already source `env.local` https://github.com/bird-house/birdhouse-deploy/blob/1719c68a8d7dafdeb88b035dc7a9bccc72c48c57/birdhouse/scripts/create-magpie-users#L72-L75

This way there is only 1 place to update the magpie image next time.

I assume we will be updating magpie a lot more frequent so we will all benefit from avoiding this duplication.

tlvu (changes requested):
Minor change requested, otherwise all good.

Thanks for the detailed list of changes in the PR description.  Please copy/paste into merge description when merge.

From that list of change, I do not see any DB change so do I need to perform a manual DB upgrade using prod DB?

tlvu:
Oh, by the way, subset.ipynb still fail https://daccs-jenkins.crim.ca/job/PAVICS-e2e-workflow-tests/job/master/413/console

And shouldn't the IAC report the test result directly here?

birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
fmigneault: I added this by request of @MatProv to fix tests that seemed to have mismatches between the created user. 
He should be able to detail more about the problems he faced.

I do agree that adding it in the file doesn't make it "secure", but the user was already referenced in other permissions before, so it wasn't secure in the first place anyway. 

The best approach would be instead to create some kind of "thredds-secure" group, to which "authtest" user is applied membership when created via [create-magpie-authtest-user](https://github.com/bird-house/birdhouse-deploy/blob/edd8ab40f783b6b724b88c1abd5c2ac1d721f8dc/birdhouse/scripts/create-magpie-authtest-user#L13-L19) instead of "anonymous". That "thredds-secure" group would replace all the `user: authtest` in that config, so "authtest" user would inherit the group permissions. Not adding the creation of "authtest" user would still make the "thredds-secure" exist, but no user would be part of it. That would also facilitate future users we want to grant access to `secure` directory, by simply adding them to that group instead of duplicating all the permissions for each one.

birdhouse/scripts/create-magpie-users
fmigneault: Good idea.

fmigneault:
> And shouldn't the IAC report the test result directly here?
@tlvu 

@MatProv told me there was an update he pushed that seemed to have broken the script. Probably the execution stopped before it could reach the reporting phase to the PR?

MatProv:
> > And shouldn't the IAC report the test result directly here?
> > @tlvu
> 
> @MatProv told me there was an update he pushed that seemed to have broken the script. Probably the execution stopped before it could reach the reporting phase to the PR?

@fmigneault Pipeline has been repaired this morning, that's not the reason why result aren't here.
@tlvu It's because currently it reports on GH to pr only when build started via jenkins, not via the hook. It'll support both soon.

MatProv:
> Oh, by the way, subset.ipynb still fail 

@tlvu Yes, seems to be notebook specific, not magpie anymore. The latest upgrade of magpie fixed the pavics_thredds.ipynb auth issue, tho.

birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
MatProv: @tlvu  So the issue we had was that the permissions were attributed to the authtest user before it has been created. The result was that magpie created a default user named authtest with a dummy password, added the permissions and trying to use authtest in the notebook failed since the username/password we specify later on in `create-magpie-authtest-user` wasn't used. Now, specifying the password directly in the cfg makes sure the password isn't a dummy one. Now the failed 401 test (`pavics_thredds.ipynb`) passes.

The `subset.ipynb` is still failing but i'm pretty sure it's unrelated to magpie.

tlvu: >  permissions were attributed to the authtest user before it has been created.

@MatProv Ah OK I understand now.  

@fmigneault I agree if we have the `authtest-group` instead, then the user creation won't happen so then the `create-magpie-authtest-user` script will be able to create the user with the expected password and we won't have to hardcode the user creation here.

The `create-magpie-authtest-user` will have to create that `authtest` user part of group `authtest-group` instead of group `anonymous` like right now.

Can you guys do this?  I am guessing just replace `user: authtest` by `group: authtest-group` in this file `secure-access-magpie-permission.cfg` and update the group in `create-magpie-authtest-user`?

tlvu:
> > Oh, by the way, subset.ipynb still fail
> 
> @tlvu Yes, seems to be notebook specific, not magpie anymore. The latest upgrade of magpie fixed the pavics_thredds.ipynb auth issue, tho.

@MatProv thanks for the update.  So not for this PR but we'll have to investigate and fix this issue so the pipeline do not report false failure.  Just FYI, that notebook do not fail on our side at the moment of this writing.

MatProv:
> @MatProv thanks for the update. So not for this PR but we'll have to investigate and fix this issue so the pipeline do not report false failure. Just FYI, that notebook do not fail on our side at the moment of this writing.

@tlvu Of course, i'm currently looking at this on our instances. Seems to be related with the nbval of the notebook, but strange thing is that errors only shows up during pavics-e2e-tests pipeline and not during manual execution of notebook cells on iac instances themselves. I still don't know the exact cause for the moment, but it's 100% unrelated to this current PR.

birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
fmigneault: @tlvu  
Sure! I will push the update with the group. 
I will call the group `thredds-secure` because there is no reason to make it specific to `authtest` (the `secure-access-magpie-permission.cfg` shouldn't have any knowledge of `authtest` on its own).

fmigneault: All done.

MatProv (approved):
All good, only need to make sure the `pavics_thredds.ipynb` still pass after the thredds-secure group change.

fmigneault:
@MatProv @tlvu 
Not sure why there still is not an auto-comment update of the test suite, but here is the result for 
198e1ce 

http://daccs-jenkins.crim.ca/job/DACCS-iac/511
http://daccs-jenkins.crim.ca/job/PAVICS-e2e-workflow-tests/job/master/417

```
12:07:07  notebooks/hummingbird.ipynb ............                                 [  8%]
12:07:16  notebooks-auth/test_thredds.ipynb ..........                             [ 15%]
12:07:23  pavics-sdi-master/docs/source/notebooks/WCS_example.ipynb .......        [ 20%]
12:07:31  pavics-sdi-master/docs/source/notebooks/WFS_example.ipynb ......         [ 24%]
12:07:38  pavics-sdi-master/docs/source/notebooks/WMS_example.ipynb ........       [ 30%]
12:07:40  pavics-sdi-master/docs/source/notebooks/WPS_example.ipynb ..........     [ 37%]
12:09:32  pavics-sdi-master/docs/source/notebooks/esgf-dap.ipynb F                 [ 37%]
12:09:32  pavics-sdi-master/docs/source/notebooks/jupyter_extensions.ipynb .       [ 38%]
12:09:35  pavics-sdi-master/docs/source/notebooks/opendap.ipynb .......            [ 43%]
12:09:41  pavics-sdi-master/docs/source/notebooks/pavics_thredds.ipynb .....       [ 46%]
12:12:21  pavics-sdi-master/docs/source/notebooks/regridding.ipynb ............... [ 57%]
12:13:16  ...............                                                          [ 67%]
12:13:28  pavics-sdi-master/docs/source/notebooks/rendering.ipynb ....             [ 70%]
12:13:35  pavics-sdi-master/docs/source/notebooks/subsetting.ipynb ......          [ 74%]
12:13:46  finch-master/docs/source/notebooks/dap_subset.ipynb ..........           [ 81%]
12:13:53  finch-master/docs/source/notebooks/finch-usage.ipynb ......              [ 86%]
12:14:21  finch-master/docs/source/notebooks/subset.ipynb ...FF....FFFFF....FF     [100%]
```

MatProv:
> Not sure why there still is not an auto-comment update of the test suite, but here is the result for

@fmigneault It's because currently it reports on GH to pr only when build started via jenkins, not via the hook. It'll support both soon.

birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
```diff
@@ -36,15 +36,18 @@ permissions:
...
+
   - service: thredds
     resource: /birdhouse/testdata/secure
     type: directory
     permission:
       name: browse
       access: allow
       scope: recursive
-    user: authtest
+    group: thredds-secure
```
tlvu, line 50: Sorry for nickpicing but `thredds-secure` is potentially a very generic name.  It would be much clearer that it is used for testing if the group is called `authtest-group`.

tlvu (changes requested):
Minor change requested, else all good for me as well.

Do I need to perform a manual DB migration test with our prod DB?  If yes, let me do the merge, else you can do the merge yourself.  Please copy/paste the excellent content of the PR description into the merge description.

birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
tlvu: > I will call the group `thredds-secure` because there is no reason to make it specific to `authtest` (the `secure-access-magpie-permission.cfg` shouldn't have any knowledge of `authtest` on its own).

@fmigneault oops missed this reply.  With the explicit naming, I just want to prevent users to re-use this group for other purposes than testing, else the `authtest` user will have access to more than it should and that would be a security leak.

birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
fmigneault: This is why I believe it should in fact be the generic name and not be `authtest-group`. 
There is nothing related to `authtest` specifically in that config, it just so happens that `authtest` user uses it because it wants to validate the same permission access this group provides. 
I would rather have a real group `thredds-secure`, that we could actually use to setup real users, that is also validated against the same user membership. We test the use case 1:1.

fmigneault:
@tlvu 
No db migration since 3.8. 
Only change outside of the code itself is the Docker command
(`gunicorn` -> `pserve` and corresponding `host = localhost` -> `host = 0.0.0.0` it requires added in this PR).

birdhouse/optional-components/secure-thredds/secure-access-magpie-permission.cfg
tlvu: > I would rather have a real group `thredds-secure`, that we could actually use to setup real users

This is exactly what I want to avoid.  This is a generic repo, we should not be implementing security for all the org deploying this repo.  And I would not want "real user" be part of this group.  Otherwise the `authtest` user will start to have more permissions as a "real user" would, which will be bad.

We are replacing the `authtest` user with a group with the exact same intention here.

fmigneault: What makes it not generic? 

If you enable *only* component `secure-thredds`, you receive a set of very specific permissions on Thredds service, and with it the corresponding group to give users membership *if you want to*. Using `authtest-group` there means you would have a random group named `authtest-group` when you are not even using `authtest` user for tests.

If instead you have the `authtest` user creation script, it is because you intend to test features, which includes testing the behaviour provided by `secure-thredds`. It makes perfect sense that the user used to test that security feature joins the relevant group created for that purpose. This is the same as any other "public" users joining the `anonymous` group to evaluate access to publicly accessible services.

Using the same group also ensures that we test everything the same way as the component defines it, regardless of who `authtest` actually is. I could call it `TheLegend27` for all we care, but it would still inherit the correct permissions to evaluate `secure-thredds` feature if he is member of `thredds-secure` group.

tlvu: I think we all agree this component is meant for testing.

So just like naming classes or variable names in a program, we want to have a descriptive name that reflect the actual role, which is for testing in the case of this `authtest-group`.  Matching the group name with the `authtest` user name will better relate each other.

On the production point of view, activating a component is a done once and forget thing.  For other orgs not heavily involved in the development like us, I think one year later, they might not remember what the group name `thredds-secure` was meant for so having a more descriptive name will help.

Or other orgs might actually want to use the fairly generic group name `thredds-secure` for something real, not for testing, thus I suggested `authtest-group` which will much less likely to cause name conflict in the future.

If we are both so uncomfortable with creating a group, then may I suggest another solution: perform a user update right after a user create so we ensure user `authtest` will always have the correct password and no need to create this group anymore.  This group was really just a work-around to ensure the user `authtest` have the correct password.  We do not actually need to have this group.

fmigneault: In the same way as you describe to be precise about the role, `authtest-group` doesn't indicate immediately what access gets granted when you are a member of it. You still need to know the specific components enabled to know its application.
So I would suggest to go all out with `thredds-secure-authtest-group` since it is so specific for that situation.

I think the group is less trouble than the user create/update handling.  
It also allows to extend tests with more users later on if we need to.

tlvu: Sold for `thredds-secure-authtest-group` thanks.

MatProv:
run tests

birdhouse/scripts/create-magpie-users
```diff
@@ -69,6 +69,9 @@ THIS_FILE="`realpath "$0"`"
 THIS_DIR="`dirname "$THIS_FILE"`"
 COMPOSE_DIR="`dirname "$THIS_DIR"`"
 
+# must retrieve MAGPIE_VERSION
+. "${COMPOSE_DIR}/default.env"
```
tlvu, line 73: Surround this with `if [ -f "${COMPOSE_DIR}/default.env" ]`, like `env.local`, so this script retain it's standalone feature.

tlvu (changes requested):
Minor change requested, plus the group name update.  Otherwise looking good now.

Oh wait, should also update the optional-componnents README.rst now that we also create an additional group.

tlvu:
@MatProv ohh nice reporting format update!  1 comment with all results is cleaner than 2 separate comments!  Nice to also post the `PAVICS_HOST`, useful when `DESTROY_INFRA_ON_EXIT : false`.

MatProv:
> @MatProv ohh nice reporting format update! 1 comment with all results is cleaner than 2 separate comments! Nice to also post the `PAVICS_HOST`, useful when `DESTROY_INFRA_ON_EXIT : false`.

@tlvu It's also now hooked to the GitHub PR, so now commit push and "run tests" will report results on the PR.

birdhouse/scripts/create-magpie-users
fmigneault: Shouldn't this file be enforced to make sure the variables exist?
It is this kind of situation that justifies why I prefer to add the "required" variables (eg Tag version of docker images her) in `VARS` as in other PRs, although they *should* always exist from loading `defaults.env`. 
Not having those variables would break the whole server deployment and will not necessarily be easy to debug opposed to a clear message "undefined variable X".

fmigneault:
**this is a fake result to test output format** 

<h1>E2E Test Results</h1>

<h2>DACCS-iac Pipeline Results</h2>
<b>Build URL</b> : <a href=http://daccs-jenkins.crim.ca:80/job/DACCS-iac/522/>http://daccs-jenkins.crim.ca:80/job/DACCS-iac/522/</a> <br>
<b>Result</b> : failure <br>
<br>
<b>BIRDHOUSE_DEPLOY_BRANCH</b> : magpie-gunicorn-bind <br>
<b>DACCS_CONFIGS_BRANCH</b> : master <br>
<b>PAVICS_E2E_WORKFLOW_TESTS_BRANCH</b> : master <br>
<b>PAVICS_SDI_BRANCH</b> : master <br>
<br>
<b>DESTROY_INFRA_ON_EXIT</b> : true <br>
<b>PAVICS_HOST</b> : <a href=https://host-140-69.rdext.crim.ca>https://host-140-69.rdext.crim.ca</a> <br>


<h2>PAVICS-e2e-workflow-tests Pipeline Results</h2>
<b>Tests URL</b> : <a href=http://daccs-jenkins.crim.ca:80/job/PAVICS-e2e-workflow-tests/job/master/439/>http://daccs-jenkins.crim.ca:80/job/PAVICS-e2e-workflow-tests/job/master/439/</a> <br>
<br>
<details>
<pre>
<code>
[2021-05-17T14:32:12.406Z] ============================= test session starts ==============================
[2021-05-17T14:32:12.406Z] platform linux -- Python 3.7.10, pytest-6.2.3, py-1.10.0, pluggy-0.13.1
[2021-05-17T14:32:12.407Z] rootdir: /home/jenkins/agent/workspace/PAVICS-e2e-workflow-tests_master
[2021-05-17T14:32:12.407Z] plugins: anyio-2.2.0, dash-1.20.0, nbval-0.9.6, tornasync-0.6.0.post2
[2021-05-17T14:32:12.407Z] collected 143 items
[2021-05-17T14:32:12.407Z]
[2021-05-17T14:32:14.458Z] notebooks/hummingbird.ipynb ............                                 [  8%]
[2021-05-17T14:32:22.905Z] notebooks-auth/test_thredds.ipynb ..........                             [ 15%]
[2021-05-17T14:32:51.035Z] pavics-sdi-master/docs/source/notebooks/WCS_example.ipynb .......        [ 20%]
[2021-05-17T14:32:58.096Z] pavics-sdi-master/docs/source/notebooks/WFS_example.ipynb ......         [ 24%]
[2021-05-17T14:33:06.886Z] pavics-sdi-master/docs/source/notebooks/WMS_example.ipynb ........       [ 30%]
[2021-05-17T14:33:08.929Z] pavics-sdi-master/docs/source/notebooks/WPS_example.ipynb ..........     [ 37%]
[2021-05-17T14:33:23.852Z] pavics-sdi-master/docs/source/notebooks/esgf-dap.ipynb .                 [ 37%]
[2021-05-17T14:33:23.852Z] pavics-sdi-master/docs/source/notebooks/jupyter_extensions.ipynb .       [ 38%]
[2021-05-17T14:33:27.715Z] pavics-sdi-master/docs/source/notebooks/opendap.ipynb .......            [ 43%]
[2021-05-17T14:33:33.850Z] pavics-sdi-master/docs/source/notebooks/pavics_thredds.ipynb .....       [ 46%]
[2021-05-17T14:36:57.622Z] pavics-sdi-master/docs/source/notebooks/regridding.ipynb ............... [ 57%]
[2021-05-17T14:37:45.008Z] ...............                                                          [ 67%]
[2021-05-17T14:38:01.870Z] pavics-sdi-master/docs/source/notebooks/rendering.ipynb ....             [ 70%]
[2021-05-17T14:38:08.927Z] pavics-sdi-master/docs/source/notebooks/subsetting.ipynb ......          [ 74%]
[2021-05-17T14:38:20.315Z] finch-master/docs/source/notebooks/dap_subset.ipynb ..........           [ 81%]
[2021-05-17T14:38:27.332Z] finch-master/docs/source/notebooks/finch-usage.ipynb ......              [ 86%]
[2021-05-17T14:39:09.321Z] finch-master/docs/source/notebooks/subset.ipynb ....................     [100%]
[2021-05-17T14:39:09.321Z]
[2021-05-17T14:39:09.321Z] ======================= 143 passed in 416.98s (0:06:56) ========================
</code>
</pre>
</details>

birdhouse/scripts/create-magpie-users
tlvu: It's very convenient to be able to run this script in standalone mode, meaning running the script outside of a proper checkout of birdhouse-deploy repo and without `env.local` matching the intended VM.  I actually use it directly on my machine to hit several of my VM, instead of having to ssh to each of my VM when I need it.  Of course, I override those env vars for the intended VM.

From a proper checkout, everything should resolve.  And even with the good `default.env`, `env.local` could still accidentally override it to an invalid value.  So at this stage, it's user error that is intentional, maybe to test an alternate version of Magpie.

So my point is even if we force sourcing the `default.env`, we still have no warranty that the var will have the same value as `default.env` and we lose the "standalone" feature.  So I'd rather keep the standalone feature.

fmigneault: I'm not against skipping sourcing `default.env` per se. 
I'm questioning the original reason why we were not validating that variables where defined because it was *assumed* that they exists since this file is always force-sourced. If this is not the case anymore because indeed they can still be defined through `env.local` instead, I believe it is in everyone's interest to have a validation check that they were indeed defined that way.

tlvu: Yes, indeed, you can copy/paste the var validation code from `pavics-compose.sh` https://github.com/bird-house/birdhouse-deploy/blob/edd8ab40f783b6b724b88c1abd5c2ac1d721f8dc/birdhouse/pavics-compose.sh#L87-L95 for the 4 vars: URL, USER, PASS, IMAGE.

To be strict, there are a bunch of other scripts that source `default.env` and `env.local` that would benefit from the same var validation code as well.  Can be done in another  PR of course.

fmigneault: Ok. Let's move it for another PR.

tlvu: Agreed, let's keep the standalone feature now and add var validation at the same time as all the other scripts in the same situation.

fmigneault: For reference #172

fmigneault:
@tlvu 
Latest tests errors related to raven only https://github.com/bird-house/birdhouse-deploy/pull/143#issuecomment-842470695 
Merge at best convenience for integration on your side.

birdhouse/scripts/create-magpie-users
tlvu: You forgot to restore standalone feature (surround the sourcing of `default.env` with an `if` statement).

tlvu (approved):
Super, all good thanks.

I have moved our prod onto our fork on birdhouse-deploy so you will be able to merge this one yourself without impacting prod and inspect the merge result on our staging (medus.ouranos.ca).

MatProv:
run tests

tlvu:
@fmigneault Thanks for the merge.  You forgot to copy/paste the content of the PR description into the merge description.  It's okay, next time.

I've launched a Jenkins against Medus after the deployment, everthing still works https://daccs-jenkins.crim.ca/job/PAVICS-e2e-workflow-tests/job/master/454/console, youppi !!!!

Curious, you've previously mentioned something about user self-registration with Magpie but the feature it not activated by default.  Is it in this version of Magpie yet?

Also I guess we can also remove this line https://github.com/bird-house/birdhouse-deploy/blob/93c1ef3c27915784be76dfa3fafa2a6434c9c6f1/birdhouse/docker-compose.yml#L324 since the new magpie has the fix?

tlvu:
@tlogan2000 This version of Magpie has the fix to create users in the UI.  It will autodeploy tonight on prod.  You should be able to create users yourself tomorrow.

fmigneault:
@tlvu  Oh sorry! I always forget to do that. I really hope we can integrate #161 sooner than later to not have to worry about this anymore. 

As for user self-registration, it is still a WIP. 
Snippet of upcoming changes:  https://github.com/Ouranosinc/Magpie/blob/register-user/CHANGES.rst.
I've been busy with other projects and couldn't progress a lot on it recently. 
Most definitely in the next version when ready.

You can try removing `MAGPIE_PORT` to make sure it works properly.

PR #149, opened 2021-04-16
Update raven and jupyter env for raven demo

tlvu:
Raven release notes PR https://github.com/Ouranosinc/raven/pull/374 + https://github.com/Ouranosinc/raven/pull/382

Jupyter env update PR https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/71

Other fixes:
* Fix intermittent Jupyter spawning error by doubling various timeouts config (it's intermittent so hard to test so we are not sure which ones of timeout fixed it)
* Fix Finch and Raven "Broken pipe" error when the request size is larger than default 3mb (bumped to 100mb) (fixes https://github.com/Ouranosinc/raven/issues/361 and Finch related comment https://github.com/bird-house/finch/issues/98#issuecomment-811230388)
* Lower chance to have "Max connection" error for Finch and Raven (bump parallelprocesses from 2 to 10). In prod, the server has the CPU needed to run 10 concurrent requests if needed so this prevent users having to "wait" after each other.

### DIFF
--- a/birdhouse/config/finch/wps.cfg.template
+++ b/birdhouse/config/finch/wps.cfg.template
@@ -2,6 +2,9 @@
 outputurl = https://${PAVICS_FQDN_PUBLIC}/wpsoutputs
 outputpath = /data/wpsoutputs
 
+# default 3mb, fix "Broken pipe" between the proxy and the wps service
+maxrequestsize = 100mb
+
 [logging]
 level = INFO
 database=postgresql://${POSTGRES_PAVICS_USERNAME}:${POSTGRES_PAVICS_PASSWORD}@postgres/finch
--- a/birdhouse/config/finch/wps.cfg.template
+++ b/birdhouse/config/finch/wps.cfg.template
@@ -5,6 +5,9 @@ outputpath = /data/wpsoutputs
 # default 3mb, fix "Broken pipe" between the proxy and the wps service
 maxrequestsize = 100mb
 
+# default 2, too low for a production server with capable CPUs
+parallelprocesses = 10
+
 [logging]
 level = INFO
 database=postgresql://${POSTGRES_PAVICS_USERNAME}:${POSTGRES_PAVICS_PASSWORD}@postgres/finch
--- a/birdhouse/docker-compose.yml
+++ b/birdhouse/docker-compose.yml
@@ -172,7 +172,7 @@ services:
     restart: always
 
   raven:
-    image: pavics/raven:0.12.0
+    image: pavics/raven:0.12.1
     container_name: raven
     ports:
       - "8096:9099"
--- a/birdhouse/docker-compose.yml
+++ b/birdhouse/docker-compose.yml
@@ -172,7 +172,7 @@ services:
     restart: always
 
   raven:
-    image: pavics/raven:0.11.1
+    image: pavics/raven:0.12.0
     container_name: raven
     ports:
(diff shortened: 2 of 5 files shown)


birdhouse/config/finch/wps.cfg.template
```diff
@@ -2,6 +2,9 @@
 outputurl = https://${PAVICS_FQDN_PUBLIC}/wpsoutputs
 outputpath = /data/wpsoutputs
 
+# default 3mb, fix "Broken pipe" between the proxy and the wps service
+maxrequestsize = 100mb
```
MatProv, line 6: Any possible side effect of this change?

MatProv (approved):

birdhouse/config/finch/wps.cfg.template
tlvu: Not sure, I am guessing if someone upload such a large `.geojson` file it will be a heavier calculation... so performance would be a side-effect I guess.  But otherwise, @huard said it's a reasonable value for production server.

huard:
We should include raven 0.12.1 here. Give me a few minutes. 
Also, I suggest making sure parallelprocesses is around 10.

birdhouse/docker-compose.yml
```diff
@@ -172,7 +172,7 @@ services:
     restart: always
 
   raven:
-    image: pavics/raven:0.11.1
+    image: pavics/raven:0.12.0
```
huard, line 175:
```suggestion
    image: pavics/raven:0.12.1
```

huard (approved):

f-PLT:
@tlvu 

Related to https://github.com/bird-house/birdhouse-deploy/commit/560a2b12428bc0e292d3be5f530e9d85c82438d1

I ran into this problem a while back on our Dev instance, but hadn't explored it much. After the timeout, if I just tried to launch the same server again it then worked. In any case, the only way I was able to reproduce the problem at the time was with newly created users or with new server images.

In my case, it happened when starting the `jupyter/datascience-notebook` for the first time. This time around, I figured I could just remove the downloaded docker image. Since it's almost 4 GB, the download time makes Jupyterhub hit the `c.Spawner.start_timeout` value. 

On my test instance, it took around 150-160 seconds for this step to complete. Setting it to 200 in `jupyter_config.py`, like in the linked commit above, took care of the problem, but it could be worthwhile to set it a bit higher just in case.

I hope this can help!

tlvu:
> On my test instance, it took around 150-160 seconds for this step to complete. Setting the `start_timeout` to 200 in `jupyter_config.py`, like in the linked commit above, took care of the problem, but it could be worthwhile to set it a bit higher just in case.

@f-PLT Hey thanks for chiming in!

For your concern that maybe we should even bump `start_timeout` to higher than in this PR, to include time to download new Jupyter images, I think it would be better solved by pre-caching them first as in https://github.com/bird-house/birdhouse-deploy/blob/86d509f8366c48f75b78b0c8a50e2b9408218300/birdhouse/env.local.example#L209-L219 and the caching happens here https://github.com/bird-house/birdhouse-deploy/blob/86d509f8366c48f75b78b0c8a50e2b9408218300/birdhouse/pavics-compose.sh#L200-L205.

See PR https://github.com/bird-house/birdhouse-deploy/pull/130 that improved the caching feature to support caching multiple images.

PR #150, opened 2021-04-19
Custom notebooks

ChaamC:
New solution for custom notebooks, related to both pull requests :
[bird-house/pavics-jupyter-base#3](https://github.com/bird-house/pavics-jupyter-base/pull/3)
[crim-ca/pavics-jupyter-images#3](https://github.com/crim-ca/pavics-jupyter-images/pull/3)

The new solutions uses a custom script that is executed directly on a specific image such as pavics/crim-jupyter-eo or pavics/crim-jupyter-nlp. This script can be executed by configuring a job for the scheduler, using the AUTODEPLOY_EXTRA_SCHEDULER_JOBS variable in the env.local file. The script uses a yaml config file found on the specific images' repo, which describes the source of the notebook files to download and the destination location where to store them.

There was also a need to restructure the tutorial-notebooks folder, in order to better control which notebooks are available depending of the selected image in JupyterHub.

We can discuss of the solution here, but I would propose putting all the current notebooks to a 'common' folder, which would contain all the notebooks that would be available to all images. There would also be other folders for each specific images.
So we would have these paths for example :
- tutorial-notebooks/common
- tutorial-notebooks/crim-eo
- tutorial-notebooks/crim-nlp
- ...
We can then mount the common folder for all images, and we can mount the folder related to our specific image.
For now, I have added the common folder for the volume mounting as an example, but the folder is empty at this current state, since we would need to move the desired notebooks there.

A restriction with my solution is that the specific image notebook folder has to have the same name as the name of the image in the whitelist from JupyterHub (see the code of the CustomDockerSpawner which uses the name of the spawned image to mount the corresponding folder.)


birdhouse/default.env
```diff
@@ -11,6 +11,9 @@ export THREDDS_IMAGE="unidata/thredds-docker:4.6.15"
 # Folder on the host to persist Jupyter user data (noteboooks, HOME settings)
 export JUPYTERHUB_USER_DATA_DIR="/data/jupyterhub_user_data"
 
+# Activates mounting a tutorial-notebooks subfolder that has the same name as the spawned image on JupyterHub
+export MOUNT_IMAGE_SPECIFIC_NOTEBOOKS=false
```
MatProv, line 15: In which case would we set this to false?

ChaamC: I simply added this option to have an easy way to activate or deactivate the mounting of a folder of the same name as the spawned image name.
I think in our case, for CRIM, we will always set this to True, but for other organizations, the option will be there, if they don't want to use this feature.

birdhouse/env.local.example
```diff
@@ -142,12 +142,33 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
...
+# Log directory used for extra autodeploy tasks
+# export AUTO_DEPLOY_LOGDIR=/var/log/PAVICS
+
 # Add more jobs to ./components/scheduler/config.yml
 #
 # Potential usages: other deployment, backup jobs on the same machine
 #
 #export AUTODEPLOY_EXTRA_SCHEDULER_JOBS=""
 #
+# Example extra job that deploys custom notebooks for a specific image
```
dbyrns, line 154: Could we just create a script that loop over DOCKER_NOTEBOOK_IMAGES env var (https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/env.local.example#L214) and perform this operation on each of them?
So, if I want to offer the eo image I add it the the available images env and voilà! I also get all its notebooks.

birdhouse/deployment/deploy-data-specific-image
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
+# Deploy data from git repo(s) to local folder(s).
+# This script is run directly on a specific image (such as pavics/crim-jupyter-eo or pavics/crim-jupyter-nlp).
+# It will be used to download and update the different tutorial notebooks associated with a specific image.
+#
+# This is meant to be run on the same host running PAVICS.
+#
+# The data details is specified in a yaml config file (TEMPLATE_CONFIG_YML), using this format :
```
dbyrns, line 8: Not clear here that the yaml config file must be in the image. This is the case right?
Also the file is expected to be located to a particular location : https://github.com/bird-house/birdhouse-deploy/pull/150/files#diff-17e9b2e274b97a022f797d1f221f2b50144c0ce3b70537a9faa2b7412b2a2cafR159

birdhouse/env.local.example
```diff
@@ -142,12 +142,33 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
...
+#export AUTODEPLOY_EXTRA_SCHEDULER_JOBS="
+#- name: notebookdeploy-eo
+#  comment: Auto-deploy tutorial notebooks for the eo image
+#  schedule: '${AUTODEPLOY_NOTEBOOK_FREQUENCY}'
+#  command: '/deploy-data-specific-image /notebook_config.yml.template'
```
dbyrns, line 159: Following my previous comment regarding this config, given that this file `/notebook_config.yml.template` is in the image, why is it a .template?

birdhouse/default.env
dbyrns: Backcompatibility I guess? Could be an option until all images take advantage of the new mecanism.

birdhouse/deployment/deploy-data-specific-image
```diff
@@ -0,0 +1,88 @@
...
+# - repo_url: path to the github repo (ex.: https://github.com/crim-ca/pavics-jupyter-images)
+#   branch: name of the branch containing the required version of the data
+#   source_path: path of the desired file or folder in the source repo
+#   dest_dir: directory where the data will be copied to in the specific image where the script is run
```
dbyrns, line 12: I don't think it should be an option for the image to choose. The host is mounting the directory $JUPYTERHUB_USER_DATA_DIR and by convention the dest_dir should be $JUPYTERHUB_USER_DATA_DIR/tutorial-notebooks/<image_name>

In fact, the host volume to be mounted should be $JUPYTERHUB_USER_DATA_DIR/tutorial-notebooks/<image_name> and not $JUPYTERHUB_USER_DATA_DIR. The image can only see it's own folder. An env var holding that folder, something like NOTEBOOK_DEST_DIR would also be set so that the script knows where to put files.

dbyrns (changes requested):
Really good job! We can see the thing takes form and it's elegant. I still see some improvements, but we are real close.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -20,7 +22,22 @@ c.JupyterHub.db_url = '/persist/jupyterhub.sqlite'
 
 c.JupyterHub.template_paths = ['/custom_templates']
 
-c.JupyterHub.spawner_class = 'dockerspawner.DockerSpawner'
+class CustomDockerSpawner(DockerSpawner):
+
+    def start(self):
+        if(os.environ['MOUNT_IMAGE_SPECIFIC_NOTEBOOKS'] == 'true'):
+            host_dir = join(os.environ['JUPYTERHUB_USER_DATA_DIR'], 'tutorial-notebooks', self.user_options.get('image'))
+            image_dir = join('/notebook_dir/tutorial-notebooks', self.user_options.get('image'))
```
dbyrns, line 30: Just some thoughts on this : Since notebooks from only one image is mounted at the same time. We do not need this neat structure :
```
/notebook_dir/tutorial-notebooks/common
                                 crim-eo
                                 crim-nlp
```
Since at most two folders will always exist. We could simply have :

```
/notebook_dir/tutorial-notebooks/common
                                 current-image-specific
```
But as of writing this, I'm not sure that having a single folder name having its content change based on the current image is really better.

tlvu: I think that code is the layout "inside" the Jupyter server.

On disk:
```
/data/tutorial-notebooks/common
                                 crim-eo
                                 crim-nlp
```

birdhouse/env.local.example
tlvu: This "cronjob" is again pretty much similar to the existing https://github.com/bird-house/birdhouse-deploy/blob/57a320c3583f804838e8c18aeb064266527a0bc9/birdhouse/components/scheduler/deploy_data_job.env.  Same question, if a feature is missing, can we add to it than forking it?

This one you might not be aware since it's under the "scheduler" component because it has to be used with that component.

Using that "generic" cronjob allows a much simplified cronjob definition like this https://github.com/bird-house/birdhouse-deploy/blob/57a320c3583f804838e8c18aeb064266527a0bc9/birdhouse/components/scheduler/deploy_raven_testdata_to_thredds.env

tlvu (changes requested):
I think we are getting somewhere in the good direction.

I just want to ensure a seemless transition and also more code re-use.

Consider this my first-pass review as I have not seen the other 2 PR that leverage this feature yet.

birdhouse/env.local.example
ChaamC: Do you mean automatically creating a job for each image found in DOCKER_NOTEBOOK_IMAGES?
I suppose we could, most of the parameters will stay the same with only the name (such as eo, nlp, etc.) changing.
But, do we always want to create a job for all of those images? For example, I don't want to have notebooks specific to pavics/workflow-tests, I don't want to have a job for this image. But I guess we could organize those images differently, like having another variable that contains the list of the images for which we want to actually run the script.

Or did you mean a single job that will run the script for each images?

birdhouse/deployment/deploy-data-specific-image
ChaamC: Well, in our case, we run this script on an image, so the config must be on the same image, yes. But, I suppose the script could be used directly on the host, without using any image. Since our use case here is to run it on a docker image, I should just put a clarification that the config file should be on the image too though. :P

And, the script has to be copied, when building the docker image, to the same path that will be written for the job command (like on the link you put here.)

birdhouse/env.local.example
ChaamC: So, I used the template word because in this version of the config, we find the variable ${JUPYTERHUB_USER_DATA_DIR}, which gets replaced by its value when running the script. 
We do a copy of the config file using `envsubst < $TEMPLATE_CONFIG_YML > $CONFIG_YML`, where the copy has the real value instead of the variable.
I agree 'template' is not the best word here though. I don't know if you have a better idea.

In regards to your comment about the dest_dir variable found in the yaml config, I could remove it, and we could always output the notebooks to the same directory, in ${JUPYTERHUB_USER_DATA_DIR}/tutorial-notebooks/{$IMAGE_NAME}.
We could remove that option of customization, which could remove my need of having a copy of the config with variables to replace.
If we decide to remove the dest_dir option, I will just remove all mentions of .template, and remove the envsubst command from the script.

birdhouse/deployment/deploy-data-specific-image
ChaamC: I agree that this could be a good option.

Also see my other comment : https://github.com/bird-house/birdhouse-deploy/pull/150#discussion_r616978832

birdhouse/env.local.example
dbyrns: I mean a single job running a script updating all images. Basically what you already have, but inside a loop. Apart the image name everything else is the same.
And yes for all images found in DOCKER_NOTEBOOK_IMAGES. Right now we have workflow-tests, but the name is misleading because it's the primary image that is used by almost everyone! In fact all images in DOCKER_NOTEBOOK_IMAGES are available to user, so yes I think that we want to keep them updated.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
dbyrns: Yes, you're right this is the structure on disk. My point was that since we would never see that layout (with all images) inside the Jupyter server (because only one image is choosen at a time + common), it is not required to use their name (no collision possible). But again using the same folder for every image could be misleading.

ChaamC: Yes it is the structure on disk.
I also use the same folder name structure on the image for volume mounting, but I could easily change the name of the sub-folder of the specific image to a more generic name such as 'current-image-specific' if we want.
But, using the name of the image as the name of the subfolder is okay too I think. The user sees easily that the folder contains the notebooks related to his environment, such as crim-eo.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -49,9 +66,10 @@ if len(host_gdrive_settings_path) > 0:
         "mode": "ro"
     }
 
+#  Mount folder containing the notebooks that should be available to all images
 host_tutorial_notebooks_dir = join(jupyterhub_data_dir, "tutorial-notebooks")
-c.DockerSpawner.volumes[host_tutorial_notebooks_dir] = {
-    "bind": join(notebook_dir, "tutorial-notebooks"),
+c.DockerSpawner.volumes[join(host_tutorial_notebooks_dir, "common")] = {
+    "bind": join(notebook_dir, "tutorial-notebooks/common"),
```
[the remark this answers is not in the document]
ChaamC: The problem was that if we leave notebooks directly in the tutorial-notebooks folder and mounting the tutorial-notebooks folder to include those notebooks, it will mount the specific image notebooks folder at the same time since they are too in the tutorial-notebooks.
We could find another way to structure it, but we have to make sure there is no resulting conflicts, and that the different folders exclude each other.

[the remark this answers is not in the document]
dbyrns: In fact, common means exactly that, the common things, core to pavics, like thredds, magpie, geoserver, eventually weaver as the entry point for every wps (but generic), the catalog. Raven, Finch and the climate birds should be moved in a climate image.

birdhouse/deployment/deploy-data-specific-image
dbyrns: Hum ok, I agree that this script doesn't mind where the config file is located, it is passed as an argument. 
In fact, what is needed is a better place to document the location of the config file in images. All images will require this config file at a specific location so that the job (my link) can find it.

birdhouse/deployment/deploy-data-specific-image
```diff
@@ -0,0 +1,88 @@
...
+
+  DEST_DIR=$( eval echo "`yq -r .[$i].dest_dir $CONFIG_YML`")
+  ensure_not_null "$DEST_DIR"
+
+  FULL_URL=$GIT_REPO/branches/$BRANCH/$SOURCE_PATH
+
+  echo "Extracting ${FULL_URL} to ${DEST_DIR}"
+
+  # Download the data from github and copy it to the destination directory
+  svn export --force $FULL_URL $DEST_DIR
```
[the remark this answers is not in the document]
ChaamC: They do have a lot of similarities. I could give it a try and try using the deploy-data script directly, but I am scared of finding a blocking point sometime if we want to add different features to one of the use case.
They are almost similar though, if not the same, for now.

Also, I see that the deploy-data script requires the use of Docker, which means we have to install Docker on the pavics-jupyter-base (this would replace the yq/jq installation actually). I don't know what is the best here, if we want to keep those images to the bare minimum.

@dbyrns I would be curious about your input on this :)

birdhouse/env.local.example
dbyrns: Ok, it solves two issues and make the whole thing simpler, so I would remove the dest_dir option from the config.
But I would not use the directory ${JUPYTERHUB_USER_DATA_DIR}/tutorial-notebooks/{$IMAGE_NAME}. This is where the notebooks need to be on the host, but in the image it could be anywhere like /tmp as long as the calling command mount the host volume at that point. My suggestion still hold : in the docker command include the volume mount and set an environment variable telling the script where to put the stuff (the dest_dir, but as a env var rather than a config)

birdhouse/config/jupyterhub/jupyterhub_config.py.template
dbyrns: Exact, let's keep that this way.

birdhouse/deployment/deploy-data-specific-image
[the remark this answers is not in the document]
dbyrns: I don't have enough information on what @tlvu would like you to reuse to make that call. If he could show us how it could easily be done I'm not against it. As long as we don't have to invest another couple of days to do that. The same apply to the other reuse request, maybe you could talk directly to each other so that it could be done fast.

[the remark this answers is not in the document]
ChaamC: I think the main difference between our scripts is that the `deploy-data` script is meant to run on a generic image that includes Docker and Git, but our new script `deploy-data-specific-image` is meant to be run directly on one of our own specific image, which includes our config.yaml file.

A benefit of using the new script is that it lets us have the yaml file directly on the specific image's repo, keeping it close to its related context. For example, this means, if a developer wants to include a new folder for the crim-eo environment, he just goes to the crim-eo's repo and changes the config there.

I am not sure if we could do this easily with the `deploy-data` script? I think the yaml files are stored directly on the birdhouse-deploy repo? Such as : [deploy_raven_testdata_to_thredds.env](https://github.com/bird-house/birdhouse-deploy/blob/57a320c3583f804838e8c18aeb064266527a0bc9/birdhouse/components/scheduler/deploy_raven_testdata_to_thredds.env) and [deploy-data-raven-testdata-to-thredds.yml](https://github.com/bird-house/birdhouse-deploy/blob/57a320c3583f804838e8c18aeb064266527a0bc9/birdhouse/deployment/deploy-data-raven-testdata-to-thredds.yml)

They are then added as a job in the env.local file : https://github.com/bird-house/birdhouse-deploy/blob/57a320c3583f804838e8c18aeb064266527a0bc9/birdhouse/env.local.example#L177

birdhouse/config/jupyterhub/jupyterhub_config.py.template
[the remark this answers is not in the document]
tlvu: For the sake of seemless backward-compat, I suggest moving the volume-mount of this `common` dir inside the `if(os.environ['MOUNT_IMAGE_SPECIFIC_NOTEBOOKS'] == 'true')` and add an `else` clause to restore the volume-mount of just the `tutorial-notebooks` folder as if this change was never there.

So if MOUNT_IMAGE_SPECIFIC_NOTEBOOKS'] == 'true', Jupyter user will see
```
/tutorial-notebooks/common/  --> /data/jupyterhub_user_data/tutorial-notebooks/common/
                    crim-eo/  --> /data/jupyterhub_user_data/tutorial-notebooks/crim-eo/
```
else
```
/tutorial-notebooks/ --> /data/jupyterhub_user_data/tutorial-notebooks/
```

birdhouse/deployment/deploy-data-specific-image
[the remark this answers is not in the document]
tlvu: Now that I have more time to digest this change. I would suggest moving this `deploy-data-specific-image` script into the jupyter-pavics-base image so the script and the config yaml (that is in the child image) is part of the same final image.

Why?  So eventually Jenkins can also checkout all the notebooks and run tests on them, like currently with the default Jupyter image.  The e2e repo should not be responsible for checking out the notebooks like currently and will enable that same e2e repo to test against different notebooks depending on the Jupyter image.  That requires implementation of https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/issues/57 but we can start to lay the ground in that direction.

For the scheduler cronjob, this means you don't even have to volume-mount the script from outside since it's already inside the image !

As for re-using the existing generic `deploy-data` script, I think it simply boils down to 
1. being able to use `yq` as `docker run` (currently) or as installed (your case), will have to add a new switch for that.
2. download the repo as a tar/zip archive (your way) instead of `git pull` (current), will have to add a new switch for that as well.
   Note that going the tar/zip archive route nullify the caching provided by `git pull`.  However, `git pull` way waste more space because basically there is a double checkout, one used for caching, one in the `/data/jupyter_user_data/tutorial-notebooks/...`.  Each side have pros and cons so for generic and re-usabilty sake, we can implement both.

In a different PR, once `deploy-data` have the additional mode of operations, the jupyter-pavics-base will wget the script part of the build and make it available inside the image as `deploy-data-specific-image` is, without being directly committed as `deploy-data-specific-image`.

birdhouse/env.local.example
```diff
@@ -142,12 +142,33 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
 # "Notebook" are all the tutorial notebooks on Jupyter.
 #export AUTODEPLOY_NOTEBOOK_FREQUENCY="@every 5m"
 
+# Log directory used for extra autodeploy tasks
+# export AUTO_DEPLOY_LOGDIR=/var/log/PAVICS
```
tlvu, line 146: This location is not specific to autodeploy.  All kinds of jobs log there so all PAVICS related stuff are at one place.

I suggest naming this PAVICS_LOGDIR instead and the default value should be in `default.env`.

birdhouse/env.local.example
tlvu: Again, now that I have more time to digest this PR ...

I would suggest moving this sample big blob of code to generate the cronjob also into jupyter-pavics-base repo so it is together with the script `deploy-data-specific-image` it wraps.
* It makes the `env.local.example` shorter and less intimidating
* You now own the "code" so if you need to fix something (like add/remove a volume-mount) you can do it and it's transparent to all the callers, all they need is always run the latest version.  You basically provide the user with a stable interface and give you control to change the implementation.
* Later, in a different PR, once we "migrate" to the generic `deploy-data` script, the matching generic cronjob `components/scheduler/deploy_data_job.env` will also have to be adapted for whatever newer options that `deploy-data` will support.  And we still keep the same consistency that the "script" and the "cronjob wrapper" are together.

Try to inspire from the existing generic cronjob wrapper, especially the part about https://github.com/bird-house/birdhouse-deploy/blob/57a320c3583f804838e8c18aeb064266527a0bc9/birdhouse/components/scheduler/deploy_data_job.env#L64-L67 to ensure it also works during autodeploy (there is a slight difference between `./pavics-compose.sh` invoked manually on the console and invoke inside the scheduler component).

birdhouse/env.local.example
```diff
@@ -208,8 +229,10 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
 
 # Jupyter single-user server images
 #export DOCKER_NOTEBOOK_IMAGES="pavics/workflow-tests:210216 \
-#                               pavics/crim-jupyter-eo:0.1.0 \
-#                               pavics/crim-jupyter-nlp:0.1.0"
+#                               ${DOCKER_EO_IMAGE} \
+#                               ${DOCKER_NLP_IMAGE}"
+#export DOCKER_EO_IMAGE="pavics/crim-jupyter-eo:0.1.0"
+#export DOCKER_NLP_IMAGE="pavics/crim-jupyter-nlp:0.1.0"
```
tlvu, line 235: These vars `DOCKER_EO_IMAGE` and `DOCKER_NLP_IMAGE` have to be declared before `DOCKER_NOTEBOOK_IMAGES` that uses them.

birdhouse/env.local.example
tlvu: Forgot to say, once you move the "cronjob generation code" out to jupyter-pavics-base, you can reference it back in `env.local.example` to "advertise" it, like https://github.com/bird-house/birdhouse-deploy/blob/57a320c3583f804838e8c18aeb064266527a0bc9/birdhouse/env.local.example#L171-L181

birdhouse/deployment/deploy-data-specific-image
```diff
@@ -0,0 +1,88 @@
...
+    fi
+}
+
+# Check installation of required package to read yaml files
+if ! command -v jq &> /dev/null; then
+  apt-get -y install jq
+fi
+if ! command -v yq &> /dev/null; then
+  pip install yq
+fi
```
MatProv, line 50: I would suggest not to install packages within the actual deploy script.
Instead, printing an issue and prematurely exiting the script would make sense. 
A clear separation between the requirements and actual logic is preferred.

birdhouse/default.env
MatProv: @ChaamC In this case, maybe we should have this option defined only in the `env.local` and not `default.env`, since it's a preference. 
Having this `MOUNT_IMAGE_SPECIFIC_NOTEBOOKS` variable defined in both files will lead to confusion.

tlvu (changes requested):
So here my "final-pass" review.

Good job, you seem to grasp the scheduler component and how to add more "jobs" to it.

Things are aligned in the right direction, that's the most important point.  There are improvements but we do not have to make them right now.

Here are what I suggest to merge this PR without too much more work:
* ensure 100% seemless transition (ie when `MOUNT_IMAGE_SPECIFIC_NOTEBOOKS == false` should behave exactly as before, as if this change never existed).
* move the custom, not generic deploy script and the matching cronjob generation to jupyter-pavics-base

All the issues about code re-use, we can handle in a "phase 2" sort of PR, iterative/incremental development style.

Final thing, remember to test this change in an autodeploy context (meaning have autodeploy callling `./pavics-compose.sh` which generate the actual `./components/scheduler/config.yml` with all the jobs fully instantiated, make sure no double job and all the variables expanded correctly).

birdhouse/deployment/deploy-data-specific-image
```diff
@@ -0,0 +1,88 @@
...
+#
+# The data details is specified in a yaml config file (TEMPLATE_CONFIG_YML), using this format :
+# - repo_url: path to the github repo (ex.: https://github.com/crim-ca/pavics-jupyter-images)
+#   branch: name of the branch containing the required version of the data
+#   source_path: path of the desired file or folder in the source repo
+#   dest_dir: directory where the data will be copied to in the specific image where the script is run
+#
+# One or more entries respecting this format can be added to the yaml file, in order to download multiple file/folders.
+#
+# An installation of jq and yq is preferred prior to executing this script.
```
MatProv, line 16: No need to specify, based on previous comment.

birdhouse/deployment/deploy-data-specific-image
```diff
@@ -0,0 +1,88 @@
...
+    exit 2
+fi
+
+# Empty value could mean typo in the keys in the config file.
+ensure_not_null() {
+    if [ "$*" = null ]; then
+        echo "ERROR: value empty" 1>&2
+        exit 1
+    fi
+}
```
MatProv, line 42: Would be great to pass the parameter as argument. 
That way, instead of `ERROR: value empty`, it would be `ERROR: value empty (GIT_REPO)`.

birdhouse/deployment/deploy-data-specific-image
```diff
@@ -0,0 +1,88 @@
...
+# Check installation of required package to read yaml files
+if ! command -v jq &> /dev/null; then
+  apt-get -y install jq
+fi
+if ! command -v yq &> /dev/null; then
+  pip install yq
+fi
+
+# Find how many entries are in the config file
+LENGTH="`yq -r '. | length' $TEMPLATE_CONFIG_YML`"
```
MatProv, line 53: Please use a more explicit naming, like `CONFIG_FILE_LENGTH`, or something alike.

birdhouse/deployment/deploy-data-specific-image
```diff
@@ -0,0 +1,88 @@
...
+
+if [ -z $LENGTH ]; then
+    echo "ERROR: empty config file" 1>&2
+    exit 1
+fi
+
+CONFIG_YML="/tmp/notebook_config.yml"
```
MatProv, line 60: I would use `CONFIG_YML_PATH`, in case we need to differentiate the path from the actual content later on.

birdhouse/deployment/deploy-data-specific-image
```diff
@@ -0,0 +1,88 @@
...
+# Replace environment variables from template to their actual values
+envsubst < $TEMPLATE_CONFIG_YML > $CONFIG_YML
+
+for ((i=0;i<$LENGTH;i++)); do
+  # Extract data from config
```
MatProv, line 65: Please keep same indent (either 2 or 4), for clarity.

birdhouse/deployment/deploy-data-specific-image
```diff
@@ -0,0 +1,88 @@
...
+  GIT_REPO="`yq -r .[$i].repo_url $CONFIG_YML`"
+  ensure_not_null "$GIT_REPO"
+
+  BRANCH="`yq -r .[$i].branch $CONFIG_YML`"
+  ensure_not_null "$BRANCH"
+
+  SOURCE_PATH="`yq -r .[$i].source_path $CONFIG_YML`"
+  ensure_not_null "$SOURCE_PATH"
+
+  DEST_DIR=$( eval echo "`yq -r .[$i].dest_dir $CONFIG_YML`")
```
MatProv, line 75: Why do we need `eval` over here, and not for previous `yq` operations?

MatProv (changes requested):
Some modifications here and there but you got some great features to add to the stack!

dbyrns:
> * ensure 100% seemless transition (ie when `MOUNT_IMAGE_SPECIFIC_NOTEBOOKS == false` should behave exactly as before, as if this change never existed).

100% agree.

> * move the custom, not generic deploy script and the matching cronjob generation to jupyter-pavics-base

For the script, it makes sense and would indeed be useful for the e2e tests. 

For the cronjob I'm not so sure about your intention. I guess that the command to launch the docker could be simplified by moving a lot of knowledge in the image so that the command to run it is very simple, but the host still need to trigger that command via a cronjob, right? Because if the cron job is in the image, every user will run wastefully the update notebooks script.

So how your idea fits with mine of having a single cron job that update every available images in DOCKER_NOTEBOOK_IMAGES? I see a basic cron job that loop over every image and run a simple docker command for each of them. The "API" exposed by the image would be a script at a specific location that only required from the host a volume where the notebooks has to be cached.

> Final thing, remember to test this change in an autodeploy context (meaning have autodeploy job in the scheduler callling `./pavics-compose.sh` which generate the actual `./components/scheduler/config.yml` with all the jobs fully instantiated, make sure no double job and all the variables expanded correctly).

100% agree too.


I invite @ChaamC to our meeting this afternoon to discuss about what remain to be done.

tlvu:
> For the cronjob I'm not so sure about your intention. I guess that the command to launch the docker could be simplified by moving a lot of knowledge in the image so that the command to run it is very simple, but the host still need to trigger that command via a cronjob, right? Because if the cron job is in the image, every user will run wastefully the update notebooks script.

Sorry, forgot to provide an example.

So Ouranos also have "custom cronjobs" (see https://github.com/bird-house/birdhouse-deploy-ouranos/tree/a8aedc98600297f726bd7a536c124e6d401129a9/scheduler-jobs) that is not generic so we did not put it into the birdhouse-deploy but it is still "integrated" into the stack at deployment time.

Our custom cronjobs are "integrated" into the stack is via `env.local` this way:

```sh
CHECKOUT_OURANOS_CONFIG="/path/to/checkout/birdhouse-deploy-ouranos"

# ensure we always use the "latest" version of the "cronjob generation code"
export AUTODEPLOY_EXTRA_REPOS="$AUTODEPLOY_EXTRA_REPOS
  $CHECKOUT_OURANOS_CONFIG
"

DEPLOY_DATA_JOB_ENV="$COMPOSE_DIR/components/scheduler/deploy_data_job.env"
DEPLOY_PAVICS_LANDING_NOTEBOOKS_JOB="$CHECKOUT_OURANOS_CONFIG/scheduler-jobs/deploy_pavics_landing_notebooks.env"

if [ -f "$DEPLOY_PAVICS_LANDING_NOTEBOOKS_JOB" -a -f "$DEPLOY_DATA_JOB_ENV" ]; then
    # default override of deploy_data_job.env, good for most hosts.
    . $DEPLOY_PAVICS_LANDING_NOTEBOOKS_JOB

    # Extra override, for the current specific host, on top of the default override above.
    # Need to override DEPLOY_DATA_JOB_CONFIG, else this error:
    # realpath: /path/to/checkout/birdhouse-deploy/birdhouse/../../birdhouse-deploy-ouranos/scheduler-jobs: No such file or directory
    DEPLOY_DATA_JOB_CONFIG="$CHECKOUT_OURANOS_CONFIG/scheduler-jobs/deploy_pavics_landing_notebooks.yml"

    # Actually generate the cronjob (write to `AUTODEPLOY_EXTRA_SCHEDULER_JOBS`).
    . $DEPLOY_DATA_JOB_ENV
fi
```
Note the similarity with how to activate cronjobs that ships with the birdhouse-deploy repo: https://github.com/bird-house/birdhouse-deploy/blob/57a320c3583f804838e8c18aeb064266527a0bc9/birdhouse/env.local.example#L171-L181

So my suggestion was for the script to be part of the the image but not the "cronjob generation code" (`deploy_data_job.env` + `deploy_pavics_landing_notebooks.env` in my example above).


> 
> So how your idea fits with mine of having a single cron job that update every available images in DOCKER_NOTEBOOK_IMAGES? I see a basic cron job that loop over every image and run a simple docker command for each of them. The "API" exposed by the image would be a script at a specific location that only required from the host a volume where the notebooks has to be cached.

In `env.local`, using my previous example above, it would probably look something like:
```sh
for NB_IMAGE in $DOCKER_NOTEBOOK_IMAGES; do    
  if [ -f "$DEPLOY_PAVICS_LANDING_NOTEBOOKS_JOB" -a -f "$DEPLOY_DATA_JOB_ENV" ]; then
    . $DEPLOY_PAVICS_LANDING_NOTEBOOKS_JOB
    # Override configs (dest_dir volume mount, jupyter image name, ...) depending on NB_IMAGE value.
    . $DEPLOY_DATA_JOB_ENV
  fi
done
```
I wonder if `NB_IMAGE` will contain enough info for all the required overrides or you will need to invent another variable to drive this loop.

birdhouse/env.local.example
```diff
@@ -226,6 +251,9 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
 #                                   }
 #"
 
+# Activates mounting a tutorial-notebooks subfolder that has the same name as the spawned image on JupyterHub
+# export MOUNT_IMAGE_SPECIFIC_NOTEBOOKS=true
```
tlvu, line 255: Can you add an explanation that `MOUNT_IMAGE_SPECIFIC_NOTEBOOKS=true`, is only useful if there are more than one image in `DOCKER_NOTEBOOK_IMAGES` and `ENABLE_JUPYTERHUB_MULTI_NOTEBOOKS` is set with proper `c.DockerSpawner.image_whitelist` matching the images in `DOCKER_NOTEBOOK_IMAGES`.

birdhouse/deployment/deploy-data-specific-image
ChaamC: I think it wasn't needed anymore, but anyway, this line will disappear since we removed the dest_dir parameter from the config.yml.

ChaamC:
@tlvu 
Thank you for the detailed explanation! I understand the general idea. It is pretty clear now, but I have a few questions after looking at the example code.

For the `deploy_data_job.env` file, should we make a copy, specialized for the use case of the `deploy-data-specific-image` script, or should we just adapt the existing `deploy_data_job.env` file? They are almost the same, but with a few differences, such as the using of Git or not, the mount of some files or not. 
The `deploy_data_job.env` seems to be closely related to the initial script deploy-data.

Also does `CHECKOUT_OURANOS_CONFIG="/path/to/checkout/birdhouse-deploy-ouranos"` implies that the repo has already been checked out somewhere on the host? So we should have a copy of the repo somewhere on our host, which corresponds to the path added to the `env.local`.

tlvu:
> For the `deploy_data_job.env` file, should we make a copy, specialized for the use case of the `deploy-data-specific-image` script, or should we just adapt the existing `deploy_data_job.env` file? They are almost the same, but with a few differences, such as the using of Git or not, the mount of some files or not.

Well ... I thought you will create your own equivalent of `deploy_data_job.env` to match your custom `deploy-data-specific-image` (basically the same content as you had previously in `env.local.example`).

> The `deploy_data_job.env` seems to be closely related to the initial script deploy-data.

Exact, these 2 are closely related so if you don't use `deploy-data`, I think it's less trouble if you roll your own equivalent of `deploy_data_job.env`.

Ideally you would have re-used `deploy-data` then you would have gotten `deploy_data_job.env` for almost free (+ additional changes for your use-case) but that would make this PR longer so we can do this in a subsequent PR.

> 
> Also does `CHECKOUT_OURANOS_CONFIG="/path/to/checkout/birdhouse-deploy-ouranos"` implies that the repo has already been checked out somewhere on the host? So we should have a copy of the repo somewhere on our host, which corresponds to the path added to the `env.local`.

Exact.  It's the same model for adding additional "components" to PAVICS.  Checkout of the repo of the external component is done once manually, then the autodeploy will keep it up-to-date automatically going forward.

ChaamC:
@dbyrns @tlvu @MatProv 
Repo has been updated with what was discussed this week.
Script has been moved to the pavics-jupyter-base repo, and automatic cronjob generation was added to make a job for each specific images.
Note that it also adds a job for pavics-workflow since it is in the list. It just creates an error in the scheduler for now, since the image doesn't have the script (since it doesn't derive from pavics-jupyter-base.)

I did a test with a local setup and it works : a notebook was updated automatically on JupyterLab after I sent a commit on the nlp image repo.
I will try to do a test with the iac too, after receiving your feedback.

birdhouse/env.local.example
```diff
@@ -226,6 +235,35 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
 #                                   }
 #"
```
ChaamC, line 237: @tlvu Let me know if you think all this following part should be in the env.local.example.
It is the way I have to implement it for it to work in the env.local, but I wasn't sure if we wanted to add all the info on the example file (such adding the extra repo to AUTODEPLOY_EXTRA_REPOS).

birdhouse/env.local.example
```diff
@@ -206,17 +208,24 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
...
+
 # allow jupyterhub user selection of which notebook image to run
 # see https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html
 #export ENABLE_JUPYTERHUB_MULTI_NOTEBOOKS="
-#c.DockerSpawner.image_whitelist = {'pavics': os.environ['DOCKER_NOTEBOOK_IMAGES'].split()[0],
-#                                   'eo-crim': os.environ['DOCKER_NOTEBOOK_IMAGES'].split()[1],
-#                                   'nlp-crim': os.environ['DOCKER_NOTEBOOK_IMAGES'].split()[2],
+#c.DockerSpawner.image_whitelist = {os.environ['JUPYTERHUB_IMAGE_SELECTION_NAMES'].split()[0]: os.environ['DOCKER_NOTEBOOK_IMAGES'].split()[0],
+#                                   os.environ['JUPYTERHUB_IMAGE_SELECTION_NAMES'].split()[1]: os.environ['DOCKER_NOTEBOOK_IMAGES'].split()[1],
+#                                   os.environ['JUPYTERHUB_IMAGE_SELECTION_NAMES'].split()[2]: os.environ['DOCKER_NOTEBOOK_IMAGES'].split()[2],
```
ChaamC, line 228: For these lines, I am aware it is not super clean. I haven't found an easier way to implement it yet, since the concerned variables have to be working both on shellscript and python.

About `DOCKER_NOTEBOOK_IMAGES` and `JUPYTERHUB_IMAGE_SELECTION_NAMES`, we could have use those names in just one variable instead of 2, using a string "formatted similarly to a dictionary" such as `"pavics;pavics/workflow-tests:210216 eo-crim;pavics/crim-jupyter-eo:0.1.0 ...."` but it would require additional string splitting, so I don't think it would be cleaner.

Another thing is we could probably simplify this by using the same name all the time, instead of having 2 different name format. We would use the image name such as `pavics/crim-jupyter-eo:0.1.0` everywhere though, meaning it would be what we see in the jupyterhub list, and it would be the name of the directories for the tutorial-notebooks, instead of the shorter version `eo-crim`. I don't think it would be necessarily cleaner either. (Potential problem with the '/' found in the image names, clashing with using it as a directory name?)

birdhouse/env.local.example
```diff
@@ -229,19 +208,24 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
 #
 #export ALLOW_UNSECURE_HTTP=""
 
-# Jupyter single-user server images
+# Jupyter single-user server images=
```
tlvu, line 211: Trailing `=`?

birdhouse/env.local.example
```diff
@@ -251,7 +235,33 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
 #                                   }
 #"
 
+# Load jobs to automatically deploy the custom notebooks from the specific images
+#
+# Ensure we always use the "latest" version of the "cronjob generation code"
+# Path to a checked out repo of "pavics-jupyter-base" which contains the config required for the cronjob generation
+#CHECKOUT_PAVICS_CONFIG="/path/to/checkout/pavics-jupyter-base"
```
tlvu, line 242: CHECKOUT_PAVICS_CONFIG ... I think a different var name is more appropriate, this is very generic.

birdhouse/env.local.example
```diff
@@ -251,7 +235,33 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
...
+#export AUTODEPLOY_EXTRA_REPOS="$AUTODEPLOY_EXTRA_REPOS $CHECKOUT_PAVICS_CONFIG"
+
+# Script for automatic cronjob generation
+#DEPLOY_DATA_SPECIFIC_IMAGE_JOB_ENV="$COMPOSE_DIR/components/scheduler/deploy_data_specific_image_job.env"
+# Config for the generation of cronjobs, found on external repo
+#DEPLOY_DATA_PAVICS_JUPYTER_ENV="$CHECKOUT_PAVICS_CONFIG/scheduler-jobs/deploy_data_pavics_jupyter.env"
```
tlvu, line 248: Oh so there is an override, not sure why you can not merge this override with the local `deploy_data_specific_image_job.env` but I haven't seen the override yet.

tlvu (commented):
Just first pass.  Have not seen the other matching PRs.  Nothing major so far, looking good.

birdhouse/components/scheduler/deploy_data_specific_image_job.env
```diff
@@ -0,0 +1,93 @@
+##############################################################################
```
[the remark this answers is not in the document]
tlvu: So I just finished reviewing the other PR, I suggest just merging this file with the file in the other PR, including the looping-part, to keep things simple.

birdhouse/env.local.example
```diff
@@ -142,6 +142,8 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
 # "Notebook" are all the tutorial notebooks on Jupyter.
 #export AUTODEPLOY_NOTEBOOK_FREQUENCY="@every 5m"
 
+#export NOTEBOOK_DEST_DIR=/tmp/notebooks
```
tlvu, line 145: I don't think this NOTEBOOK_DEST_DIR would ever need to be manually set in this `env.local` at all.

It is always set in the "loop" no?  If you need default value for the loop, at least keep it really close to the loop, not alone like here.

birdhouse/env.local.example
tlvu: Agreed it's ugly, but also agree other alternative either complexify the implemenation a lot or look less great.  Open to other alternative as well, if someone see something we both missed ...

birdhouse/env.local.example
tlvu: Put the github link to pavics-jupyter-base so user can perform the checkout without guessing/searching.

birdhouse/env.local.example
```diff
@@ -226,6 +235,35 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
...
+#export AUTODEPLOY_EXTRA_REPOS="$AUTODEPLOY_EXTRA_REPOS $CHECKOUT_PAVICS_CONFIG"
+
+# Script for automatic cronjob generation
+#DEPLOY_DATA_SPECIFIC_IMAGE_JOB_ENV="$COMPOSE_DIR/components/scheduler/deploy_data_specific_image_job.env"
+# Config for the generation of cronjobs, found on external repo
+#DEPLOY_DATA_PAVICS_JUPYTER_ENV="$CHECKOUT_PAVICS_CONFIG/scheduler-jobs/deploy_data_pavics_jupyter.env"
+
+# Generates a cronjob for each image found in DOCKER_NOTEBOOK_IMAGES
+#if [ -f "$DEPLOY_DATA_PAVICS_JUPYTER_ENV" -a -f "$DEPLOY_DATA_SPECIFIC_IMAGE_JOB_ENV" ]; then
+#    for i in $( seq 1 $(echo $DOCKER_NOTEBOOK_IMAGES | wc -w)) ; do
```
tlvu, line 252: I think it's possible to move this loop into `deploy_data_pavics_jupyter.env` as well so all you're left is
```
if [ -f "$DEPLOY_DATA_PAVICS_JUPYTER_ENV" ]; then
  . $DEPLOY_DATA_PAVICS_JUPYTER_ENV
fi
```
Much cleaner for the user.

birdhouse/env.local.example
```diff
@@ -226,6 +235,35 @@ export POSTGRES_MAGPIE_PASSWORD=postgres-qwerty
...
+#
+#        . $DEPLOY_DATA_PAVICS_JUPYTER_ENV
+#        . $DEPLOY_DATA_SPECIFIC_IMAGE_JOB_ENV
+#    done
+#fi
+
+# Activates mounting a tutorial-notebooks subfolder that has the same name as the spawned image on JupyterHub
+# This variable is only useful if there are more than one image in DOCKER_NOTEBOOK_IMAGES
+# and ENABLE_JUPYTERHUB_MULTI_NOTEBOOKS is set with a proper c.DockerSpawner.image_whitelist
+# matching the images in DOCKER_NOTEBOOK_IMAGES.
```
tlvu, line 264: New var to keep in sync as well: `JUPYTERHUB_IMAGE_SELECTION_NAMES`

birdhouse/env.local.example
tlvu: Yes these should be in `env.local.example` but simplified since the loop is out.

tlvu (changes requested):
Great work.  I think you got all the pieces.  Just simplify them a bit since you are not aiming for something generic (it's all specific to notebooks deployment).

birdhouse/components/scheduler/deploy_data_specific_image_job.env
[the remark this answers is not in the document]
dbyrns: I agree, a loop doesn't look like a config someone should add to get this working. I was hoping for a switch that can be activated in the env to get this working. What @tlvu proposes seem pretty streamline.

dbyrns (commented):
I feel bad for the long journey, but we are almost there.

tlvu:
> I feel bad for the long journey, but we are almost there.

Yes sorry @ChaamC I was the one who lead you into having the loop outside `env.local` in comment https://github.com/bird-house/birdhouse-deploy/pull/150#issuecomment-824363029.

At the time, not knowing how things will turn out, I was simply thinking with what I have currently.

birdhouse/components/scheduler/deploy_data_specific_image_job.env
```diff
@@ -0,0 +1,93 @@
...
+fi
+
+# Job description.
+if [ -z "$DEPLOY_DATA_JOB_JOB_DESCRIPTION" ]; then
+    DEPLOY_DATA_JOB_JOB_DESCRIPTION="$DEPLOY_DATA_JOB_JOB_NAME"
+fi
+
+# Log file location.  Default location under /var/log/PAVICS/ has built-in logrotate.
+if [ -z "$DEPLOY_DATA_JOB_LOGFILE" ]; then
+    DEPLOY_DATA_JOB_LOGFILE="/var/log/PAVICS/${DEPLOY_DATA_JOB_JOB_NAME}.log"
```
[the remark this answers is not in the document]
ChaamC: I will replace it here, but if I search for the "/var/log/PAVICS" path in the repo, I see that it is referenced many times, in different other jobs or configs (logrotate, autodeploy, trigger-deploy-notebook, ...) Should I also take the liberty of replacing those with the new variable? I am not sure if I want to change anything in parts of the code that I don't know about.

birdhouse/components/scheduler/deploy_data_specific_image_job.env
```diff
@@ -0,0 +1,93 @@
...
+    export AUTODEPLOY_EXTRA_SCHEDULER_JOBS="
+$AUTODEPLOY_EXTRA_SCHEDULER_JOBS
+
+- name: $DEPLOY_DATA_JOB_JOB_NAME
+  comment: $DEPLOY_DATA_JOB_JOB_DESCRIPTION
+  schedule: '$DEPLOY_DATA_JOB_SCHEDULE'
+  command: '/deploy_data_specific_image /notebook_config.yml'
+  dockerargs: >-
+    --rm --name $DEPLOY_DATA_JOB_JOB_NAME
+    --volume /var/run/docker.sock:/var/run/docker.sock:ro
```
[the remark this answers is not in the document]
ChaamC: I think you're right. It seems to work without it, and it would make sense since we don't need docker for this script. 
I'll remove it!

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -23,18 +23,30 @@ c.JupyterHub.db_url = '/persist/jupyterhub.sqlite'
...
-
     def start(self):
         if(os.environ['MOUNT_IMAGE_SPECIFIC_NOTEBOOKS'] == 'true'):
-            host_dir = join(os.environ['JUPYTERHUB_USER_DATA_DIR'], 'tutorial-notebooks', self.user_options.get('image'))
-            image_dir = join('/notebook_dir/tutorial-notebooks', self.user_options.get('image'))
+            host_dir = join(os.environ['JUPYTERHUB_USER_DATA_DIR'], 'tutorial-notebooks')
+            image_dir = join('/notebook_dir/tutorial-notebooks')
+
+            #  Mount folder containing the notebooks that should be available to all images
+            self.volumes[join(host_dir, "common")] = {
```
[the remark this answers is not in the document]
ChaamC: Just to be clear here, should I just remove the `common` dir volume here? Or should I also implement a way to easily customize the volume mounts for the specific images?
I don't think right now it is easy to have a `common` dir just for the eo/nlp images, since those volumes are handled by the jupyterhub_config.py.template, not by the env.local that we use to customize the environment in the daccs-configs repo. Unless this `common` folder is found directly in the specific image directory (ex.: tutorial-notebooks/crim-eo/common ), but it makes less sense.

We use the [ENABLE_JUPYTERHUB_MULTI_NOTEBOOKS](https://github.com/bird-house/birdhouse-deploy/blob/02d1e9db0cfc0ca0c2ca382e379228a48d4ca182/birdhouse/env.local.example#L228) variable right now to send a custom image whitelist from the env.local to the jupyterhub_config. So we could use a similar mechanic for custom volumes.

Anyway, let me know what I should do here, if I just remove the common volume, or if you need more.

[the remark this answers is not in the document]
ChaamC: A solution was discussed with @dbyrns. We will put back a destination folder in the config.yml files, to add a way to customize the organization of a directory such as "tutorial-notebooks/crim-eo".
More to follow :)

birdhouse/env.local.example
ChaamC: Moved to the env file in the pavics-jupyter-base repo

ChaamC:
@dbyrns @tlvu @MatProv 
Updated PR : moved the env file to the jupyter-pavics-base repo, and fixed other concerns pointed in your feedback

Let me know if it works!

birdhouse/components/scheduler/deploy_data_specific_image_job.env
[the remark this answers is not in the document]
ChaamC: @tlvu Let me know what I should do for this :)

tlvu (approved):
Looks good on this side.

MatProv (approved):

dbyrns (approved):
Perfect! I really like how it turns out :)

birdhouse/components/scheduler/deploy_data_specific_image_job.env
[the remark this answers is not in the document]
ChaamC: @tlvu ?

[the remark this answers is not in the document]
tlvu: Sorry, missed this question.  Yes indeed, will be nice to clean up all existing hardcode but I'd suggest to just finish this PR that's been open for a while already.  You can open another PR to fix the remaining of reference.

ChaamC:
@dbyrns @tlvu 

After testing on the daccs-iac, I found one bug with the current solution.
It seems the existing 'notebookdeploy' scheduler job is in conflict with the new specific image jobs we are trying to add.
I did not see it at first on a local setup, since the 'notebookdeploy' scheduler job was only happening one time per hour, but I started seeing weird behaviors with the volume mounts on the daccs-iac, with the 'notebookdeploy' job happening at each 5 minutes on the daccs-iac setup.

At each 5 minutes, I first see the image-specific folders getting updated as expected, but the 'notebookdeploy' job finishes after, and erases the eo-crim/nlp-crim folders. I was testing the specific image job at each minute, so the following minutes, the eo-crim/nlp-crim come back and get updated as expected. The problem then reoccurs each 5 minutes.

I think the problem is that the 'notebookdeploy' job seems to erase everything in the 'tutorial-notebooks' folder, when running the trigger-deploy-notebook script : https://github.com/bird-house/birdhouse-deploy/blob/8dd34d6d0b61e94c1c21423b8f8b7c9be4e5d9fb/birdhouse/deployment/trigger-deploy-notebook#L82

I am not sure what would be to best solution here. Maybe we need to restructure those directories. 
I would be curious about your feedback on this.
Should we close this PR first, and attack this problem in a new one?
We can still keep the feature disabled in the configs for now while we fix this part.

tlvu:
> the 'notebookdeploy' job seems to erase everything in the 'tutorial-notebooks' folder

@ChaamC 
Yes indeed, I forgot again about this little detail.  It deletes everything so that when an existing notebook is renamed or deleted, it does not leave behind the old name.

A quick solution I can propose to you is to deploy your notebooks to a different location on disk, ex: `/data/jupyterhub_user_data/tutorial-notebooks-multi-image/`.  So on disk it's different location but you can still volume-mount to the same `/notebook_dir/tutorial-notebooks/` inside the Jupyter container.

So the feature toggle `MOUNT_IMAGE_SPECIFIC_NOTEBOOKS` now will also look for the new location on disk.

I had to do the exact same trick (deploy to another location on disk `/data/jupyterhub_user_data/pavics-homepage`) to deploy our extra tutorial notebook from our new landing page in PR https://github.com/bird-house/birdhouse-deploy-ouranos/pull/8.

Good thing you tested in the real IAC env that replicate the prod env !

ChaamC:
I checked for the svn command and it doesn't seem to override the folder. If I put some random file in the folder tutorial-notebooks/eo-crim/eo, and run the deploy job with those settings :
```
source_dir: eo/notebooks
dest_sub_dir: eo
```
It just takes every file from the source_dir, and sends them to the dest_sub_dir, updating any file that already exists with the same name.
Any other file found in the dest_sub_dir will just stay there for now.

ChaamC:
Updated with new fix. 
tutorial-notebooks files related to specific images are not found on another folder on the host.
I will do a quick check in the iac too, and I think I can merge this after, if it's good for everyone!

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -86,7 +101,7 @@ c.Spawner.disable_user_config = True
 c.DockerSpawner.default_url = '/lab'
 c.DockerSpawner.remove = True  # delete containers when servers are stopped
 ${ENABLE_JUPYTERHUB_MULTI_NOTEBOOKS}    # noqa
-c.DockerSpawner.pull_policy = "always"  # for images not using pinned version
+# c.DockerSpawner.pull_policy = "always"  # for images not using pinned version
```
tlvu, line 104: Is there a reason to comment out this line?

This is there for the case when user use "latest" flavor of image so new update will not be detected.

tlvu (approved):
Looks good for me with the new location on disk for notebooks.  But I'd rather we play safe and revert the `c.DockerSpawner.pull_policy = "always"` change unless there is a compelling reason for it.

Make sure to re-test in a real IAC setup.

ChaamC:
Did more tests in IAC setup, and new fix seems to be working fine now!

I see that the E2E tests are failing, but they seems to have failed in other recent PRs too, so I suppose it is correct for me to ignore this, and to go on with the merge.

PR #151, opened 2021-04-21
update magpie v3.9.0

fmigneault:
[CHANGES](https://github.com/Ouranosinc/Magpie/blob/master/CHANGES.rst#390-2021-04-06)

Provides multiple updates related to authentication headers metadata that can be used by browsers to locate the endpoint where to authenticate.

### DIFF
--- a/birdhouse/docker-compose.yml
+++ b/birdhouse/docker-compose.yml
@@ -286,7 +286,7 @@ services:
     restart: always
 
   magpie:
-    image: pavics/magpie:3.8.0
+    image: pavics/magpie:3.9.0
     container_name: magpie
     ports:
       - "2001:2001"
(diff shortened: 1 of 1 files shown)


dbyrns (approved):
LGTM

huard (approved):

tlvu:
Just curious, why not go direct to 3.10.0 since it already exists?  This is peanuts since we just went from 1.7.3 to 3.8.0 :D

No DB change this time so I will not need to test with prod DB?

Jenkins failed, known error?

fmigneault:
@tlvu 
I would like to try with a working test from the bot (checking with @MatProv to resolve that). 
When that passes, the 3.10 can be bumped right after. 
Also, 3.10 requires a small change of config.

fmigneault:
@tlvu 
Since this PR only changes the tags, I will close it and refer directly to #143 that does the same + small patch of the gunicorn command required.

PR #155, opened 2021-04-26
proxy: allow homepage (location /) to be configurable

tlvu:
For each org to have their own custom homepage.

Existing default to redirect to JupyterHub login page is preserved for when no custom homepage is required.

See matching PR for deploying Ouranos custom homepage https://github.com/bird-house/birdhouse-deploy-ouranos/pull/11.

Test server: https://lvupavics.ouranos.ca

Fixes https://github.com/Ouranosinc/PAVICS-landing/issues/20.

### DIFF
--- a/birdhouse/config/proxy/conf.d/all-services.include.template
+++ b/birdhouse/config/proxy/conf.d/all-services.include.template
@@ -1,5 +1,5 @@
     location / {
-        return 302 https://$host/jupyter/hub/login;
+        ${PROXY_ROOT_LOCATION}
     }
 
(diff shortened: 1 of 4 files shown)


huard (approved):

dbyrns (approved):

MatProv (approved):

tlvu:
@tlogan2000 FYI, as agreed, this change will not take effect immediately on prod.  Will wait for your OK before activating it.  It is safe to merge now since the activation is via config, not in the code so we can merge the code now.

PR #161, opened 2021-05-04
🚧 changelog

### LINKED ISSUE #157: Tracking CHANGES
Sorry, I didn't think about that before `bird-house/contributions-guideline` (#156) was merged...  I would like to amend something.

What came up and made it very obvious during PRs #152, #154, and probably more to come... is that it is not easy to know what changed between versions. There are effectively no easy way to retrieve full changelogs. With upcoming releases by tagged versions and different forks across organizations, these will become ever increasingly important.

Also, there is no specific guidelines about how changelogs should be tracked, so it is very free-for-all in the commits. 
I know that @tlvu does very detailed descriptions in his PRs, but any other commit/PR contributors doesn't (including myself). 
I don't think that is the usual (natural?) way for most developers also, and I find the process of opening individual commits to read descriptions very time wasting/frustrating when I want to find when a change was introduced. 

For this reason, I would like to propose to have a standard CHANGES file, as most repos do, and have a proper listing of versions and relevant changes each time. The guideline would also add a requirement to list whatever was changed by each PR, making the info still easy to retrieve within individual merge commits. 

The biggest advantages would be: 
1. a standard format defined in CHANGES, normalizing how everyone reports them 
2. a common place where to look for them, making it easy to retrieve what happened between each version by looking through a single file.
3. easy auto doc in the ReadTheDocs, allowing to also expose the CHANGES in the published/official documentation references
4. possibility to add `bumpversion` and other useful tools if the need arises

Thanks for feedback and discussions. 
Whomever it concerns: @huard @tlvu @dbyrns @MatProv  
(tag any others as needed)

fmigneault:
# Overview

Adds CHANGES history listing extracted by existing release/tags for past versions. 
Generated from this script: https://github.com/fmigneault/script-utils/blob/git-tools/git-tools/changes.py 

See the results: https://github.com/bird-house/birdhouse-deploy/tree/changes

## Changes

**Non-breaking changes**
- Adds CHANGES.md (generated from above script)
- Add auto-bump for future changes at deploy/tag time
- Add tags and shields in readme

**Breaking changes**
n/a

## Related Issue / Discussion

- Resolves #157 

## TODO

- [x] Re-run script / sync with latest updates to CHANGES.rst since more merges do occur until this PR is integrated


tlvu:
Good stuff here recreating the CHANGES.rst from tagged commit.

tlvu:
Sorry I didn't know you were waiting on this PR since it's in draft mode, I thought it was not ready for a proper review.

So what's the procedure to "refresh" the generated changelog?   A few newer tags have been since this was generated and need a refresh.  I can do this refresh and quick review of the generated content.

Are there more automation required with bumping tags now?  I think you probably have a working recipe from Magpie you can backport here?

fmigneault:
We only need to add a config similar to this one: 
https://github.com/Ouranosinc/Magpie/blob/master/setup.cfg#L1-L30

Then, anything you where normally doing `git tag "<tag>"`, you will run instead one of the following: 
```shell
bump2version --new-version <tag> patch  # bump current version to specific new version 
bump2version <part>                     # bump the specific <part> (major|minor|patch) of current version by 1
```

That will update the items defined in `setup.cfg`, generate the new changelog entry for future *unreleased* changes, make a `bump <current-version> -> <new-version>` message and tag `<new-version>`. 

Devs only need to provide their "changes" items in that *unreleased* section for them to be included in the next version bump.

dbyrns (approved):
This is a good thing to have here.
Just take a look at the m2r extension issue.

CHANGES.md
```diff
@@ -0,0 +1,2049 @@
+# Changes
+
+[//]: # **DEFINE LATEST CHANGES UNDER BELOW 'Unreleased' SECTION - THEY WILL BE INTEGRATED IN NEXT RELEASE**
+
+## [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
+
+[//]: # list changes here, using '-' for each new entry (remove this when items are added)
```
tlvu, line 7: So should you be summarizing this PR in here so next time we run the bump2release command the changelog is already up-to-date?

Going forward, when opening a PR we should also include changelog update in the same PR if I understand correctly?

Another "procedure" question.  Before, I used to tag after the PR is merged, since there is no new commit involved.  Now with bump2version, it will modify a few files, so should the bump happen when the PR is approved and right before the PR is merged?  Otherwise, it will cost another PR to merge the modified files!

tlvu (approved):
Great job!  I have a few clean up in the changelog to do but I'll do it in a different PR so this one can be merged quickly so we do not have to regenerate the changelog file again.

I have a few "procedure question" going forward, otherwise all good, merge when you are ready (and as usual copy/paste the PR description into the merge commit description, thanks).

tlvu:
Went to check RtD, it's awsome that the changelog shows up there as well.

CHANGES.md
fmigneault: Yes, I actually need to add my changelog update here. 
It should be added in the same PR. I will add it too in the following commit.

For the tag, this is actually up to us to define how we want to proceed. Some possible alternatives: 

1) We bump in the PR itself as final commit before merge. The tag would be applied one commit before the merge-commit. 
Personally, I don't like this too much, because if we had another commit to the PR after bumping by mistake (or to patch something), it will be technically left out of the tagged version.

2) We modify `.bumpversion.cfg` option `tag = true` to `tag = false`. We then run `bump2version` as normal toward the end of the PR, but the tag is not created. After merge, we push the corresponding tag manually onto the merge-commit.
The downside of this could be that 2 PRs being worked on at the same time and about to be merged will cause some version conflict. The 2nd PR would have to push another commit to patch the erroneous version to one above.

3) We proceed like I do in Magpie, the PR is merged as is without bump. Everything remains  in "unreleased" until we decide to bump, which can be following every single PR, or after an aggregation of a few features as desired. When we decide to "release" a tag with one or more features, we run bump2version on the merged master and push directly the update. Since only versioning items in README and CHANGES files are modified, I don't think it is too critical to do this, but I can understand if there is opposition to do it this way.

Let me know which (or other?) method you would prefer since you will be doing it most of the time.

tlvu: Option 3 (all release activities after the merge) is most similar to what I do now.  But given this repo directly impact deployment, we have set a restriction to not allow direct push to `master`, requiring PR at all time and at least one approuver to merge the PR.  This option means another PR !

Option 1, like you said, might put the tag not at the end of the PR, diffing between tags later will be error prone.

I guess we are left with Option 2.  By the time we have too frequent PR merge race we can re-consider Option 3.

fmigneault: Sounds good. I'll push the update for option 2.

CONTRIBUTING.rst
```diff
@@ -23,35 +23,64 @@ Policy objectives
...
+
+6. Each organization wishing to review the changes has the duty to do so within a reasonable period of time (7 days)
+   or to indicate its intention to do so later with reasonable reasons (e.g., vacation). After this time, its implicit
+   support will be considered. It will be assumed that the organization agrees to the changes, and they will get merged
+   without further notice.
+
+7. To encourage review in a timely manner, contributions should be simple and focused (do not mix multiple goals) and
+   well explained (describe the "why" and the "impact/behavior change" on existing production deployment, as requested
+   in the contribution template).
+   They should also include a description of provided modifications and fixes under the active ``Unreleased`` section
```
tlvu, line 58: If possible, can you perform "reformating" in a separate commit so it's easier to spot what has changed by the reformatting and what has really changed by manual update.

tlvu (approved):
All good thanks!  Minor comment for next time.

fmigneault:
@matprov 
Maybe you can check. I'm not sure what is the problem: 
http://daccs-jenkins.crim.ca/job/PAVICS-e2e-workflow-tests/job/master/479/console
Some esgf test started failing recently, but wasn't a few hours ago. 
There is nothing different in master for the last 12 days and this PR doesn't change anything to the stack itself.

fmigneault:
For reference, sample output using dry-run: 

```
❯ bump2version --dry-run --verbose patch
Reading config file .bumpversion.cfg:
[bumpversion]
current_version = 1.13.9
commit = True
# don't tag automatically, do it manually after PR is merged and approved to apply directly on merge-commit
# see: https://github.com/bird-house/birdhouse-deploy/pull/161#discussion_r661746230
tag = False
tag_name = {new_version}

# following moves items listed under the ``Unreleased`` section under the generated version release,
# and generates a fresh one for future PR updates
# note:
#       because '#' at the start of line (markdown header at level 1) is interpreted as comments here, must use equals
#       (see: https://github.com/c4urself/bump2version/issues/99)
[bumpversion:file:CHANGES.md]
search =
        ==================================================================================================================
replace =
        ==================================================================================================================

        [{new_version}](https://github.com/bird-house/birdhouse-deploy/tree/{new_version}) ({now:%%Y-%%m-%%d})
        ==================================================================================================================

[bumpversion:file:README.rst]
search = {current_version}
replace = {new_version}

Attempting to increment part 'patch'
Values are now: major=1, minor=13, patch=10
Dry run active, won't touch any files.
New version will be '1.13.10'
Asserting files CHANGES.md, README.rst contain the version string...
current_version=1.13.9
commit=True
tag=False
tag_name={new_version}
new_version=1.13.10
Would write to config file .bumpversion.cfg:
[bumpversion]
current_version = 1.13.10
commit = True
tag = False
tag_name = {new_version}

[bumpversion:file:CHANGES.md]
search = 
        ==================================================================================================================
replace = 
        ==================================================================================================================
        
        [{new_version}](https://github.com/bird-house/birdhouse-deploy/tree/{new_version}) ({now:%%Y-%%m-%%d})
        ==================================================================================================================

[bumpversion:file:README.rst]
search = {current_version}
replace = {new_version}


Would prepare Git commit
Would add changes in file 'CHANGES.md' to Git
Would add changes in file 'README.rst' to Git
Would add changes in file '.bumpversion.cfg' to Git
Would commit to Git with message 'Bump version: 1.13.9 → 1.13.10'
Would tag '1.13.10' with message 'Bump version: 1.13.9 → 1.13.10' in Git and not signing

```

matprov:
> Some esgf test started failing recently, but wasn't a few hours ago.
> There is nothing different in master for the last 12 days and this PR doesn't change anything to the stack itself.

@fmigneault I've seen ESGF tests failing randomly over the past months - they aren't merge blockers.
Their host might sometimes be unreachable for few minutes here and there. 
Only thing we can do is retrigger the test suite when it happens. Otherwise, simply ignore.

tlvu:
Yes ESGF servers sometime not responding.



Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

tlvu:
@fmigneault 

Follow new release process for PR https://github.com/bird-house/birdhouse-deploy/pull/180, works perfectly !

Notice whitespace change in `.bumpversion.cfg` in commit https://github.com/bird-house/birdhouse-deploy/pull/180/commits/1353d263064e83208ecbef440be25b11c7eb25a8, is that expected?

fmigneault:
@tlvu 
Yes, bumpversion rewrites it differently than what most IDE/linters generate after striping trailing spaces.

Issue #67, opened 2020-09-03
Idle Jupyter server cause open file limit exhaustion


tlvu:
This morning, none of the docker commands was responding because we have exhausted the open file limit on the user that runs the PAVICS platform (the user that does `./pavics-compose.sh up -d`).

```
$ ./pavics-compose.sh ps
-bash: fork: retry: No child processes
-bash: fork: retry: No child processes
-bash: fork: retry: No child processes
-bash: fork: retry: No child processes
-bash: fork: Resource temporarily unavailable
```

The immediate work-around is to increase the soft and hard nofile limit for the corresponding user in `/etc/security/limits.conf` file and apply that new limit immediately `ulimit -n NEW_LIMIT` (`ulimit -n` show the current effective limit).  Find the current number of open file this way `sudo lsof -u $USER|wc -l` and put something higher than that in the `limits.conf` file. Reference https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/5/html/tuning_and_optimizing_red_hat_enterprise_linux_for_oracle_9i_and_10g_databases/chap-oracle_9i_and_10g_tuning_guide-setting_shell_limits_for_the_oracle_user

Then using a different user than the user that usually runs PAVICS because that user can not do anything else and restart the docker daemon (`sudo systemctl restart docker`).  Back to the regular user running PAVICS, ff the containers have problems to restart, destroy them and re-create them from scratch (`./pavics-compose.sh down && sleep 10 && ./pavics-compose.sh up -d`).

For a more permanent solution than keep increasing the limit each time we burst it is to setup culling idle Jupyter server as describe here https://discourse.jupyter.org/t/jupyterhub-doesnt-kill-processes-and-threads-when-notebooks-are-closed-or-user-log-out/2244/2

Also need to add monitoring for open file limit to we are alerted in advance of near exhaustion to avoid having to restart the entire docker daemon.

Ping @moulab88 @tlogan2000 if you guys have anything to add.

Edit:
* add command to see current effective limit and set new limit immediately
* add reference to redhat docs

tlogan2000:
[...]

PR #177, opened 2021-06-09
jupyterhub: allow config override via env.local

tlvu:
## Overview

This is basically the same as `ENABLE_JUPYTERHUB_MULTI_NOTEBOOKS` but at the bottom of the file so it can override everything.

`ENABLE_JUPYTERHUB_MULTI_NOTEBOOKS` is kept for backward-compat.

First useful application is to enable server culling for auto shutdown of idle kernels and idle jupyter single server, hopefully fixes https://github.com/bird-house/birdhouse-deploy/issues/67.

The culling settings will only take effect the next time user restart their personal Jupyter server because it seems that the Jupyter server is the one culling itself.  JupyterHub do not perform the culling, it simply forward the culling settings to the Jupyter server.

```sh
$ docker inspect jupyter-lvu --format '{{ .Args }}'
[run -n birdy /usr/local/bin/start-notebook.sh --ip=0.0.0.0 --port=8888 --notebook-dir=/notebook_dir --SingleUserNotebookApp.default_url=/lab --debug --disable-user-config --NotebookApp.terminals_enabled=False --NotebookApp.shutdown_no_activity_timeout=180 --MappingKernelManager.cull_idle_timeout=180 --MappingKernelManager.cull_connected=True]
```

## Changes

**Non-breaking changes**
- jupyterhub: allow config override via env.local

## Tests

Deployed to https://lvupavicsdev.ouranos.ca/jupyter (timeout set to 5 mins)

FYI @tlogan2000 @dbyrns @nikola-rados @ChaamC


matprov (approved):

tlvu:
FYI: The culling settings will only take effect the next time user restart their personal Jupyter server because it seems that the Jupyter server is the one culling itself. JupyterHub do not perform the culling, it simply forward the culling settings to the Jupyter server.

PR #205, opened 2021-11-02
fix mongodb network

fmigneault:
## Overview

Move `mongodb` network to `mongodb` image directly and update `phoenix`/`weaver` images accordingly.

## Changes

**Non-breaking changes**
- Apply ``mongodb`` network to ``mongodb`` image in order to allow ``phoenix`` to properly reference it.
- Remove ``mongodb`` definition from ``./components/weaver`` since the extended ``mongodb`` network is already provided.

**Breaking changes**
- n/a

## Related Issue / Discussion

- Resolves https://github.com/bird-house/birdhouse-deploy/pull/114#discussion_r740666666

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.17.0
+current_version = 1.17.1
 commit = True
 tag = False
 tag_name = {new_version}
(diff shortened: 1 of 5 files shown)


tlvu:
FYI `weaver_example.ipynb` is failing, is that normal?

fmigneault:
@tlvu  

> FYI `weaver_example.ipynb` is failing, is that normal?

That's very weird because the collected notebook reach the end with all outputs with "success" contents.
I'll have to investigate why some asserts don't succeed...

PR #225, opened 2021-12-03
bump weaver 4.5.0 and relevant mongodb changes

fmigneault:
## Changes
- Upgrade default `Weaver` version to [4.5.0](https://github.com/crim-ca/weaver/blob/master/CHANGES.rst#450-2021-11-25)   (from [4.2.1](https://github.com/crim-ca/weaver/blob/master/CHANGES.rst#421-2021-10-20)) for new features and fixes.

**Non-breaking changes**
  - Adds support of `X-WPS-Output-Context` header to define the WPS output nested directory (for user context).
  - Adds support of `X-Auth-Docker` header to define a private Docker registry authentication token when the 
    referenced Docker image in the deployed Application Package requires it to fetch it for Process execution. 
  - Require `MongoDB==5.0` Docker image for Weaver's database.
  - Fixes related to handling `dismiss` operation of job executions and retrieval of their results.
  - Fixes related to fetching remote files and propagation of intermediate results between Workflow steps.

**Breaking changes**
- New MongoDB version 5.0 (as separate image for Weaver, not bump of the 3.4 used by Phoenix)

	Because of the new `MongoDB==5.0` database requirement for Weaver that uses (potentially) distinct version from other 
	birds (notably `phoenix` with `MongoDB==3.4`), a separate Docker image is employed only for Weaver. If some processes, 
	jobs, or other Weaver-related data was already defined on one of your server instances, manual transfer between the 
	generic `${DATA_PERSIST_ROOT}/mongodb_persist` to new  `${DATA_PERSIST_ROOT}/mongodb_weaver_persist` directory must 
	be accomplished. The data in the new directory should then be migrated to the new version following the procedure 
	described in [Database Migration](https://pavics-weaver.readthedocs.io/en/latest/installation.html?#database-migration).

**Legal Notice**
- While migrating from ``MongoDB==3.4`` to ``MongoDB==5.0``, its license changes from AGPL to SSPL
(reference: [mongodb/mongo@6ea81c8/README#L89-L95](https://github.com/mongodb/mongo/blob/6ea81c88/README#L89-L95)).
This should not impact users using the platform for public and Open Source uses, but should be considered otherwise.

## Related Issue / Discussion

- Resolves https://www.crim.ca/jira/browse/DAC-187 *Update weaver so that outputs can be saved into the user workspace*

### DIFF
--- a/birdhouse/components/weaver/docker-compose-extra.yml
+++ b/birdhouse/components/weaver/docker-compose-extra.yml
@@ -47,9 +47,9 @@ services:
     #env_file:
     #  - ./config/mongodb/credentials.env
     depends_on:
-      - mongodb
+      - weaver-mongodb
     links:
-      - mongodb
+      - weaver-mongodb
     networks:
       - default
       - mongodb
@@ -74,11 +74,11 @@ services:
     #  - ./config/mongodb/credentials.env
     depends_on:
       # no 'default' network here such that the worker is not accessible from proxy/http
-      - mongodb
+      - weaver-mongodb
       - weaver  # if not started first, sometimes celery misbehaves and will not pick jobs in queue
       - docker-proxy
     links:
-      - mongodb
+      - weaver-mongodb
     networks:
       # WARNING: *DO NOT* provide 'default' network
       #   Not providing 'default' network ensure that only connection with other images (birds, proxy) can be
@@ -113,6 +113,19 @@ services:
       interval: 60s
       timeout: 10s
       retries: 2
+
+  # Dedicated database for Weaver, since other 'mongodb' image does not employ the same version.
+  weaver-mongodb:
+    image: mongo:5.0
+    container_name: weaver-mongodb
+    networks:
+      - mongodb
+    volumes:
+      - ${DATA_PERSIST_ROOT}/mongodb_weaver_persist:/data/db
+    # Mongodb crash with permission denied errors if the command is not overridden like this
+    command: bash -c 'chown -R mongodb:mongodb /data && chmod -R 755 /data && mongod'
+    restart: always
+    logging: *default-logging
 
   # Image 'docker-proxy' is the connector between docker image 'worker' jobs and their sibling docker executions
   docker-proxy:
--- a/birdhouse/components/weaver/docker-compose-extra.yml
+++ b/birdhouse/components/weaver/docker-compose-extra.yml
@@ -116,7 +116,7 @@ services:
 
   # Dedicated database for Weaver, since other 'mongodb' image does not employ the same version.
   weaver-mongodb:
-    image: mongo:5.0
+    image: mongo:${WEAVER_MONGODB_VERSION}
     container_name: weaver-mongodb
     networks:
       - weaver-mongodb
--- a/birdhouse/components/weaver/docker-compose-extra.yml
+++ b/birdhouse/components/weaver/docker-compose-extra.yml
@@ -52,7 +52,7 @@ services:
       - weaver-mongodb
     networks:
       - default
-      - mongodb
+      - weaver-mongodb
     volumes:
       - ./components/weaver/config/weaver/weaver.ini:/opt/local/src/weaver/config/weaver.ini:ro
       - ./components/weaver/config/weaver/data_sources.yml:/opt/local/src/weaver/config/data_sources.yml:ro
@@ -84,7 +84,7 @@ services:
       #   Not providing 'default' network ensure that only connection with other images (birds, proxy) can be
       #   accomplished through Weaver API which cannot run sibling dockers by itself. Isolate Dockers execution.
       - docker-proxy
-      - mongodb
+      - weaver-mongodb
     volumes:
       #- wps_private:/data/wps_private
       - ${DATA_PERSIST_ROOT}/wps_private:/data/wps_private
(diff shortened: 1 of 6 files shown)


birdhouse/components/weaver/docker-compose-extra.yml
```diff
@@ -114,6 +114,19 @@ services:
       timeout: 10s
       retries: 2
 
+  # Dedicated database for Weaver, since other 'mongodb' image does not employ the same version.
+  weaver-mongodb:
+    image: mongo:5.0
```
tlvu, line 119: Want to put that mongo image version into weaver/default.env as well so it is configurable and follow the same convention as weaver.  Is that a specific version or a "rolling" version?  Let's use a specific version for reproductibility.

birdhouse/components/weaver/docker-compose-extra.yml
```diff
@@ -114,6 +114,19 @@ services:
...
+    container_name: weaver-mongodb
+    networks:
+      - mongodb
```
tlvu, line 122: I think this is the same network name as the other mongodb !  Better not clash with the other one.

tlvu (approved):
LGTM with minor change suggested.  Happy to see the Weaver component is completely independent now.

You'll have to `bump2version minor` for this one since technically for those who had previously enable Weaver, they will need a manual data migration.

birdhouse/components/weaver/docker-compose-extra.yml
fmigneault: I'm not sure what you mean by rolling version, but it is a minimum requirement for Weaver.
I didn't add a default.env parameter for this reason.

birdhouse/components/weaver/docker-compose-extra.yml
fmigneault: I'm not sure if it would clash or not, but I'll separate it just to be sure.

birdhouse/components/weaver/docker-compose-extra.yml
tlvu: On https://hub.docker.com/_/mongo?tab=tags, currently there is `5.0.4`,  `5.0` and `5`.  I suspect `5.0` and `5` are rolling tag (like the `latest` tag) and only `5.0.4` is a specific tag that will not change with time.

For reproductibility, we want to avoid rolling tags, like we avoided the `latest` tag.

Adding to `default.env` will allow you to override that tag for testing a newer version easier, if needed.

fmigneault: I see what you mean now. 
I can add the variable to allow more specific version overrides, but I believe those patch version are often for security fixes that are better to obtain as soon as possible.
Note that all other dbs `mongo:3.4` and `postgres:9.6` (generic and magpie one) also use those rolling versions.

tlvu: > I believe those patch version are often for security fixes that are better to obtain as soon as possible.

The problem is docker do not automatically detect tag override and re-download them so we are falsely leading the user to believe they are getting rolling release update.

Also using rolling release means no reproductibility.

> Note that all other dbs `mongo:3.4` and `postgres:9.6` (generic and magpie one) also use those rolling versions.

Indeed, old code.  New code, we should be more disciplined.  Pretty sure in the past I've mistaken rolling tag for exact tag as well so it's something we have to be careful.

Not reproducible deployment means we can lose lots of time chasing false leads, like I've learned the hard way with Geoserver upgrade, see https://github.com/kartoza/docker-geoserver/issues/232

dbyrns:
Look good to me, but I wonder if we should include a notice about the new mongo licence for releases since October 16, 2018 (3.4.0 was still under AGPL, but 5.0.4 is under SSPL). Birdhouse is meant to be public so it should not be a problem. 
Reference : https://github.com/mongodb/mongo/blob/6ea81c883e7297be99884185c908c7ece385caf8/README#L89-L95

dbyrns (approved):

fmigneault:
> Look good to me, but I wonder if we should include a notice about the new mongo licence for releases since October 16, 2018 (3.4.0 was still under AGPL, but 5.0.4 is under SSPL). Birdhouse is meant to be public so it should not be a problem. Reference : [mongodb/mongo@`6ea81c8`/README#L89-L95](https://github.com/mongodb/mongo/blob/6ea81c883e7297be99884185c908c7ece385caf8/README#L89-L95)

I'm trying to see what this change of license implies. 
Referring to this: https://opensource.stackexchange.com/questions/8025/difference-between-mongodb-sspl-and-gnu-agpl
It looks like the change only makes the license even more public, so it shouldn't be a problem.

dbyrns:
> It looks like the change only makes the license even more public, so it shouldn't be a problem.

Exact, for us it doesn't change anything as everything is open. But for a potential Weaver or Birdhouse users it could be a deal breaker as they couldn't use the current stack along with private components. But maybe some other components are already under this kind of licence, I'm just hyper vigilent about mongo since I read about this change.

So, maybe a simple notice in the changes.md along the new mongo version would be enough.

tlvu:
> I'm just hyper vigilent about mongo since I read about this change.
> 
> So, maybe a simple notice in the changes.md along the new mongo version would be enough.

@dbyrns Thanks for catching this.  Legal stuff is not my strong suit.  But it's something we should care about.

PR #236, opened 2022-02-28
update weaver 4.5.0 -> 4.12.0

fmigneault:
## Overview

Weaver: update `weaver` component from [4.5.0](https://github.com/crim-ca/weaver/tree/4.5.0) to [4.12.0](https://github.com/crim-ca/weaver/tree/4.12.0).

## Changes

**Non-breaking changes**
  - Adds `WeaverClient` and *Weaver CLI*. Although not strictly employed by the platform itself to offer *Weaver* as a
    service, these can be employed to interact with *Weaver* using Python or shell commands, providing access to all 
    WPS *birds* offered by the platform using the common [OGC API - Processes][ogcapi-proc] interface through 
    [Weaver Providers](https://pavics-weaver.readthedocs.io/en/latest/processes.html#remote-provider).
  - Adds [Vault](https://pavics-weaver.readthedocs.io/en/latest/processes.html#vault) functionality allowing temporary
    and secure storage to upload files for single-use process execution.
  - Various bugfixes and conformance resolution related to [OGC API - Processes][ogcapi-proc].
  - Fix `weaver-mongodb` link references for `weaver-worker`. New default variables `WEAVER_MONGODB_[HOST|PORT|URL]`
    are defined to construct different INI configuration formats employed by `weaver` and `weaver-worker` images.
  - Fix missing `EXTRA_VARS` variables in [Weaver's default.env](./birdhouse/components/weaver/default.env).

**Breaking changes**
- n/a

## Related Issue / Discussion

- Resolves https://www.crim.ca/jira/browse/DAC-455

### DIFF
--- a/birdhouse/components/weaver/config/weaver/weaver.ini.template
+++ b/birdhouse/components/weaver/config/weaver/weaver.ini.template
@@ -18,8 +18,8 @@ pyramid.debug_routematch = false
 pyramid.default_locale_name = en
 
 # mongodb
-mongodb.host = weaver-mongodb
-mongodb.port = 27017
+mongodb.host = ${WEAVER_MONGODB_HOST}
+mongodb.port = ${WEAVER_MONGODB_PORT}
 mongodb.db_name = weaver
 
 # --- Weaver Configuration ---
@@ -102,8 +102,8 @@ weaver.extra_options =
 ###
 [celery]
 use_celeryconfig = false
-broker_url = mongodb://mongodb:27017/celery
-result_backend = mongodb://mongodb:27017/
+broker_url = ${WEAVER_MONGODB_URL}/celery
+result_backend = ${WEAVER_MONGODB_URL}/
 
 # following is technically redundant, but Weaver-API sporadically doesn't correctly discover tasks
 # enforce the import location to make sure they are detected and tasks submission can be registered in workers
--- a/birdhouse/components/weaver/config/weaver/weaver.ini.template
+++ b/birdhouse/components/weaver/config/weaver/weaver.ini.template
@@ -85,6 +85,10 @@ weaver.wps_email_notify_ssl = true
 weaver.wps_email_notify_template_dir =
 weaver.wps_email_notify_template_default =
 
+# --- Weaver vault configuration ---
+weaver.vault = true
+weaver.vault_dir = /tmp/vault
+
 # --- Weaver other configurations ---
 # default WPS processes to always preload
 weaver.wps_processes_file = wps_processes.yml
--- a/birdhouse/components/weaver/default.env
+++ b/birdhouse/components/weaver/default.env
@@ -6,6 +6,11 @@
 # single quotes are important in below list to keep variable names intact until 'pavics-compose' parses them
 EXTRA_VARS='
   $WEAVER_CONFIG
+  $WEAVER_VERSION
+  $WEAVER_MONGODB_VERSION
+  $WEAVER_MONGODB_HOST
+  $WEAVER_MONGODB_PORT
+  $WEAVER_MONGODB_URL
   $WEAVER_MANAGER_NAME
   $WEAVER_WORKER_NAME
   $WEAVER_WPS_NAME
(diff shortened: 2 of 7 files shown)


matprov (approved):

f-PLT:
I spotted something that we missed earlier on:

https://github.com/bird-house/birdhouse-deploy/blob/weaver-update/birdhouse/components/weaver/config/weaver/weaver.ini.template#L105-L106

The `weaver.ini.template` still uses the main `mongodb`, but it should be switched to the new `weaver-mongodb`

Trying it out, I also had to add `--bind_ip 0.0.0.0` in weaver's `docker-compose-extra.yml` to the container's command : https://github.com/bird-house/birdhouse-deploy/blob/weaver-update/birdhouse/components/weaver/docker-compose-extra.yml#L126. Without it, `weaver-worker` had a connection refused error trying to access `weaver-mongodb`

fmigneault:
@f-PLT 

> I spotted something that we missed earlier on:
> 
> [https://github.com/bird-house/birdhouse-deploy/blob/weaver-update/birdhouse/components/weaver/config/weaver/weaver.ini.template#L105-L106](https://github.com/bird-house/birdhouse-deploy/blob/weaver-update/birdhouse/components/weaver/config/weaver/weaver.ini.template?rgh-link-date=2022-02-28T20%3A10%3A34Z#L105-L106)

Well spotted. 

Regarding the `command`, I think it can be removed entirely. The override was for an issue specific in `mongo:3.4` (or maybe even older?), but with more recent mongo, the entrypoint is a script that setups a lot of extra stuff before launching mongo.

fmigneault:
run tests

f-PLT (approved):
`celery-healthcheck` succeeds, but the docker-compose's timeout for that healthcheck is too low on my deployed instance.

I still got an `unhealthy` result from the the container. I tried with a 15 secs instead of 10 and finally got a `healthy` output.

(https://github.com/bird-house/birdhouse-deploy/blob/weaver-update/birdhouse/components/weaver/docker-compose-extra.yml#L114)

If this is also an issue on other deployments, I'd consider setting this timeout higher.

Otherwise, I think this can be merged once the E2E issue is fixed.

birdhouse/components/weaver/default.env
```diff
@@ -6,6 +6,11 @@
 # single quotes are important in below list to keep variable names intact until 'pavics-compose' parses them
 EXTRA_VARS='
   $WEAVER_CONFIG
+  $WEAVER_VERSION
+  $WEAVER_MONGODB_VERSION
```
tlvu, line 10: `WEAVER_VERSION` and `WEAVER_MONGODB_VERSION` not used in `weaver.ini.template` in this PR.  Are they used outside of this PR for any sort of templating expansion?  If not then they do not have to be added to `EXTRA_VARS`, please remove.

birdhouse/components/weaver/config/weaver/weaver.ini.template
```diff
@@ -85,6 +85,10 @@ weaver.wps_email_notify_ssl = true
 weaver.wps_email_notify_template_dir =
 weaver.wps_email_notify_template_default =
 
+# --- Weaver vault configuration ---
+weaver.vault = true
+weaver.vault_dir = /tmp/vault
```
tlvu, line 90: The default `weaver.vault_dir` is under `/tmp`?

Shouldn't it be somewhere else that is persisted to disk?  Just my 2 cents. I do not really know what this does.

tlvu (approved):
LGTM, minor comments.  Not testing to deploy this for real since weaver is not enabled in production for Ouranos.

birdhouse/components/weaver/default.env
fmigneault: They are used in the extra compose of weaver component.

birdhouse/components/weaver/config/weaver/weaver.ini.template
fmigneault: No. In it in fact better to not have it persisted because it is intended for single use and forget situations to transfer secrets or other sensible data as input for some process execution.

PR #237, opened 2022-02-28
update magpie 3.19.1 -> 3.21.0

fmigneault:
## Overview

Magpie: update `magpie` service from [3.19.1](https://github.com/Ouranosinc/Magpie/tree/3.19.1) to [3.21.0](https://github.com/Ouranosinc/Magpie/tree/3.21.0).

## Changes

**Non-breaking changes**
  - Update *WFS*, *WMS* and *WPS* related services to properly implement the relevant *Permissions* and *Resources*
    according to their specific implementation details. For example, *GeoServer*-based *WMS* implementation supports
    *Workspaces* and additional operations that are not offered by standard *OGC*-based *WMS*. Some of these
    implementation specific operations can be taken advantage of with improved *Permissions* and *Resources* resolution.
  - Add multi-*Resource* effective access resolution for *Services* that support it.
    For example, accessing multiple *Layers* under a permission-restricted *WFS* with parameters that allow multiple
    values within a single request is now possible, if the user is granted to all specified *Resources*.
    Previously, users would require to access each *Layer Resource* individually with distinct requests.
  - Magpie's API and UI are more verbose about supported hierarchical *Resource* structure under a given *Service* type.
    When creating *Resources*, specific structures have to be respected, and only valid cases are proposed in the UI.
  - Minor UI fixes.

**Breaking changes**
- None expected.

  _**Note:**_ 
  Modified services by the new version add permissions that were previously missing, or remove permissions that would yield an error anyway since the corresponding operations are undefined in the underlying service. 
  If issues are faced due to those changes, it is most probably because incorrect service types were employed, or unexpected/unintended behaviour were taken advantage of.

## Related Issue / Discussion

- Resolves https://www.crim.ca/jira/browse/DAC-151

PR #242, opened 2022-04-14
Add secure Geoserver access

francisPLT:
## Overview

I added a new provider and location as an optional component, `test-geoserver-secured-access`, in order to test secured access to Geoserver before we completely move Geoserver behind Twitcher (undetermined date).

The old `/geoserver` path is still available, so current workflows should not be affected.

The new `/geoserver-secured` path is now available for testing once the optional component is activated.

To test the `geoserver-secured` service through Magpie, each workspace needs to be added to the new service and then permissions can be set on a per-workspace or even layer basis.

Service configuration:

image

Ex: for the Anonymous Group configuration:

image

From the above example, a `GetFeature` request for a layer in the `public` workspace will succeed for any user using any of these two request types:

- {BASE_URL}/geoserver-secured/wfs?version=2.0.0&request=GetFeature&typeNames=**public**:{LAYER_NAME}
- {BASE_URL}/geoserver-secured/**public**/wfs?version=2.0.0&request=GetFeature&typeNames={LAYER_NAME}

Whereas access to the `private` workspace will require a user or group be given explicit permissions.

Mind you, I'm not an expert with Nginx, so I based these changes from the existing configurations.

## Changes

**Non-breaking changes**
- New Provider (Magpie) : `geoserver-secured`
- New Location (Proxy)  : `/geoserver-secured`
- Copied current WFS GetCapabilities and DescribeFeatureType permissions to new Provider

**Breaking changes**
- None anticipated

## Related Issue / Discussion

- This PR in essentially an optional component to test and take advantage of the changes made by this PR : https://github.com/bird-house/birdhouse-deploy/pull/237

## Future considerations

- Moving Geoserver behind Twitcher will affect how it's admin web console can be accessed (`{BASE_URL}/geoserver/web`). I thought I had found a way a quick way to manage permissions to it, but further testing proved me wrong. All the internal redirections while navigating the web console will probably create issues, and the UI itself did not appreciate the proxy/redirect. I have not looked into this further for the time being.
- We will also need to evaluate if performance is impacted with Geoserver behind Twitcher.


dbyrns (approved):
Given that it's a non-breaking change and that it's goal is to test the geoserver services protection I don't see any issue in this PR.

tlvu (changes requested):
I would like to ask for a favor, could you make this change an optional component (under https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse/optional-components dir)?

The reason is I already planned to eventually break out every components into separate components to allow for each org to enable only the parts they need so I'd rather have any new changes aligned with the future plan.

You can look into https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse/optional-components/generic_bird for sample how it's done. It has hooks into the front Nginx proxy and Magpie as well so it's a pretty complete example to based upon to save you time.

Document your new optional component the README.rst at the root of the `optional-components/` dir.

Also, do not forget to prepare the https://github.com/bird-house/birdhouse-deploy/blob/master/CHANGES.md file so the PR is complete for eventual release.

The full release procedure is documented here https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse#release-procedure

Sorry for the last minute refactoring request.  DavidB has briefly mentioned this upcoming feature during our previous bi-weekly but I forgot to inquire about the implementation whether it will be a "component".

tlvu (approved):
Oh sorry I did not see you had done the optional-component.

Everything looks good, one minor thing missing is updating the doc of the optional-components https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/optional-components/README.rst to list and describe your component.

See many example there for the format.

Should do this after merging with master. I think FrancisM has added a new optional-component lately as well.

francisPLT:
@tlvu Just to be sure, since I don't do merges in this repo often, all that's left is to run bumpversion before merging the pull request?

tlvu:
> @tlvu Just to be sure, since I don't do merges in this repo often, all that's left is to run bumpversion before merging the pull request?

https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/README.rst#release-procedure  (added recently after you opened this PR so you might not be aware).

You are basically at this last step "Only when you are ready to merge the PR immediately, ...".  Follow from there.  If something unclear, let me know.

fmigneault:
@francisPLT 
Remove the `1.22.0` tag. 
Update the branch with latest I just merged.
Then, once merged, push a tag directly on the merged commit.

francisPLT:
> @francisPLT Remove the `1.22.0` tag. Update the branch with latest I just merged. Then, once merged, push a tag directly on the merged commit.

Yes, I just saw that, working on it

birdhouse/config/canarie-api/docker_configuration.py.template
```diff
@@ -242,7 +242,7 @@ PLATFORMS = {
         'info': {
             'name': 'PAVICS',
             'synopsis': 'The PAVICS (Power Analytics for Visualization of Climate Science) platform is a collection of climate analysis services served through Open Geospatial Consortium (OGC) protocols. These services include data access, processing and visualization. Both data and algorithms can be accessed either programmatically, through OGC-compliant clients such as QGIS or ArcGIS, or a custom web interface.',
-            'version': '1.21.1',
+            'version': '1.22.0',
             'releaseTime': '2022-08-24T15:43:51Z',
```
tlvu, line 246: @fmigneault release time did not update.  Is there a special option we need to run `bumpversion` with, for release time to update?  If yes, we need to doc this special option.  I just checked one of my previous release, the release time did not update for me as well.  Or need a special version of `bumpversion`?

tlvu:
@francisPLT FYI, the tag should be on the merge commit created by merging this PR, not on the bump version commit.  It's okay, just for next time.

tlvu:
> @francisPLT FYI, the tag should be on the merge commit created by merging this PR, not on the bump version commit. It's okay, just for next time.

@francisPLT Just to be clear, do not touch the tag at the wrong location anymore (FrancisM said to remove the tag at the wrong location and redo it but I've merged this to our production fork already and sync the tags, so too late to fix the wrong location of the tag).

birdhouse/config/canarie-api/docker_configuration.py.template
fmigneault: Seems like the portion https://github.com/bird-house/birdhouse-deploy/blob/master/.bumpversion.cfg#L28-L29 cannot update itself because bumpversion rewrites the `.bumpversion.cfg` file after applying changes using what it initially read. So the update works the first time when datetime match, but fail on following cases because the new datetime was not applied. 

I think we could hack around this using a separate file that contains only the release time. Bumpversion could modify that file without problem. It would need to be mounted in the docker image for the python config to read it.

fmigneault:
@francisPLT Will need to apply this new component to `EXTRA_CONF_DIRS` in the daccs-configs repository to apply them on new instances.

PR #244, opened 2022-05-12
magpie/twitcher/weaver requests hooks for WPS outputs in user workspace

fmigneault:
## Overview

Prepare magpie/twitcher/weaver requests hooks for WPS outputs in user workspace directory 

## Changes

**Non-breaking changes**
- When a job is run on weaver while being logged in, the output results will be nested under a user-workspace directory rather than at the root of WPS outputs. 

**Breaking changes**
- n/a

## Related Issue / Discussion

- Depends on: 
   - https://github.com/Ouranosinc/Magpie/pull/517  (Magpie [3.25.0](https://github.com/Ouranosinc/Magpie/tree/3.25.0))
   - https://github.com/bird-house/twitcher/pull/114  (Twitcher [0.7.0](https://github.com/bird-house/twitcher/tree/v0.7.0))

PR #245, opened 2022-05-21
add magpie/twticher response hook for filtering weaver processes

fmigneault:
## Changes

**Non-breaking changes**
- Processes returned by `Weaver` during listing will be filtered with read permissions to those processes.
- Update `Magpie` to 3.26.0

**Breaking changes**
- n/a

## Related Issue / Discussion

- Depends on https://github.com/bird-house/birdhouse-deploy/pull/244
- Depends on https://github.com/Ouranosinc/Magpie/pull/519 (Magpie 3.26.0)
- Addresses [DAC-178](https://crim-ca.atlassian.net/browse/DAC-178).

## Details

Permissions applied for `anonymous` group and `authtest` user respectively (only `weaver` service matters here). 

image

Responses from `weaver` placed behind twitcher with the magpie filtering hook of this PR.

Response for `admin`:
image

Response for `authtest` user:
image

Response for `anonymous` (not logged in):
image

### DIFF
--- a/birdhouse/components/weaver/config/magpie/config.yml.template
+++ b/birdhouse/components/weaver/config/magpie/config.yml.template
@@ -91,9 +91,10 @@ permissions:
     action: create
 
   # Process deployment (write) and listing (read)
+  # use 'read-match' to allow only listing, and not describe underlying processes (require 'read' on them individually)
   - service: ${WEAVER_MANAGER_NAME}
     resource: /processes    # GET is processes listing, POST is deploy: only allow view by anonymous
-    permission: read        # under '/processes/...', JSON 'DescribeProcess', POST job submit, GET results, etc.
+    permission: read-match  # under '/processes/...', JSON 'DescribeProcess', POST job submit, GET results, etc.
     group: anonymous
     action: create
 
--- a/birdhouse/components/weaver/config/magpie/config.yml.template
+++ b/birdhouse/components/weaver/config/magpie/config.yml.template
@@ -31,6 +31,10 @@ providers:
         path: "/jobs"
         method: POST
         target: /opt/birdhouse/src/magpie/hooks/weaver_hooks.py:add_x_wps_output_context
+      - type: response
+        path: "/processes"
+        method: GET
+        target: /opt/birdhouse/src/magpie/hooks/weaver_hooks.py:filter_allowed_processes
 
   # FIXME: remove when https://github.com/Ouranosinc/Magpie/issues/360 implemented, see 'default.env'
   ${WEAVER_WPS_NAME}:
--- a/birdhouse/components/weaver/config/magpie/config.yml.template
+++ b/birdhouse/components/weaver/config/magpie/config.yml.template
@@ -10,6 +10,7 @@ providers:
     c4i: false
     type: api   # FIXME: 'ades' when https://github.com/Ouranosinc/Magpie/issues/360 implemented
     sync_type: api
+    # hook locations should be relative to mounted Twitcher location as they are run within that container
     # see following for hooks details:
     # - https://github.com/Ouranosinc/Magpie/blob/master/config/providers.cfg
     # - https://pavics-magpie.readthedocs.io/en/latest/configuration.html#service-hooks
@@ -21,15 +22,15 @@ providers:
       - type: request
         path: "/providers/[\\w_-]+/processes/[\\w_-]+/jobs"
         method: POST
-        target: /opt/local/src/magpie/hooks/adapter_hooks.py:add_x_wps_output_context
+        target: /opt/birdhouse/src/magpie/hooks/weaver_hooks.py:add_x_wps_output_context
       - type: request
         path: "/processes/[\\w_-]+/jobs"
         method: POST
-        target: /opt/local/src/magpie/hooks/adapter_hooks.py:add_x_wps_output_context
+        target: /opt/birdhouse/src/magpie/hooks/weaver_hooks.py:add_x_wps_output_context
       - type: request
         path: "/jobs"
         method: POST
-        target: /opt/local/src/magpie/hooks/adapter_hooks.py:add_x_wps_output_context
+        target: /opt/birdhouse/src/magpie/hooks/weaver_hooks.py:add_x_wps_output_context
 
   # FIXME: remove when https://github.com/Ouranosinc/Magpie/issues/360 implemented, see 'default.env'
   ${WEAVER_WPS_NAME}:
--- a/birdhouse/components/weaver/config/magpie/config.yml.template
+++ b/birdhouse/components/weaver/config/magpie/config.yml.template
@@ -10,6 +10,26 @@ providers:
     c4i: false
     type: api   # FIXME: 'ades' when https://github.com/Ouranosinc/Magpie/issues/360 implemented
     sync_type: api
+    # see following for hooks details:
+    # - https://github.com/Ouranosinc/Magpie/blob/master/config/providers.cfg
+    # - https://pavics-magpie.readthedocs.io/en/latest/configuration.html#service-hooks
+    hooks:
+      # when a job is created in weaver, apply the header that will nest output results under user's context directory
+      # see also:
+      # - https://pavics-weaver.readthedocs.io/en/latest/processes.html?highlight=x-wps-output-context#outputs-location
+      # each path below are equivalents, but with more or less specific reference to the requested service/process
+      - type: request
+        path: "/providers/[\\w_-]+/processes/[\\w_-]+/jobs"
+        method: POST
+        target: /opt/local/src/magpie/hooks/adapter_hooks.py:add_x_wps_output_context
+      - type: request
+        path: "/processes/[\\w_-]+/jobs"
+        method: POST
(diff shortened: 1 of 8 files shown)


birdhouse/components/weaver/config/magpie/weaver_hooks.py
```diff
@@ -0,0 +1,112 @@
...
+                if proc_name not in known_processes:
+                    continue  # do not bother checking missing resource
+                child_proc = known_processes[proc_name]["node"]
+                perms = context.service.effective_permissions(request_user, child_proc, [Permission.READ])
+                if perms[0].access == Access.ALLOW:
+                    proc = processes[proc_name]
+                    allowed_processes.append(proc)
+
+            # override collected and permitted processes access by user
+            body["processes"] = allowed_processes
```
dbyrns, line 101: Since the response doesn't contain any Weaver version information, I would at least make a basic check of the `processes` structure before overriding it in case that the format has evolved. If the structure doesn't fit, a warning could be raised so that we know that the hook doesn't support this version of Weaver anymore.

dbyrns (approved):
Apart from my future compatibility comment, verythings looks good.

matprov (approved):

birdhouse/components/weaver/docker-compose-extra.yml
```diff
@@ -30,6 +30,15 @@ services:
       - ./components/weaver/config/magpie/config.yml:/opt/local/src/magpie/config/permissions/weaver-permissions.cfg:ro
       - ./components/weaver/config/magpie/config.yml:/opt/local/src/magpie/config/providers/weaver-provider.cfg:ro
 
+  # extend twitcher with MagpieAdapter hooks employed for weaver proxied requests
+  twitcher:
+    environment:
+      MAGPIE_CONFIG_PATH: /opt/birdhouse/src/magpie/config.yml
```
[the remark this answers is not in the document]
tlvu: MagpieAdapter has always been installed with twitcher in the past, did not need this `MAGPIE_CONFIG_PATH` before?  

Now weaver is using this `MAGPIE_CONFIG_PATH`, if another service needs to also "customize" twitcher or the MagpieAdapter, it is stuck because `MAGPIE_CONFIG_PATH` already used?  Not saying we should fix this limitation now.  I just wanted to understand if this limitation exists.

birdhouse/components/weaver/config/magpie/config.yml.template
```diff
@@ -10,6 +10,31 @@ providers:
...
+      - type: request
+        path: "/providers/[\\w_-]+/processes/[\\w_-]+/jobs"
+        method: POST
+        target: /opt/birdhouse/src/magpie/hooks/weaver_hooks.py:add_x_wps_output_context
+      - type: request
+        path: "/processes/[\\w_-]+/jobs"
+        method: POST
+        target: /opt/birdhouse/src/magpie/hooks/weaver_hooks.py:add_x_wps_output_context
+      - type: request
+        path: "/jobs"
```
tlvu, line 31: `path: "/jobs"` here will match to `https://PAVICS_HOST/ows/proxy/twitcher/jobs` ?

birdhouse/components/weaver/config/magpie/weaver_hooks.py
```diff
@@ -0,0 +1,112 @@
...
+    header = get_header("X-WPS-Output-Context", request.headers)
+    # if explicitly provided, ensure it is permitted (admin allow any, otherwise self-user reference only)
+    if header is not None:
+        if request.user is None:
+            header = "public"
+        else:
+            if not is_admin(request):
+                # override disallowed writing to other location
+                # otherwise, up to admin to have writen something sensible
+                header = "user-" + str(request.user.id)
```
tlvu, line 47: Just curious, if it is indeed an admin user, what other values than `"user-" + str(request.user.id)` can `X-WPS-Output-Context` have?

birdhouse/components/weaver/config/magpie/weaver_hooks.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def filter_allowed_processes(response, context):
+    # type: (Response, HookContext) -> Response
+    """
+    Filter processes returned by Weaver response according to allowed resources by user.
+    """
+    if "application/json" in response.content_type:
+        body = response.json
+        if "processes" in body:
```
tlvu, line 64: Can add as comment a sample json body?  So it's easier to follow the code below that extract processes id or verbose string, and seems to do something with the children.

tlvu: To follow up on @dbyrns comment about supported `processes` structure, showing a sample json body here will also clarify what is the expected `processes` structure.

tlvu (commented):
Sorry for the bunch of questions, it's from a "fresh" eyes that do not know the code.

Also, since this brings a new version of twitcher, let me test on a real staging server with production config for regression before merging this.

birdhouse/components/weaver/config/magpie/weaver_hooks.py
fmigneault: I think it is more secure that way. If the structure evolved and processes were not overridden, this could make previously private processes to be leaked. The hook would need to be updated regardless to support the new structure, but processes would at least not be shown by mistake with current implementation.

birdhouse/components/weaver/docker-compose-extra.yml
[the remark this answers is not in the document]
fmigneault: Yes, the rest is not used by Twitcher. Twitcher never loaded the config because it was not needed. It actually doesn't know anything about this config. It is only MapgieAdapter that knows what to do with it.

That's a good point about other configs. I will adjust it to use the directory approach to allow loading many configs per-service as done in Magpie.

birdhouse/components/weaver/config/magpie/config.yml.template
fmigneault: It will match `https://PAVICS_HOST/ows/proxy/twitcher/ows/proxy/weaver/jobs`

birdhouse/components/weaver/config/magpie/weaver_hooks.py
fmigneault: It can be any valid directory name. 
Directories will be formed (and created by Weaver if missing) using this format:

`<wps_output_dir>/<wps_output_context>/<job_id>/<output-files>`

If `X-WPS-Output-Context` is not present instead (current behavior), it is:

`<wps_output_dir>/<job_id>/<output-files>`

So basically, the hook enforces outputs to:
- `<wps_output_dir>/user-<id>/<job_id>/<output-files>` for a non-admin
- `<wps_output_dir>/public/<job_id>/<output-files>` for no-auth (process will have to be public as well to allow execute)
- `<wps_output_dir>/**anywhere**/<job_id>/<output-files>` for an admin (possibly using literal `user-<id>` if desired)

birdhouse/components/weaver/config/magpie/weaver_hooks.py
fmigneault: Added in 28ad95c

birdhouse/components/weaver/config/magpie/config.yml.template
fmigneault: To clarify, the `path` in the context of Twitcher/Magpie hooks is whatever is defined after the resolved protected service. 
I also added a link in the python file to the reference doc in Magpie if you need more details.

birdhouse/components/weaver/config/magpie/weaver_hooks.py
dbyrns: You're right!

birdhouse/components/weaver/config/magpie/weaver_hooks.py
tlvu: Oh, I did not realized `X-WPS-Output-Context` will be used part of the wps output dir, thanks.

fmigneault:
Magpie/Twitcher docker version `3.26.0` are now deployed. This PR should start working along #244  when the IaC is patched.

fmigneault:
run tests

matprov:
run tests

fmigneault:
@matprov 
Test pipeline is now working (as in not crashing), but settings don't seem to propagate correctly since Weaver tests were not executed (DACCS-IaC env shows `TEST_PAVICS_SDI_WEAVER=true` but pipeline later shows `TEST_PAVICS_SDI_WEAVER=false`, and test job receives this `false` value).

tlvu (approved):
Deployed this branch to my staging host, without Weaver enabled.

Jenkins all passed, from `docker logs magpie` I see 1 DB migration (`Running upgrade cb92ff1f81bb -> 0c6269f410cd, Move anonymous user to anonymous group permissions.`) without any errors.  The DB was from our production.

fmigneault:
Only thing blocking me now for both #244 and #245 is for the weaver test status to reach pavics-sdi. 
@matprov Any ETA on that?

matprov:
run tests

PR #248, opened 2022-06-07
deploy-data: new env var DEPLOY_DATA_RSYNC_USER_GRP to avoid running cronjobs as root

tlvu:
When `deploy-data` is used by the `scheduler` component, it is run as
  `root`.  This new env var will force the rsync process to run as a regular user to
  follow security best practice to avoid running as root when not needed.

  Note that the `git checkout` step done by `deploy-data` is still run as root.
  This is because `deploy-data` is currently still run as root so it can
  execute `docker` commands (ex: spawning the `rsync` command above in its own
  docker container).

  To fix this limitation, the regular user inside the `deploy-data` container
  need to have docker access inside the container and outside on the host at
  the same time.  If we make that regular user configurable so the script
  `deploy-data` is generic and can work for any organisations, this is tricky
  for the moment so will have to be handle in another PR.

  So for the moment we have not achieved full non-root user in cronjobs
  launched by the `scheduler` compoment but the most important part, the part
  that perform the actual job (rsync or execute custom command using an
  external docker container) is running as non-root.

  See PR https://github.com/bird-house/birdhouse-deploy-ouranos/pull/18 that
  make use of this new env var.

  When `deploy-data` is invoking an external script that itself spawn a new
  `docker run`, then it is up to this external script to ensure the proper
  non-root user is used by `docker run`.
  See PR https://github.com/Ouranosinc/pavics-vdb/pull/50 that handle that
  case.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.18.12
+current_version = 1.18.13
 commit = True
 tag = False
 tag_name = {new_version}
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,11 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
+[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+
+[1.18.13](https://github.com/bird-house/birdhouse-deploy/tree/1.18.13) (2022-06-07)
+------------------------------------------------------------------------------------------------------------------
+
 ## Changes:
 
 - deploy-data: new env var DEPLOY_DATA_RSYNC_USER_GRP to avoid running cronjobs as root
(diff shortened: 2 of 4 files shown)


matprov (approved):

moulab88 (approved):

tlvu:
> For the record, the `avoid-running-cronjobs-as-root` branch has been merged and deleted in the meantime the pipeline created the iac VM for E2E tests.

@matprov yes, I already assumed the same reason for the funny IAC empty result.  Thanks for the clarification.

Also thanks for excluding weaver notebook in "standard" run to make the IAC pass.  So the weaver nb is only enabled when the PR creator is Francis?

matprov:
@tlvu In the pipeline there is the `TEST_PAVICS_SDI_WEAVER` that is set to false by default. The iac pipeline is supposed to read this value from our test `env.loval` in which `TEST_PAVICS_SDI_WEAVER` is true, so the default behaviour will be that weaver tests will start by default. Once we have repaired the iac pipeline, via jenkins we'll still be able to deactivate the execution.

PR #256, opened 2022-08-15
add cowbird service to the stack

fmigneault:
## Overview

Add new service [Ouranosinc/cowbird](https://github.com/Ouranosinc/cowbird/) to the stack.

This PR focuses more on the integration part (services running, starting, configs found with no error) rather than the actual cowbird configuration portion for services to work together. 
Following PRs should validate configs and make sure items sync accordingly with desired use cases.

### Relevant changes
* Cowbird can be integrated to the instance using [components/cowbird](./birdhouse/components/cowbird) 
when added to in ``EXTRA_CONF_DIRS`` in the ``env.local`` variable definitions.
* Offers syncing operations between various other *birds* in order to apply user/group permissions between
corresponding files, granting access to them seamlessly through distinct services.
* Allows event and callback triggers to sync permissions and volume paths between API endpoints and local storages.

### Breaking changes
- n/a

## Related Issue / Discussion

- Relates to [DAC-145](https://crim-ca.atlassian.net/browse/DAC-145) *Create private directory for new user for climate data [Cowbird/Thredds]*   
  (required to sync permissions between Thredds/GeoServer user-workspaces)
- Relates to [DAC-469](https://crim-ca.atlassian.net/browse/DAC-469) *Secure WPS-Outputs served by Nginx*
  (will be required to grant access to generated files by the user that requested the WPS execution) 
- Resolves [DAC-470](https://crim-ca.atlassian.net/browse/DAC-470) *Intégrer Cowbird à birdhouse*

### Future TODOs:
- [ ] Test endpoints between services considering how they will be employed. 
- [ ] Adjust Cowbird configs accordingly to services/resources to be synced
- [x] Adjust Cowbird config according to https://github.com/Ouranosinc/cowbird/issues/22
- [ ] Add Cowbird component to `EXTRA_CONF_DIRS` in DACCS-IaC config to include its service definition in E2E Test instances
  (relates to https://gitlab.com/crim.ca/clients/daccs/daccs-configs/-/merge_requests/15)


ChaamC (approved):

birdhouse/components/cowbird/default.env
```diff
@@ -0,0 +1,29 @@
+# -------
+# Cowbird
+# -------
+
+# add any new variables not already in 'VARS' or 'OPTIONAL_VARS' that must be replaced in templates here
+# single quotes are important in below list to keep variable names intact until 'pavics-compose' parses them
+EXTRA_VARS='
```
tlvu, line 7: This extra housekeeping blocks about VARS and EXTRA_VARS could be at the bottom of this file so we immediately see the most important values first.

birdhouse/components/cowbird/docker-compose-extra.yml
```diff
@@ -0,0 +1,82 @@
...
+  &default-logging
+  driver: "json-file"
+  options:
+    max-size: "50m"
+    max-file: "10"
+
+services:
+  cowbird:
+    image: pavics/cowbird:${COWBIRD_VERSION}-webservice
+    container_name: cowbird-web
```
tlvu, line 13: name the container_name to match the service name ... easier for the users follow

birdhouse/components/cowbird/default.env
```diff
@@ -0,0 +1,29 @@
...
+# Cowbird Configuration
+# =====================
+
+export COWBIRD_VERSION="0.5.0"
+
+# reuse the mongo instance provided by the core docker-compose
+# the 'host' is the mongo 'service' inter-docker network connection in this case
+# if another service should be used (e.g.: dedicated mongodb), adjust the variable
+# the selected 'mongo' service should be member of a network of same name
+export COWBIRD_MONGODB_SERVICE=mongodb
```
tlvu, line 27: Is this piggyback into the "default" mongodb that is very old?  I remember you had to roll your own mongodb for Weaver so you can have a newer mongodb.  Why not roll your own mongodb container for this component so you have full control on the version again?

tlvu (approved):
Nice integration into the stack via components.

Minor changes suggested but none blocking since this do not affect the production deployment at all.

Just curious, if I enable cowbird on my test VM, what kind of operations I can do to "see" it working?  Any plan later for an automated test notebook like the Weaver notebook?

birdhouse/components/cowbird/config/cowbird/config.yml.template
```diff
@@ -0,0 +1,155 @@
...
+    admin_password: ${MAGPIE_ADMIN_PASSWORD}
+  Geoserver:
+    active: true
+    url: http://${PAVICS_FQDN_PUBLIC}:8765/geoserver
+    workspace_dir: ${WORKSPACE_DIR}
+    admin_user: ${GEOSERVER_ADMIN_USER}
+    admin_password: ${GEOSERVER_ADMIN_PASSWORD}
+  Catalog:
+    active: true
+    url: https://${HOSTNAME}/twitcher/ows/proxy/catalog
```
francisPLT, line 24: Since it's been changed above, shouldn't `HOSTNAME` here also have been changed to `PAVICS_FQDN_PUBLIC` ?

birdhouse/components/cowbird/config/cowbird/config.yml.template
```diff
@@ -0,0 +1,155 @@
...
+  Magpie:
+    active: true
+    url: https://${PAVICS_FQDN_PUBLIC}/magpie
+    admin_user: ${MAGPIE_ADMIN_USERNAME}
+    admin_password: ${MAGPIE_ADMIN_PASSWORD}
+  Geoserver:
+    active: true
+    url: http://${PAVICS_FQDN_PUBLIC}:8765/geoserver
+    workspace_dir: ${WORKSPACE_DIR}
+    admin_user: ${GEOSERVER_ADMIN_USER}
```
francisPLT, line 20: Side note, as this PR is for integration and not 100% functionality. Happened on a brand-new IaC instance too.

`${GEOSERVER_ADMIN_USER}` somehow is not being templated correctly.

The other variables, except `WORKSPACE_DIR` which has not been defined yet, are replaced as expected in `config.yml`. 

I had to add `GEOSERVER_ADMIN_USER` to Cowbird's own `.../components/cowbird/default.env` for it to be replaced properly, which is weird considering there's no problem with `GEOSERVER_ADMIN_PASSWORD` and `PAVICS_FQDN_PUBLIC`

Might be a side effect from something else in the configuration? I'm bringing this up in case it is reproduced in subsequent PRs where configurations will be fine-tuned.

francisPLT (approved):
Same as tlvu, none of my comments are blocking. 

Cowbird component initiates well on my end. With a bit of quick tweaking of the configurations I was able to test some of Cowbird's services, so everything is a go for now on my end.

huard (approved):

birdhouse/components/cowbird/config/cowbird/config.yml.template
tlvu: Good catch @francisPLT `GEOSERVER_ADMIN_PASSWORD` is templated expanded properly because it is in the `VARS` list here https://github.com/bird-house/birdhouse-deploy/blob/d1ecc63284ec9d2940bfa2b1b4baca3fbe1308b3/birdhouse/pavics-compose.sh#L31

`GEOSERVER_ADMIN_USER` is not in that `VARS` list so that's why it has not been template expanded properly.  The reason it is not in that list is because it currently is not needed be template expanded anywhere.

The fix is simple, just add it to the same location as `GEOSERVER_ADMIN_PASSWORD`.

birdhouse/components/README.rst
```diff
@@ -311,3 +317,63 @@ Customizing the Component
...
+.. _cowbird-diagram: https://github.com/Ouranosinc/cowbird/blob/master/docs/_static/cowbird_components.png
+.. |cowbird-repo| replace:: Ouranosinc/cowbird
+.. _cowbird-repo: https://github.com/Ouranosinc/cowbird
+.. |cowbird-rtd| replace:: ReadTheDocs
+.. _cowbird-rtd: https://pavics-cowbird.readthedocs.io/
+
+Operations Performed by Cowbird
+-------------------------------
+
+- Synchronize `Magpie` user and group permissions between "corresponding files" located under different services.
```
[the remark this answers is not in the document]
fmigneault: Purposely single tick to format with italic.

birdhouse/components/cowbird/config/cowbird/cowbird.ini.template
```diff
@@ -0,0 +1,101 @@
...
+###
+# app configuration
+# http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/environment.html
+###
+
+[composite:main]
+use = egg:Paste#urlmap
+/ = cowbird_app
+#/api = api_app
+#/cowbird/api = api_app
```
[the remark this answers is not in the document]
fmigneault: This is an older configuration that needed swagger to run in parallel. 
Now `cowbird_app` directly serves the OpenAPI contents on `/api` relative to its root. 
Once in PAVICS, `/cowbird/*` will redirect to `/twitcher/ows/proxy/cowbird/*` which then redirects to `/*` internally.

birdhouse/components/cowbird/config/cowbird/config.yml.template
```diff
@@ -0,0 +1,155 @@
...
+    # When using an implicit format, if the access and/or scope are not specified, it will use the default
+    # access `allow` and/or the default scope `recursive`.
+    # For more info on Magpie permissions :
+    #     https://pavics-magpie.readthedocs.io/en/latest/permissions.html#permission-definition-and-modifiers
+    #     https://pavics-magpie.readthedocs.io/en/latest/permissions.html#permissions-representation
+    # See documentation 'Configuration' for more details on the mapping and synchronization features
+    permissions_mapping:
+      # A `read` permission for `thredds_workspace` creates/deletes all specified `geoserver_workspace` permissions.
+      # And a specified `geoserver_workspace` permission creates/deletes a `read` permission on `thredds_workspace`.
+      - "geoserver_workspace : [getCapabilities, getFeature, getFeatureInfo] <-> thredds_workspace : read"
```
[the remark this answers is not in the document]
fmigneault: Indeed. Cowbird does a lot of schema and regex validation of those definitions. 
The one-liner syntax was preserved over nested definitions because they are more readable at a glance, especially when a lot of rules are stacked in a list.

birdhouse/components/cowbird/default.env
fmigneault: I've added it at the top to actually produce the inverse effect. Something like a reminder because you see it first as soon as you open the file. When at the bottom, it is very easy to add new variables to the file and forget to add them to this list for them to become effective.

birdhouse/components/cowbird/default.env
fmigneault: In this case, it is "ok" to use the old one. The example https://github.com/Ouranosinc/cowbird/blob/master/docker/docker-compose.example.yml uses 3.4 also for testing. In the case of Weaver, there was an actual requirement of 5.0 for a new mongodb feature.

I've actually though about this possibility to have its own mongodb version. This is why I defined this service variable that allows one to override if they want (they would have to provide a `docker-compose.override` with that service as well). Otherwise, I think it is better to avoid wasting more resources for this case.

birdhouse/components/cowbird/config/cowbird/config.yml.template
fmigneault: Variable added.

birdhouse/components/README.rst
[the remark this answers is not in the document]
tlvu: Oh you wanted italics?  It's single star and not single backtick, see https://docutils.sourceforge.io/docs/user/rst/quickref.html#inline-markup

Single backtick is "interpreted text", not sure how to use it.

birdhouse/components/cowbird/default.env
tlvu: Ah I see, makes sense.

birdhouse/components/cowbird/default.env
tlvu: I usually follow the saying "Premature Optimization Is the Root of All Evil". I usually choose ease of maintenance, simplicity and correctness over performance and resource optimization.  An idle mongodb is probably not going to suck that much resources.  But it's your component so your call.

birdhouse/components/README.rst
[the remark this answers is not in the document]
fmigneault: Both work if no underscores are placed after, it is interpreted as italic. 
I'm using them all the time, but I don't mind adjusting it to make it clearer.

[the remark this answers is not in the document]
tlvu: Oh I didn't know single backtick is also italics, interesting.  Or maybe the editor you use choose to renderer "interpreted text" as italics while github choose to render them normal.

fmigneault:
Validation screenshots of the service and configs on a test instance with added component.

image
image

fmigneault:
@tlvu 
Re-review just FYI about 0f97ccd change.

Without added nginx config headers, the endpoint `https://PAVICS_HOST/cowbird` and `https://PAVICS_HOST/twitcher/ows/proxy/cowbird` were both indicating `url: http://cowbird:7000` (i.e.: docker network link referenced in Magpie). 

Added headers  in `cowbird` config fixed  `https://PAVICS_HOST/cowbird`. 
Added headers in `twitcher` config fixed `https://PAVICS_HOST/twitcher/ows/proxy/cowbird`

tlvu (approved):
LGTM

fmigneault:
No. It is an old dupe. It can be removed.

PR #259, opened 2022-08-24
fix invalid `canarie-api-full-monitoring` endpoints adding double `/` when substituting variables

fmigneault:
## Overview

- fix invalid `canarie-api-full-monitoring` endpoints adding double `/` when substituting variables

## Changes

**Non-breaking changes**
- fix invalid `canarie-api-full-monitoring` endpoints adding double `/` when substituting variables

**Breaking changes**
- n/a


tlvu (approved):
Nice catch!  Thanks for catching this.  It did not cause any problems in the past.  Is it causing a problem now? Just curious, what problem is it causing?

PR #260, opened 2022-08-25
add RELEASE.txt and Makefile to control bumpversion releaseTime auto-versioning

fmigneault:
## Overview

Because `bumpversion` cannot update its own `.bumpversion.cfg` file to apply the new `releaseTime` value (since it rewrites the config with its own config handler once changes were applied), use a workaround to produce the same effect.

Use Makefile to parse the current release time, then run `bumpversion` as normal, retrieve the new release time that was applied by `bumpversion`, then replace the old->new release time in the config for future calls.

To test, a few bumps were done (note: second tag on the line is the "active version" on that commit)
```
* eb537ef Bump version: 1.22.3 → 1.22.4
* f41fb55 Bump version: 1.22.2 → 1.22.3
* 81470bd Bump version: 1.22.1 → 1.22.2
```

The diffs are as follows, showing correct behaviour to replace new `releaseTime` everywhere: 

<details><summary> diff 1.22.2 → 1.22.3 </summary>
<p>

```diff

diff --git a/.bumpversion.cfg b/.bumpversion.cfg
index 36e736a..8a68e80 100644
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,23 +1,23 @@
 [bumpversion]
-current_version = 1.22.2
+current_version = 1.22.3
 commit = True
 tag = False
 tag_name = {new_version}
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+) (?P<releaseTime>.*)?
-serialize =
+serialize = 
 	{major}.{minor}.{patch}
 	{utcnow:%Y-%m-%dT%H:%M:%SZ}
 
 [bumpversion:file:CHANGES.md]
-search =
+search = 
 	[Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 	------------------------------------------------------------------------------------------------------------------
-replace =
+replace = 
 	[Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 	------------------------------------------------------------------------------------------------------------------
-
+	
 	[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
-
+	
 	[{new_version}](https://github.com/bird-house/birdhouse-deploy/tree/{new_version}) ({utcnow:%Y-%m-%d})
 	------------------------------------------------------------------------------------------------------------------
 
@@ -30,11 +30,11 @@ search = {current_version}
 replace = {new_version}
 
[...]

PR #261, opened 2022-08-26
fix doc missing links to repo files

fmigneault:
Based on #260
See diff: https://github.com/bird-house/birdhouse-deploy/compare/fix-release-time-bumpversion...fix-doc-refs

## Changes

**Non-breaking changes**
- Provides multiple fixes to properly redirect missing links to non-RST/MD files referenced in the HTML docs.

**Breaking changes**
- n/a


docs/_ext/doc_redirect.py
```diff
@@ -0,0 +1,169 @@
...
+    In other words, when ``file1.rst`` is viewed from GitHub, the literal relative path is used an the file is found,
+    while on ``Readthedocs`` (or when viewing locally in a browser), ``file1.html`` will contain a raw relative path to
+    where the pointed ``file2.html`` *should* be using corresponding base directories. This is demonstrated below:
+
+    .. code-block:: text
+
+        '<root>/docs/file1.rst' ===> '<build-html>/file1.html' (ref [file2]) ---> (raw '<root>/docs/file2.rst')
+                                                                                               |
+                                                                                               |
+        '<root>/docs/file2.rst' ===> '<build-html>/file2.html'   <------------------------------
```
tlvu, line 83: Very interesting !

tlvu (commented):
This looks very interesting and seem to solve a long pain in the butt with viewing .rst directly on Github vs the generated files on ReadTheDocs.  If it works, we can remove all the "download" links?   I'll have to play this this further.

docs/_ext/doc_redirect.py
fmigneault: The only problem I have add so far is to try to make it work correctly under ReadTheDocs. 
Locally, it redirects fine, but ReadTheDocs triggers a file download of the "RST" with HTML contents. 
Still exploring...

docs/source/conf.py
```diff
@@ -42,8 +48,47 @@
...
+md_parse_relative_links = True
+
+# references to RST files in 'docs' dir redirect to corresponding HTML
+# references to RST files in repo root (README/CHANGES) redirect to their equivalent HTML in 'docs' dir
+DOC_REDIRECT_DIRS = [
+    os.path.join(DOC_SRC_ROOT, "birdhouse"),
+]
+
+# files that will be copied over from root repo or linked birdhouse directory
+# under build directory (not copied over with same nested dir structure, directly under '_build/html')
```
tlvu, line 63: is there a way to keep the nested dir structure?  I just found a link that do not work, probably need nested dir structure:
https://birdhouse-deploy--261.org.readthedocs.build/en/261/optional-components/emu/docker-compose-extra.yml

docs/_ext/doc_redirect.py
tlvu: You might want to ask the ReadTheDocs developper.  They might hardcode some behavior on ReadTheDocs and let you know how to undo it.  They might even suggest a different solution to our problem that could be more compatible.

If they have a compatible solution that allow us to avoid those ugly looking duplicate download link, this would be huge.

birdhouse/README.rst
```diff
@@ -17,22 +17,29 @@ Requirements:
...
+If the file `env.local` is somewhere else, symlink it here, next to `docker-compose.yml <docker-compose.yml>`_
+(:download:`download </birdhouse/docker-compose.yml>`) because many scripts assume this location.
 
 To follow infrastructure-as-code, it is encouraged to source control the above
 `env.local` file and any override needed to customized this PAVICS deployment
-for your organization.  For an example of possible override, see how the `emu service <optional-components/emu/docker-compose-extra.yml>`_ (:download:`download </birdhouse/optional-components/emu/docker-compose-extra.yml>`)
-(`README <optional-components/README.rst#emu-wps-service-for-testing>`_) can be optionally added to the deployment via the `override mechanism <https://docs.docker.com/compose/extends/>`_.
-Ouranos specific override can be found in this `birdhouse-deploy-ouranos <https://github.com/bird-house/birdhouse-deploy-ouranos>`_ repo.
+for your organization.  For an example of possible override,
+see how the `emu service <optional-components/emu/docker-compose-extra.yml>`_
```
tlvu, line 34: Here's the link that did not resolve on RtD.

birdhouse/README.rst
```diff
@@ -17,22 +17,29 @@ Requirements:
...
+(:download:`download </birdhouse/optional-components/emu/docker-compose-extra.yml>`)
+(`README <birdhouse/optional-components/README.rst#emu-wps-service-for-testing>`_) can be optionally added to the
```
tlvu, line 36: This one resolve fine (https://birdhouse-deploy--261.org.readthedocs.build/en/261/birdhouse/optional-components/README.rst#emu-wps-service-for-testing)

docs/_ext/doc_redirect.py
fmigneault: Good idea.

birdhouse/README.rst
fmigneault: It doesn't work in different manners per client/browser.
In Chrome (in my case at least), it shows the raw code of the HTML instead of displaying it. 
In Firefox, it simply downloads the file.

docs/source/conf.py
fmigneault: Will look into it. It will help keeping track of which file comes from it that can work. 

The hard case is that some references resolve automatically with the right nesting via symlink `docs/source/birdhouse -> birdhouse`, but others don't (ie: anything at the root above `birdhouse`), and there are "conflicting cases" like `CHANGES.md` that should not be auto-redirected since it is indirectly integrated via `docs/source/changes.rst`.

docs/_ext/doc_redirect.py
fmigneault: Maybe an approach would simply to define redirects specifically in ReadTheDocs: 
https://docs.readthedocs.io/en/latest/user-defined-redirects.html
It's somewhat manual configuration though. 

Others have tried to automate it, but it's not trivial (cannot define it via config yet).
https://github.com/readthedocs/readthedocs.org/issues/2904

PR #267, opened 2022-11-08
Various platform improvements to support deployment of new portraits climatiques backend

tlvu:
## Changes:

**Non-breaking changes**

- autodeploy: allow repos to optionally decide if a deploy is required

  Useful when only a subset of file changes in a repo will actually impact deployment.

  Without this mechanism any file changes in a repo will trigger a deployment, which would cost a full platform restart for no reason.

  Var `GIT_CHANGED_FILES` is given to optional script `<repo_root>/autodeploy/conditional-trigger` and only an exit code 0 will trigger deploy.

- fix-geoserver-data-dir-perm: allow overriding data dir to use on another instance of Geoserver

## Related Issue / Discussion

- Required for PR https://github.com/Ouranosinc/portraits_climatiques/pull/41


birdhouse/deployment/triggerdeploy.sh
```diff
@@ -93,17 +106,46 @@ should_trigger() {
         AHEAD_CNT="`git rev-list --left-only --count ${local}...${remote}`"
         echo "BEHIND_CNT=$BEHIND_CNT" >> $TMP_SAVE_FILE
         echo "AHEAD_CNT=$AHEAD_CNT" >> $TMP_SAVE_FILE
+        echo "local=$local" >> $TMP_SAVE_FILE
```
matprov, line 109: Why does `local` and `remote` doesn't use the same case as `BEHIND_CNT` and `AHEAD_CNT`?

matprov (approved):

birdhouse/deployment/triggerdeploy.sh
tlvu: Oh yeah, that's because they are variables inside the while loop that do not survive outside the loop since data is feed in by a pipe.  It's a new shell for each loop, see https://github.com/bird-house/birdhouse-deploy/blob/4d9b1ae6009a79bf0a3e0a135b3f8420d3d8cc65/birdhouse/deployment/triggerdeploy.sh#L99-L100

I usually use lowercase for transient variables like that.  They actually survive simply because I write it out to a tmp file and read it back in at the end of the loop https://github.com/bird-house/birdhouse-deploy/blob/4d9b1ae6009a79bf0a3e0a135b3f8420d3d8cc65/birdhouse/deployment/triggerdeploy.sh#L109-L114

Issue #270, opened 2022-12-08
:bug: [BUG]: Regression handling `DATA_PERSIST_ROOT` setting


fmigneault:
## Summary

Regression in handling `DATA_PERSIST_ROOT` setting does not produce the intended effect if anything other than defaults are used.

## Details

Regression introduced in https://github.com/bird-house/birdhouse-deploy/commit/d4f2f4401b6535c7401a170c0c0794e2a19e479a regarding the use of `DATA_PERSIST_ROOT`. 

New variable `MAGPIE_PERSIST_DIR` ~(there might be others doing similar things in other commits [?] )~ have been defined in `default.env` making use of the (up to that point) default value of `DATA_PERSIST_ROOT=/data`. 

> **edit** @fmigneault 
> I can confirm that all modified directories for jupyter images, geoserver data and magpie's postgres and misbehaving in the data direcotory paths using a quick local test. I personnally use `DATA_PERSIST_ROOT=/data/daccs` and they have all been generated as `/data/magpie_persists`, `/data/geoserver` and `/data/jupyterhub_user_data` instead of their usual location nested under `/data/daccs`. 

If the user defined instead their own `DATA_PERSIST_ROOT` in the `env.local` for an alternate storage location, all those derived variables do not consider the new value because they are already set.

This produces inconsistent directory storage locations and does not respect the expected behavior defined by the user.

### To Reproduce

1. Define `DATA_PERSIST_ROOT=/my-data`
2. `pavics-compose up -d`

The volumes that make direct reference to `${DATA_PERSIST_ROOT}` in `docker-compose,yml` (and component overrides) will have the correct path.
When using derived directory variables, they will still be under `/data/...`. 

## Environment

| Information           | Value
| --------------------- | --------------------------------------------------------------
| Server/Platform URL   | all that define `DATA_PERSIST_ROOT` with another value
| Version Tag/Commit    | since [1.19.1](https://github.com/bird-house/birdhouse-deploy/releases/tag/1.19.1) 
| Related issues/PR     | https://github.com/bird-house/birdhouse-deploy/pull/249
| Related components    | magpie, postgres-magpie
| Custom configuration  | any with custom `DATA_PERSIST_ROOT` and no explicit `MAGPIE_PERSIST_DIR` 

## Concerned Organizations

@matprov @tlvu

tlvu:
In this case, I can only see 2 solutions:

[...]

Issue #266, opened 2022-10-25
:bug: [BUG]: Silent error when  EXTRA_CONF_DIRS is invalid


aulemahal:
When adding  a directory to EXTRA_CONF_DIRS, if there is a typo in its name, the error goes unnoticed.

Feel free to close this if not important enough, but I felt that it would be nice if `pavics-compose.sh` would fail when encountering an invalid directory name.

tlvu:
We do not validate those dirs because all the files (docker-compose-extra.yml, default.env) under those dirs are optional, but probably we need to validate the dir itself, without validating the files under.

PR #272, opened 2022-12-15
pavics compose.sh: overriding `DATA_PERSIST_ROOT` in `env.local` do not take effect and warn when a dir in `EXTRA_CONF_DIRS` does not exist

tlvu:
## Fixes:

- Overriding `DATA_PERSIST_ROOT` in `env.local` do not take effect for
  `JUPYTERHUB_USER_DATA_DIR`, `MAGPIE_PERSIST_DIR`, and `GEOSERVER_DATA_DIR`.

  These 3 vars will have to be delayed evaluated for override in `env.local` to
  take effect.

  For a variable to be delayed evaluated, it has to be defined using
  single-quote and be added to the list of `DELAYED_EVAL` in `default.env`.

  If those steps are forgotten in `env.local`, it will still work since
  `env.local` is the last file to be read.  However those steps should not be
  forgotten in any `default.env` for all components.

  So the impact or burden is on the developpers to write their `default.env`
  file properly, not on the users that only modify the `env.local` file.

  All `default.env` files header have been updated with notice about this new
  delayed evaluation feature.

  Fixes https://github.com/bird-house/birdhouse-deploy/issues/270

## Changes:

- Warn when a dir in `EXTRA_CONF_DIRS` does not exist.

  Most likely a typo in a new dir.  Just warn and not exit directly to avoid
  leaving the entire platform down during an unattended autodeploy since no
  one is around to take immediate action.

  Fixes https://github.com/bird-house/birdhouse-deploy/issues/266


tlvu:
@fmigneault I changed my mind and go ahead with single-quote and delayed evaluation instead.

The impact on users is actually minimum since forgetting to use single-quote in `env.local` has no problem.  Basically the complexity burden is on the developer only.  I've believe I've left enough documentation comments for the developers.

birdhouse/default.env
```diff
@@ -25,7 +27,8 @@ export BASH_IMAGE="bash:5.1.4"
 export DATA_PERSIST_ROOT="/data"
 
 # Folder on the host to persist Jupyter user data (noteboooks, HOME settings)
-export JUPYTERHUB_USER_DATA_DIR="${DATA_PERSIST_ROOT}/jupyterhub_user_data"
+# Must use single-quote for delayed eval.
+export JUPYTERHUB_USER_DATA_DIR='$DATA_PERSIST_ROOT/jupyterhub_user_data'
```
fmigneault, line 31: Can we leave the `{}` for better readability (here and everywhere else where quotes were changed)? It is easier to spot variables with them (`$` by itself is too easily confused with `S` depending on font style/size).

CHANGES.md
```diff
@@ -14,7 +14,39 @@
...
+  file properly, not on the users that only modify the `env.local` file.
+
+  All `default.env` files header have been updated with notice about this new
+  delayed evaluation feature.
+
+  Fixes https://github.com/bird-house/birdhouse-deploy/issues/270.
+
+## Changes:
+
+- Warn when a dir in `EXTRA_CONF_DIRS` do not exist.
```
fmigneault, line 42: typo "does"

CHANGES.md
```diff
@@ -14,7 +14,39 @@
...
+
+  Most likely a typo in a new dir.  Just warn and not exit directly to avoid
+  leaving the entire platform down during an unattended autodeploy since no
+  one around to take immediate action.
```
fmigneault, line 46: typo "is around"

birdhouse/pavics-compose.sh
```diff
@@ -95,6 +95,13 @@ COMPOSE_DIR="`pwd`"
 [ -f env.local ] && . ./env.local
 
 for adir in ${EXTRA_CONF_DIRS}; do
+  if [ ! -e "$adir" ]; then
+    # Do not exit to not break unattended autodeploy since no human around to
+    # fix immediately.
+    # The new adir with typo will not be active but at least all the existing
+    # will still work.
+    echo "WARNING: EXTRA_CONF_DIRS '$adir' do not exist" 1>&2
```
fmigneault, line 103: Same "does" typo.

Maybe reword also because is it one item of `EXTRA_CONF_DIRS` that is flagged here.
Current wording could be misinterpreted as `EXTRA_CONF_DIRS = $adir` (*only*).

birdhouse/pavics-compose.sh
```diff
@@ -106,6 +113,12 @@ done
 # components.
 [ -f env.local ] && . ./env.local
 
+for i in ${DELAYED_EVAL}; do
+  v="`eval "echo \\$${i}"`"
+  eval 'export ${i}="`eval "echo ${v}"`"'
+  echo "delayed eval '`env |grep ${i}=`'"
+done
```
fmigneault, line 120: I'm not sure if I understand the need for the separate `DELAYED_EVAL` definition. 
Wouldn't developers forgetting to specify a new variable name in it be more prone to error? 
Couldn't all variables in `${VARS}` could do the ```v="`eval "echo \\$${i}"`"``` step, and this way simply using the single quotes would be sufficient to achieve the same result?

birdhouse/default.env
tlvu: done

CHANGES.md
tlvu: done

birdhouse/pavics-compose.sh
tlvu: done

CHANGES.md
tlvu: done

birdhouse/pavics-compose.sh
tlvu: Well, if you think for simplicity, all vars in `default.env` should be delayed eval then `VARS` do not contain all of them.  `VARS` currently do not contain the 3 broken vars in `DELAYED_EVAL`.

Then it looks like delayed eval is an exception rather than the norm.  So I'd rather not forced the single quote on more vars than needed.

birdhouse/default.env
```diff
@@ -1,5 +1,7 @@
 # All env in this default.env can be overridden by env.local.
-# All env in this default.env must not depend on any env in env.local.
+
+# All env in this default.env must NOT depend on any other env.  If they do,
+# must use single quote and must add to the list of DELAYED_EVAL.
```
fmigneault, line 4: I think it would be worth adding the reason here that "[...] must use single quote *to avoid early expansion before overrides in local.env are applied* [...]".

birdhouse/pavics-compose.sh
tlvu: Ouch, the grep trick might too magic for me.  How do you get the name of each var after?

birdhouse/default.env
tlvu: done in 407c348e576b917657bc8b2de1e3b5e5f8c45fe5

birdhouse/pavics-compose.sh
fmigneault: Using this:
 cat birdhouse/default.env | grep -E "export [A-Z0-9_]+=" | cut -d ' ' -f 2 | cut -d '=' -f 1`

fmigneault: And an extra ` | sed -r -e 's/(.*)/\$\1/g'` at the end for adding the `$`.

tlvu: Oh wow, think I'll pass on this magic for now.  We have to handle all the `default.env` from components from any other external repos as well.  I don't have time to think of all the corner cases with this magic for now.

Do you see anything blocking with the current approach?

fmigneault: No. Good to go with the current alternative.

fmigneault (approved):
Current method sufficient for now. 
Alternative to avoid `DELAYED_EVAL` to be considered as some other time as per https://github.com/bird-house/birdhouse-deploy/pull/272#discussion_r1055829225.

For reference: https://github.com/bird-house/birdhouse-deploy/issues/274

PR #275, opened 2023-01-11
support display image names with version on jupyterhub

ChaamC:
Fix to support JupyterHub image selection names that use the `<name>:<version>` format. 

Initially, we were only displaying the name of the images on the JupyterHub page, but we added the version to give more information to the user. This requires the adjustment found in this PR, since we rely on those values to find a corresponding directory to mount when starting the image. The directory should only be named using the `name` part, and should exclude the `version`. 

For example, if we display an option `eo-crim:0.3.0` on the JupyterHub selection page, the image used to start JupyterLab should mount the directory `eo-crim`, which should contain its corresponding tutorial notebooks.

I also added some more comments to clarify the relationship between the names displayed on the JupyterHub page and directory names used for the notebooks of each corresponding image.

This should be backward compatible, still working if we only use the name instead of the `<name>:<version>` format.

Related to [pavics-jupyter-base PR](https://github.com/bird-house/pavics-jupyter-base/pull/7)

### DIFF
--- a/birdhouse/env.local.example
+++ b/birdhouse/env.local.example
@@ -237,6 +237,8 @@ export GEOSERVER_ADMIN_PASSWORD=geoserverpass
 # Name of the images displayed on the JupyterHub image selection page
 # The name order must correspond to the order of the DOCKER_NOTEBOOK_IMAGES variable,
 # and both variables should have the same number of entries.
+# Note that the names here (or the names preceding each ':', if the version is included) are also used as directory
+# names for the tutorial-notebooks directories mounted when starting the corresponding image.
 #export JUPYTERHUB_IMAGE_SELECTION_NAMES="pavics \
 #                                         eo-crim \
 #                                         nlp-crim"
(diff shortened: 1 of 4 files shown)


birdhouse/env.local.example
```diff
@@ -237,6 +237,9 @@ export GEOSERVER_ADMIN_PASSWORD=geoserverpass
 # Name of the images displayed on the JupyterHub image selection page
 # The name order must correspond to the order of the DOCKER_NOTEBOOK_IMAGES variable,
 # and both variables should have the same number of entries.
+# Note that the selection names are also used as directory names for the tutorial-notebooks directories mounted when
+# starting the corresponding image. The name can use the '<name>' or the '<name>:<version>' format. The version will be
+# excluded when mounting the corresponding directory.
 #export JUPYTERHUB_IMAGE_SELECTION_NAMES="pavics \
 #                                         eo-crim \
 #                                         nlp-crim"
```
tlvu, line 245: So basically here we can write `eo-crim` or `eo-crim:0.3.0` ?  If yes, then maybe juste provide directly as example so it is clear to the user?  Some users just do not read the comments :(

tlvu (approved):
Looks good to me.  Ouranos do not use this feature of different tutorial notebooks depending on the jupyter image but I can see how it can be useful.

I suggested a small improvement.

Don't forget to write the changelog.

birdhouse/env.local.example
ChaamC: Yes, both formats are indeed supported. 
I added an update with an example of selection name that uses the format with the version, as suggested. :)

matprov (approved):

PR #280, opened 2023-01-24
Jupyter: Restore terminal and allow recursive folder deletion

tlvu:
## Changes:
- Jupyter: allow recursive directory deletion

  This was not possible before since non-empty dir deletion was not possible.

- Jupyter: re-enable terminal for all users

  It was disabled to avoid malicious usage but with the monitoring in
  place and the demo account restricted to limited resources, it's
  probably safe to try enabling this again.

  For legitimate users, not having the terminal is pretty annoying.
  Should not penalize legit users for some rogue users.

Deployed to https://lvupavicsmaster.ouranos.ca for testing.


fmigneault (commented):
The changes look OK, but I wonder if it would not be better to have a separate `.sh` that does the full `process_delayed_eval` definition + `. env.local` + `. default.env` steps in the appropriate order. In other words, move the logic of sourcing ENV definitions together instead of duplicated in each script. It will be easier to maintain in case other approaches like in https://github.com/bird-house/birdhouse-deploy/pull/272#discussion_r1055829225 are attempted.

tlvu:
> The changes look OK, but I wonder if it would not be better to have a separate `.sh` that does the full `process_delayed_eval` definition + `. env.local` + `. default.env` steps in the appropriate order. In other words, move the logic of sourcing ENV definitions together instead of duplicated in each script. It will be easier to maintain in case other approaches like in [#272 (comment)](https://github.com/bird-house/birdhouse-deploy/pull/272#discussion_r1055829225) are attempted.

Agreed.  I think a further "logic grouping" is required.

Retrospectively, maybe I should not have supported delayed eval in the first place.  It's a bit more complex than initially planned for not a very big gain (only 3 variables so far).

So I am not sure I want to put more effort into this feature.  Let's keep it as-is for now and see how it evolves.  But you are right, next time we leverage this feature again and decide to keep this feature, the "logic grouping" should be done at the same time.

fmigneault:
> > The changes look OK, but I wonder if it would not be better to have a separate `.sh` that does the full `process_delayed_eval` definition + `. env.local` + `. default.env` steps in the appropriate order. In other words, move the logic of sourcing ENV definitions together instead of duplicated in each script. It will be easier to maintain in case other approaches like in [#272 (comment)](https://github.com/bird-house/birdhouse-deploy/pull/272#discussion_r1055829225) are attempted.
> 
> Agreed. I think a further "logic grouping" is required.
> 
> Retrospectively, maybe I should not have supported delayed eval in the first place. It's a bit more complex than initially planned for not a very big gain (only 3 variables so far).
> 
> So I am not sure I want to put more effort into this feature. Let's keep it as-is for now and see how it evolves. But you are right, next time we leverage this feature again and decide to keep this feature, the "logic grouping" should be done at the same time.


I think it would be better to group them right away, but without modifying the DELAYED_VARS approach yet. If this new method is applied right away and spreads across scripts (including user-specific overrides), it will have an even bigger impact with more restrictions to modify it later.
We should take the opportunity to adjust it while it is fairly new and not widely adopted.

tlvu:
> I think it would be better to group them right away, but without modifying the DELAYED_VARS approach yet. If this new method is applied right away and spreads across scripts (including user-specific overrides), it will have an even bigger impact with more restrictions to modify it later.
> We should take the opportunity to adjust it while it is fairly new and not widely adopted.

Then I'll remove the fix in this PR and open a different PR to avoid delaying this PR.

huard (approved):

fmigneault (approved):
Ok with scope change.

huard:
Looking at the nb failures.
- I ran `climex` on the Alpha instance and it runs, so not clear what is going on with this nb. 
- Also ran `pavics_thredds` without problem.

tlvu:
> Looking at the nb failures.
> 
>     * I ran `climex` on the Alpha instance and it runs, so not clear what is going on with this nb.

I think it's because it looks for the  `.json` file on the test server while it should look on PAVICS instead.  

Must be a precommit autoformat that breaks that line on multple lines.   But I don't understand, I thought we escaped all the lines with `# TEST_USE_PROD_DATA` already so it should not break it on multiple lines.  Will ask Trevor for the escape.

```
 cat = intake.open_esm_datastore(
      "https://host-140-154.rdext.crim.ca/catalog/climex.json"
   cat = intake.open_esm_datastore(
10:36:34      "https://host-140-154.rdext.crim.ca/catalog/climex.json"
10:36:34  )  # TEST_USE_PROD_DATA
10:36:34  cat.df.head()
  cat.df.head()
```
```
JSONDecodeError: Expecting value: line 4 column 1 (char 3)
```
> 
>     * Also ran `pavics_thredds` without problem.

Seems to be test env configuration problem.  User `authtest`, passwd `authtest1234` should have been already created before the test is launched.  @matprov can you double check problems on your side.

```
  AUTH_USR = os.getenv("TEST_MAGPIE_AUTHTEST_USERNAME", "authtest")
  AUTH_PWD = os.getenv("TEST_MAGPIE_AUTHTEST_PASSWORD", "authtest1234")
```

```
HTTPError: 401 Client Error: Unauthorized for url: https://host-140-154.rdext.crim.ca/magpie/signin
```

tlvu:
Merged this PR anyways since the 2 failed notebooks are absolutely unrelated to my JupyterHub config change.

tlvu:
> I think it's because it looks for the `.json` file on the test server while it should look on PAVICS instead.
> 
> Must be a precommit autoformat that breaks that line on multple lines. But I don't understand, I thought we escaped all the lines with `# TEST_USE_PROD_DATA` already so it should not break it on multiple lines. Will ask Trevor for the escape.

The escape for the precommit autoformater to avoid wrapping the line was already done in this commit: https://github.com/Ouranosinc/pavics-sdi/commit/75652fbe9751bc1f2b85109daf44402533639e8b

Now sure how this escape got lost.  I'll just restore it.

PR #281, opened 2023-01-25
add components representation and JSON response endpoint

fmigneault:
## Overview

New endpoint for reporting platform components.

## Changes

**Non-breaking changes**
- Proxy: add `/components` endpoint that provides a JSON list of loaded components by the platform.

  Prior to this functionality, it was impossible to know which potential capabilities, services or behaviors were to be
  expected by a given DACCS/PAVICS/birdhouse instance. Using this endpoint, nodes can obtain minimal machine-readable
  details about their supported capabilities, allowing better interoperability.

  Furthermore, developers maintaining distinct stacks and integrating different features can have a better
  understanding of behaviors by the various web services when performing requests against a given node.

**Breaking changes**
- n/a (if the operation fails to load/transform components into JSON, the list will simply be empty - silent failure)

## Related Issue / Discussion

- n/a

## Additional Information

The sample JSON output in the script is the actual response received when requesting `/components` on a test instance.


birdhouse/pavics-compose.sh
```diff
@@ -103,6 +104,7 @@ COMPOSE_DIR="`pwd`"
 # we source local configs, if present
 # we don't use usual .env filename, because docker-compose uses it
 [ -f env.local ] && . ./env.local
+. ./scripts/get-components-json.sh
```
tlvu, line 107: Given this is not a script that can be executed as all other script in folder `scripts/`, I think a more appropriate place for it would be right next to `pavics-compose.sh` with the name `get-components-json.inc.sh`.

`inc.sh` suffix to indicate it is meant to be included and not executed.  If you know other convention than `inc.sh` suffix to indicate it has to be sourced/included, please let me know.  I simply made that up !

birdhouse/scripts/get-components-json.sh
```diff
@@ -0,0 +1,41 @@
...
+#    "bird-house/birdhouse-deploy:components/monitoring",
+#    "bird-house/birdhouse-deploy:optional-components/canarie-api-full-monitoring",
+#    "bird-house/birdhouse-deploy:optional-components/all-public-access",
+#    "bird-house/birdhouse-deploy:optional-components/wps-healthchecks",
+#    "bird-house/birdhouse-deploy:optional-components/secure-thredds",
+#    "bird-house/birdhouse-deploy:optional-components/testthredds",
+#    "bird-house/birdhouse-deploy:optional-components/test-weaver",
+#    "bird-house/birdhouse-deploy:components/weaver",
+#    "bird-house/birdhouse-deploy:components/cowbird",
+#    "custom:daccs-env"
```
tlvu, line 16: I think showing all the built-in components is fine, they are already public anyways.  I am a bit uneasy to show other custom and potentially private components.

Let me try on PAVICS staging to see what leaks out.  We do have a few private extra components.

birdhouse/scripts/get-components-json.sh
```diff
@@ -0,0 +1,41 @@
...
+# each component that starts by './' gets replaced with the below birdhouse prefix to provide contextual information
+# other component locations are considered 'custom' and marked as such to provide contextual information
+BIRDHOUSE_DEPLOY_COMPONENTS_BASE="bird-house/birdhouse-deploy:"
+BIRDHOUSE_DEPLOY_COMPONENTS_LIST=$( \
+  echo "${EXTRA_CONF_DIRS}" \
+  | sed '/^[[:space:]]*$/d' \
+  | sed -E 's/^\s*([A-Za-z0-0./_-]+)\s*$/"\1",/g' \
+  | sed -E "s|^\"((\./)?(\.\./)+)+(.+)\"|\"custom:\\4\"|g" \
+  | sed -E "s|^\"\./(.*)\"|\"${BIRDHOUSE_DEPLOY_COMPONENTS_BASE}\\1\"|g" \
+  | sed '/^\n*$/d' \
```
tlvu, line 38: geez, you really like string parsing aren't you !

tlvu (changes requested):
The idea to show the various capabilities to help interoperability is very useful.

If we keep the current form, rename the included script.

But in the current form, maybe some private config might leak out.

Maybe each component can declare how it wants to be shown instead so then it can decide to not show anything as well, if it is a private component.

birdhouse/pavics-compose.sh
fmigneault: I'm not aware of any specific convention, but I think it is a good idea to have some "include" indication.

I prefer to leave it in `scripts/` though to avoid cluttering the `birdhouse` base directory. Otherwise, users will have a harder time figuring out which script is the entrypoint.

birdhouse/scripts/get-components-json.sh
fmigneault: Sure. Let me know if you see any way to improve it easily. 

I want to avoid adding too much complexity, if possible. I don't want to impose each component to define a "display" variable, as it goes against auto-detecting which components are active. If it is absolutely necessary, a "hide me" variable could be added for individual components, but this will require doing some `cat default.env | grep ... ` on each component. 
Note that the contents of components are not sourced to avoid breaking how the stack loads them.

birdhouse/scripts/get-components-json.sh
fmigneault: I'm starting to be somewhat agile with regexes 😅

huard (approved):
I like the idea, can't really judge how it's done.

birdhouse/scripts/get-components-json.sh
fmigneault: It is this part that replaces `| sed -E "s|^\"((\./)?(\.\./)+)+(.+)\"|\"custom:\\4\"|g" \`.
If there is some logic in the paths to define what should be captured by the "custom" namespace, I can add it. 
Alternatively, I could limit the inclusion of components only under `birdhouse-deploy/birdhouse`.

tlvu:
> Maybe each component can declare how it wants to be shown instead so then it can decide to not show anything as well, if it is a private component.

I think I have a cheap solution to this: add a new var `EXTRA_CONF_DIRS_PRIVATE`.  So any components that want to stay private goes there so anything in the original `EXTRA_CONF_DIRS` can freely be public.

birdhouse/scripts/get-components-json.sh
tlvu: My path is an absolute path like `/path/to/repo/config-dir`.  What does that `sed` line above do exactly?  Any paths not starting with a dot (`.`) is replaced with `custom`?  That should have handled by absolute path normally.

Alternatively, yes, let's just limit to listing components under `birdhouse-deploy/birdhouse` since those matter most.  Listing any custom repos is more or less useless since we might not know what these custom repos/components do anyways.  This will also avoid leaking private components out.

fmigneault:
@tlvu 
> I think I have a cheap solution to this: add a new var `EXTRA_CONF_DIRS_PRIVATE`. So any components that want to stay private goes there so anything in the original `EXTRA_CONF_DIRS` can freely be public.

We can either filter out `EXTRA_CONF_DIRS` with items in `EXTRA_CONF_DIRS_PRIVATE`, or work the other way around, include `EXTRA_CONF_DIRS_PRIVATE` in `EXTRA_CONF_DIRS` after `get-components-json.include.sh` is called. The first doesn't modify `EXTRA_CONF_DIRS`, but the second method avoids possible mismatching issues. Any approach you prefer?

birdhouse/scripts/get-components-json.sh
fmigneault: Ok. I'll adjust to consider only local birdhouse-deploy components. The regex was looking for something starting with at least 1 dot and any number of relative parent dirs with `../`, but it did not consider other absolute paths.

tlvu: Yes, just keep it simple and only broadcast local components for now.  Then we do not have to worry about external private components.

fmigneault: The result from https://github.com/bird-house/birdhouse-deploy/pull/281/commits/1efb04bd7e5e5115a1e7da06dbe8da16183f87dc should contain only the birdhouse-deploy components.

tlvu (changes requested):
After 122f196c3572cda79dce82cfdc5b0551a5b12e84 I still see my private components in external repos listed with their full path.

```
$ curl --silent https://medus.ouranos.ca/components/
{"components": ["bird-house/birdhouse-deploy:optional-components/canarie-api-full-monitoring",
"bird-house/birdhouse-deploy:optional-components/wps-healthchecks",
"bird-house/birdhouse-deploy:optional-components/testthredds",
"bird-house/birdhouse-deploy:optional-components/generic_bird",
"bird-house/birdhouse-deploy:components/weaver",
"bird-house/birdhouse-deploy:components/scheduler",
"bird-house/birdhouse-deploy:components/monitoring",
"/path/to/repo1/config-dir",
"/path/to/repo2/config-dir"]}
```

Maybe you want to add some absolute path in your `env.local` for testing.

birdhouse/scripts/get-components-json.include.sh
```diff
@@ -0,0 +1,64 @@
...
+fi
+
+# resolve path considering if sourced or executed, and whether from current dir, pavics-compose include or another dir
+BIRDHOUSE_DEPLOY_COMPONENTS_ROOT=$(dirname -- "$(realpath "$0")")
+if [ "$(echo "${BIRDHOUSE_DEPLOY_COMPONENTS_ROOT}" | grep -cE "/birdhouse/?\$" 2>/dev/null)" -eq 1 ]; then
+  BIRDHOUSE_DEPLOY_COMPONENTS_ROOT=.
+else
+  BIRDHOUSE_DEPLOY_COMPONENTS_ROOT="${BIRDHOUSE_DEPLOY_COMPONENTS_ROOT}/.."
+fi
+chdir "${BIRDHOUSE_DEPLOY_COMPONENTS_ROOT}"
```
tlvu, line 34:
```
$ ./pavics-compose.sh restart proxy
./scripts/get-components-json.include.sh: line 34: chdir: command not found
```
`cd` :)

birdhouse/scripts/get-components-json.include.sh
```diff
@@ -0,0 +1,64 @@
...
+# note: no quotes in 'ls' on purpose to expand glob patterns
+BIRDHOUSE_DEPLOY_COMPONENTS_LIST_KNOWN="$( \
+  ls -d1 ./*components/*/ \
+  | sed -E "s|\./(.*)/|\1|" \
+)"
+if [ -z "${BIRDHOUSE_DEPLOY_COMPONENTS_LIST_KNOWN}" ]; then
+  echo "[ERROR]" \
+    "Could not resolve known birdhouse-deploy components." \
+    "Aborting to avoid potentially leaking sensible details."
+  exit 1
```
tlvu, line 45: I'd rather just put a warning here and do not `exit 1` to avoid leaving the entire platform down during unattended autodeploy.

tlvu (approved):
This works on my staging machine.

However, `chdir` is a typo and please do not `exit` directly in this script to avoid leaving the entire platform down during unattended autodeploy.

birdhouse/scripts/get-components-json.include.sh
fmigneault: Ok, but in that case I will enforce the components to be empty as failing to retrieve those will not let us perform the filtering.

birdhouse/scripts/get-components-json.include.sh
```diff
@@ -0,0 +1,68 @@
...
+BIRDHOUSE_DEPLOY_COMPONENTS_LIST_KNOWN="$( \
+  ls -d1 ./*components/*/ \
+  | sed -E "s|\./(.*)/|\1|" \
+)"
+if [ -z "${BIRDHOUSE_DEPLOY_COMPONENTS_LIST_KNOWN}" ]; then
+  echo "[WARNING]" \
+    "Could not resolve known birdhouse-deploy components." \
+    "Aborting to avoid potentially leaking sensible details." \
+    "Components will not be reported on the platform's JSON endpoint."
+  exit 0
```
tlvu, line 49: Please do not `exit` at all !

This is being sourced by `pavics-compose.sh` so an `exit` here will exit in `pavics-compose.sh` and we do not want this since we do not want to leave the platform down during a unattended autodeploy.

birdhouse/scripts/get-components-json.include.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# default value in case of error or missing definitions
+export BIRDHOUSE_DEPLOY_COMPONENTS_JSON='{"components": []}'
+if [ -z "${EXTRA_CONF_DIRS}" ]; then
+  echo "No components in EXTRA_CONF_DIRS. Components JSON list will be empty!"
+  exit 0
+fi
+
+# resolve path considering if sourced or executed, and whether from current dir, pavics-compose include or another dir
+BIRDHOUSE_DEPLOY_COMPONENTS_ROOT=$(dirname -- "$(realpath "$0")")
```
tlvu, line 28: I am not sure why you need to discover `BIRDHOUSE_DEPLOY_COMPONENTS_ROOT`.  Since this script is being source by `pavics-compose.sh` its `PWD` will always be the same location of `pavics-compose.sh` since `pavics-compose.sh` needs to enforce that so it can find the `docker-compose.yml`.

So no need to perform any `cd` in the first place.

tlvu (changes requested):
Please do not use `exit` in this script and no need for any `cd`.

birdhouse/scripts/get-components-json.include.sh
fmigneault: Just to avoid getting this issue if we ever end up calling it instead of sourcing it. This is the actual reason why it didn't filter your absolute paths the first time, while it was working on my side. I was sourcing it directly from another dir to test.

tlvu: Ah ok.  Just curious, in what other scenario would you want to just source this one outside of `pavics-compose.sh`?

In fact, testing it outside of `pavics-compose.sh` is dangerous since you will forget to test it with `pavics-compose.sh` !

fmigneault: I tested both to be sure afterward, but I prefer to test it directly than having to go though pavics-compose and quickly breaking out of it. Anyway. they should both work for now and simply remain empty if anything fails.

birdhouse/scripts/get-components-json.include.sh
```diff
@@ -0,0 +1,65 @@
...
+#    "bird-house/birdhouse-deploy:components/cowbird"
+#  ]
+# }
+#
+
+# default value in case of error or missing definitions
+export BIRDHOUSE_DEPLOY_COMPONENTS_JSON='{"components": []}'
+if [ -z "${EXTRA_CONF_DIRS}" ]; then
+  echo "No components in EXTRA_CONF_DIRS. Components JSON list will be empty!"
+  return
```
tlvu, line 24: Why `return`?  You are not in a function !  Just do nothing.

birdhouse/scripts/get-components-json.include.sh
```diff
@@ -0,0 +1,65 @@
...
+BIRDHOUSE_DEPLOY_COMPONENTS_LIST_KNOWN="$( \
+  ls -d1 ./*components/*/ 2>/dev/null \
+  | sed -E "s|\./(.*)/|\1|" \
+)"
+if [ -z "${BIRDHOUSE_DEPLOY_COMPONENTS_LIST_KNOWN}" ]; then
+  echo "[WARNING]" \
+    "Could not resolve known birdhouse-deploy components." \
+    "Aborting to avoid potentially leaking sensible details." \
+    "Components will not be reported on the platform's JSON endpoint."
+  return
```
tlvu, line 46: Again, no need to `return` here, just put nothing.

tlvu (changes requested):

birdhouse/scripts/get-components-json.include.sh
fmigneault: No, I need to return otherwise the rest of the script resumes and defines stuff that leak the information.

birdhouse/scripts/get-components-json.include.sh
fmigneault: Return stops processing of that file at that point, but does not exit the calling script sourcing it. I've tried it with a quick test.

tlvu: Oh I didn't know `return` stopped processing of a sourced file !

I though once a file is sourced, it behaves as if the code is part of the sourcing script so `return` would break out of any functions or loop.

So I tried in `pavics-compose.sh` as well and it works fine.  So let's keep as-is, merge when ready.

tlvu (approved):

tlvu:
@fmigneault I found out on pour production curl https://pavics.ouranos.ca/components/ is empty, it was deployed by the autodeploy.  Have you tested this feature under the autodeploy environment?

fmigneault:
@tlvu 
> I found out on pour production curl [pavics.ouranos.ca/components](https://pavics.ouranos.ca/components/) is empty, it was deployed by the autodeploy. Have you tested this feature under the autodeploy environment?

No. We are not using autodeploy anywhere on our side. 
Is it possibly related to a similar problem as identified in https://github.com/bird-house/birdhouse-deploy/pull/288? 
The components script only uses `EXTRA_CONF_DIRS`, so it must not be properly defined only in that situation, as I expect your prod to have many components enabled.

PR #273, opened 2022-12-22
Weaver: fix post script to be compatible with autodeploy

tlvu:
Autodeploy runs inside its own docker container and `curl` is not available.  Therefore Weaver post script should be using `curl` from a docker image instead of locally installed flavor.

Fixes https://github.com/crim-ca/weaver/issues/508.

Output from running inside autodeploy, notice all the `tput: not found` warnings since `tput` not available inside autodeploy container either.  We can ignore these `tput` warnings.
```
executing './components/weaver/post-docker-compose-up'
+ set +o
+ old_state='set +o errexit
set +o noglob
set +o ignoreeof
set +o interactive
set +o monitor
set +o noexec
set +o stdin
set -o xtrace
set +o verbose
set +o noclobber
set +o allexport
set +o notify
set +o nounset
set +o vi
set +o pipefail'
+ set +x
Running: ./components/weaver/post-docker-compose-up
  Requested Weaver WPS providers: [
    catalog
    finch
    flyingpigeon
    hummingbird
    malleefowl
    raven
]
  Will retry requests at most for 120s
[Weaver] Wait for response from Magpie to login [https://lvupavicsmaster.ouranos.ca/magpie]. OK!
Validate Magpie token... OK!
[Weaver] Wait for response from Weaver [https://lvupavicsmaster.ouranos.ca/twitcher/ows/proxy/weaver]. OK!
[Weaver] Using URL: [https://lvupavicsmaster.ouranos.ca/twitcher/ows/proxy/weaver]
[Weaver] Wait for response from remote WPS provider [catalog] on [https://lvupavicsmaster.ouranos.ca/twitcher/ows/proxy/catalog].                                                                                                                                                                                                                                                                                                                                                                                               
[Weaver] Got valid response from remote WPS provider [catalog].
[Weaver] Unregistering any remote WPS provider matching [catalog].
[Weaver] Delete [catalog] response: 204
[Weaver] Registering remote WPS provider [catalog] on [https://lvupavicsmaster.ouranos.ca/twitcher/ows/proxy/catalog]... OK!
[Weaver] Wait for response from remote WPS provider [finch] on [https://lvupavicsmaster.ouranos.ca/twitcher/ows/proxy/finch].
[Weaver] Got valid response from remote WPS provider [finch].
[Weaver] Unregistering any remote WPS provider matching [finch].
[Weaver] Delete [finch] response: 204
[...]

PR #282, opened 2023-01-26
fixes for post-docker-compose-up with weaver component

fmigneault:
## Overview

Fixes for issues identified when starting the stack on a test instance following #273.

## Changes

**Non-breaking changes**
- Weaver: fixes for running `post-docker-compose-up` operation.

  - When the target `curl` image was not already available on the machine (each time for ephemeral test instances), the 
    docker pull outputs on the first call would mangle the monitoring messages. An initial pull is done to avoid it.

  - When running on `sh` (as expected by the script's shebang), the utility variable `$RANDOM` is missing.
    A POSIX portable equivalent is used if `$RANDOM` could not be resolved.

**Breaking changes**
-  n/a

## Related Issue / Discussion

The output looked like the following, with `docker pull` outputs in between the initial part and step update of the message trying to obtain the login response. 

```log
[Weaver] Wait for response from Magpie to login [https://10.0.2.15/magpie].Unable to find image 'curlimages/curl:7.87.0' locally
7.87.0: Pulling from curlimages/curl
213ec9aee27d: Already exists
307ed28f4c25: Pulling fs layer
<rest of docker pull lines>
<rest of Magpie login message>
```


birdhouse/components/weaver/post-docker-compose-up
```diff
@@ -96,12 +96,18 @@ if [ -z "$WEAVER_CURL_IMAGE" ]; then
     WEAVER_CURL_IMAGE="curlimages/curl:7.87.0"
 fi
 
+# POSIX portable RNG if RANDOM does not exist on the current shell
+RANDOM_NUMBER=${RANDOM:-$(tr -dc 0-9 < /dev/urandom | head -c 5)}
```
tlvu, line 100: nice catch

birdhouse/components/weaver/post-docker-compose-up
```diff
@@ -96,12 +96,18 @@ if [ -z "$WEAVER_CURL_IMAGE" ]; then
...
+
 # To know when a docker run was started in case it hangs.
-DOCKER_RUN_TAG="`date -Isecond | sed 's/:/_/g' | sed 's/+/p/g'`_$RANDOM"
+DOCKER_RUN_TAG="weaver_post_curl_`date -Isecond | sed 's/:/_/g' | sed 's/+/p/g'`_${RANDOM_NUMBER}"
```
tlvu, line 103: I intentionally not put `weaver_post_curl_` in `DOCKER_RUN_TAG` so the same tag can be used for another docker run invocation in the future, if this script needs more than `curl` later.

But can leave this as-is for now.  Later if more than `curl` via `docker run` is needed, you can make this `DOCKER_RUN_TAG` more generic.

tlvu (approved):
All good for me.

I suspect the mangling of the output is because of output buffering.  I assume this is in Jenkins build output?

If the `docker pull` trick works then fine.  Otherwise, investigate buffering or stderr interleaving with stdout instead.

fmigneault:
> I suspect the mangling of the output is because of output buffering. I assume this is in Jenkins build output?

No, it was in my local terminal. I expect the same on Jenkins though. It is because of the first `docker run curl ...`, the image is not found locally, so it triggers its `docker pull` with the layers output, but this happens between multiple `printf` statements used to do iterative `....` printing when attempting requests until the response is received (until magpie finished starting).

birdhouse/components/weaver/post-docker-compose-up
fmigneault: I thought that might be the case, but since the random value was fixed in `DOCKER_RUN_TAG`, the full date/rng command would have to be duplicated as well anyway.

tlvu: I intended one `DOCKER_RUN_TAG` per invocation of each script.  The `RANDOM` part was in case 2 scripts start exactly at the same second so the timestamp is not different enough.  But this is small details.  Either way is fine with me.

PR #283, opened 2023-01-31
optional-component secure-data-proxy + related configs in cowbird/magpie/weaver

fmigneault:
Multiple updates to support secured WPS outputs.

## Changes

**Non-breaking changes**
- secure-data-proxy: add new [`secure-data-proxy`][secure-data-proxy] optional component.

  When enabled, this component will enforce authentication and authorization to be resolved against the `/wpsoutputs`
  endpoint prior to accessing the results produced by WPS executions. A Magpie service named `secure-data-proxy` is
  created to define the resource and permission hierarchy of directories and files the users and groups can access.
  When disabled, the original behavior to provide open access to `/wpsoutputs` is employed.

  A variable named `SECURE_DATA_PROXY_AUTH_INCLUDE` is dynamically assigned based on the activation or not of this
  component. Corresponding validation of optional/mandatory/delayed-eval variables used by this component are also
  applied dynamically, as well as mounting the necessary `nginx` and `docker-compose` extended configurations.

- Weaver: adjust user-context output directory hooks and permissions for [`secure-data-proxy`][secure-data-proxy].

  When a process defined in Weaver (either a WPS provider or a local definition) is executed by a user that was granted
  authorization to run a job, the corresponding user-context directory under `/wpsoutputs/users/{user-id}` will be used
  for storing the execution outputs and will have the appropriate permissions set for that user to grant them access to
  those outputs.

- Magpie/Twitcher: update minimum version `magpie>=3.31.0` to employ `twitcher>=0.8.0` in `MapgieAdatepr`.

  - Resolve an issue where `response.request` references were not set in OWS proxy responses when handled by Twitcher.
    This caused `MapgieAdatepr` response hooks to fail, which in turn caused failing requests for any non-WPS
    service that defined any proxy request hook, such as in the case of [`weaver`][weaver-component] component.

  - Adds the Twitcher ``/ows/verify/{service_name}[/{extra_path}`` endpoint employed for validating authorized access
    to Magpie service/resources, in the same fashion as the protected proxy endpoint, but without performing the proxied
    request toward the target service. This is mandatory for using the new [`secure-data-proxy`][secure-data-proxy] 
    optional component, otherwise the proxy endpoint triggers data download twice, once for authorization and another
    for actually accessing the data.

**Breaking changes**
- n/a (technically)
  New feature  [`secure-data-proxy`][secure-data-proxy] with optional component should default to using the original methodology of public access if not enabled.

## Related Issue / Discussion

[...]

Issue #214, opened 2021-11-03
Migrate existing pieces to new pluggable component architecture, part 1 clean up the unused pieces


tlvu:
The following are scheduled to be removed:

```
  frontend:
    image: pavics/pavics-frontend:1.0.5
  project-api:
    image: pavics/pavics-project-api:0.9.0
  catalog:
    image: pavics/pavics-datacatalog:0.6.11
  solr:
    image: pavics/solr:5.2.1
  ncwms2:
    image: pavics/ncwms2:2.0.4
```

Includes finding and removing all references from
* configs of those components above, and remove the following unused directory `birdhouse/config/ncops`
* canarie-api, magpie and other remaining components if exist
* notebooks tested by Jenkins only.  Other notebooks are not included.

mishaschwartz:
Hi @tlvu. I'm working on rethinking the deployment strategy for DACCS/PAVICS over at UofT. Do you mind if I take over this issue as well as the other "Migrate existing pieces to new pluggable component architecture" issues in this repository?

tlvu:
@mishaschwartz not at all.  Please let me know if you need any further info.

You can see the pluggable component architectures already in place for new `components/` and `optional-components/`already.  It also works for external repos, ex: https://github.com/bird-house/birdhouse-deploy-ouranos

The plan is to break out all the currently enabled components by default the same way.

The tricky part is to keep backward compatibility with the current default when user do not specify any extra components.

The other tricky part is inter-dependencies between the various components.  New WPS services (birds) are self-contained with their own DB (ex: https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse/optional-components/generic_bird), but all current WPS services (Finch, Raven, Hummingbird, ...) all share the same DB !

There are also some tightly coupled components, ex: Twitcher and Magpie.  These 2 versions have to always match and they probably do not make sense to exist as separate components, ex: activating Magpie alone without Twitcher make a little sense.

There are currently 3 PAVICS deployments: Ouranos, CRIM and PCIC.  I will tag the persons responsible for CRIM and PCIC below.

Please share your thoughts and code change early with us so we can plan the upgrade path on our side.

Ideally all changes are small, incremental and backward compatible so upgrade is painless.  For sure this is not 100% possible so we have a find a middle ground.  I intentionally break the task into many smaller parts to hopefully make it small and incremental.

[...]

PR #284, opened 2023-02-01
fix CanarieAPI config to support Python 3.x

fmigneault:
## Overview

Please include a summary of the changes and which issues are fixed. 

Please also include relevant motivation and context. 

List any dependencies that are required for this change.

## Breaking changes

- CanarieAPI: update to `0.7.1`.

  - The Docker running `CanarieAPI` is now using Python 3 (since `0.4.x` tags).
    Configurations need to be updated if any specific Python 2 definitions were used.
    See [2to3](https://docs.python.org/3/library/2to3.html) to help migrate configurations automatically if necessary.
  - Update the [CanarieAPI configuration](birdhouse/config/canarie-api/docker_configuration.py.template) to use
    Python 3.x executable code.

## Changes

- CanarieAPI: update to `0.7.1`.

  - The server node now provides a generic ``server`` configuration for the current ``platform`` definition.
  - Added multiple missing docuementation references for all the services included within `CanarieAPI` configurations.
  - With new `CanarieAPI` version, a slightly improved UI with more service details are provided for the active server:

image

- Add optional variables witht defaults to define reference Docker image version tags.

  Following optional variables are defined by default. These are used as reference in the respective Docker compose
  service definition of these components, as well as in their `CanarieAPI` configuration to retrieve the release time
  of the tag, and refer to relevant URL references as needed.

  - `CATALOG_VERSION`
  - `FINCH_VERSION`
  - `FLYINGPIGEON_VERSION`
  - `GEOSERVER_VERSION`
  - `HUMMINGBIRD_VERSION`
  - `MALLEEFOWL_VERSION`
  - `RAVEN_VERSION`

## Fixes:

- CanarieAPI: update to `0.7.1`.

  - Fixes an `AttributeError` raised due to misconfiguration of the Web Application with Flask 2.x definitions
    (relates to [Ouranosinc/CanarieAPI#10](https://github.com/Ouranosinc/CanarieAPI/pull/10)).
  - Skip over `0.4.x`, `0.5.x`, `0.6.x`  versions to avoid issue related to `cron` job monitoring and log parser
    command failures in order to collect configured service statistics and statuses
    (see also [Ouranosinc/CanarieAPI#14](https://github.com/Ouranosinc/CanarieAPI/pull/14)).

- Weaver: update CanarieAPI monitoring definitions
  - Move monitoring of public endpoint under [optional-components/canarie-api-full-monitoring][canarie-monitor].
  - Add monitoring of private endpoint by default when using Weaver component.

- Cowbird: update CanarieAPI monitoring definitions
  - Add monitoring of public endpoint under [optional-components/canarie-api-full-monitoring][canarie-monitor].
  - Add public Magpie permission on Cowbird entrypoint only to allow its monitoring.

## Additional Information

Resolves the following log error.

```
proxy              | [2023-01-31 19:37:01 +0000] [37] [DEBUG] GET /canarie/
proxy              | [2023-01-31 19:37:01,708] [37] [INFO] app_object : Disconnecting from database
proxy              | [2023-01-31 19:37:01,709] [37] [DEBUG] app_object : Using db filename : /opt/local/src/CanarieAPI/stats.db
proxy              | [2023-01-31 19:37:01 +0000] [37] [DEBUG] Closing connection. 
proxy              | [2023-01-31 19:37:02 +0000] [37] [DEBUG] GET /canarie/background.jpg
proxy              | [2023-01-31 19:37:02,176] [37] [INFO] app_object : Disconnecting from database
proxy              | [2023-01-31 19:37:02,176] [37] [DEBUG] app_object : Using db filename : /opt/local/src/CanarieAPI/stats.db
proxy              | [2023-01-31 19:37:02 +0000] [37] [ERROR] Error handling request /canarie/background.jpg
proxy              | Traceback (most recent call last):
proxy              |   File "/usr/local/lib/python2.7/dist-packages/gunicorn/workers/async.py", line 56, in handle
proxy              |     self.handle_request(listener_name, req, client, addr)
proxy              |   File "/usr/local/lib/python2.7/dist-packages/gunicorn/workers/ggevent.py", line 152, in handle_request
proxy              |     super(GeventWorker, self).handle_request(*args)
proxy              |   File "/usr/local/lib/python2.7/dist-packages/gunicorn/workers/async.py", line 107, in handle_request
proxy              |     respiter = self.wsgi(environ, resp.start_response)
proxy              |   File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1997, in __call__
proxy              |     return self.wsgi_app(environ, start_response)
proxy              |   File "/opt/local/src/CanarieAPI/canarieapi/reverse_proxied.py", line 33, in __call__
proxy              |     return self.app(environ, start_response)
proxy              |   File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1985, in wsgi_app
proxy              |     response = self.handle_exception(e)
proxy              |   File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1532, in handle_exception
proxy              |     handler = self._find_error_handler(InternalServerError())
proxy              |   File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1449, in _find_error_handler
proxy              |     .get(code))
proxy              |   File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1440, in find_handler
proxy              |     handler = handler_map.get(cls)
proxy              | AttributeError: 'function' object has no attribute 'get'
```

The above problem lead to unresponsive CanarieAPI, which in turn caused the platform to fail responding with successful monitoring statuses for requests toward the configured services, which in turn, caused the stack to never completely boot. When the stack failed to boot, the *End2End Test Results* (example: https://github.com/bird-house/birdhouse-deploy/pull/283#issuecomment-1410825258) could not run due to the unresponsive instance, which is the cause of the incomplete output:

`````
Tests URL :

NOTEBOOK TEST RESULTS
   
````
</code>
``
`````

# To Do (in follow-up PRs)

- remove deprecated configs deleted in #287 
- remove deprecated configs deleted in #291 
- remove deprecated configs deleted in #292

### CODE REVIEW

**Comment by fmigneault** [comment]
run tests 

using required https://github.com/Ouranosinc/CanarieAPI/releases/tag/0.4.4 now deployed

File: birdhouse/config/canarie-api/docker_configuration.py.template [hunk 1]
```diff
@@ -279,15 +280,36 @@ PLATFORMS = {
...
+        })
+        with open(filepath, 'rb') as file:
+            exec(compile(file.read(), filepath, 'exec'), globals, locals)
+
 
 if os.path.exists(CANARIE_MONITORING_EXTRA_CONF_DIR):
     # alphabetically sorted for reproducible override precedence
     for extra_conf in sorted(os.listdir(CANARIE_MONITORING_EXTRA_CONF_DIR)):
-        extra_conf_full_path = "{CANARIE_MONITORING_EXTRA_CONF_DIR}/{extra_conf}".format(**locals())
+        extra_conf_full_path = "{}/{}".format(CANARIE_MONITORING_EXTRA_CONF_DIR, extra_conf)
```
tlvu, line 309: Why change this?  This is the python2 way to simulate the awesome python3 f-string.  f-strings are way more readable.

birdhouse/config/canarie-api/docker_configuration.py.template
```diff
@@ -279,15 +280,36 @@ PLATFORMS = {
...
         # only handle files ending with .py
         if os.path.isfile(extra_conf_full_path) and extra_conf_full_path.endswith(".py"):
-            logger.info("canarie-api: loading extra config '{extra_conf_full_path}'".format(**locals()))
-            execfile(extra_conf_full_path)
+            logger.info("canarie-api: loading extra config '%s'", extra_conf_full_path)
```
tlvu, line 312: Same comment here and below about keeping f-string emulation in python2.

birdhouse/config/canarie-api/docker_configuration.py.template
```diff
@@ -279,15 +280,36 @@ PLATFORMS = {
...
+            exec_file(extra_conf_full_path, locals=locals())
```
tlvu, line 313: Passing locals is the fix to AttributeError?

Does `execfile()` in python2 accept passing in locals?

tlvu (changes requested):
> The above problem lead to unresponsive CanarieAPI, which in turn caused the platform to fail responding with successful monitoring statuses for requests toward the configured services, which in turn, caused the stack to never completely boot.

It's funny that the error happens during the E2E test.  I don't remember seeing this error on our production instance.

Missing CHANGES.md update, small changes to revert and the PR description needs a little review I think.

fmigneault:
@tlvu While testing, I identified another issue where CanarieAPI is monitored in Jenkins to check that the stack is "ready", but the value it looks for does not exist. There's a PR open that @matprov needs to review and integrate beforehand. Should be fixed soon enough 🤞

birdhouse/config/canarie-api/docker_configuration.py.template
fmigneault: It does, but seems like it doesn't like getting them by keyword. I'll adjust the call accordingly.

birdhouse/config/canarie-api/docker_configuration.py.template
fmigneault: In this case, it is not f-string or string formatting (directly), but the `logging` module arguments if logging level is active.

birdhouse/config/canarie-api/docker_configuration.py.template
fmigneault: I can revert, but expanding `locals` on each call instead of just using the variables themselves is not that great. 
Alternatively, we could also do:
```python
"{dir}/{conf}".format(dir=CANARIE_MONITORING_EXTRA_CONF_DIR, conf=extra_conf)
```
or simply 
```python
CANARIE_MONITORING_EXTRA_CONF_DIR + "/" + extra_conf
```

fmigneault:
run tests

matprov:
@fmigneault Jenkins was restarting while running latest E2E tests, therefore I trigger them again.

run tests

matprov:
run tests

fmigneault:
Working on a new version to provide more capabilities: https://github.com/Ouranosinc/CanarieAPI/pull/11

With changes in https://github.com/bird-house/birdhouse-deploy/pull/284/commits/76743a4fb0328ccc4c2aeb9cbb843dc2148f9566, I get the following result on a local instance :
image

birdhouse/config/canarie-api/docker_configuration.py.template
tlvu: `CANARIE_MONITORING_EXTRA_CONF_DIR + "/" + extra_conf` is less bad on the eye yes.

birdhouse/config/canarie-api/docker_configuration.py.template
tlvu: Ah, I read too quick, you are using logging arguments, okay keep your version, which is slightly more optimized when the logging level is not reached.

birdhouse/config/canarie-api/docker_configuration.py.template
```diff
@@ -284,23 +284,23 @@ CANARIE_MONITORING_EXTRA_CONF_DIR = os.environ.get(
...
 
-        See https://stackoverflow.com/a/41658338/5936364.
-        """
-        if globals is None:
-            globals = {}
-        globals.update({
+    See https://stackoverflow.com/a/41658338/5936364.
+    """
+    _globals = kwargs.get("globals") or globals()
+    _locals = kwargs.get("locals") or _globals  # as per exec/execfile documentation
```
tlvu, line 638: Shouldn't this be `kwargs.get("locals") or locals()` ?

tlvu: Sorry, you are right.

fmigneault:
If #284 is integrated first, this PR will have to be updated to remove the extra configs that will otherwise make the platform "not ready" as they would not respond anymore.

tlvu:
> If #284 is integrated first, this PR will have to be updated to remove the extra configs that will otherwise make the platform "not ready" as they would not respond anymore.

Huh ... #284 is this PR, is this comment in the wrong PR?

tlvu:
It seems this PR modify a few same files as the secure-data-proxy PR #283.  I think we should get the secure-data-proxy PR in first, then we can properly test for this one.

fmigneault:
Oups.
I meant #287 that removes catalog. Same thing for #291 and #292 since the "service" configurations within CanarieAPi would not be relevant anymore.

birdhouse/config/canarie-api/docker_configuration.py.template
```diff
@@ -12,11 +16,77 @@ DATABASE = {
     'log_pid': '/var/run/nginx.pid'
 }
 
+
+def _get_release_time_from_github_tag(repository_reference, tagged_version):
+    # type: (str, str) -> str
+
+    repo = repository_reference.split("github.com/")[-1]
+    url = "https://api.github.com/repos/{}/tags".format(repo)
```
tlvu, line 24: `url = f'https://api.github.com/repos/{repo}/tags'`

This is f-string, since python 3.6, much more natural and readable.  Try to use this if you are 3.6 or higher for all your strings.

birdhouse/config/canarie-api/docker_configuration.py.template
```diff
@@ -31,84 +101,22 @@ SERVICES = {
...
     },
     'indices': {
         'info': {
             'name': 'Climate indices',
             'synopsis': 'Users of climate data are interested in specific indices such as the number of freeze-thaw cycles, the number of degree-days of cooling, the duration of heatwaves, etc. This returns annual values of the most popular climate indices.',
-            'version': '0.7.1',
+            'version': FINCH_VERSION,
             'institution': 'Ouranos',
-            'releaseTime': '2021-03-25T00:00:00Z',
+            'releaseTime': FINCH_RELEASE_TIME,
```
tlvu, line 119: woooh fancy !

birdhouse/config/canarie-api/docker_configuration.py.template
```diff
@@ -234,14 +242,351 @@ SERVICES = {
...
+            'support': 'https://github.com/ouranosinc/pyramid-phoenix/issues',
+            'source': 'https://github.com/ouranosinc/pyramid-phoenix',
+            'tryme': 'https://${PAVICS_FQDN_PUBLIC}/geoserver/web/',  # FIXME: original value doesn't make sense
+            'licence': 'https://github.com/ouranosinc/pyramid-phoenix/blob/master/LICENSE.txt',
+            'provenance': 'https://ouranosinc.github.io/pavics-sdi/provenance/index.html'
+        },
+        "monitoring": {
+            "Phoenix": {
+                'request': {
+                    'url': 'https://${PAVICS_FQDN_PUBLIC}/geoserver/web/'  # FIXME: original value doesn't make sense
```
tlvu, line 376: Yeah that's wierd geoserver monitoring url for phoenix in the original value.

birdhouse/config/canarie-api/docker_configuration.py.template
```diff
@@ -234,14 +242,351 @@ SERVICES = {
...
+            'support': 'https://github.com/bird-house/finch/issues',
+            'source': 'https://github.com/bird-house/finch',
+            'tryme': 'https://ouranosinc.github.io/pavics-sdi/notebooks/climate_indices.html',
+            'licence': 'https://github.com/bird-house/finch/blob/master/LICENSE.txt',
+            'provenance': 'https://ouranosinc.github.io/pavics-sdi/provenance/index.html'
+        },
+        'monitoring': {
+            'Finch': {
+                'request': {
+                    'url': 'https://${PAVICS_FQDN_PUBLIC}${TWITCHER_PROTECTED_PATH}/finch?service=WPS&version=1.0.0&request=GetCapabilities'
```
tlvu, line 475: This now changes from `http` with internal port to `https`?  This public access is already tested in `optional-components/canarie-api-full-monitoring/`

The idea is we can potentially restrict birds from public access so only the internal access is guarantied to work everytime.

So keep all internal access here and only components not behind Twitcher can directly be tested with their public access.

birdhouse/config/canarie-api/docker_configuration.py.template
```diff
@@ -279,15 +624,52 @@ PLATFORMS = {
...
-            logger.info("canarie-api: ignoring extra config '{extra_conf_full_path}'".format(**locals()))
+            logger.info("canarie-api: ignoring extra config '%s'", extra_conf_full_path)
+
+
+# For  historical reasons, the "node" service was used to contain all "services" of the current "platform".
+# However, those services should instead be listed directly at the root as individual services of the current platform,
+# to directly obtain their status instead of going through the "node" service sub-page. This also allows us to provide
+# the individual links (docs, release, etc.) for the services themselves. The services can also be represented nested
+# under the "Platform" section as well. This "Platform" section could be extended with other known platforms part of a
+# DACCS network. For backward-compatibility, repopulate the "node" service using moved definitions.
```
tlvu, line 667: nice

tlvu (changes requested):

tlvu:
> Working on a new version to provide more capabilities: [Ouranosinc/CanarieAPI#11](https://github.com/Ouranosinc/CanarieAPI/pull/11)
> 
> With changes in [76743a4](https://github.com/bird-house/birdhouse-deploy/commit/76743a4fb0328ccc4c2aeb9cbb843dc2148f9566), I get the following result on a local instance : image

You might want to include this services list screenshot to better show the refactoring you did.

tlvu:
Also remove monitoring of ncWMS2 and Catalog that we discussed in https://github.com/bird-house/birdhouse-deploy/pull/293#issuecomment-1430512372

Might as well remove everything that would eventually be removed by https://github.com/bird-house/birdhouse-deploy/issues/214

tlvu:
Many components do not resolve release time:
```
canarie-api: failed retrieving version [0.9.2] release time from [github: bird-house/finch]                                                                                                                  
  File "/config/docker_configuration.py", line 63, in get_release_time_from_repo_tag                                                                                                                         
    release = _get_release_time_from_github_tag(repository_reference, tagged_version)                                                                                                                        
  File "/config/docker_configuration.py", line 27, in _get_release_time_from_github_tag                                                                                                                      
    raise ValueError("HTTP status from tags list: {}".format(resp.status_code))                                                                                                                              
ValueError: HTTP status from tags list: 403                                                                                                                                                                  
canarie-api: failed retrieving version [3.31.0] release time from [github: Ouranosinc/Magpie]                                                                                                                
  File "/config/docker_configuration.py", line 63, in get_release_time_from_repo_tag                                                                                                                         
    release = _get_release_time_from_github_tag(repository_reference, tagged_version)                                                                                                                        
  File "/config/docker_configuration.py", line 27, in _get_release_time_from_github_tag                                                                                                                      
    raise ValueError("HTTP status from tags list: {}".format(resp.status_code))                                                                                                                              
ValueError: HTTP status from tags list: 403                                                                                                                                                                  
canarie-api: failed retrieving version [0.8.0] release time from [github: bird-house/twitcher]                                                                                                               
  File "/config/docker_configuration.py", line 63, in get_release_time_from_repo_tag                                                                                                                         
    release = _get_release_time_from_github_tag(repository_reference, tagged_version)                                                                                                                        
  File "/config/docker_configuration.py", line 27, in _get_release_time_from_github_tag                                                                                                                      
    raise ValueError("HTTP status from tags list: {}".format(resp.status_code))                                                                                                                              
ValueError: HTTP status from tags list: 403                                                                                                                                                                  
canarie-api: found version [2.19.0-kartoza-build20210329-r2-with-snakeyaml] release time [2022-08-12T00:03:54.549806Z] from [docker: pavics/geoserver]                                                       
canarie-api: found version [pavics-0.2.3] release time [2018-05-09T15:54:39.405697Z] from [docker: pavics/pyramid-phoenix]                                                                                   
canarie-api: found version [0.6.11] release time [2019-11-21T19:28:55.787132Z] from [docker: pavics/pavics-datacatalog]                                                                                      
canarie-api: failed retrieving version [pavics-0.3.5] release time from [github: Ouranosinc/malleefowl]                                                                                                      
  File "/config/docker_configuration.py", line 63, in get_release_time_from_repo_tag                                                                                                                         
    release = _get_release_time_from_github_tag(repository_reference, tagged_version)                                                                                                                        
  File "/config/docker_configuration.py", line 27, in _get_release_time_from_github_tag                                                                                                                      
    raise ValueError("HTTP status from tags list: {}".format(resp.status_code))                                                                                                                              
ValueError: HTTP status from tags list: 403                                                                                                                                                                  
canarie-api: found version [5.2.1] release time [2020-01-21T19:35:14.093976Z] from [docker: pavics/solr]                                                                                                     
canarie-api: found version [4.6.18] release time [2021-12-18T21:31:59.768806Z] from [docker: unidata/thredds-docker]                                                                                         
canarie-api: found version [1.4.0-20210506] release time [2021-05-06T22:12:51.363362Z] from [docker: pavics/jupyterhub]                                                                                      
canarie-api: failed retrieving version [4.22.0] release time from [github: crim-ca/weaver]                                                                                                                   
  File "/config/docker_configuration.py", line 63, in get_release_time_from_repo_tag                                                                                                                         
    release = _get_release_time_from_github_tag(repository_reference, tagged_version)                                                                                                                        
  File "/config/docker_configuration.py", line 27, in _get_release_time_from_github_tag                                                                                                                      
    raise ValueError("HTTP status from tags list: {}".format(resp.status_code))                                                                                                                              
ValueError: HTTP status from tags list: 403
```

tlvu:
I think the cronjob that actually perform the monitoring might be broken with this refactoring, there are no result after 10 minutes of uptime !
Screenshot from 2023-02-15 13-57-21

birdhouse/components/cowbird/config/canarie-api/cowbird_config.py.template
```diff
@@ -0,0 +1,38 @@
...
+        'support': 'https://github.com/Ouranosinc/cowbird//issues',
+        'source': 'https://github.com/Ouranosinc/cowbird/',
+        'tryme': 'https://${PAVICS_FQDN_PUBLIC}/cowbird/',
+        'licence': 'https://github.com/Ouranosinc/cowbird//blob/${COWBIRD_VERSION}/LICENSE',
+        'provenance': 'https://github.com/Ouranosinc/cowbird/'
+    },
+    "monitoring": {
+        "Cowbird": {
+            'request': {
+                'url': 'http://cowbird:7000/'
```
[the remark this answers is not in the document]
fmigneault: It has a protected endpoint on `/cowbird/`, but admin permissions are required since cowbird is only for synching permissions/services. We could however add  `/cowbird/` and  `/cowbird/version` with exact path match as permitted publicly in Magpie.

fmigneault:
> Many components do not resolve release time:

I think that might be related to GitHub rate-limiting. I will need to investigate further...

> I think the cronjob that actually perform the monitoring might be broken with this refactoring, there are no result after 10 minutes of uptime !

Can you check if the `/canarie/node/service/status` has anything on your side? 
The `server` only replicates the contents of `node` that should be the same endpoints/config as the originals.
I have noticed that the status updates after calling `/canarie/test` once.

birdhouse/config/canarie-api/docker_configuration.py.template
fmigneault: Should we consider `phoenix` as another once to be removed in scope of #214 ?

tlvu: Oh maybe, I am not aware anyone on Ouranos side use it, you probably know on CRIM side, but we probably should ask PCIC, just in case.

tlvu:
> Can you check if the `/canarie/node/service/status` has anything on your side? The `server` only replicates the contents of `node` that should be the same endpoints/config as the originals. I have noticed that the status updates after calling `/canarie/test` once.

Oops I switched my VM to a different branch to test something else already.  I'll keep an eye on this next time I come back to your branch.

I guess you can also spin up a throwaway instance in your Jenkins, and ssh into it to debug.

birdhouse/config/canarie-api/docker_configuration.py.template
fmigneault: Ok, I'll remove the config since its not monitoring it anyway. The actual component removal can be considered later.

fmigneault: Well... I cannot remove just the monitoring part as it fails schema validation. I'll simply adjust the URL to match where it actually is accessible, and the config can be removed entirely when the component is removed.

birdhouse/config/canarie-api/docker_configuration.py.template
```diff
@@ -205,7 +213,7 @@ SERVICES = {
     'slicer': {
```
fmigneault, line 213: @tlvu @huard 
I'd like validations about this item. 
The `stats` refer to `finch`, but the `monitoring` name is `Flyingpigeon`, and `finch` is already added elsewhere in the config.
I cannot find `slicer` in both of these birds' processes. 
Which process is being targeted so I can update the config accordingly?

birdhouse/config/canarie-api/docker_configuration.py.template
fmigneault: Now using Python 3 f-string syntax exclusively.

mishaschwartz:
@fmigneault do you want me to translate this to the new pluggable architecture that was just merged?

fmigneault:
@mishaschwartz Yes, that would be really helpful. Thanks.

birdhouse/config/canarie-api/docker_configuration.py.template
tlvu: Since Finch is taking on more and more of Flyingpigeon, maybe the slicer functionality is now rightfully Finch?  David would be able to confirm.  Just my guess.

mishaschwartz:
@fmigneault Ok I've done a first pass on the merge with master and I've updated to version 0.6.0. If anything is looking weird still after the tests run I can look at it again tomorrow.

mishaschwartz:
@fmigneault it looks like the test failure is due to canarie not getting an "OK" status for some services. But, it looks like all the services that are failing are ones that are due to be removed in #302 so I don't know if you want to wait for that PR to get pulled in first or if you want to fix it here first. Have a look but it might not be worth your time if we can get #302 in before this one.

fmigneault:
The monitoring shows the following: 
```
proxy              | [2023-03-23 15:20:55,435] [20] [INFO] canarieapi.app_object : node.Thredds-public : Ok
proxy              | [2023-03-23 15:20:55,575] [20] [INFO] canarieapi.app_object : node.Catalog-public : Ok
proxy              | [2023-03-23 15:20:55,710] [20] [INFO] canarieapi.app_object : node.Malleefowl-public : Ok
proxy              | [2023-03-23 15:20:56,178] [20] [INFO] canarieapi.app_object : node.Finch-public : Ok
proxy              | [2023-03-23 15:20:56,400] [20] [INFO] canarieapi.app_object : node.Raven-public : Ok
proxy              | [2023-03-23 15:20:56,490] [20] [INFO] canarieapi.app_object : node.Hummingbird-public : Ok
proxy              | [2023-03-23 15:20:56,523] [20] [WARNING] canarieapi.app_object : Bad return code from https://host-140-88.rdext.crim.ca/twitcher/ows/proxy/ncWMS2/wms?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.3.0 (Expecting 200, Got 401
proxy              | [2023-03-23 15:20:56,523] [20] [INFO] canarieapi.app_object : node.ncWMS2-public : Up but returning unexpected response
proxy              | [2023-03-23 15:20:56,552] [20] [INFO] canarieapi.app_object : node.testthredds-public : Ok
proxy              | [2023-03-23 15:20:56,562] [20] [INFO] canarieapi.app_object : node.testthredds : Ok
proxy              | [2023-03-23 15:20:56,569] [20] [INFO] canarieapi.app_object : node.Cowbird : Ok
proxy              | [2023-03-23 15:20:56,612] [20] [INFO] canarieapi.app_object : node.Cowbird-public : Ok
proxy              | [2023-03-23 15:20:56,619] [20] [INFO] canarieapi.app_object : node.Weaver : Ok
proxy              | [2023-03-23 15:20:56,698] [20] [INFO] canarieapi.app_object : node.Catalog : Ok
proxy              | [2023-03-23 15:20:56,914] [20] [INFO] canarieapi.app_object : node.Finch : Ok
proxy              | [2023-03-23 15:20:57,026] [20] [INFO] canarieapi.app_object : node.GeoServer : Ok
proxy              | [2023-03-23 15:20:57,060] [20] [INFO] canarieapi.app_object : node.Jupyter : Ok
proxy              | [2023-03-23 15:20:57,097] [20] [INFO] canarieapi.app_object : node.Magpie : Ok
proxy              | [2023-03-23 15:20:57,194] [20] [INFO] canarieapi.app_object : node.Malleefowl : Ok
proxy              | [2023-03-23 15:20:57,227] [20] [INFO] canarieapi.app_object : node.Phoenix : Ok
proxy              | [2023-03-23 15:20:57,398] [20] [INFO] canarieapi.app_object : node.Raven : Ok
proxy              | [2023-03-23 15:20:57,410] [20] [WARNING] canarieapi.app_object : Bad response content from http://host-140-88.rdext.crim.ca:8983/solr/birdhouse/select?q=CMIP5&fq=model:MPI-ESM-MR&fq=experiment:rcp45&fq=variable:tasmax&fq=institute:MPI-M&fq=frequency:mon&wt=json (Expecting : 
proxy              | .*catalog_url":".+/testdata/flyingpigeon/cmip5.*/catalog.xml.*
proxy              | Got : 
proxy              | {"responseHeader":{"status":0,"QTime":1,"params":{"q":"CMIP5","fq":["model:MPI-ESM-MR","experiment:rcp45","variable:tasmax","institute:MPI-M","frequency:mon"],"wt":"json"}},"response":{"numFound":0,"start":0,"docs":[]}}
```

Seems like there is something incorrectly configured for `node.ncWMS2-public` in Magpie permissions because the exact same request after login works. I think that https://github.com/bird-house/birdhouse-deploy/blob/37fe32ebba87a8c6afe3aa83852bfb0dfbd1f265/birdhouse/optional-components/canarie-api-full-monitoring/default.env should include https://github.com/bird-house/birdhouse-deploy/tree/37fe32ebba87a8c6afe3aa83852bfb0dfbd1f265/birdhouse/optional-components/all-public-access as it assumes that those services are publicly accessible. 
Also, `all-public-access` would be missing the specific permission for `ncWMS2`. 
Something similar to:
https://github.com/bird-house/birdhouse-deploy/blob/37fe32ebba87a8c6afe3aa83852bfb0dfbd1f265/birdhouse/optional-components/all-public-access/config/catalog/permissions.cfg#L1-L5

For `solr`, I agree it could be dropped since unused.

fmigneault:
run  tests

fmigneault:
run tests

mishaschwartz:
The Jenkins build is failing because the thredds-docker:4.6.18 no longer exists and so canarie-api does not start (it fails trying to get the docker image release date from the tag). This is fixed in #313 but is currently blocked. See [comment here](https://github.com/bird-house/birdhouse-deploy/pull/313#issuecomment-1503802979)

mishaschwartz:
It looks like the infrastructure deployment failed because the canarie cron jobs didn't run so the component status was never checked: 

https://host-140-20.rdext.crim.ca/canarie/server/platform/stats

image

On my local test deployment, this seems to run fine though so I'm tempted to merge this in as well. Any objections @tlvu @fmigneault ?

fmigneault:
I prefer to investigate the cause before merge. Otherwise, any subsequent PR will not run any validation of E2E tests.

fmigneault:
run tests

fmigneault:
@mishaschwartz @tlvu 
Seems to be behaving correctly now. I will merge when I get approvals. 
The remaining error is related to Weaver failing to obtain a response from Hummingbird.
Not sure why it's taking so long to boot, will investigate in a separate PR since non blocking for this.

fmigneault:
@mishaschwartz @tlvu 
I've made PR #318 that seems to address the issue from https://github.com/bird-house/birdhouse-deploy/pull/284#issuecomment-1512001189 (Weaver WPS providers).
For now, it merges into this PR for quickly testing, but I can merge this one separately, and then #318 on its own.

mishaschwartz (approved):
I'm happy to approve this! I've tested this one out locally a bunch already so I'm confident in the changes.

tlvu:
FYI, trying this PR on my VM ...

fmigneault:
@tlvu 
I added the description + image to the changelog. 
Let me know once you've finished validating.

fmigneault:
run tests

fmigneault:
run tests

birdhouse/optional-components/canarie-api-full-monitoring/default.env
```diff
@@ -2,5 +2,6 @@
 
 # add any component that this component requires to run
 COMPONENT_DEPENDENCIES="
+  ./optional-components/all-public-access
```
tlvu, line 5: Should not be here.  Not all org want to make everything public so activating all-public-access component should be explicit by each org and not implicit like this.

This brings another question.  Misha all-public-access will grant public access only if the WPS component is enabled.  Does this canarie-api-full-monitoring having handling for not enabled components?

birdhouse/config/canarie-api/entrypoint
```diff
@@ -5,5 +5,12 @@
 # remove if cause problem with newer version of canarieapi
 touch /etc/cron.d/canarieapi-cron
 
+pip install requests-cache
```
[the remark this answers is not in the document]
fmigneault: It could, but it is not used anywhere in it.

birdhouse/config/catalog/config/canarie-api/canarie_api_monitoring.py.template
```diff
@@ -1,5 +1,34 @@
...
+        'supportEmail': '${SUPPORT_EMAIL}',
+        'category': 'Catalog',
+        'tags': ['Catalog', 'Data']
+    },
+    'stats': {
+        'method': '.*',
+        'route': 'http://${PAVICS_FQDN}:8086/.*'
+    },
+    'redirect': {
+        'doc': 'http://pyramid-phoenix.readthedocs.io/en/latest/index.html',
```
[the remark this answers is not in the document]
fmigneault: Oups! Will update.

birdhouse/config/canarie-api/entrypoint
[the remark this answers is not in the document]
tlvu: Or have a Pavics specific Canarie-API image?  Any other suggestions?  It just feels so wrong to install in the entrypoint like this.  This is usually a trick for short term work-around only.

birdhouse/config/finch/default.env
```diff
@@ -4,11 +4,8 @@
 # must use single quotes to avoid early expansion before overrides in env.local
 # are applied and must be added to the list of DELAYED_EVAL.
 
-export FINCH_IMAGE="birdhouse/finch:version-0.9.2"
-
-# When canarie-api is monitoring twitcher, at least one WPS service protected by twitcher
-# needs to be available. This sets this component as the service used to monitor twitcher.
-export WPS_FOR_TWITCHER_MONITORING=${WPS_FOR_TWITCHER_MONITORING:-finch}
```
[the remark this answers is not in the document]
fmigneault: It is not used anymore. 
I replaced the Twitcher monitoring by its own API. 
There is no need to ping a Twitcher-protected WPS bird anymore, as their own CanarieAPI should add the same endpoint when enabling them.

birdhouse/config/twitcher/default.env
```diff
@@ -2,6 +2,7 @@
 
 export TWITCHER_LOG_LEVEL=INFO
 export VERIFY_SSL="true"
+export TWITCHER_VERSION=0.8.0
```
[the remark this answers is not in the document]
fmigneault: Sadly no, because Twitcher uses the Magpie-build version that matches with it.
This was only for the CanarieAPI config, but that is not very robust either, as we must remember to update it along Magpie (which we won't 😅).

I'll see if I can parse the Magpie Dockerfile link directly to retrieve it instead.

birdhouse/optional-components/canarie-api-full-monitoring/default.env
fmigneault: When activating `canarie-api-full-monitoring`, minimal permission required would be `GetCapabilities` for WPS-based services, and `read-match` on the top-most resource for RESTful-based services.

Those permissions could be set directly in the `canarie-api-full-monitoring` definition, so the access requirements are activated at the same time. However, it would be important that those permissions be loaded first (load order of Magpie permission config files) to allow broader `read` (if set anywhere) to override the `read-match`. 

Regarding the handling, the configs use a loop over injected birds in:
`birdhouse/optional-components/canarie-api-full-monitoring/config/canarie-api/canarie_api_full_monitoring.py.template`
Those bird configs are injected according to enabled components. So permissions should only be applied for enabled services.

tlvu: Are you proposing for the `canarie-api-full-monitoring` to handle its own read permissions?

I prefer to leave it like this to avoid surprise for the user (user don't know where the permission are coming from and do not expect it to be from `canarie-api-full-monitoring`).

It is an optional components after all.  It should not start fiddling with Magpie permissions.

tlvu:
From docker logs:
```
canarie-api: found version [0.6.11] release time [2019-11-21T19:28:55.787132Z] from [docker: pavics/pavics-datacatalog]                                   
canarie-api: found version [0.9.2] release time [2022-07-19T21:42:25Z] from [github: bird-house/finch]
canarie-api: found version [2.22.2-kartoza-build20230226-r7-allow-change-context-root-and-fix-missing-stable-plugins-and-avoid-chown-datadir] release time [2023-03-31T02:31:49.229204Z] from [docker: pavics/geoserver]                                                                   
canarie-api: found version [1.4.0-20210506] release time [2021-05-06T22:12:51.363362Z] from [docker: pavics/jupyterhub]        
canarie-api: found version [3.31.0] release time [2023-02-07T18:18:27Z] from [github: Ouranosinc/Magpie]
canarie-api: found version [pavics-0.3.5] release time [2018-05-16T19:18:11Z] from [github: Ouranosinc/malleefowl]
canarie-api: found version [pavics-0.2.3] release time [2018-05-09T15:54:39.405697Z] from [docker: pavics/pyramid-phoenix]
canarie-api: found version [5.2.1] release time [2020-01-21T19:35:14.093976Z] from [docker: pavics/solr]
canarie-api: found version [4.6.18-unidata-2022-01] release time [2023-04-11T18:46:40.742406Z] from [docker: pavics/thredds-docker]
canarie-api: found version [0.8.0] release time [2023-02-01T15:48:01Z] from [github: bird-house/twitcher]
```

It's missing flyingpigeon, raven, hummingbird.

birdhouse/config/solr/config/canarie-api/canarie_api_monitoring.py.template
```diff
@@ -1,8 +1,39 @@
...
+        'source': 'https://github.com/Ouranosinc/PAVICS/tree/master/birdhouse/docker/solr',
+        'tryme': 'http://${PAVICS_FQDN}:8983/solr/',
+        'licence': 'https://github.com/bird-house/finch/blob/master/LICENSE.txt',
+        'provenance': 'https://ouranosinc.github.io/pavics-sdi/arch/backend.html#indexation'
+    },
+    "monitoring": {
+        'Solr': {
+            'request': {
+                # FIXME: remove port by design (https://github.com/bird-house/birdhouse-deploy/issues/222)
+                'url': 'http://${PAVICS_FQDN}:8983/solr/birdhouse/select'
```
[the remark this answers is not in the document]
mishaschwartz: I would prefer not to have it check for a specific file. Just because the file is added in bootstrap-instance-for-testsuite doesn't mean that a deployment of PAVICS will necessarily have added that data. I don't want canarie failing because we haven't set things up exactly as the test suite expects (when we're not running tests)

birdhouse/optional-components/canarie-api-full-monitoring/default.env
mishaschwartz: What if we set the required permissions for canarie-api-full-monitoring in `canarie-api-full-monitoring/config/magpie` so that it is clear that these permissions are required in order to make it work (without having to rely on all-public-access)

fmigneault:
@tlvu 
> It's missing flyingpigeon, raven, hummingbird.

We thought the same. 
flyingpigeon and hummingbird didn't even have a canarieapi config!
see [08041e6](https://github.com/bird-house/birdhouse-deploy/pull/284/commits/08041e65a514a681af3768ef4f44824fa51dd778)

fmigneault:
> Each component seems to appear in 3 locations, ex: node.Raven, raven.Raven, server.Raven. Are their monitoring url being hit just once or 3 times?

All 3 times. CanarieAPI doesn't use any logic to compare requests configurations and quick resolve equivalent ones.

birdhouse/config/canarie-api/docker_configuration.py.template
```diff
@@ -12,11 +18,97 @@ DATABASE = {
...
+    if resp.status_code != 200:
+        raise ValueError(
+            f"HTTP status from tag info: {resp.status_code} from [{url}] with detail:\n{resp.text!s}"
+        )
+
+    updated = resp.json()["last_updated"]
+    return updated
+
+
+def get_release_time_from_repo_tag(repository_type: str, repository_reference: str, tagged_version: str) -> str:
```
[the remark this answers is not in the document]
fmigneault: Since the links for the "source", "releasenotes", etc. point more often at the GitHub repository, I felt the "releaseTime" was better aligned with it if using the same reference. I'm not against adjusting them for their corresponding docker tag if those feel more adequate. 

I used Docker only for cases where there was a strong divergence in the docker tag vs source tag, such as in the case of GeoServer.

birdhouse/config/solr/config/canarie-api/canarie_api_monitoring.py.template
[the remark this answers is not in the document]
fmigneault: The URL I posted above is directly from one of the builds from this PR. See the `response.numFound == 0`. 
Most probably the crawler did not complete its operation yet, or the test data is not inserted in those builds.
Either way, this regex mismatch causes the initial monitoring to loop indefinitely and never run tests.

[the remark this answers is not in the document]
fmigneault: > It must have worked before, we have had many passing builds in the pipeline.

Maybe before monitoring was introduced in the CI to check that the instance was ready, so it silently failed?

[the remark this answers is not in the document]
fmigneault: I agree with @mishaschwartz 
Instances are free to extend further the number of items monitored for more reporting/robustness, but we shouldn't include a test requirement on the core configuration of the service. It could be a `optional-component/testdata-canarie-monitoring`.

birdhouse/optional-components/canarie-api-full-monitoring/default.env
fmigneault: > Are you proposing for the canarie-api-full-monitoring to handle its own read permissions?

I think that would be the most logical approach. Avoid extra steps from users.
The `canarie-api-full-monitoring` already states this, so there shouldn't be a surprise: 

https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse/optional-components#monitor-all-components-in-canarie-node-both-public-and-internal-url
> This assume all the WPS services are public. If not the case, make a copy of this config and adjust accordingly.

I think users would be even more surprised to see errors of something they just enabled (it should work out-of-the-box), and asking them to correctly understand the Magpie permission configuration to correct accesses, which require understanding each service type/permission scheme, makes it much more complicated for them.

tlvu: I think I need to explain why that `canarie-api-full-monitoring` component came to be.

We all agree that ensuring the public access works is a good idea.  So why this component is optional and there is a notice saying "If not the case, make a copy of this config and adjust accordingly".  Probably that notice is not descriptive enough, my bad.

It was chosen to put all the WPS birds behind Twitcher for access control.  Furthermore, it is a very fine grained access control.  Each specific WPS action (GetCapabilities, DescribeProcess, ...) can be toggle independently for each WPS service.

I was not there at the time it was designed that way, but I assume we want a strict control over which WPS service and capabilities we want to allow.  So maybe an org can deploy all the birds but only want to allow complete public access for some only, but not all of them. 

So that's why this very useful `canarie-api-full-monitoring` component is optional and not integrated into to main `canarie-api`.  The notice "If not the case, make a copy of this config and adjust accordingly" means make a copy of that config and remove the birds you do not want to make public.

At Ouranos, all the birds are public.  But if we take PCIC for example, I know they do not use any of the birds.  They have their own birds they are plugin in via external repos.  So they can make a copy of this config, re-adjust for their birds, and deploy together with their birds via their external repo.  They do not need to activate this `canarie-api-full-monitoring` as-is.

Furthermore, Magpie permissions once given via a component, is not revoked by simply disabling that component.  You have to go to Magpie UI to manually remove the access.  So if someone want just to try out this component, then decide they do not want it, all their birds will still be wide open to the internet, even after they disabled the component.

The `all-public-access` component https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse/optional-components#give-public-access-to-all-resources-for-testing-purposes state that very clearly "Once enabled, if you need to revert the change, you have to do it manually by logging into Magpie. Just disabling this component will not revert the change."

So putting our sysadmin hat, we really do not want a silent surprise that all the birds are public on the internet without us explicitly knowingly allow it.

So that's why I much prefer this `canarie-api-full-monitoring` component stay a bit "dumb" and relies on either manual explicit intervention on Magpie WebUI or explicit activation of `all-public-access` if on a staging or test system.  Either ways, the sysadmin knowingly did it.  It's not a surprise.

I understand this is not the "it just works out of the box" experience we are looking for, but otherwise it would be a false sense of security to put Twitcher there but then activate public access everywhere and the sysadmin has to manually remember to go disable the accesses they do not want.

So instead of this notice for `canarie-api-full-monitoring` 
  "This assume all the WPS services are public. If not the case, make a copy of this config and adjust accordingly", 
would this be better 
  "This assume all the WPS services are public. If not the case, because not all WPS services are required to be publicly accessible or additional custom WPS services are enabled, you can make a copy of this config, or even this entire component, and add or remove WPS services according to the services that are public only."?

birdhouse/config/solr/config/canarie-api/canarie_api_monitoring.py.template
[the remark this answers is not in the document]
tlvu: So here it's true it breaks the "it just works out of the box" experience when a system is not properly configured for the testsuite.  Someone just want to try out PAVICS probably do not want to immediately configure for the testsuite.

So yes leave out the data query for Solr in the canarie-api.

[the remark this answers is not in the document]
tlvu: > Maybe before monitoring was introduced in the CI to check that the instance was ready, so it silently failed?

Since day one of the pipeline, Mathieu used that canarie-api monitoring link so make sure the system is properly provisioned before launching Jenkins testsuite on it (there is a notebook that query Solr as well).  So this monitoring is there since day one, it did not fail silently before monitoring is added.

I think this provisioning flakiness might be related to the same root cause that randomly sometime the stack do not boot up.

birdhouse/config/canarie-api/docker_configuration.py.template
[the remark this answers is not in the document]
tlvu: Ah ok, I was just curious.  Now that you already invested into github releasetime, and dealt with github rate limiting, let's keep it.

tlvu:
> All 3 times. CanarieAPI doesn't use any logic to compare requests configurations and quick resolve equivalent ones.

Ouch, that's more un-necessary hit for nothing.  Not helping that we are currently have an intermittent access problem and performance problem in https://github.com/bird-house/twitcher/issues/123.

If there's no easy fix then let's keep this and see how it goes.

birdhouse/config/canarie-api/docker_configuration.py.template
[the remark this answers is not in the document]
mishaschwartz: Maybe this is a weird question but is there a reason why we need to report these release times by querying the github or dockerhub urls? If we give the user a link to the source and a version number they should be able to go to that website and see the release time for themselves.

I guess that including the release time was a requirement of the canarie folks? Other than that, what is the use-case for displaying the release time of specific components through canarie-api?

fmigneault:
@tlvu 
> Ouch, that's more un-necessary hit for nothing. Not helping that we are currently have an intermittent access problem and performance problem in https://github.com/bird-house/twitcher/issues/123.
> 
> If there's no easy fix then let's keep this and see how it goes

I agree, it could be improved, and it's not particularly hard to do it. It's just not implemented yet.
You can create an issue in CanarieAPI to keep track of this potential improvement. 
Still, it is not "that much" of a performance hit. 
The stress-test does way more requests/sec than the few requests this does every minute.

birdhouse/config/canarie-api/docker_configuration.py.template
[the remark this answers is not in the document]
fmigneault: The fields that are returned were specifically required by Canarie with a specific format.
It is also good to have the release time be displayed at a glance for all components, rather than manually checking them one by one digging down Github links.
Updating them with this PR, I've also seen that almost every component was not reporting the correct version number and release time of what was actually deployed. This approach with requesting the tag release times removes the burden of remembering to update this information each time we update a service.

birdhouse/config/solr/config/canarie-api/canarie_api_monitoring.py.template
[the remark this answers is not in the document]
fmigneault: Until about 1-2 months ago, it was only pinging the `/canarie/` endpoint and checking that there is "Ouranos" somewhere in the output. It did not guarantee that any service in the stack was ready for tests. The advanced monitoring of all services as "up" to resume tests was added by myself very recently. If they were ready, it was just coincidence with other docker services taking longer to boot, and proxy depending on them to restart.

The stack not booting up seems to be related to available docker-pool data usage. Whenever it's too high, they just start failing until we manually flush some items.

birdhouse/optional-components/canarie-api-full-monitoring/default.env
tlvu: Please remove this dependency on `all-public-access` component because this is too sneaky on the users.  They do not realized they are opening up to the wide internet and the only way out is manual edit in Magpie UI.

fmigneault:
@tlvu 
[f4da986](https://github.com/bird-house/birdhouse-deploy/pull/284/commits/f4da986e705f4c7670ab6a3a288c8f41d4cb8713) resolves the hummingbird tag 

Not sure about the `lastStatusUpdate` though. Any errors in logs regarding a failing operation?

fmigneault:
run tests

tlvu:
> Not sure about the `lastStatusUpdate` though. Any errors in logs regarding a failing operation?

Nothing in the logs, I think bug in the DB update somewhere, not in the monitoring config in this birdhouse-deploy repo.

tlvu:
> ```
> canarie-api: failed retrieving version [0.5_dev] release time from [github: bird-house/hummingbird]                                                                                                          
> Traceback (most recent call last):                                                                                                                                                                           
>   File "/config/docker_configuration.py", line 83, in get_release_time_from_repo_tag                                                                                                                         
>     release = _get_release_time_from_github_tag(repository_reference, tagged_version)                                                                                                                        
>   File "/config/docker_configuration.py", line 54, in _get_release_time_from_github_tag                                                                                                                      
>     raise ValueError(f"No matching version in tagged releases under [{url}].")                                                                                                                               
> ValueError: No matching version in tagged releases under [https://api.github.com/repos/bird-house/hummingbird/tags]. 
> ```

Confirm fixed

```
canarie-api: found version [0.5_dev] release time [2020-01-21T20:20:21.953117Z] from [docker: pavics/hummingbird] 
```

birdhouse/optional-components/canarie-api-full-monitoring/default.env
fmigneault: >Magpie permissions once given via a component, is not revoked by simply disabling that component. You have to go to Magpie UI to manually remove the access. So if someone want just to try out this component, then decide they do not want it, all their birds will still be wide open to the internet, even after they disabled the component.

This is true. 
The right way to handle this automatically would be to have a component that lists all known permissions for "opening" services publicly, and use the `remove` action instead. 

I agree that `all-public-access` is too open to be used as dependency, this is why `canarie-api-full-monitoring` should set only the permissions it needs (basically just the `read` or `GetCapabilities` of the top most monitored resource.

fmigneault: @mishaschwartz 
> - create a dedicated user for canarie-api in magpie
> - allow that user limited access permissions to the routes we want to monitor
> - canarie-api then can access these routes as that user
>
> I know you mentioned in the past that you don't love the idea of having internal users registered in magpie [...]

That would work. I don't recall mentioning that internal users are bad, as I also think the approach you propose is an ideal way to handle this issue. Some considerations though. 
All requests in CanarieAPI will have to be updated to use `requests-magpie` auth hook to perform the login/cookie session handling if credentials are provided. Also, `proxy` (or eventually `canarieapi` if separated out) will need to depend on Magpie to ensure it is ready to register users and perform the login.

fmigneault: @tlvu @mishaschwartz 
For now, I reverted the `all-public-access` without any extra permission handling. 
On the CI side, `all-public-access` is added explicitly, so everything should still work for PR tests.
Users will have to add custom permissions in the meantime if they want to provide access to CanarieAPI.

birdhouse/config/canarie-api/entrypoint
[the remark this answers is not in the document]
fmigneault: Seems overkill to generate a separate image just for that.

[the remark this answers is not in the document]
tlvu: Not overkill at all.  Reproductibility is the reason docker image was invented.  It gives a snapshot of all the binaires required to run a product.

We have a custom JupyterHub image simply to have the magpie_authenticator and the dockerspawner pre-installed, see https://github.com/Ouranosinc/jupyterhub/blob/master/Dockerfile

My custom Geoserver build is to have the 2 missing plugin pre-installed into the image instead of downloading them on the fly in the entrypoint, see https://github.com/tlvu/docker-geoserver/blob/2e5dbb99effa75abe818cdf551532f2b1de7739c/Dockerfile.custom

tlvu (approved):
Good enough for now.  Please open an issue to not forget to address the issue about performing an install in the entrypoint of canarie-api which defeat the reproducibility feature of a docker image.

tlvu:
Forget to say, this one needs a minor bump instead of a patch bump because of all the image format change.

birdhouse/config/canarie-api/entrypoint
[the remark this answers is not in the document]
fmigneault: https://github.com/bird-house/birdhouse-deploy/issues/322

PR #288, opened 2023-02-07
Fix `DELAYED_EVAL` not working for other than `pavics compose.sh`

tlvu:
## Fixes

There are other scripts sourcing `default.env` and `env.local` and all those scripts have to expand the vars in `DELAYED_EVAL` list to have their actual values.

Only scripts using the 3 variables in `DELAYED_EVAL` list are broken.

`DELAYED_EVAL` was previously introduced in PR https://github.com/bird-house/birdhouse-deploy/pull/272.

### Sample errors

`fix-geoserver-data-dir-perm` (called at the end of `pavics-compose.sh`):
```
fix GeoServer data dir permission on first run only, when data dir do not exist yet.
+ DATA_DIR='${DATA_PERSIST_ROOT}/geoserver'
+ '[' -n  ]
+ docker run --rm --name fix-geoserver-data-dir-perm --volume '${DATA_PERSIST_ROOT}/geoserver:/datadir' --env FIRST_RUN_ONLY bash:5.1.4 bash -xc 'if [ -z "$FIRST_RUN_ONLY" -o ! -f /datadir/global.xml ]; \
    then chown -R 1000:10001 /datadir; else echo "No execute."; fi'
docker: Error response from daemon: create ${DATA_PERSIST_ROOT}/geoserver: "${DATA_PERSIST_ROOT}/geoserver" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
```
`trigger-deploy-notebook` (broke notebook deploy job):
```
+ TMP_SCRIPT=/tmp/notebookdeploy/notebookdeploy.XXXXXXIfafFK/deploy-notebook
+ cat
+ chmod a+x /tmp/notebookdeploy/notebookdeploy.XXXXXXIfafFK/deploy-notebook
+ docker run --rm --name deploy_tutorial_notebooks -u root -v /tmp/notebookdeploy/notebookdeploy.XXXXXXIfafFK/deploy-notebook:/deploy-notebook:ro -v /tmp/notebookdeploy/notebookdeploy.XXXXXXIfafFK/tutorial-notebooks:/tutorial-notebooks:ro -v '${DATA_PERSIST_ROOT}/jupyterhub_user_data:/notebook_dir:rw' --entrypoint /deploy-notebook bash:5.1.4
docker: Error response from daemon: create ${DATA_PERSIST_ROOT}/jupyterhub_user_data: "${DATA_PERSIST_ROOT}/jupyterhub_user_data" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
```
### Explanation of the fix

All scripts have to remember to call function `process_delayed_eval` in order to obtain the real value of each vars in `DELAYED_EVAL` list.

Centralized all logic about reading configs (config files reading order, remember to call `process_delayed_eval`) to avoid mistake and to ease updating logic in the future.  Too many scripts were reading the configs  themselves and some are not doing it properly, ex: forgot to hide password when reading `env.local`.

### All scripts should do this going forward

```sh
# Set variable COMPOSE_DIR to the dir containing pavics-compose.sh and docker-compose.yml.

# Source the script providing function read_configs.
# read_configs uses COMPOSE_DIR to find default.env and env.local.
. $COMPOSE_DIR/read-configs.include.sh

# Call function read_configs to read the various config files in the appropriate order and process delayed eval vars properly.
read_configs
```


birdhouse/read-configs.include.sh
```diff
@@ -0,0 +1,131 @@
...
+        elif [ -e "../../birdhouse-deploy/birdhouse/pavics-compose.sh" ]; then
+            # Case of subdir of sibling checkout at same level as birdhouse-deploy.
+            COMPOSE_DIR="`realpath "../../birdhouse-deploy/birdhouse"`"
+        elif [ -e "../../../birdhouse-deploy/birdhouse/pavics-compose.sh" ]; then
+            # Case of sub-subdir of sibling checkout at same level as birdhouse-deploy.
+            COMPOSE_DIR="`realpath "../../../birdhouse-deploy/birdhouse"`"
+        elif [ -e "./birdhouse/pavics-compose.sh" ]; then
+            # Child dir is COMPOSE_DIR
+            COMPOSE_DIR="`realpath birdhouse`"
+        fi
```
fmigneault, line 37: For readability, a for loop seems a better to try/break various options.

Also, is there a technical (older shell support?) reason to use ``` `` ``` instead of `$()`? Using a shell-checker on this file flags those calls. `$()` is better as it allows nesting calls and are more readable.

birdhouse/read-configs.include.sh
```diff
@@ -0,0 +1,131 @@
...
+        export COMPOSE_DIR
```
fmigneault, line 38: Do we want to add `echo ${COMPOSE_DIR}`? 
This is a good way to pass the function "output" (`DIR=$(discover_compose_dir)`) without passing by an external variable that could be modified at any time.

birdhouse/read-configs.include.sh
```diff
@@ -0,0 +1,131 @@
...
+}
+
+
+read_components_default_env() {
+    # EXTRA_CONF_DIRS normally set by env.local so should read_env_local() first.
+
+    # EXTRA_CONF_DIRS relative paths are relative to COMPOSE_DIR.
+    if [ -e "$COMPOSE_DIR" ]; then
+        cd "$COMPOSE_DIR"
+    fi
```
fmigneault, line 75: Use `-d` to avoid error on badly overridden value with file or empty.

birdhouse/deployment/certbotwrapper
```diff
@@ -50,21 +50,10 @@ THIS_FILE="`realpath "$0"`"
...
-if [ -e "$ENV_LOCAL_FILE" ]; then
-    # Override default values
-    . "$ENV_LOCAL_FILE"
-else
-    echo "WARNING: '$ENV_LOCAL_FILE' not found."
-fi
-
-set -x  # resume logging/tracing
+# Get PAVICS_FQDN_PUBLIC, PAVICS_FQDN, SUPPORT_EMAIL, SSL_CERTIFICATE.
+read_configs
```
fmigneault, line 56: In all files where this is done, consider removing the blank lines between the statements, as they are strongly connected. Having the call right after the include line highlights that it is defined in that file.

fmigneault (changes requested):
Very nice to see all this logic finally refactored together.

birdhouse/pavics-compose.sh
```diff
@@ -99,37 +99,11 @@ cd $(dirname $(readlink -f $0 || realpath $0))
...
 COMPOSE_DIR="`pwd`"
 
-. ./default.env
+. "$COMPOSE_DIR/read-configs.include.sh"
 
-# we source local configs, if present
-# we don't use usual .env filename, because docker-compose uses it
-[ -f env.local ] && . ./env.local
-. ./scripts/get-components-json.include.sh
+read-configs
```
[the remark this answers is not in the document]
fmigneault: This must be related to the previous ```COMPOSE_DIR="`pwd`"```. If I recall correctly, the CI pipeline calls `pavics-compose` that is an alias to `/home/ubuntu/birdhouse-deploy/birdhouse/pavics-compose.sh` defined in the user install directories. Therefore, `pwd` is something like `/usr/bin` or similar where the rest of the project cannot be found.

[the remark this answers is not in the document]
tlvu: Ah geez typo!  Should be `read_configs` (underscore and not dash) !

birdhouse/deployment/certbotwrapper
tlvu: The empty line between the include and the call is because I always comment which vars I try to get with the call so it would look weird to not have an empty line.  Ex:

```
. read-configs.include.sh
# get VAR1, VAR2, VAR3
read_configs
```

I usually put an empty line before a comment.

Also the file is `read-config.include.sh` and the function is `read_configs` so I think the connection is fairly obvious even with a empty line in between.

Still I did remove the empty line when no comment in between, see 2e16242503e2d3471525016cd95a97d011706691

birdhouse/read-configs.include.sh
tlvu: A for-loop to try/break various options, could you give an example?  I am not sure how to write a for-loop in this case.

And yes I used backticks instead of `$()` for portability reason.  I try to stay within the limited features of the original Bourne shell (/bin/sh) and not Bourne Again shell (/bin/bash).

Now that you asked, I tried to search for differences between sh and bash and I could not find anything definitive which syntax is which shell.  So to play safe, I'd rather stay with backticks for now if you don't mind.

tlvu:
@eyvorchuk FYI, you are tagged in this PR since you might write custom scripts that might need to read the config files.  This PR provided a unified way to read config files.

birdhouse/read-configs.include.sh
fmigneault: Something similar to this:
```shell
TESTS="
  test1
  ../test2
  ../../test3
  ./test4
"
for path in ${TESTS}; do
  if [ -d "${path}" ]; then
    echo "Found ${path}"
    COMPOSE_DIR="${path}"
    break
  fi
done

```

fmigneault: `$()` are ok in `sh`
I'm using it (and have to because it's nested) here: 
https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/scripts/get-components-json.include.sh#L57-L63

fmigneault (commented):
Changes are good and seem to work. 
Only items in https://github.com/bird-house/birdhouse-deploy/pull/288#discussion_r1100436265 to consider, but ok to integrate.

birdhouse/read-configs.include.sh
tlvu: I switched to `$()` except in `process_delayed_eval` since I could not get the `eval` to work properly.  Looks like in some corner cases backticks still have its use.

As for the for-loop try/catch, it's not applicable to this case because there is a different mapping for each case:
```
./pavics-compose.sh -> realpath .
../pavics-compose.sh -> realpath ..
./birdhouse/pavics-compose.sh -> realpath birdhouse
../birdhouse-deploy/birdhouse/pavics-compose.sh -> realpath ../birdhouse-deploy/birdhouse
...
```

tlvu:
> Changes are good and seem to work. Only items in [#288 (comment)](https://github.com/bird-house/birdhouse-deploy/pull/288#discussion_r1100436265) to consider, but ok to integrate.

I still need an approval to merge this.

tlvu:
https://daccs-jenkins.crim.ca/job/DACCS-iac-birdhouse/1225/console
```
[CHECK] Waiting for canarie-api [https://host-140-36.rdext.crim.ca/canarie/node/service/status]...
...................................................................................................................................................................................................................................
[WARNING] Timeout of [1200] seconds reached. Exiting, but leaving the instance running.
```

I think it's probably not my change from using backticks to `$()` causing this, `pavics-compose.sh up -d` works for me.  Can someone from CRIM confirm?

tlvu:
@fmigneault Do you see anything blocking? I'd like to merge this PR soon because it is actually fixing a regression.

mishaschwartz:
Have we considered avoiding the need to avoid a `DELAYED_EVAL` list entirely by using parameter expansions?
So instead of defining variables in the `default.env` files as (for example):

```sh
export MAGPIE_DB_NAME="magpiedb"
export MAGPIE_PERSIST_DIR='${DATA_PERSIST_ROOT}/magpie_persist'

# and then add MAGPIE_PERSIST_DIR to the DELAYED_EVAL list
```

we do

```sh
export MAGPIE_DB_NAME=${MAGPIE_DB_NAME:-"magpiedb"}
export MAGPIE_PERSIST_DIR="${DATA_PERSIST_ROOT}/magpie_persist"
```

or if we want to exit early if required variables are not set:

```sh
export MAGPIE_PERSIST_DIR="${DATA_PERSIST_ROOT:?DATA_PERSIST_ROOT must be set}/magpie_persist" || exit 1
```

and then if we source the `env.local` file first the `default.env` files will not override those and everything will be set in order.

This will make the `default.env` files a bit more verbose but will reduce the complexity of having to keep track of `DELAYED_EVAL` variables.

tlvu:
@mishaschwartz The delayed eval trick was needed for `env.local` to override variables previously defined in `default.env` when these variables are used in other variables.  Ex:

`default.env`:
```
DATA_PERSIST_ROOT=/data
MAGPIE_PERSIST_DIR=$DATA_PERSIST_ROOT/magpie_persist  # /data/magpie_persist
```
`env.local`:
```
DATA_PERSIST_ROOT=/my/custom/dataroot
```
Because `default.env` is read before `env.local`, `MAGPIE_PERSIST_DIR` is already set at the moment we read `default.env` so it is still `/data/magpie_persist` after `env.local` is read.

With delayed eval, `MAGPIE_PERSIST_DIR` is expanded only after `env.local` is read, so we have the good value `/my/custom/dataroot/magpie_persist`.

Even without this bug caused by the introduction of delayed eval, it is still a good idea to centralize the configs reading logic to avoid errors when each script does it manually, ex: forgot to hide password when reading `env.local`.

mishaschwartz:
@tlvu I completely agree that centralizing the config reading functionality is a great idea. 

I'm sorry I don't think I explained the idea very well. What I was trying to say is that if we source env.local first then we can use parameter expansion in the default.env files (which are sourced afterwards):

```sh
# env.local
DATA_PERSIST_ROOT=/my/custom/dataroot

# default.env
DATA_PERSIST_ROOT="${DATA_PERSIST_ROOT:-/data}" # /my/custom/dataroot
MAGPIE_PERSIST_DIR="${MAGPIE_PERSIST_DIR:-${DATA_PERSIST_ROOT}/magpie_persist}" # /my/custom/dataroot/magpie_persist
```

I'm not suggesting we change this for this PR and the current method of delaying evaluation definitely works. I'm just suggesting an alternative to think about so that users don't have to worry about knowing whether to single quote or double quote variable values in `env.local`.

fmigneault:
One issue I can see with what @mishaschwartz proposes is that multiple configurations of existing instances assume that variables defined in `default.env` are available before `env.local` is source. So, for example, sourcing `env.local` before `default.env` would make instances that use: 

```env
# env.local
MAGPIE_PERSIST_DIR="${DATA_PERSIST_ROOT}/magpie"  
``` 
Resolve to `/magpie` since `DATA_PERSIST_ROOT` is not yet defined in that case. 
For backward compatibility, we cannot change the sourcing order of those files.

fmigneault (approved):

tlvu:
> For backward compatibility, we cannot change the sourcing order of those files.

It's not just `env.local` that can use variables defined in the main `default.env`, each component `default.env` can also use variables defined in the main `default.env` as well so can also result in wrong value if we inverse the reading order.

However it's good to have thinking outside the box like this.  Had this trick worked for all cases, it would make config reading much simpler.  This trick would work if vars can only reference other vars defined in the same config file, but then there is no way to enforce this.

PR #297, opened 2023-02-24
Add STAC catalog

matprov:
## Overview

- Adds STAC to the stack (optional) when ``./components/stac`` is added to ``EXTRA_CONF_DIRS``.

## Changes

  * Service ``stac`` (API) gets added with endpoints ``/twitcher/ows/proxy/stac`` and ``/stac``.

  * STAC catalog can be explored via the ``stac-browser`` component, available under ``/stac-browser``.

  * Image ``crim-ca/stac-app`` is a STAC implementation based on ``stac-utils/stac-fastapi``.

  * Image ``crim-ca/stac-browser`` is a fork of ``radiantearth/stac-browser``.

  * Adds `Magpie` permissions and service for `stac` endpoints.

  * Uses [stac-populator](https://github.com/crim-ca/stac-populator) to populate STAC catalog with sample collection 
  items via [CEDA STAC Generator](https://github.com/cedadev/stac-generator), employed in sample 
  [CMIP Dataset Ingestion Workflows](https://github.com/cedadev/stac-generator-example/tree/master/conf).

## Demo Instance

STAC API : https://stac-dev.crim.ca/stac/
STAC Browser : https://stac-dev.crim.ca/stac-browser/

Note that by default STAC API will return 10 items to reduce payload size. It is however possible to change this limitation by adding `?limit=200` to the URL in order to query 200 items. In the response payload you'll have a link referring to the `next` items, adding a token to the query params in order for STAC API to return next results.

**Sample STAC API collection query using a CLI**

Remove the `-c` flag for global query across any collection.

```
pip install pystac-client
stac-client search https://stac-dev.crim.ca/stac -c c604ffb6d610adbb9a6b4787db7b8fd7 --query "variable_id=txgt_32" "scenario=ssp585"
```

**Sample STAC API global query using CQL via cURL call**

Note that the operators are describe here : https://portal.ogc.org/files/96288

```
curl --location --globoff 'https://stac-dev.crim.ca/stac/search' \
--header 'Content-Type: application/json' \
--data '{
   "filter":{
      "and":[
         {
            "eq":[
               {
                  "property":"freq"
               },
               "MS"
            ]
         },
         {
            "like":[
               {
                  "property":"variable_id"
               },
               "tr_%"
            ]
         }
      ],
      "intersects":[
         {
            "property":"geometry"
         },
         {
            "type":"Polygon",
            "coordinates":[
               [
                  [
                     -140.99778,
                     41.6751050889
                  ],
                  [
                     -140.99778,
                     83.23324
                  ],
                  [
                     -52.6480987209,
                     83.23324
                  ],
                  [
                     -52.6480987209,
                     41.6751050889
                  ],
                  [
                     -140.99778,
                     41.6751050889
                  ]
               ]
            ]
         }
      ],
      "anyinteracts":[
         {
            "property":"datetime"
         },
         [
            "2010-05-03T13:21:30.040Z",
            "2022-05-03T13:21:30.040Z"
         ]
      ]
   }
}'
```

**Get the queryables of the CMIP6 collection, statically created at collection creation**

https://stac-dev.crim.ca/stac/collections/c604ffb6d610adbb9a6b4787db7b8fd7

**Get the queryables of the CMIP6 collection, dynamically created at query time**

https://stac-dev.crim.ca/stac/collections/c604ffb6d610adbb9a6b4787db7b8fd7/queryables

**Get the queryables of the union of the CMIP5 and CMIP6 collections, dynamically created at query time**

https://stac-dev.crim.ca/stac/queryables?collections=0798aa197d54eb4332767a5a4077fb0f,c604ffb6d610adbb9a6b4787db7b8fd7

daccs_configs_branch: stac_populator
daccs_skip_ci: true

fyi @huard @mishaschwartz

### DIFF
--- a/birdhouse/components/README.rst
+++ b/birdhouse/components/README.rst
@@ -286,7 +286,7 @@ How to Enable the Component
 
 - Edit ``env.local`` (a copy of `env.local.example`_)
 
-  - Add ``"./components/weaver"`` to ``EXTRA_CONF_DIRS``.
+  - Add ``./components/weaver`` to ``EXTRA_CONF_DIRS``.
 
   - Component ``birdhouse/optional-components/all-public-access`` should also be enabled to ensure that `Weaver`_
     can request ``GetCapabilities`` of every WPS provider to be registered. Publicly inaccessible services will not
@@ -364,7 +364,7 @@ How to Enable the Component
 ---------------------------
 
 - Edit ``env.local`` (a copy of `env.local.example`_)
-- Add ``"./components/cowbird"`` to ``EXTRA_CONF_DIRS``.
+- Add ``./components/cowbird`` to ``EXTRA_CONF_DIRS``.
 
 Customizing the Component
 -------------------------
@@ -414,4 +414,4 @@ How to Enable the Component
 ---------------------------
 
 - Edit ``env.local`` (a copy of `env.local.example`_)
-- Add ``"./optional-components/stac"`` to ``EXTRA_CONF_DIRS``.
+- Add ``./optional-components/stac`` to ``EXTRA_CONF_DIRS``.
--- a/birdhouse/optional-components/README.rst
+++ b/birdhouse/optional-components/README.rst
@@ -307,7 +307,7 @@ except time required to feed the catalog.
 To enable this optional-component:
 
 - Edit ``env.local`` (a copy of `env.local.example`_)
-- Add ``"./optional-components/stac-populator"`` to ``EXTRA_CONF_DIRS``.
+- Add ``./optional-components/stac-populator`` to ``EXTRA_CONF_DIRS``.
 
 
 Allow public access to STAC catalog
@@ -318,4 +318,4 @@ STAC Public Access allows STAC catalog to be accessed by anyone, without authent
 To enable this optional-component:
 
 - Edit ``env.local`` (a copy of `env.local.example`_)
-- Add ``"./optional-components/stac-public-access"`` to ``EXTRA_CONF_DIRS``.
+- Add ``./optional-components/stac-public-access`` to ``EXTRA_CONF_DIRS``.


birdhouse/components/stac/populator/populate.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
fmigneault, line 1: I'm not quite sure if there is any benefice of using this separate bash script over `argparse` and `subprocess`/`multiprocess`/`threading` calls in the same Python file that could combine `stac_generator.scripts.stac_generator` and `collection_processorCollectionProcessor`.

It seems working all under Python would reduce the number of input validations, would simplify chaining them and makes it easier to extend with more arguments or functionalities.

birdhouse/components/stac/populator/populate.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# replace STAC host name of asset generators
+sed -i "/host:/c\      host: $STAC_HOST" conf/thredds-cmip6-asset-generator.yaml
+sed -i "/host:/c\      host: $STAC_HOST" conf/thredds-cmip5-asset-generator.yaml
+
+# add items
+timeout $STAC_ASSET_GENERATOR_TIMEOUT bash -c "python3 -m stac_generator.scripts.stac_generator conf/thredds-cmip6-asset-generator.yaml" || true &
+PID_CMIP6=$!
+timeout $STAC_ASSET_GENERATOR_TIMEOUT bash -c "python3 -m stac_generator.scripts.stac_generator conf/thredds-cmip5-asset-generator.yaml" || true &
+PID_CMIP5=$!
```
fmigneault, line 34: Can't scale with more collections. A basic loop with a YAML files within a separate directory would be an easy fix.

fmigneault (changes requested):
The main STAC components look fine, but the populator needs refactorings.

Nice work for making it work in the stack! 🎉

birdhouse/components/stac/populator/collection_processor.py
```diff
@@ -0,0 +1,163 @@
+# encoding: utf-8
+"""
+STAC API collection processor
```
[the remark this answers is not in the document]
fmigneault: image

The `optional-components/stac-populator` would insert the docker-compose and config with a pre-built docker from a separate repository.

birdhouse/components/stac/populator/collection_processor.py
```diff
@@ -0,0 +1,163 @@
...
+-------------
+
+Create a STAC collection or update its summaries based on its queryables.
+"""
+__author__ = "Mathieu Provencher"
+__date__ = "20 Apr 2022"
+__copyright__ = "Copyright 2022 Computer Research Institute of Montreal"
+__license__ = "BSD - see LICENSE file in top-level package directory"
```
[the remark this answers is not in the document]
fmigneault: To align with our other licenses for open source items, you can use the [MIT](https://opensource.org/licenses/MIT) licence for this tool.

mishaschwartz:
@matprov @fmigneault @tlvu 

I'd like to get this component into the master branch since I'm excited to start using it! I understand that this is momentarily stalled as @matprov is working on other things. 
@dchandan is able to work on this to make the final few changes and factor out the stac-populator into its own repo. Do you mind giving him write permissions for this repo so that he can work on this in the meantime?

fmigneault:
@mishaschwartz @dchandan

I'd also like to have this integrated as it would be useful for other projects as well on our side.
@matprov is not planned to work on this until at least May, so if @dchandan can dive into it, go for it! 
I remain available to answer to questions as needed.

dchandan:
@fmigneault sounds good! It's great to hear that you will be available for discussions. I already have some questions; I'll be in touch to discuss them at some point next week.

CHANGES.md
```diff
@@ -14,7 +14,24 @@
...
+  [STAC Component](https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/components/README.rst#STAC)
+  Following happens when enabled:
+    
+  * Service ``stac`` (API) gets added with endpoints ``/twitcher/ows/proxy/stac`` and ``/stac``.
+    
+  * STAC catalog can be explored via the ``stac-browser`` component, available under ``/stac-browser``.
+      
+  * Image ``crim-ca/stac-app`` is a STAC implementation based on ``stac-utils/stac-fastapi``.
+
+  * Image ``crim-ca/stac-browser`` is a fork of ``radiantearth/stac-browser``.
```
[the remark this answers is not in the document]
fmigneault: Is there actually any reason to use forks? 
Both crim-ca repositories indicate they are behind by many commit, and nothing ahead. 
If there is no real reason, and we can avoid maintaining an already diverging fork, I would prefer we refer to the originals.
If there are any relevant changes, consider contributing them back to the originals.

CHANGES.md
```diff
@@ -14,7 +14,24 @@
...
+- Adds [STAC](https://github.com/crim-ca/stac-app) to the stack (optional) when ``./components/stac`` 
+  is added to ``EXTRA_CONF_DIRS``. For more details, refer to 
+  [STAC Component](https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/components/README.rst#STAC)
+  Following happens when enabled:
+    
+  * Service ``stac`` (API) gets added with endpoints ``/twitcher/ows/proxy/stac`` and ``/stac``.
+    
+  * STAC catalog can be explored via the ``stac-browser`` component, available under ``/stac-browser``.
+      
+  * Image ``crim-ca/stac-app`` is a STAC implementation based on ``stac-utils/stac-fastapi``.
```
fmigneault, line 26: Doesn't seem to be needed anymore. 
The added queryables extension seems to already exist: 
https://github.com/radiantearth/stac-api-spec/tree/master/fragments/filter#queryables

fmigneault (changes requested):

birdhouse/components/stac/conf.extra-service.d/stac.conf.template
```diff
@@ -0,0 +1,26 @@
...
+        proxy_pass https://${PAVICS_FQDN_PUBLIC}${TWITCHER_PROTECTED_PATH}/stac/stac;
+        proxy_set_header Host $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Host $host:$server_port;
+        proxy_buffering off;
+        include /etc/nginx/conf.d/cors.include;
+    }
+
+    location /stac-browser/ {
+        proxy_pass http://stac-browser:8080/;
```
mishaschwartz, line 16: Is there a reason not to put this behind twitcher as welL

birdhouse/components/stac/config/magpie/config.yml.template
```diff
@@ -0,0 +1,20 @@
...
+    public: true
+    c4i: false
+    type: api
+    sync_type: api
+
+permissions:
+  # Administrator permissions over whole service
+  - service: stac
+    permission: read
+    group: anonymous
```
mishaschwartz, line 15: the comment above says "Administrator permissions over whole service" but this is giving anonymous permissions. It seems like a bad idea to give anonymous write permissions here.

birdhouse/components/stac/docker-compose-extra.yml
```diff
@@ -0,0 +1,71 @@
...
+    volumes:
+      - stac-db:/var/lib/postgresql/data
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+
+  # extend proxy with endpoint and config for STAC API access
+  proxy:
```
mishaschwartz, line 54: please move these definitions to files in `birdhouse/config/stac/config/proxy` 

(see other components in `birdhouse/config` for examples)

birdhouse/components/stac/docker-compose-extra.yml
```diff
@@ -0,0 +1,71 @@
...
+    volumes:
+      - ./components/stac/conf.extra-service.d:/etc/nginx/conf.extra-service.d/stac:ro
+    links:
```
mishaschwartz, line 57: I'm pretty sure this isn't necessary anymore. All docker containers on the same network can reach each other using the service name by default without the need for an explicit `links` directive.

`links` is mostly used to define aliases these days: https://docs.docker.com/compose/networking/#link-containers

birdhouse/config/twitcher/config/proxy/conf.extra-service.d/twitcher.conf.template
```diff
@@ -4,4 +4,5 @@
         proxy_set_header X-Forwarded-Proto $real_scheme;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Host $host:$server_port;
+        proxy_set_header Forwarded "proto=https;host=${PAVICS_FQDN}";    # Helps the STAC component to craft URLs containing the full PAVICS_FQDN
```
mishaschwartz, line 7: If this is specific to the stac component, why not add the header in the `location /stac` block instead?

birdhouse/optional-components/stac-populator/docker-compose-extra.yml
```diff
@@ -0,0 +1,21 @@
+version: "3.4"
+
+x-logging:
+  &default-logging
+  driver: "json-file"
+  options:
+    max-size: "50m"
+    max-file: "10"
+
+services:
```
mishaschwartz, line 10: Correct me if I'm wrong but doesn't this automatically populate the stac catalog with cmip5 and cmip6 by default: 

https://github.com/crim-ca/stac-populator/blob/master/populate.sh#L40-L43

That seems reasonable to demonstrate the catalog but not very useful for actual usage. We should be able to configure what entries are in the catalog (raised in https://github.com/crim-ca/stac-populator/issues/2).

mishaschwartz: Also, right now, this component will run once when we call `./pavics-compose.sh up` and then exit when its done. That goes against the idea of a "service" in compose, which should really be a long running service, not a single-run process. 

It's also not that intuitive how to use this to re-populate the stac instance once the birdhouse stack is already up and running.

I think that this makes more sense as a script  in (`birdhouse/scripts`) that calls:

`docker run --rm --network container:stac --name stac-populator -e STAC_ASSET_GENERATOR_TIMEOUT=30 -e STAC_HOST=http://stac:8000/stac ghcr.io/crim-ca/stac-populator:master  ./populate.sh`

mishaschwartz (changes requested):

CHANGES.md
```diff
@@ -82,6 +65,28 @@
 
 - Licence: update copyright line with year and ownership
 
+- Adds [STAC](https://github.com/crim-ca/stac-app) to the stack (optional) when ``./components/stac``
```
fmigneault, line 68: Changes need to be moved up to latest section.

CHANGES.md
```diff
@@ -82,6 +65,28 @@
...
+  * Service ``stac`` (API) gets added with endpoints ``/twitcher/ows/proxy/stac`` and ``/stac``.
+    
+  * STAC catalog can be explored via the ``stac-browser`` component, available under ``/stac-browser``.
+      
+  * Image ``crim-ca/stac-app`` is a STAC implementation based on ``stac-utils/stac-fastapi``.
+
+  * Image ``crim-ca/stac-browser`` is a fork of ``radiantearth/stac-browser`` in order to have the capacity to build 
+  the Docker container. The image reference will change when the 
+  [stac-browser PR related to Dockerfile](https://github.com/bird-house/birdhouse-deploy/issues/346) will have been 
+  merged.
```
fmigneault, line 82: References to the repositories still missing.

birdhouse/components/README.rst
```diff
@@ -381,29 +381,35 @@ define your custom values in ``env.local`` directly.
...
+
+  export PROXY_ROOT_LOCATION="return 302 https://\$host/stac-browser;"
 
-STAC API comes pre-loaded with sample data ingested by the ``stac-populator`` component. On stack initialization, 
-STAC item generation workflows will run for ``STAC_ASSET_GENERATOR_TIMEOUT`` seconds.
+Here is a sample search query using a CLI::
 
-Here is a sample search query using a CLI:
+  pip install pystac-client
+  stac-client search $PAVIS_FQDN/stac -q "variable_id=txgt_32" "scenario=ssp585"
```
fmigneault, line 405: Place this inside a 

```
.. code-block:: shell
    <code>
```

birdhouse/components/README.rst
```diff
@@ -381,29 +381,35 @@ define your custom values in ``env.local`` directly.
...
 
-```
-pip install pystac-client
-stac-client search $PAVIS_FQDN/stac -c c604ffb6d610adbb9a6b4787db7b8fd7 -q "variable_id=txgt_32" "scenario=ssp585"
-```
+It is important to note that STAC API is not protected by any authorization mecanism. Calls to the STAC API pass
```
fmigneault, line 407: This is contradictory, as it is mentioned after that this can be adjusted.
It can be protected, but is by default open to anyone.
The STAC Browser however is not protected. This could be added.

birdhouse/components/README.rst
```diff
@@ -381,29 +381,35 @@ define your custom values in ``env.local`` directly.
...
-Here is a sample search query using a CLI:
+  pip install pystac-client
+  stac-client search $PAVIS_FQDN/stac -q "variable_id=txgt_32" "scenario=ssp585"
 
-```
-pip install pystac-client
-stac-client search $PAVIS_FQDN/stac -c c604ffb6d610adbb9a6b4787db7b8fd7 -q "variable_id=txgt_32" "scenario=ssp585"
-```
+It is important to note that STAC API is not protected by any authorization mecanism. Calls to the STAC API pass 
+through Twitcher in order to do authentication. An anonymous user can have read-only access to STAC resources while
```
fmigneault, line 408:
```diff
- to do authentication
+ to validate authorization
```
(authentication is Magpie)

```diff
- An anonymous user can have read-only access to STAC resources
+ Unauthenticated users will have read-only access by default to STAC API resources
``` 
Use more generic terms that doesn't require the reader to know the specific implementation details.
Specify when talking about STAC API vs STAC Browser as they don't share the same protection setup.

birdhouse/components/stac/config/proxy/conf.extra-service.d/stac.conf
```diff
@@ -0,0 +1,26 @@
...
+        proxy_pass https://10.0.0.9/twitcher/ows/proxy/stac/stac;
+        proxy_set_header Host $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Host $host:$server_port;
+        proxy_buffering off;
+        include /etc/nginx/conf.d/cors.include;
+    }
+
+    location /stac-browser/ {
+        proxy_pass http://stac-browser:8080/;
```
fmigneault, line 16: Missing a note about what we decided for protecting the browser as discussed (protection through STAC API sufficient).

birdhouse/optional-components/stac-populator/default.env
```diff
@@ -0,0 +1 @@
+export STAC_ASSET_GENERATOR_TIMEOUT=30
```
fmigneault, line 1: This is missing the variable registration for expansion in templates. 
See example: 
https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/optional-components/secure-data-proxy/default.env#L9-L17

Also, this component should probably have the `components/stac` as a dependency.
See example: 
https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/optional-components/secure-data-proxy/default.env#L30-L36

birdhouse/optional-components/stac-populator/docker-compose-extra.yml
```diff
@@ -0,0 +1,21 @@
...
+  # populates STAC catalog with sample collection items
+  stac-populator:
+    container_name: stac-populator
+    image: ghcr.io/crim-ca/stac-populator:master
+    environment:
+      - STAC_ASSET_GENERATOR_TIMEOUT=${STAC_ASSET_GENERATOR_TIMEOUT}
+      - STAC_HOST=https://${PAVICS_FQDN_PUBLIC}/stac
+    command: >
+      bash -c "./wait-for-it.sh stac:8000 -t 30 && ./populate.sh"
```
fmigneault, line 19: Should the  `-t 30` use `-t ${STAC_ASSET_GENERATOR_TIMEOUT}`?

birdhouse/components/stac/default.env
```diff
@@ -0,0 +1,4 @@
+export STAC_POSTGRES_USER=${POSTGRES_PAVICS_USERNAME}
+export STAC_POSTGRES_PASSWORD=${POSTGRES_PAVICS_PASSWORD}
+export STAC_PGUSER=${POSTGRES_PAVICS_USERNAME}
+export STAC_PGPASSWORD=${POSTGRES_PAVICS_PASSWORD}
```
fmigneault, line 4: Missing variable registration for expansion and the dependant components (only `twitcher`, `magpie`, `proxy` in this case). 
See other comment for details.

birdhouse/components/stac/config/magpie/config.yml.template
```diff
@@ -0,0 +1,19 @@
...
+    title: STAC
+    public: true
+    c4i: false
+    type: api
+    sync_type: api
+
+permissions:
+  - service: stac
+    permission: read
+    group: anonymous
```
mishaschwartz, line 14: Is there a reason why this is added here instead of in the `all-public-access` optional component? 

For almost every other service, the default is that it is not accessible through magpie by default unless `all-public-access` is enabled.

birdhouse/components/stac/config/magpie/docker-compose-extra.yml
```diff
@@ -0,0 +1,6 @@
+version: "3.4"
+services:
+  magpie:
+    volumes:
+    - ./components/stac/config/magpie/config.yml:/opt/local/src/magpie/config/permissions/stac.yml:ro
```
mishaschwartz, line 5: Please use the `MAGPIE_PERMISSIONS_CONFIG_PATH` and `MAGPIE_PROVIDERS_CONFIG_PATH` environment variables here instead of hard-coding the paths.

birdhouse/components/stac/config/proxy/conf.extra-service.d/stac.conf
mishaschwartz, line 1: This file shouldn't have been committed. Please remove it and add it to the .gitignore file

mishaschwartz (changes requested):

birdhouse/components/stac/config/magpie/config.yml.template
mishaschwartz: I understand. My point is that we don't assign anonymous permissions by default for any other component. Is there a reason why stac is different from the other components and needs to have the permissions handled differently?

mishaschwartz: I thought the idea was that the person administering the node chooses what permissions to enable (by logging in to magpie). 

For example, if you're using finch in your stack right now (without `all-public-access`) then users can't query the `getcapabilities` from finch out of the box, the permissions have to be set.

If this is not the pattern we want to follow with stac, that's fine. I'm just pointing out that it's different from what other components are doing right now.

mishaschwartz:
If you go to your demo page (https://stac-dev.crim.ca/stac-browser/) and click either of the following links (under "Additional resources"):

-  [OpenAPI service description](https://stac-dev.crim.ca/api)
-  [OpenAPI service documentation](https://stac-dev.crim.ca/api.html)

It links to the equivalent of `https://${PAVICS_FQDN}/api` which is not a valid route for the birdhouse stack. 

You can see the relevant links in the `"links"` values if you go to [https://stac-dev.crim.ca/stac](https://stac-dev.crim.ca/stac):
image

Is there a way to change or remove these links so that they're not confusing to the end-user?

birdhouse/components/stac/config/magpie/config.yml.template
fmigneault: I tend to agree with @mishaschwartz (*somewhat* 😅)

The node administrator should indeed add a custom Magpie `config.yml` or `permissions.cfg` that applies any relevant permissions to match their needs for their own platform.

Adding this permission to https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse/optional-components/all-public-access will document on how to do this specifically for the STAC service.  It doesn't mean we rely on it necessarily, since each platform has to choose to opt in to the `stac` component anyway.

However, considering that STAC (specifically the `stac-browser`) is eventually intended to be an alternative to JupyterHub as server entrypoint, it could be considered to have an alternate optional component `stac-browser-entrypoint` specifically for setting this up. In that component, it would make sense to have the public permission applied automatically, since the server would fail immediately if accessed publicly otherwise. 

Also, given that STAC is essentially only for metadata search, I am not that much worried if the permission was left there either. Similar public setup is assumed and defined for other components, (e.g.: https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/components/weaver/config/magpie/config.yml.template). Therefore, this rule of custom permissions is not *strictly* applied for all components.

fmigneault: Yes, but add it as a separate `optional-components/stac-public-access`. 
I can see STAC being made public by itself more often while keeping the rest protected.

Also, `optional-components/all-public-access` would depend on `optional-components/stac-public-access`.

birdhouse/components/stac/config/canarie-api/_docker-compose-extra.yml
fmigneault, line 1: Not sure about the reason of this rename? Feels like something wrong.

CHANGES.md
```diff
@@ -15,7 +15,30 @@
...
+  [radiantearth/stac-browser](https://github.com/radiantearth/stac-browser) in order to have the capacity to build 
+  the Docker container. The image reference will change when the 
+  [stac-browser PR related to Dockerfile](https://github.com/bird-house/birdhouse-deploy/issues/346) will have been 
+  merged.
+      
+  * Adds `Magpie` permissions and service for `stac` endpoints.
+  
+- Adds [stac-populator](https://github.com/crim-ca/stac-populator) to populate STAC catalog with sample collection 
+  items via [CEDA STAC Generator](https://github.com/cedadev/stac-generator), employed in sample 
+  [CMIP Dataset Ingestion Workflows](https://github.com/cedadev/stac-generator-example/tree/master/conf).
```
fmigneault, line 42: Missing details about the new optional service to make public STAC.

birdhouse/components/stac/.gitignore
```diff
@@ -1,2 +1 @@
-conf.extra-service.d/stac.conf
-config/magpie/config.yml
```
fmigneault: The files generated from the templates should remain listed here, to avoid them being re-added back.

birdhouse/optional-components/README.rst
```diff
@@ -292,3 +292,19 @@ See |geoserver_secured_pr|_. for more details.
...
+
+Once enabled in the stack, this component will run automatically on stack boot time in order to populate the catalog. 
+On stack initialization, STAC item generation workflows will run for ``STAC_ASSET_GENERATOR_TIMEOUT`` seconds in order 
+to populate the catalog with sample data. Change this timeout as needed, as there are no impact on the stack boot, 
+except time required to feed the catalog.
+
+To enable this optional-component:
+
+- Edit ``env.local`` (a copy of `env.local.example`_)
+- Add ``"./optional-components/stac-populator"`` to ``EXTRA_CONF_DIRS``.
```
fmigneault, line 310: Need details about the `optional-components/stac-public-access` as well.

fmigneault (changes requested):
Quick review pass. 
Will review again once the platform works with those configs.

birdhouse/optional-components/README.rst
```diff
@@ -308,3 +308,14 @@ To enable this optional-component:
...
+
+Allow public access to STAC catalog
+--------------------------------------------------------
+
+STAC Public Access allows STAC catalog to be accessed by anyone, without authentication.
+
+To enable this optional-component:
+
+- Edit ``env.local`` (a copy of `env.local.example`_)
+- Add ``"./optional-components/stac-public-access"`` to ``EXTRA_CONF_DIRS``.
```
fmigneault: Remove the extra string double-quote around the `./optional-components/stac[...]` items.

fmigneault (approved):
Just a minor edit left to fix. 

Is there anything still blocking this PR?

mishaschwartz (approved):

fmigneault:
@matprov 
The bump versions were not applied. Are they planned in a follow-up PR?

birdhouse/components/stac/docker-compose-extra.yml
```diff
@@ -0,0 +1,63 @@
...
+  &default-logging
+  driver: "json-file"
+  options:
+    max-size: "50m"
+    max-file: "10"
+
+services:
+  stac:
+    container_name: stac
+    image: ghcr.io/crim-ca/stac-app:main
```
tlvu, line 13: Not using exact version for reproductibility?

birdhouse/components/stac/docker-compose-extra.yml
```diff
@@ -0,0 +1,63 @@
...
+    depends_on:
+      - stac-db
+    environment:
+      - POSTGRES_USER=${STAC_POSTGRES_USER}
+      - POSTGRES_PASS=${STAC_POSTGRES_PASSWORD}
+      - POSTGRES_DBNAME=postgis
```
tlvu, line 19: Just curious, is `postgis` hardcoded somewhere?  Why not call the DB simply `stac`?

birdhouse/config/twitcher/config/proxy/conf.extra-service.d/twitcher.conf.template
tlvu, line 7: Better use `PAVICS_FQDN_PUBLIC` for anything public facing.

tlvu (commented):
First off, totally sorry for my late review.  I had too much on my plate lately and I completely forgot about this PR given it has been opened for like 6 months and movement seems to only be picked up lately.

I found a few things to fix here but it's alright since all are in the new codes so no regression at all.

birdhouse/optional-components/stac-public-access/docker-compose-extra.yml
```diff
@@ -0,0 +1,6 @@
+version: "3.4"
```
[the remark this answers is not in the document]
mishaschwartz: This works... but if we want to follow the pattern we use elsewhere we should really add this to `birdhouse/optional-components/stac-public-access/config/magpie/docker-compose-extra.yml` so that it will only be set if magpie is enabled as well.

We're almost certainly going to make magpie a required component but it would be nicer to keep the pattern we've already established. 

Thanks for finding that @tlvu

birdhouse/components/stac/docker-compose-extra.yml
mishaschwartz: https://github.com/crim-ca/stac-app/issues/1

birdhouse/optional-components/stac-populator/docker-compose-extra.yml
```diff
@@ -0,0 +1,21 @@
...
+services:
+  # populates STAC catalog with sample collection items
+  stac-populator:
+    container_name: stac-populator
+    image: ghcr.io/crim-ca/stac-populator:master
+    environment:
+      - STAC_ASSET_GENERATOR_TIMEOUT=${STAC_ASSET_GENERATOR_TIMEOUT}
+      - STAC_HOST=http://stac:8000/stac  # STAC API internally accessed to avoid Twitcher authentication
+    command: >
+      bash -c "./wait-for-it.sh stac:8000 -t 30 && ./populate.sh"
```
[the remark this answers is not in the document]
dchandan: @tlvu The pipeline for populating STAC is being handled in a separate repo (https://github.com/crim-ca/stac-populator). I think this is just for testing.

PR #298, opened 2023-03-01
Adjust workspace for cowbird

ChaamC:
## Overview

Update to adjust `birdhouse-deploy` for the use of workspaces with Cowbird. JupyterHub now mounts the `WORKSPACE_DIR` path, which defaults to the usual `/data/jupyterhub_user_data` path. The variable is overriden, if using Cowbird, to use Cowbird's workspace instead (`/data/user_workspaces`), which contains for each user a symlink to his own notebook dir found in `/data/jupyterhub_user_data`. The symlink is created by Cowbird upon a user creation via Magpie.

## Changes

- Cowbird: Updated Cowbird config for user workspaces and for working callbacks to Magpie.

  When enabling Cowbird, the config will now mount a different working directory with JupyterHub, which corresponds to the user workspaces created with Cowbird. These workspaces will use symlinks to the Jupyterhub data directories.

  
  For example, we have the original directory, which is still mounted by default by JupyterHub, which contains the user's notebooks :
image

  If Cowbird is enabled, JupyterHub mounts Cowbird's workspace instead, which has a symlink to the other dir :
image
  Cowbird's workspace can also contain other files related to other services.
  Cowbird's workspace directory is defined by the added environment variable `USER_WORKSPACES`.

- JupyterHub: Updated config to support Cowbird, which uses a different working directory.

  JupyterHub now mounts the variable `WORKSPACE_DIR` when starting a JupyterLab instance. It will refer to the original JupyterHub data directory by default, and if Cowbird is activated, it will be overridden to refer to Cowbird's workspace instead.

  In JupyterHub with Cowbird enabled, the `writable-workspace` is the Cowbird user's workspace :
image
  When we open the `notebooks` dir, it displays the files found at the symlink's source :
image

Relates to https://github.com/Ouranosinc/cowbird/pull/32


ChaamC:
Note : We should determine how we want to define the structure of directories and files for the workspaces. I put the same directory for JupyterHub and Cowbird, but it should probably be defined differently.

We should probably refine this before merging this PR.

fmigneault:
@tlvu 
This is the structure we envisioned. Please share your thoughts or concerns.
```
JUPYTER SIDE       MOUNT
----------------------------------------------------------------------------
workspace/         -> /data/user_workspaces/<USER>
  public/:ro       -> /data/user_workspaces/<USER>/public/
  shared/:ro       -> /data/user_workspaces/<USER>/shared/
    weaver/:ro     -> /data/user_workspaces/<USER>/shared/weaver/
      <JOB>/:ro    -> /data/user_workspaces/<USER>/shared/weaver/<JOB>/
  user/:rw         -> /data/user_workspaces/<USER>/user/
    notebooks:rw   -> /data/user_workspaces/<USER>/user/notebooks



SERVER SIDE                 LINK (TYPE IN ACCORDANCE WITH CASE STUDY RESULT)
----------------------------------------------------------------------------
/data/
  datasets/
    testdata/
      xclim/
        .../
  jupyterhub_user_data/
    <USER>/
      .../
  wps_outputs/
    weaver/ 
    finch/
    ...
  user_workspaces/
    public/
      datasets/         
        testdata/
          xclim/            -> dr--r--r-- /data/datasets/testdata/xclim/
    shared/
      weaver/
        <JOB>/              -> dr--r----- /data/wps_outputs/weaver/<JOB>/
    user/
      notebooks/            -> drw------- /data/jupyterhub_user_data/<USER>/
```

The links themselves would be managed by Cowbird using directories and hard-links synchronized with applied Magpie permissions. Hard-links are used as it resolves for multi-volume mounts and respects `:ro`/`:rw` applied to Dockers, as evaluated a long time ago by @dbyrns.

birdhouse/components/cowbird/docker-compose-extra.yml
```diff
@@ -30,7 +30,7 @@ services:
     volumes:
       - ./components/cowbird/config/cowbird/config.yml:/opt/local/src/cowbird/config/cowbird.yml
       - ./components/cowbird/config/cowbird/cowbird.ini:/opt/local/src/cowbird/config/cowbird.ini
-      - "${DATA_PERSIST_ROOT}/user_workspaces:/user_workspaces"
+      - "${DATA_PERSIST_ROOT}/${USER_WORKSPACES}:/${USER_WORKSPACES}"
```
fmigneault, line 33: Maybe Cowbird will actually need `${DATA_PERSIST_ROOT}/` mount as well in order to generate the links. To be tested...

birdhouse/components/cowbird/docker-compose-extra.yml
```diff
@@ -74,8 +74,10 @@ services:
   # extend Magpie permissions to grant access to Cowbird API via secured Twitcher proxy
   magpie:
     links:
-      # must have link to send webhook requests directly though internal network
+      # must have link to send webhook requests directly through internal network
       - cowbird
+    environment:
+      MAGPIE_WEBHOOKS_CONFIG_PATH: "/opt/local/src/magpie/config/webhooks"
```
fmigneault, line 80: This is now defined in the core `magpie` service, so no need to override anymore.

birdhouse/default.env
```diff
@@ -32,9 +32,12 @@ export BASH_IMAGE="bash:5.1.4"
 # Root directory under which all data persistence should be nested under
 export DATA_PERSIST_ROOT="/data"
 
+# Subdirectory containing the user workspaces
+export USER_WORKSPACES="jupyterhub_user_data"
+
 # Folder on the host to persist Jupyter user data (noteboooks, HOME settings)
 # Must use single-quote for delayed eval.
-export JUPYTERHUB_USER_DATA_DIR='${DATA_PERSIST_ROOT}/jupyterhub_user_data'
+export JUPYTERHUB_USER_DATA_DIR='${DATA_PERSIST_ROOT}/${USER_WORKSPACES}'
```
fmigneault, line 40: To revert based on previous discussions.

Using `${DATA_PERSIST_ROOT}/` mount in `cowbird`/`cowbird-worker`, the link to this directory would be managed by user creation and/or permission sync with `magpie`.

birdhouse/env.local.example
```diff
@@ -285,7 +285,7 @@ export GEOSERVER_ADMIN_PASSWORD=geoserverpass
 # For example, a user named "bob" will have his data in $JUPYTERHUB_USER_DATA_DIR/bob
 # and this folder will be mounted when he logs into JupyterHub.
 # (note: if using 'DATA_PERSIST_ROOT', it must be defined earlier, either in this file or from 'default.env')
-#export JUPYTERHUB_USER_DATA_DIR="$DATA_PERSIST_ROOT/jupyterhub_user_data"
+#export JUPYTERHUB_USER_DATA_DIR="$DATA_PERSIST_ROOT/${USER_WORKSPACES}"
```
fmigneault, line 288: also revert

birdhouse/scripts/backup-jupyterhub-notebooks.sh
```diff
@@ -6,7 +6,7 @@ if [ -z "$BACKUP_OUT_DIR" ]; then
 fi
 
 if [ -z "$JUPYTERHUB_USER_DATA_DIR" ]; then
-    JUPYTERHUB_USER_DATA_DIR=${DATA_PERSIST_ROOT:-/data}/jupyterhub_user_data
+    JUPYTERHUB_USER_DATA_DIR=${DATA_PERSIST_ROOT:-/data}/${USER_WORKSPACES}
```
fmigneault, line 9: also revert

fmigneault (changes requested):

ChaamC:
@fmigneault @tlvu 
PR has been updated. Instead of both JupyterHub and Cowbird using the same workspace, they now use a different workspace. If Cowbird is enabled, its workspaces will include symlinks to the user's JupyterHub data.

The PR is ready for review. More details in the overview description.

birdhouse/docker-compose.yml
```diff
@@ -409,6 +409,7 @@ services:
       JUPYTERHUB_IMAGE_SELECTION_NAMES: ${JUPYTERHUB_IMAGE_SELECTION_NAMES}
       DOCKER_NETWORK_NAME: jupyterhub_network
       JUPYTERHUB_USER_DATA_DIR: ${JUPYTERHUB_USER_DATA_DIR}
+      WORKSPACE_DIR: ${JUPYTERHUB_USER_DATA_DIR}
```
fmigneault, line 412: I think the override strategy of this variable by Cowbird works at the moment only because components are *always* loaded after this "base" docker-compose. 

However, with eventual PR #296 that will make JupyterHub also a component, it could potentially break depending on load order. It think we should find a way to define some kind of `USER_WORKSPACE_DIR` variable that is resolved conditionally considering all loaded components, and the `jupyter` service would simply use it without need to override in `cowbird`'s docker-extra. So, in other words, the override should happen at the `pavics-compose.sh` or the JupyterHub config level (if not already the case as suggested by the `if os.environ['WORKSPACE_DIR'] != jupyterhub_data_dir:` check?) instead of `docker-compose` load order.

@mishaschwartz Tagging you as this item might have some considerations with your PR. Any ideas on how to best handle it so solutions align after both PRs are merged.

birdhouse/env.local.example
```diff
@@ -231,8 +231,8 @@ export GEOSERVER_ADMIN_PASSWORD=geoserverpass
 
 # Jupyter single-user server images
 #export DOCKER_NOTEBOOK_IMAGES="pavics/workflow-tests:210216 \
-#                               pavics/crim-jupyter-eo:0.2.0 \
-#                               pavics/crim-jupyter-nlp:0.2.0"
+#                               pavics/crim-jupyter-eo:0.3.0 \
+#                               pavics/crim-jupyter-nlp:0.4.0"
```
fmigneault, line 235: Add these changes to changelog.

birdhouse/env.local.example
```diff
@@ -334,7 +334,6 @@ export GEOSERVER_ADMIN_PASSWORD=geoserverpass
 #
 ### public-read paths
 #
-## /data/jupyterhub_user_data/public-share/
```
fmigneault, line 337: Instead of removing, update with the variable? 
The example value helps understand the purpose of that variable.

birdhouse/env.local.example
```diff
@@ -347,7 +346,6 @@ export GEOSERVER_ADMIN_PASSWORD=geoserverpass
 #
 ### public-share paths
 #
-## /data/jupyterhub_user_data/public-share/{username}-public
```
fmigneault, line 350: Will the shared/public aspect presented in https://github.com/bird-house/birdhouse-deploy/pull/298#issuecomment-1452189327 be defined in a subsequent PR? It looks like these variable might need conditional activation/update based on `cowbird` being enabled or not.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -99,12 +107,21 @@ if readme != "":
...
 
-    subprocess.call(["chown", "-R", "1000:1000", user_dir])
+    if os.environ['WORKSPACE_DIR'] != jupyterhub_data_dir:
+        # Case for cowbird setup. The workspace directory should also have the user's ownership,
+        # to have working volume mounts with the DockerSpawner.
+        workspace_user_dir = join(os.environ['WORKSPACE_DIR'], username)
+        if not os.path.exists(workspace_user_dir):
+            raise FileNotFoundError(f"The user {username}'s workspace doesn't exist in the workspace directory, "
+                                    "but should have been created by Cowbird already.")
+        subprocess.call(["chown", "1000:1000", workspace_user_dir])
```
fmigneault, line 124: Was removing the `-R` for recursive intentional?
If the directory is not freshly created by Cowbird, could nested contents with different permissions cause problems?

fmigneault (changes requested):

birdhouse/docker-compose.yml
mishaschwartz: I hate to sound like a broken record here but why not use parameter expansion?

```sh
# default.env
export JUPYTERHUB_USER_DATA_DIR='${DATA_PERSIST_ROOT}/${JUPYTER_HUB_USER_DATA_BASE_DIR:-jupyterhub_user_data}'

# components/cowbird/default.env
export JUPYTERHUB_USER_DATA_BASE_DIR="user_workspaces"
```

Also, maybe I'm missing something here but why does the volume mounted at `JUPYTERHUB_USER_DATA_DIR` need to be at the same path as on the host?

birdhouse/config/jupyterhub/jupyterhub_config.py.template
ChaamC: Also, currently, when mounting the cowbird user workspace on JupyterHub, the user is able to create notebooks at the root of the `writable-workspace`. I am wondering if there will be confusions between having notebooks in the actual jupyterhub dir (`/writable-workspace/notebooks`, when mounting with cowbird), vs notebooks at the root of `writable-workspace` or other subdirectories.

fmigneault: The user should be able to create/modify/delete whatever they want under `notebooks/`, but not under other directories at the same level. Permissions items under those directories would be set individually or recursively (if dirs) based on Magpie permissions.

birdhouse/docker-compose.yml
fmigneault: It is possible that some `env.local` could already define `export JUPYTERHUB_USER_DATA_DIR` explicitly, so the required path to be adjusted by cowbird to make this feature work would never be applied. The `WORKSPACE_DIR` variable however is new and specific to this feature, so if it is overridden and things don't work, it is more understandable.

---
For volumes.
**TL;DR** : not required, but way easier / less error prone

They don't "need" (from a technical point) to be the same locations, but having different paths implies that Cowbird/Magpie services always have to perform some kind of reverse mapping of the location to generate the correct links that would be seen from inside the docker images when they mount these volumes. The `cowbird-worker` resolves paths/permissions across all users, so it needs the top `/data/user_workspaces` directory, JupyerHub, however, needs only the user-specific workspace. If we did the following:

```yaml
services:
  jupyter:
    volumes:
      - "/data/jupyterhub_user_data_dir/userX:/notebooks"
      - "/data/user_workspaces/userX:/workspace"
  cowbird-worker:
    volumes:
      - "/data/jupyterhub_user_data_dir:/notebooks"
      - "/data/user_workspaces:/user_workspaces"
```

Then a link `/user_workspaces/userX/notebooks -> /notebooks/userX` created from inside `cowbird-worker` will not work inside `jupyter`, since under its in-docker context, the target path would not exist (directly `/notebooks` is mounted, not `/notebooks/userX`). Cowbird would have to assume that `/notebooks` and `/workspace` will be the mount points in `jupyter`, to create `/user_workspaces/userX/notebooks -> /notebooks`, although `/notebooks` from Cowbird's point of view is an entirely different location. This can lead to a lot of confusion, and incorrect file permission configuration, when trying to manage which links to create/delete under the different `jupyter`/`cowbird` in-docker contexts.  

If instead of `/data/jupyterhub_user_data_dir:/notebooks`, we used let say `/data/jupyterhub_user_data_dir:/all-users-notebooks` to avoid the `/notebooks` location confusion, then Cowbird would instead have to "blindly" create `/user_workspaces/userX/notebooks -> /notebooks`, although for Cowbird, that target `/notebooks` wouldn't exist. All OS links/chmod/etc. calls would have to do a lot of error handling to ignore "missing" paths and force link creation.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
ChaamC: Then I suppose it's alright not to add the `-R` option here, since we don't want to overwrite whatever permission/ownership Magpie would have applied on other directories. 
The `notebooks` directory will also be correct since there is already a `chown` specifically to apply the user `1000` ownership.

birdhouse/env.local.example
ChaamC: Yes, I think this path could change when enabling `cowbird`. I will revert this change (and the other comment example above) for now, since this PR doesn't include changes for the shared/public aspects.

Let me know if you'd prefer me to implement and add the shared/public aspects directly in this PR.

birdhouse/docker-compose.yml
ChaamC: @mishaschwartz 
Since this PR has been ready to merge for a few days, we were wondering if you were OK with me merging it on `master` and you could check to integrate the changes in your current PR [#296](https://github.com/bird-house/birdhouse-deploy/pull/296)?

Would that be okay, or would you have another proposition? Thank you!

mishaschwartz: @ChaamC That's fine with me. Once this is merged in to master I'll update my branch

birdhouse/env.local.example
fmigneault: A following PR is ok.

fmigneault (approved):
Good to go after https://github.com/Ouranosinc/cowbird/pull/32 is merged + deployed. 
Need to update https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/components/cowbird/default.env#L27 accordingly.

Don't forget to include the contents of the PR description in the merge commit message.

tlvu:
Sorry for the very late review.  Got hit by some operational problems on Ouranos side.

Is there a reason you did not perform a release (a tag basically) for this PR?  This PR effectively enhance cowbird functionality.  See https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/README.rst#release-procedure

But do not open a new PR just for this.  Just keep in mind next time if you change some user facing behavior, it is worth a release.  Updating documentations or code comments do not need a release of course.

Below just some question:


> SERVER SIDE                 LINK (TYPE IN ACCORDANCE WITH CASE STUDY RESULT)
> ----------------------------------------------------------------------------
> /data/
>   datasets/
>     testdata/
>       xclim/
>         .../
>   jupyterhub_user_data/
>     <USER>/
>       .../
>   wps_outputs/
>     weaver/ 
>     finch/
>     ...
>   user_workspaces/
>     public/
>       datasets/         
>         testdata/
>           xclim/            -> dr--r--r-- /data/datasets/testdata/xclim/

All data on our Thredds right now are mostly public.  Does it means the entire Thredds hierarchy will be replicated in this `public/datasets` section?

>     shared/
>       weaver/
>         <JOB>/              -> dr--r----- /data/wps_outputs/weaver/<JOB>/

FYI `wps_outputs` is currently a data-volume, not a direct mount from disk.  

This `shared` section is for birds output only or other users sharing with the current user will also appear here?  Read-only and writable share both supported?

>     user/
>       notebooks/            -> drw------- /data/jupyterhub_user_data/<USER>/

Inside the Jupyter env, `workspaces/user/notebooks` will be the same as `writable-workspace`?

ChaamC:
@tlvu 
> Is there a reason you did not perform a release (a tag basically) for this PR? 

No reason in particular. I didn't know if birdhouse was bumped systematically at each PR.  Sorry about this, will do next time.

> All data on our Thredds right now are mostly public. Does it means the entire Thredds hierarchy will be replicated in this `public/datasets` section?

Not sure about this one. Maybe @fmigneault has an answer.

> This `shared` section is for birds output only or other users sharing with the current user will also appear here? Read-only and writable share both supported?

It is not yet implemented, so it might change, but I believe the `shared` section will indeed contain files shared from other users to the current user, with whatever permission is applied (read-only, writable).

> > ```
> > user/
> >   notebooks/            -> drw------- /data/jupyterhub_user_data/<USER>/
> > ```
> 
> Inside the Jupyter env, `workspaces/user/notebooks` will be the same as `writable-workspace`?

The `user_workspaces/user/notebooks` dir is a simlink to the original jupyterhub user data dir (`/data/jupyterhub_user_data/<USER>/`). 
So if Cowbird is not activated, the `writable-workspace` remains the same as the original jupyterhub user data dir.
If Cowbird is activated, the `writable-workspace` is instead mounted with the `user_workspaces/user` path.

fmigneault:
@ChaamC @tlvu 
For now, THREDDS is not replicated. However, it is on our todo list of items to investigate. 

One thing that could be considered is to have each sync method/service combination defined in a respective optional-component. 
(this is the same for GeoServer, as discussed with @ChaamC)

Therefore, an instance could decide to provide Cowbird Notebooks share/permission syncs in user workspaces, while omitting THREDDS-specific syncs, and/or GeoServer-specific syncs, and so on.

tlvu:
> For now, THREDDS is not replicated. However, it is on our todo list of items to investigate.

@fmigneault 

You should make this configurable (which part of Thredds to replicate, for which users/groups) and allow an option to toogle this off altogether.

We have tons of contents on Thredds and 99% are public.  If the entire public content is replicated to all Jupyter users (we have a lot), this might potentially be a mess to manage.  This large content might also overwhelm the user?

What worries me is that you said the links are hardlinks.  Which means if I update the source file on Thedds, the old version will never be released since there is still a hardlink pointing to it.   And now we will have tons of hardlinks pointing to the old version !

Is there a detection if the source change, re-create all the hardlinks?

So Cowbird should trigger hardlinks sync when permission change in Magpie but also when the source file simply changed/deleted.

tlvu:
Another potential problem with hardlink in the case of writable share: how can both sides edit the same file since hardlink will not let them see the other side update?  I assume a writable share is for collaborating on the same file, like collaborative editing on a google docs.

fmigneault:
@tlvu 

The specific items to sync or not between THREDDS and the user workspace can be defined with the cowbird configuration (https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/components/cowbird/config/cowbird/config.yml.template#L37). 
It should be possible to have different configuration over multiple optional components for distinct use cases (sync `public` dir, sync `shared` dir between users) as cowbird can load multiple configurations like magpie does.
Going deeper than this level of fragmentation of the configs however would not be reasonable for this shared deployment stack. If more fined-grained control is needed, it should be a custom configuration.

The file system monitoring (https://github.com/Ouranosinc/cowbird/blob/master/cowbird/monitoring/monitor.py) is in charge of detecting the changes of links to apply relevant updates with its callback operations defined by cowbird. We need to test this more, but I'm guessing it should be able to handle changes to file revisions to call the link re-creation.

I think I have found a way to do it with soft links, though, so no updates would be needed if files remain the same. They would only need creation/cleanup if files are added/removed.

PR #301, opened 2023-03-08
Pre commit hooks for automated code style cleanup

mishaschwartz:
## Overview

This adds a configuration for automated code cleanup using the pre-commit (https://pre-commit.com/) tool. 

The .pre-commit-config.yaml file defines a few hooks that are run whenever a commit is made locally and edits changed files to do things like:

- ensure there is a single newline at the end of each file
- ensure that lines do not contain trailing whitespaces
- cleans up yaml files

There are additional hooks that could be added later if needed (see the website for details).

To install locally:

```python
pip install pre-commit
pre-commit install # inside the root directory of this repo
```

This can also be added as a CI service on github (https://pre-commit.ci/) and it's free for open source projects. 

I would highly recommend we enable this as a CI service to enforce code style.

Note: As part of this PR I have run these hooks on all files in this repo. I added the commit hash where these changes were made to a `.git-blame-ignore-revs` file so that these changes don't pollute the git blame results.

## Changes

**Non-breaking changes**

- adds a config file for git-commit and clean up newlines and whitespace

**Breaking changes**

None

## Additional Information

Since this does not change the functionality of the code at all do I need to add to the CHANGES.md file?


.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
```
Zeitsperre, line 1:
```suggestion
default_language_version:
  python: python3

repos:
```

I've found that this helps prevent issues where pre-commit hooks were installed with a particular Python version and the user later changed the Python version. Just a suggestion.

mishaschwartz: Sounds good. I don't think this applies to any of the current hooks but if we add some python specific ones in the future this will make sure that we're enforcing python3 syntax.

.bumpversion.cfg
```diff
@@ -4,20 +4,20 @@ commit = True
...
 	------------------------------------------------------------------------------------------------------------------
-replace = 
+replace =
 	[Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 	------------------------------------------------------------------------------------------------------------------
-	
+
 	[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
-	
+
```
fmigneault, line 20: This will always be changed back and forth. The added spaces are applied by `bump2version` called on each new tag.
Better to have an ignore entry to avoid the noise those changes will generate.

birdhouse/components/weaver/images/component-diagram.svg
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Do not edit this file with editors other than diagrams.net -->
```
fmigneault, line 2: Just to validate. The actual diff is just the newline? Not sure if it's better to ignore `.svg`.

fmigneault (commented):
Consider adding a `.github/workflows/linting.yaml` that performs similar checks.
While `pre-commit` is useful, it assumes it was installed by the developer, which might not be the case. We should double-check and cause a required test to fail PRs if anything was pushed bypassing `pre-commit`.

mishaschwartz:
> While `pre-commit` is useful, it assumes it was installed by the developer

Right, which is why I recommend adding the CI plugin so that the changes are enforced upstream as well.

This is just a suggestion, if there is another tool you'd prefer, such as a linting workflow, then we can do that instead. The one advantage that pre-commit has is that if a developer chooses to install the hooks locally, they don't get surprised *after* they push.

.bumpversion.cfg
mishaschwartz: Sure, we can add as many files in the ignore list as we want....

birdhouse/components/weaver/images/component-diagram.svg
mishaschwartz: Yeah it looks like the only change is the newline...

Ok, so we should add the following files to the ignore regex:

- birdhouse/components/weaver/images/component-diagram.svg
- .bumpversion.cfg

Any other ones you can think of?

Zeitsperre:
In case you're curious, here's an example that works for us: https://github.com/Ouranosinc/xclim/blob/master/.github/workflows/main.yml

mishaschwartz:
Thanks @Zeitsperre, that's really useful!

tlvu:
Nice !

Can you add in birdhouse/README.md how to setup local env to replicate the same pre-commit?

Add something to CHANGES.md, but I think this PR do not need a tag on merge since no code change.

I noticed a .svg file change, I safer to exclude since it's graphics file.

Zeitsperre:
> Can you add in birdhouse/README.md how to setup local env to replicate the same pre-commit?

This might be a bit much, but if we wanted to, we could suggest that users use the `git-init` hooks so that this check is always performed, regardless of repo: https://pre-commit.com/#automatically-enabling-pre-commit-on-repositories

mishaschwartz:
> * Enable pre-commit CI with the auto-commit option to effect changes on pushes to branches
>       
>       * Leverage `styfle/cancel-workflow-action` in order to stop builds when pre-commit pushes changes
> 
>     * Instead of creating a separate linting workflow, add linting checks as a first step in your testing workflows.
>       
>       * Have your testing steps follow from your linting step so that failures prevent needless testing.

@tlvu how do you recommend we enable this for the Jenkins tests?

tlvu:
> > * Enable pre-commit CI with the auto-commit option to effect changes on pushes to branches
> >   ```
> >   * Leverage `styfle/cancel-workflow-action` in order to stop builds when pre-commit pushes changes
> >   ```
> >   
> >   
> >       
> >         
> >       
> >   
> >         
> >       
> >   
> >       
> >     
> >   
> >   * Instead of creating a separate linting workflow, add linting checks as a first step in your testing workflows.
> >     
> >     * Have your testing steps follow from your linting step so that failures prevent needless testing.
> 
> @tlvu how do you recommend we enable this for the Jenkins tests?

@mishaschwartz Do you mean run these pre-commit on Jenkins?  Not sure it's needed if it runs on Github already for this repo.

For any changes to the deployment, we already have Jenkins run by CRIM on each PR.

For me Jenkins is for tests that have sensitive info (ex: internal hostname/config, ...) or too heavy for public platform (ex: spinning up an full PAVICS stack).

Otherwise, keeping all the tests on public platform is easier for sharing test results.

mishaschwartz:
@tlvu I was mostly asking about the line: "...  in order to stop builds when pre-commit pushes changes" but now I remember Mathieu mentioning that canceling tests was not possible for your Jenkins set up. So ignore this for now... 

I suggest we just enable pre-commit CI for now as a standalone check and we can figure out if we want to tie it into the full CI pipeline with the tests later on.

mishaschwartz:
I'm going to close this for now as it is not needed at the moment and will be easy enough to recreate in the future if needed. Or if another solution is preferred we can go with that one instead.

PR #304, opened 2023-03-24
Enforce the load order of components defined in env.local

### LINKED ISSUE #303: :bug: [BUG]: Load order of `EXTRA_CONF_DIRS` not respected by `COMPOSE_CONF_LIST`
## Summary

Load order of `EXTRA_CONF_DIRS` is not respected by `COMPOSE_CONF_LIST` to apply items defined in a critical order.

The `EXTRA_CONF_DIRS` should be considered as user-critical definitions to be respected at all costs, and therefore, loaded as much as possible toward the end to preserve its overrides. 

## Details

In my `env.local`, I define the following: 
(note the last `daccs-env` which contains custom docker-compose overrides, defined outside `birdhouse-deploy`)

```
export EXTRA_CONF_DIRS="
    ./components/monitoring
    ./optional-components/test-weaver
    ./optional-components/secure-data-proxy
    ./components/weaver
    ./components/cowbird
    ../../daccs-env
"
```

Following are the exact load order reported by `COMPOSE_CONF_LIST` (only reformatted line-by-line for readability) 
when running `pavics-compose up -d` 

_**note**: using a diff only to highlight relevant items, not an actual diff_
($\color{green}{green}$: items matching above extras, $\color{red}{red}$: items of interest)

```diff
docker-compose.yml
./config/proxy/docker-compose-extra.yml
./config/phoenix/docker-compose-extra.yml
./config/geoserver/docker-compose-extra.yml
./config/flyingpigeon/docker-compose-extra.yml
./config/finch/docker-compose-extra.yml
./config/raven/docker-compose-extra.yml
./config/hummingbird/docker-compose-extra.yml
./config/thredds/docker-compose-extra.yml
./config/portainer/docker-compose-extra.yml
./config/magpie/docker-compose-extra.yml
./config/twitcher/docker-compose-extra.yml
./config/jupyterhub/docker-compose-extra.yml
./config/frontend/docker-compose-extra.yml
./config/project-api/docker-compose-extra.yml
./config/catalog/docker-compose-extra.yml
./config/malleefowl/docker-compose-extra.yml
./config/solr/docker-compose-extra.yml
+ ./components/monitoring/docker-compose-extra.yml
+ ./optional-components/testthredds/docker-compose-extra.yml
+ ./optional-components/test-weaver/docker-compose-extra.yml
+ ./components/weaver/docker-compose-extra.yml
+ ./components/cowbird/docker-compose-extra.yml
+ ../../daccs-env/docker-compose-extra.yml
./config/mongodb/docker-compose-extra.yml
./config/postgres/docker-compose-extra.yml
./config/wps_outputs-volume/docker-compose-extra.yml
- ./config/data-volume/docker-compose-extra.yml  # loaded after custom env!
./config/ncwms2/docker-compose-extra.yml
./config/phoenix/config/proxy/docker-compose-extra.yml
./config/geoserver/config/magpie/docker-compose-extra.yml
./config/geoserver/config/proxy/docker-compose-extra.yml
./config/flyingpigeon/config/magpie/docker-compose-extra.yml
./config/flyingpigeon/config/wps_outputs-volume/docker-compose-extra.yml
./config/finch/config/magpie/docker-compose-extra.yml
./config/finch/config/proxy/docker-compose-extra.yml
./config/finch/config/wps_outputs-volume/docker-compose-extra.yml
./config/raven/config/magpie/docker-compose-extra.yml
./config/raven/config/proxy/docker-compose-extra.yml
./config/raven/config/wps_outputs-volume/docker-compose-extra.yml
./config/hummingbird/config/data-volume/docker-compose-extra.yml
./config/hummingbird/config/magpie/docker-compose-extra.yml
./config/hummingbird/config/wps_outputs-volume/docker-compose-extra.yml
./config/thredds/config/proxy/docker-compose-extra.yml
./config/portainer/config/proxy/docker-compose-extra.yml
./config/magpie/config/proxy/docker-compose-extra.yml
./config/twitcher/config/proxy/docker-compose-extra.yml
./config/jupyterhub/config/magpie/docker-compose-extra.yml
./config/jupyterhub/config/proxy/docker-compose-extra.yml
./config/frontend/config/proxy/docker-compose-extra.yml
./config/project-api/config/proxy/docker-compose-extra.yml
./config/catalog/config/magpie/docker-compose-extra.yml
./config/catalog/config/proxy/docker-compose-extra.yml
./config/malleefowl/config/data-volume/docker-compose-extra.yml
./config/malleefowl/config/magpie/docker-compose-extra.yml
./config/malleefowl/config/proxy/docker-compose-extra.yml
./config/malleefowl/config/wps_outputs-volume/docker-compose-extra.yml
./config/solr/config/proxy/docker-compose-extra.yml
+ ./optional-components/wps-healthchecks/config/catalog/docker-compose-extra.yml
+ ./optional-components/wps-healthchecks/config/finch/docker-compose-extra.yml
+ ./optional-components/wps-healthchecks/config/flyingpigeon/docker-compose-extra.yml
+ ./optional-components/wps-healthchecks/config/hummingbird/docker-compose-extra.yml
+ ./optional-components/wps-healthchecks/config/malleefowl/docker-compose-extra.yml
+ ./optional-components/wps-healthchecks/config/raven/docker-compose-extra.yml
+ ./optional-components/secure-thredds/config/magpie/docker-compose-extra.yml
+ ./optional-components/testthredds/config/proxy/docker-compose-extra.yml
+ ./optional-components/secure-data-proxy/config/magpie/docker-compose-extra.yml
+ ./optional-components/secure-data-proxy/config/proxy/docker-compose-extra.yml
+ ./components/weaver/config/magpie/docker-compose-extra.yml
+ ./components/weaver/config/proxy/docker-compose-extra.yml
+ ./components/weaver/config/twitcher/docker-compose-extra.yml
+ ./components/cowbird/config/geoserver/docker-compose-extra.yml
+ ./components/cowbird/config/jupyterhub/docker-compose-extra.yml
+ ./components/cowbird/config/magpie/docker-compose-extra.yml
+ ./components/cowbird/config/proxy/docker-compose-extra.yml
./config/canarie-api/config/proxy/docker-compose-extra.yml
./config/wps_outputs-volume/config/proxy/docker-compose-extra.yml
./config/ncwms2/config/magpie/docker-compose-extra.yml
./config/ncwms2/config/proxy/docker-compose-extra.yml
./config/ncwms2/config/wps_outputs-volume/docker-compose-extra.yml
```

As displayed above, load order seems somewhat random. The docker-compose from `weaver`, `cowbird`, etc. are loaded high up, and remaining configs from the same components are then loaded almost at the end. Meanwhile, multiple other components are loaded in between.

In this specific case, the last `../../daccs-env` from `EXTRA_CONF_DIRS` is explicitly required at the end. 
Amongst other things, it overrides the `volumes.data: {}` definition with an actual location on the machine.
Because `./config/data-volume/docker-compose-extra.yml` is loaded after `../../daccs-env/docker-compose-extra.yml`, it "overrides-back" my original override.
I end up getting the following error due to mismatching definitions with the existing volume: 
```
ERROR: Configuration for volume data specifies "o" driver_opt None, but a volume with the same name uses a different "o" driver_opt (bind). If you wish to use the new configuration, please remove the existing volume "birdhouse_data" first:
$ docker volume rm birdhouse_data
```

Furthermore, the `./optional-components/canarie-api-full-monitoring` item from `EXTRA_CONF_DIRS` is missing entirely. 
Not quite sure why, but this is breaking #284 and the E2E that depends on it.

### To Reproduce

Steps to reproduce the behaviour:

See description. 
Load order hard to predict.

## Environment

| Server/Platform URL   |  n/a
| Version Tag/Commit    | https://github.com/bird-house/birdhouse-deploy/tree/1.24.0
| Related issues/PR     | #284
| Related components    | `pavics-compose`
| Custom configuration  | *see description*

## Concerned Organizations

All

mishaschwartz:
## Overview

  Extra components defined in the `EXTRA_CONF_DIRS` variables were being loaded before the dependant components
  defined in the `COMPONENT_DEPENDENCIES` variables in each default.env file. This meant that if an extra component
  was meant to override some setting defined in a dependant component, the setting would not be overridden by the
  extra component. 

  This change enforces the following load order rules:

  - components defined in `DEFAULT_CONF_DIRS` are loaded before those in `EXTRA_CONF_DIRS`
  - components are loaded in the order they appear in either `DEFAULT_CONF_DIRS` or `EXTRA_CONF_DIRS`
  - components that appear in `COMPONENT_DEPENDENCIES` variable are immediately loaded unless they have already been
    loaded

  For example, with the following files in place:

  ```shell
  # env.local
  DEFAULT_CONF_DIRS="
    ./config/twitcher
    ./config/project-api
    ./config/magpie
  "
  EXTRA_CONF_DIRS="
    ./optional-components/generic_bird
    ./components/cowbird
  "
  
  # config/twitcher/default.env
  COMPONENT_DEPENDENCIES="
    ./config/magpie
  "
  # optional-components/generic_bird/default.env
  COMPONENT_DEPENDENCIES="
    ./config/wps_outputs-volume
  "
  ```

  the load order is:

  - ./config/magpie (loaded as a dependency of twitcher, not loaded a second time after project-api)
  - ./config/twitcher
  - ./config/project-api
  - ./config/wps_outputs-volume (loaded as a dependency of generic_bird)
  - ./optional-components/generic_bird
  - ./components/cowbird

  This load order also applies to the order that docker-compose-extra.yml files are specified. If a component also
  includes an override file for another component (eg: ./config/finch/config/proxy/docker-compose-extra.yml overrides 
  ./config/proxy/docker-compose-extra.yml), the following additional load order rules apply:

  - if the component that is being overridden has already been loaded, the override file is loaded immediately
  - otherwise, the override files will be loaded immediately after the component that is being overridden has been loaded

  For example, with the following files in place:

  ```shell
  # env.local
  DEFAULT_CONF_DIRS="
    ./config/finch
    ./config/proxy
  "
  ```
  ```yaml
  # config/proxy/docker-compose-extra.yml
    ...
  # config/finch/docker-compose-extra.yml
    ...
  # config/finch/config/proxy/docker-compose-extra.yml
    ...
  ```

  the docker compose files will be loaded in the following order: 

  - config/finch/docker-compose-extra.yml
  - config/proxy/docker-compose-extra.yml
  - config/finch/config/proxy/docker-compose-extra.yml
## Changes

**Non-breaking changes**
- make load order of components closer to how it was prior to version 1.24

**Breaking changes**
- None

## Related Issue / Discussion

- Resolves #303
- Closes #306 

## Additional Information

Note that commit [remove unnecessary variable extension](https://github.com/bird-house/birdhouse-deploy/commit/25daaea486c5ee7946020480e36fd5051bdf3266) is not strictly necessary for this fix. It simplifies the env files a bit though.


fmigneault (changes requested):
I'm happy to see that explicit `$COMPONENT_DEPENDENCIES` references are not needed anymore. That is much easier to maintain and less error-prone for new users!

I think dependencies should probably be loaded _**before**_ the component itself. 
By this, I refer to the similar behaviour when importing modules in Python or other language.
For example: 

```python
# file a.py
from b import B
```
```python
# file b.py
from c import C
```
```python 
# main
from a import A
```
This will start loading `a`, but in order to complete loading `A`, `b` will have to be loaded first.
When starting to load `b` to get `B`, `C` will be loaded until the end of `c` module. 
Import of `b` then completes, then `a`, and finally `main` finishes doing its thing.

This is to avoid an issue where let's say I use `weaver` variables to define other `cowbird` variables relevant to `weaver`. Naturally, `weaver` parameters must be loaded first (and all its dependencies that can themselves be need by `weaver`) before `cowbird` can use them. 

It would make sense to have `docker-compose.yml` files loaded in the same order that reflects that logic, such that what is reported as `COMPOSE_CONF_LIST` follows the same dependency resolution order.

I also think that `EXTRA_CONF_DIRS` should be loaded completely at the end, after all the `COMPONENT_DEPENDENCIES`, to define "anything left" that wasn't already loaded by other components. By that point, most of the time only explicit overrides will remain, or components that are not a dependency of anything, so it's fine to load last.

I tried running `pavics-compose` with a merge of #284 + PR + my custom overrides. 
I got the following loading order. 

```shell
COMPOSE_CONF_LIST=
-f docker-compose.yml 
-f ./config/proxy/docker-compose-extra.yml 
-f ./config/phoenix/docker-compose-extra.yml 
-f ./config/malleefowl/docker-compose-extra.yml 
-f ./config/postgres/docker-compose-extra.yml 
-f ./config/wps_outputs-volume/docker-compose-extra.yml 
-f ./config/data-volume/docker-compose-extra.yml 
-f ./config/flyingpigeon/docker-compose-extra.yml 
-f ./config/catalog/docker-compose-extra.yml 
-f ./config/mongodb/docker-compose-extra.yml 
-f ./config/geoserver/docker-compose-extra.yml 
-f ./config/finch/docker-compose-extra.yml 
-f ./config/raven/docker-compose-extra.yml 
-f ./config/hummingbird/docker-compose-extra.yml 
-f ./config/thredds/docker-compose-extra.yml 
-f ./config/portainer/docker-compose-extra.yml 
-f ./config/magpie/docker-compose-extra.yml 
-f ./config/twitcher/docker-compose-extra.yml 
-f ./config/jupyterhub/docker-compose-extra.yml 
-f ./config/frontend/docker-compose-extra.yml 
-f ./config/ncwms2/docker-compose-extra.yml 
-f ./config/project-api/docker-compose-extra.yml 
-f ./config/solr/docker-compose-extra.yml 
-f ./components/monitoring/docker-compose-extra.yml 
-f ./optional-components/testthredds/docker-compose-extra.yml 
-f ./optional-components/test-weaver/docker-compose-extra.yml 
-f ./components/weaver/docker-compose-extra.yml 
-f ./components/cowbird/docker-compose-extra.yml 
-f ../../daccs-env/docker-compose-extra.yml 
-f ./config/canarie-api/config/proxy/docker-compose-extra.yml 
-f ./config/phoenix/config/proxy/docker-compose-extra.yml 
-f ./config/malleefowl/config/data-volume/docker-compose-extra.yml 
-f ./config/malleefowl/config/magpie/docker-compose-extra.yml 
-f ./config/malleefowl/config/proxy/docker-compose-extra.yml 
-f ./config/malleefowl/config/wps_outputs-volume/docker-compose-extra.yml 
-f ./config/wps_outputs-volume/config/proxy/docker-compose-extra.yml 
-f ./config/flyingpigeon/config/magpie/docker-compose-extra.yml 
-f ./config/flyingpigeon/config/wps_outputs-volume/docker-compose-extra.yml 
-f ./config/catalog/config/magpie/docker-compose-extra.yml 
-f ./config/catalog/config/proxy/docker-compose-extra.yml 
-f ./config/geoserver/config/magpie/docker-compose-extra.yml 
-f ./config/geoserver/config/proxy/docker-compose-extra.yml 
-f ./config/finch/config/magpie/docker-compose-extra.yml 
-f ./config/finch/config/proxy/docker-compose-extra.yml 
-f ./config/finch/config/wps_outputs-volume/docker-compose-extra.yml 
-f ./config/raven/config/magpie/docker-compose-extra.yml 
-f ./config/raven/config/proxy/docker-compose-extra.yml 
-f ./config/raven/config/wps_outputs-volume/docker-compose-extra.yml 
-f ./config/hummingbird/config/data-volume/docker-compose-extra.yml 
-f ./config/hummingbird/config/magpie/docker-compose-extra.yml 
-f ./config/hummingbird/config/wps_outputs-volume/docker-compose-extra.yml 
-f ./config/thredds/config/proxy/docker-compose-extra.yml 
-f ./config/portainer/config/proxy/docker-compose-extra.yml 
-f ./config/magpie/config/proxy/docker-compose-extra.yml 
-f ./config/twitcher/config/proxy/docker-compose-extra.yml 
-f ./config/jupyterhub/config/magpie/docker-compose-extra.yml 
-f ./config/jupyterhub/config/proxy/docker-compose-extra.yml 
-f ./config/frontend/config/proxy/docker-compose-extra.yml 
-f ./config/ncwms2/config/magpie/docker-compose-extra.yml 
-f ./config/ncwms2/config/proxy/docker-compose-extra.yml 
-f ./config/ncwms2/config/wps_outputs-volume/docker-compose-extra.yml 
-f ./config/project-api/config/proxy/docker-compose-extra.yml 
-f ./config/solr/config/proxy/docker-compose-extra.yml 
-f ./optional-components/canarie-api-full-monitoring/config/proxy/docker-compose-extra.yml 
-f ./optional-components/all-public-access/config/catalog/docker-compose-extra.yml 
-f ./optional-components/all-public-access/config/finch/docker-compose-extra.yml 
-f ./optional-components/all-public-access/config/flyingpigeon/docker-compose-extra.yml 
-f ./optional-components/all-public-access/config/hummingbird/docker-compose-extra.yml 
-f ./optional-components/all-public-access/config/malleefowl/docker-compose-extra.yml 
-f ./optional-components/all-public-access/config/ncwms2/docker-compose-extra.yml 
-f ./optional-components/all-public-access/config/raven/docker-compose-extra.yml 
-f ./optional-components/all-public-access/config/secure-data-proxy/docker-compose-extra.yml 
-f ./optional-components/all-public-access/config/thredds/docker-compose-extra.yml 
-f ./optional-components/wps-healthchecks/config/catalog/docker-compose-extra.yml 
-f ./optional-components/wps-healthchecks/config/finch/docker-compose-extra.yml 
-f ./optional-components/wps-healthchecks/config/flyingpigeon/docker-compose-extra.yml 
-f ./optional-components/wps-healthchecks/config/hummingbird/docker-compose-extra.yml 
-f ./optional-components/wps-healthchecks/config/malleefowl/docker-compose-extra.yml 
-f ./optional-components/wps-healthchecks/config/raven/docker-compose-extra.yml 
-f ./optional-components/secure-thredds/config/magpie/docker-compose-extra.yml 
-f ./optional-components/testthredds/config/proxy/docker-compose-extra.yml 
-f ./components/weaver/config/magpie/docker-compose-extra.yml 
-f ./components/weaver/config/proxy/docker-compose-extra.yml 
-f ./components/weaver/config/twitcher/docker-compose-extra.yml 
-f ./optional-components/secure-data-proxy/config/magpie/docker-compose-extra.yml 
-f ./optional-components/secure-data-proxy/config/proxy/docker-compose-extra.yml 
-f ./components/cowbird/config/geoserver/docker-compose-extra.yml 
-f ./components/cowbird/config/jupyterhub/docker-compose-extra.yml 
-f ./components/cowbird/config/magpie/docker-compose-extra.yml 
-f ./components/cowbird/config/proxy/docker-compose-extra.yml
```

Although it *works*, I think there are a lot of chances that any component after my `../../daccs-env/docker-compose-extra.yml ` override could break my setup at any time, if they just so happen to modify the set of dependencies or rewrite a `docker-compose` item I also override.

I would also expect all `./components/weaver/configs/[...]` to be loaded somewhat close to the `./components/weaver`, since they are themselves only dependencies of that component.

birdhouse/read-configs.include.sh
```diff
@@ -119,22 +120,9 @@ read_components_default_env() {
...
-    # Note: source_conf_files may update COMPONENT_DEPENDENCIES when sourcing the default.env
-    # files in each component's directory.
-    source_conf_files "$requested_conf_dirs" 'EXTRA_CONF_DIRS or DEFAULT_CONF_DIRS'
-    while [ "$current_dependencies" != "$COMPONENT_DEPENDENCIES" ]; do
-      # if additional component dependencies are added by sourcing configuration files, then
-      # source the new dependencies as well.
-      current_dependencies=$COMPONENT_DEPENDENCIES
-      source_conf_files "$COMPONENT_DEPENDENCIES" 'COMPONENT_DEPENDENCIES'
-    done
+
```
tlvu, line 123: Very nice simplification, much cleaner logic !!!  In the previous review https://github.com/bird-house/birdhouse-deploy/pull/296#discussion_r1138013218 I was lost at this part of the logic too.

birdhouse/read-configs.include.sh
```diff
@@ -108,6 +108,7 @@ source_conf_files() {
       if [ -f "$COMPONENT_DEFAULT_ENV" ]; then
           echo "reading '$COMPONENT_DEFAULT_ENV'"
           . "$COMPONENT_DEFAULT_ENV"
+          source_conf_files "$COMPONENT_DEPENDENCIES" "a dependency of $adir"
```
tlvu, line 111: I order to make this config loader order respect the intended override precedence, I would suggest the following logic change.

Current logic in `source_conf_files()`:
```
for adir in list_of_conf_dirs:
  add adir to ALL_CONF_DIRS
  read adir/default.env  # get value of COMPONENT_DEPENDENCIES
  recursive call source_conf_files on COMPONENT_DEPENDENCIES
    add COMPONENT_DEPENDENCIES to ALL_CONF_DIRS
    read defaut.env of COMPONENT_DEPENDENCIES  # get value of COMPONENT_DEPENDENCIES and turtle down
```

If `comp2` depends on `comp1`, and only `comp2` is enabled, `ALL_CONF_DIRS` will list `comp2` before `comp1` so if `comp2` has a fragment to override `comp1` compose file, the fragment might not work since `comp1` compose file has not been read yet.

Furthermore, if `comp2 default.env` has override for `comp1 default.env`, given `comp2 default.env` is read before `comp1 default.env`, it will not be able to override it properly.

Also, since the ordering of `ALL_CONF_DIRS` is reversed, we will also executing all the pre/post docker-compose-up scripts in the reverse order.

To be safe, maybe I would change the logic to something like below (untested, I might be wrong).

Proposed logic in source_conf_files():
```
for adir in list_of_conf_dirs:
  read adir/default.env  # get value of COMPONENT_DEPENDENCIES

  recursive call source_conf_files on COMPONENT_DEPENDENCIES
    read adir of COMPONENT_DEPENDENCIES default.env   # COMPONENT_DEPENDENCIES variable name clashing here, might need to save previous value

    recursive call source_conf_files on COMPONENT_DEPENDENCIES  # and turtle down 

    read adir COMPONENT_DEPENDENCIES default.env  # read again to be able to override dependent default.env
    append adir COMPONENT_DEPENDENCIES to ALL_CONF_DIRS

  read adir/default.env  # again to override dependent default.env
  append adir to ALL_CONF_DIRS
```

So the suggested logic above might fix the loading order for "hard" dependencies (listed in `COMPONENT_DEPENDENCIES`).

For "soft" dependencies (ex: `all-public-access` has magpie config to make all the birds public but do not list all the birds in its `COMPONENT_DEPENDENCIES`), the user still need to ensure `all-public-access` is listed after all the birds (all its soft dependencies) in `EXTRA_CONF_DIRS`.

tlvu (changes requested):
I suggested a logic change to restore the original intended override capabilities.

Not tested so I am not sure it does what I think it should do or can be implemented (recursive call is usually more tricky to code).  Let me know how it goes.

tlvu:
To further preserve the loading order from before, this section of code that generate `COMPOSE_CONF_LIST` var
https://github.com/bird-house/birdhouse-deploy/blob/a91b7dc4673b62907a460ce7d0d14b7bb0628dc2/birdhouse/pavics-compose.sh#L107-L130

Should also be changed to this so all extra fragments modifying other components will load immediately after its parent component:

```
loop on ALL_CONF_DIRS to create CONFIGURED_COMPONENTS

for adir in ALL_CONF_DIRS:
  append adir/docker-compose-extra.yml to COMPOSE_CONF_LIST
  for conf_dir in "$adir"/config/*:
    append adir/config/conf_dir/docker-compose-extra.yml to COMPOSE_CONF_LIST if conf_dir in CONFIGURED_COMPONENTS
```

This change, coupled with my suggested change in https://github.com/bird-house/birdhouse-deploy/pull/304#discussion_r1148424454, should hopefully result in this kind of load order:

```
COMPOSE_CONF_LIST=

# proxy should be the first one as it front everything else
-f ./config/proxy/docker-compose-extra.yml 

# all of magpie
-f ./config/magpie/docker-compose-extra.yml 
-f ./config/magpie/config/proxy/docker-compose-extra.yml 

# twitcher depends on magpie so loaded after magpie, all of twitcher loaded at the same time
-f ./config/twitcher/docker-compose-extra.yml
-f ./config/twitcher/config/proxy/docker-compose-extra.yml 

# all of weaver which modify magpie so should be loaded after magie (due to ordering in EXTRA_CONF_DIR since weaver do not directly depend on magpie)
-f ./components/weaver/docker-compose-extra.yml 
-f ./components/weaver/config/magpie/docker-compose-extra.yml 
-f ./components/weaver/config/proxy/docker-compose-extra.yml 
-f ./components/weaver/config/twitcher/docker-compose-extra.yml

# this last item in EXTRA_CONF_DIR will stay last
-f ../../daccs-env/docker-compose-extra.yml  
```

mishaschwartz:
@fmigneault @tlvu Ok I believe I've updated the load order as expected. See https://github.com/bird-house/birdhouse-deploy/pull/304/files#diff-d975bf659606195d2165918f93e1cf680ef68ea3c9cab994f033705fea8238b2R18-R95 for an explanation of the current behaviour

birdhouse/read-configs.include.sh
```diff
@@ -88,29 +88,38 @@ read_env_local() {
...
-          source_conf_files "$COMPONENT_DEPENDENCIES" "a dependency of $adir"
       fi
+      if echo "$ALL_CONF_DIRS" | grep -qE "^\s*$adir\s*$"; then
+          # check again in case a dependency has already added this to the ALL_CONF_DIRS variable
+          continue
+      fi
+      ALL_CONF_DIRS="$ALL_CONF_DIRS
+        $adir
+      "
+      COMPONENT_DEFAULT_ENV=$_OLD_COMPONENT_DEFAULT_ENV
```
tlvu, line 120: Say we have 3 or more levels of dependencies, example: magpie <== twitcher <== weaver <== cowbird (not 100% sure this dependency graph is actually true, just for illustration).

`_OLD_COMPONENT_DEFAULT_ENV` will be local to each invocation of `source_conf_files`?

birdhouse/read-configs.include.sh
```diff
@@ -88,29 +88,38 @@ read_env_local() {
...
           # The new adir with typo will not be active but at least all the existing
           # will still work.
           echo "WARNING: '$adir' in $conf_locations does not exist" 1>&2
       fi
+      _OLD_COMPONENT_DEFAULT_ENV=$COMPONENT_DEFAULT_ENV # a posix-compliant way to make this a local variable
       COMPONENT_DEFAULT_ENV="$adir/default.env"
       if [ -f "$COMPONENT_DEFAULT_ENV" ]; then
+          # Source config settings of dependencies first if they haven't been sourced previously.
+          # Note that this will also define the order that docker-compose-extra.yml files will be loaded.
+          source_conf_files "$(. "$COMPONENT_DEFAULT_ENV" && echo "$COMPONENT_DEPENDENCIES")" "a dependency of $adir"
```
tlvu, line 109: Woo fancy, sourcing and echo in same line.  This helps for variable scoping or just to make the code more compact?

birdhouse/read-configs.include.sh
```diff
@@ -88,29 +88,38 @@ read_env_local() {
 source_conf_files() {
   dirs=$1
   conf_locations=$2
+  _old_adir=$adir # a posix-compliant way to make this a local variable'
```
tlvu, line 91: `adir` is a very non specific var.  If it is used in another function before calling this one, would it mess this one up?

birdhouse/pavics-compose.sh
```diff
@@ -105,25 +105,28 @@ if [ x"$1" = x"up" ]; then
...
+  COMPOSE_CONF_LIST="${COMPOSE_CONF_LIST} ${previous_overrides}"
+  # Load overrides for other components unless the component to be overridden hasn't been loaded yet. If the component
+  # hasn't been loaded yet, store a reference to it so that it can be added in as soon as the component is loaded.
   for conf_dir in "$adir"/config/*; do
-    service_name=$(basename "$conf_dir")
+    override_service_name=$(basename "$conf_dir")
     extra_compose="$conf_dir/docker-compose-extra.yml"
     if [ -f "$extra_compose" ]; then
-      if echo "$CONFIGURED_COMPONENTS" | grep -q "$service_name"; then
+      if printf '%b' "${LOADED_COMPONENTS}" | grep -q "^$override_service_name$"; then
```
tlvu, line 126: Nice trick `print '%b'`

CHANGES.md
```diff
@@ -56,13 +56,44 @@
...
   - ./config/wps_outputs-volume (loaded as a dependency of generic_bird)
+  - ./optional-components/generic_bird
   - ./components/cowbird
 
+  This load order also applies to the order that docker-compose-extra.yml files are specified. If a component also
+  includes an override file for another component (eg: ./config/finch/config/proxy/docker-compose-extra.yml overrides 
+  ./config/proxy/docker-compose-extra.yml), the following additional load order rules apply:
+
+  - if the component that is being overridden has already been loaded, the override file is loaded immediately
+  - otherwise, the override files will be loaded immediately after the component that is being overridden has been loaded
```
tlvu, line 71: Very nice delayed load !  Components do not have to know the load ordering amount themselves.

tlvu (commented):
Nice load ordering fix.  Minor comments for extra robustness.

birdhouse/read-configs.include.sh
mishaschwartz: No, it isn't actually, this is one of the edge cases that the test helped uncover :smile: . A better solution is to use a stack and push/pop the adir to it.

birdhouse/read-configs.include.sh
mishaschwartz: It's for variable scoping. Since `$(...)` runs in a subshell then we can get the value of `$COMPONENT_DEPENDENCIES` without populating the environment of the main process with the variables sourced in `COMPONENT_DEFAULT_ENV` (since we want to source the dependencies first).

birdhouse/read-configs.include.sh
mishaschwartz: It shouldn't matter... but this has been changed anyway with the new stack implimentation

birdhouse/pavics-compose.sh
mishaschwartz: Thanks! I never remember how to use printf so I'm always looking it up

mishaschwartz:
@tlvu @fmigneault I know you guys are busy with other work but this PR is ready to go and I need at least 1 approved review to merge it in. If you're ok with me merging this in can I get an approval so that github will let me proceed. Thanks!

tlvu (approved):
LGTM, minor comments, sorry for the delay.

birdhouse/read-configs.include.sh
```diff
@@ -66,17 +66,19 @@ read_default_env() {
 read_env_local() {
     # we don't use usual .env filename, because docker-compose uses it
 
-    if [ -e "$COMPOSE_DIR/env.local" ]; then
+    echo "Using local environment file at: ${BIRDHOUSE_LOCAL_ENV:="$COMPOSE_DIR/env.local"}"
```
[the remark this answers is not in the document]
mishaschwartz: Good point. I intended it as a hidden feature but there's no reason why someone couldn't use it. I'll add some documentation!

birdhouse/README.rst
```diff
@@ -202,6 +202,18 @@ Starting and managing the lifecycle of the VM:
...
 
 
+Framework tests
+---------------
+
+Core features of the platform has tests to prevent regressions.
+
+To run the tests:
+
+.. code-block::
```
fmigneault, line 212: Add `.. code-block:: shell` for highlight in RST.

CHANGES.md
```diff
@@ -56,13 +56,50 @@
...
+    ...
+  # config/finch/config/proxy/docker-compose-extra.yml
+    ...
+  ```
+
+  the docker compose files will be loaded in the following order: 
+
+  - config/finch/docker-compose-extra.yml
+  - config/proxy/docker-compose-extra.yml
+  - config/finch/config/proxy/docker-compose-extra.yml
```
fmigneault, line 95: This seems to contradict what the test evaluates, or there is something I missed? 
I'm expecting `proxy` to be first as `finch` depends on it to provide its extra compose YAML.

tests/test_read_configs_include.py
```diff
@@ -0,0 +1,265 @@
...
+
+def get_read_config_stdout(proc: subprocess.CompletedProcess) -> str:
+    return proc.stdout.split(ENV_SPLIT_STR)[0]
+
+
+def get_command_stdout(proc: subprocess.CompletedProcess) -> str:
+    return proc.stdout.split(ENV_SPLIT_STR)[1]
+
+
+class TestReadConfigs:
```
fmigneault, line 40: Maybe only one test case missing for "other extras not within birdhouse".
They should be last, after components and optional-components, as final overrides.

fmigneault (commented):
Just docs to validate. Behaviour seems ok based on tests.

CHANGES.md
mishaschwartz: Finch doesn't have proxy as an explicit dependency (it is not included in the COMPONENT_DEPENDENCIES variable defined in config/finch/default.env) so proxy is not loaded before finch. Because of this, the order that the components appear in the DEFAULT_CONF_DIRS list is respected (finch is before proxy). However, the file that overrides proxy settings (config/finch/config/proxy/docker-compose-extra.yml) can't be loaded until _after_ both finch and proxy are loaded, since it could potentially override settings in either one, so it is loaded as soon as both of finch an proxy are loaded. 

From the CHANGES.md file, here is the relevant explanation:

>   This load order also applies to the order that docker-compose-extra.yml files are specified. If a component also
  includes an override file for another component (eg: ./config/finch/config/proxy/docker-compose-extra.yml overrides 
  ./config/proxy/docker-compose-extra.yml), the following additional load order rules apply:
>
>  - if the component that is being overridden has already been loaded, the override file is loaded immediately
>  - otherwise, the override files will be loaded immediately after the component that is being overridden has been loaded

tests/test_read_configs_include.py
mishaschwartz: Sure I can do that but I'm not sure what that test adds? Anything in EXTRA_CONF_DIRS is treated the same regardless of whether its "in birdhouse" or a separate component. So it will be last if it's last in the list, there is no special treatment other than that.

mishaschwartz: I added an additional test that just makes sure we can include components in some other arbitrary location

CHANGES.md
fmigneault: I see. So in the tests, `proxy` is loaded before `finch` because of some other component that defined it needed `proxy` as dependency, but considering only `proxy`/`finch` by themselves in this example, they have no particular order. Only the extra should be last, as it needs both.

fmigneault (approved):

mishaschwartz:
@fmigneault yes that is a good way to think about it. In this case the example with just proxy and finch was meant to be a minimal example for explanatory purposes. I don't think that you'd ever have a case where only finch and proxy are running on your stack :shrug:

fmigneault:
> I don't think that you'd ever have a case where only finch and proxy are running on your stack

I don't think anyone would do that, but actually thinking about it, the `proxy` would probably be a required dependency of `finch` such that it becomes accessible. Every web-service should probably have `proxy` as a dependency.

mishaschwartz:
@fmigneault 

> Every web-service should probably have proxy as a dependency.

Yes that will eventually be the case where there will be some components that are always required (at least in production) and proxy will be one of them, that's for the next PR though.

PR #306, opened 2023-03-26
Add tests for component load order

tlvu:
This PR should merge to @mishaschwartz PR https://github.com/bird-house/birdhouse-deploy/pull/304.

Proper component load order is a core feature allowing the user to customize all default behavior of the stack.  This is worth to have tests to prevent accidental regressions.

Currently I made the tests passed with the current wrong load order that should be fixed.

Running the tests look like this:
```
$ ./bash_unit test_read_configs.sh
Running tests in test_read_configs.sh
        Running test_simple_all_conf_dirs ... SUCCESS ✓
        Running test_simple_compose_conf_list ... SUCCESS ✓
Overall result: SUCCESS ✓
```

The test framework is from https://github.com/pgrange/bash_unit/blob/master/bash_unit.

I also looked at https://github.com/bats-core/bats-core, which seems more popular but also more complex to use and setup.  bash_unit is one single file to add while bats-core requires multiple git submodules!  KISS for the win.

In order to test the `COMPOSE_CONF_LIST`, I had to move its logic to `read-configs.include.inc`.  Probably the wrong location and a better location is a brand new `.include.inc` file.  Name suggestion?

I also updated the "Release Instructions" in the README to use `make bump major|minor|patch` command instead of directly calling `bump2version` to harmonize with the section "Tagging policy" right above.


birdhouse/tests/test_read_configs.sh
```diff
@@ -0,0 +1,189 @@
...
+          
+            ./config/phoenix
+          
+            ./config/malleefowl
+          
+            ./config/postgres
+          
+            ./config/wps_outputs-volume
+          
+            ./config/data-volume
```
fmigneault, line 57: Should be before the proxy and the birds since they depend on these?

fmigneault (commented):
I'd suggest defining a requirements file with `pytest` + `pytest-shell-utilities`.
Then, it is easier to extend advanced tests and checks with everything `pytest` supports natively (filtering by markers, collect, report statistics, output, colors, etc.), without worrying about parsing `sh` specific calls. It also allows us to combine test suites in a language agnostic manner (run python test + bash tests at the same time). 

```python
import os
import pytest
from pytestshellutils.shell import Subprocess

CUR_DIR = os.path.abspath(os.path.dirname(__file__))
ROOT_DIR = os.path.dirname(os.path.dirname(CUR_DIR))


@pytest.mark.functional
@pytest.mark.shell
def test_make_config_disabled(shell: Subprocess) -> None:
    env = {"SOME_VAR": ""}
    ret = shell.run("run-some-script.sh", "[param1]", env=env, cwd=ROOT_DIR)
    assert ret.returncode == 0
    assert ret.stderr == ""
    assert "expected output" in ret.stdout
    lines = ret.stdout.splitlines()
    assert len(lines) == 123
    assert lines[-1] == "expected last line"
```


Consider adding the test suite in some `.github/workflows/tests.yml` definition to auto-validate it each time.

birdhouse/tests/test_read_configs.sh
tlvu: Yes that should hopefully be fixed if my suggested algo change does what I think it does.

tlvu:
> I'd suggest defining a requirements file with `pytest` + `pytest-shell-utilities`.
> Then, it is easier to extend advanced tests and checks with everything `pytest` supports natively (filtering by markers, collect, report statistics, output, colors, etc.), without worrying about parsing `sh` specific calls. It also allows us to combine test suites in a language agnostic manner (run python test + bash tests at the same time).

@fmigneault 
Can that pytest-shell-utilities test shell function separately? I basically only unit test the new function `source_conf_files` and `gen_compose_conf_list`. 

Note we do not want to test the entire `pavics-compose.sh` script because that would be too heavy (it spins up the entire PAVICS stack).  That's what Jenkins is doing.  We just want to unit test some core functions to quickly catch regressions.  So we just want something lightweight without introducing new requirements (python).  Or are you thinking to add python code to the platform code, that is entirely shell script right now?

I see "platform code" as `pavics-compose.sh`, and `read-configs.include.sh` for the moment.  All the rest are just plugins into multiple hooks that the platform override feature allows.

fmigneault:
@tlvu  
Yes, each test can run its own `shell.run("run-some-script.sh" ...`, and you can also combine them with `pytest.mark.parametrize` to run multiple variations of a same test. 

I agree we should focus on minimal tests without booting the stack here. We can add more or less items to that test suite as we encounter new problems or add new features.

tlvu:
> Yes, each test can run its own `shell.run("run-some-script.sh" ...`, and you can also combine them with `pytest.mark.parametrize` to run multiple variations of a same test.

@fmigneault 
Just to be clear are you suggesting replacing `bash_unit` with `pytest` or have `pytest` call `bash_unit`?



> I agree we should focus on minimal tests without booting the stack here. We can add more or less items to that test suite as we encounter new problems or add new features.

`bash_unit` is very simple and does everything we need at this moment.  I suggest we adopt it to quickly moving onto other issues and change it later if needed.

I still have an issue, I had to factor out the `gen_compose_conf_list` function. I put it in `read-configs.include.sh` for now but it does not really belongs there.  How about a new `platform-utils.include.sh`?

@mishaschwartz anything you would like to change before merging this into your PR and we will have tests for your PR.

mishaschwartz:
@tlvu @fmigneault 

I agree with @fmigneault that pytest is the better option and should replace bash_unit. 

I have taken your the tests from this PR and I'm adapting them into pytests which I will include in #304. I'm mostly done but I won't have time to finish up the tests fully by the end of the day. I'll likely be pushing the changes to the load-components-in-defined-order branch tomorrow morning.

PR #300, opened 2023-03-06
Add a zenodo configuration

Zeitsperre:
## Overview

Please include a summary of the changes and which issues are fixed. 

Please also include relevant motivation and context. 

List any dependencies that are required for this change.

## Changes

**Non-breaking changes**
- Adds a Zenodo configuration 

## Related Issue / Discussion

- Resolves [#299](https://github.com/bird-house/birdhouse-deploy/issues/299) 

## Additional Information

 - ~Zenodo has not yet been enabled on the repository.~ Enabled
 - This project also does not contain a License file, so I have not added any license or access rights to the configuration.

PR #310, opened 2023-04-10
Canarie-api should not be a mandatory component

mishaschwartz:
## Overview

Canarie-api is currently deployed in the same container as the nginx reverse proxy
service meaning that it is not possible to deploy nginx without including canarie-api.

This means that it is currently not possible to run this deployment without canarie-api
or use a different monitoring application. This change fully separates the configuration
for canarie-api and nginx so that a user can choose to run nginx with or without canarie-api.

Canarie-api has been kept on the DEFAULT_CONF_DIRS list so that canarie-api is included by
default, for backwards-compatibility. In order to run nginx without canarie-api, remove the
`./conf/canarie-api` line from the DEFAULT_CONF_DIRS environment variable. 

A user can also choose a specific version of the nginx docker image to use by specifying 
the PROXY_IMAGE environment variable (default is "nginx"). Note that if canarie-api is used
(by including the `./conf/canarie-api` line in DEFAULT_CONF_DIRS), then the PROXY_IMAGE 
variable will be ignored.

## Changes

**Non-breaking changes**
Makes it possible to run the stack without canarie-api

**Breaking changes**
None

## Related Issue / Discussion

Related to the various "pluggable component" issues. This makes canarie-api a "pluggable component".


tlvu (approved):
Totally approve the spirit of this PR.  However I probably will not be able to perform a proper review of this PR and your 2 other PR as well this week.  I have to give a hand for the paper David is alluring about in this comment https://github.com/bird-house/birdhouse-deploy/pull/300#issuecomment-1499650688

However I do not want to hold you back either.  So I propose that as long as your PR is back-compat (no manual intervention required to roll-out, to clean up after roll-out, to ensure subsequent autodeploy works) and Francis is okay with it, then go ahead and merge.

If later we find some problems, ex: the component load order changed and it's a regression, we can simply just fix it in a subsequent PR, just like the component load order.  There is no requirement a PR is perfect on first try.  Software have bugs all the time.

birdhouse/config/proxy/default.env
```diff
@@ -1,7 +1,6 @@
 # All env in this default.env can be overridden by env.local.
 
-# Folder inside "proxy" container to drop extra monitoring config
-export CANARIE_MONITORING_EXTRA_CONF_DIR="/conf.d"
+export PROXY_IMAGE="nginx"
```
tlvu, line 3: I would rather pin the specific latest version at this point in time.  Because `latest` version is not reproducible.  It changes on each build.  

Reproducibility is very important if we want to be able to diagnose problem across multiple different deployments of PAVICS.

For a given PAVICS tag, we should know precisely which version of the binary is being deployed.  We do allow image override via `env.local` but that's mostly for testing or the staging env.  For production deployment, I think and hope most users will want to stick to the tested versions coming from the stack.

birdhouse/config/canarie-api/default.env
```diff
@@ -0,0 +1,11 @@
+# Folder inside "proxy" container to drop extra monitoring config
+export CANARIE_MONITORING_EXTRA_CONF_DIR="/conf.d"
+
+export PROXY_IMAGE="pavics/canarieapi:0.3.5"
```
tlvu, line 4: Thanks for keeping this old behavior for backward compat.

tlvu (commented):
Quick review, just minor comment.

birdhouse/config/proxy/default.env
mishaschwartz: Sure and I agree. This is literally just a copy from https://github.com/Ouranosinc/CanarieAPI/blob/master/Dockerfile#L1 to maintain the existing settings but I'll explicitly pin it to whatever is the latest going forward.

mishaschwartz:
Thanks @tlvu I appreciate it. Does that apply to #304 as well since I would like to pull that in so that I can apply my changes on top of that as well?

fmigneault:
@mishaschwartz 
Can you backport any relevant changes from this PR into #284?

fmigneault:
@mishaschwartz It only needs some final validation to confirm that the monitoring is working correctly, but I did not get the chance to invest more time into it lately.

birdhouse/config/canarie-api/.gitignore
```diff
@@ -1,2 +1,5 @@
 docker_configuration.py
-conf.extra-service.d/canarie-api.conf
```
tlvu, line 2: @mishaschwartz Our staging env found this file became "new" file for git and broke the subsequent autodeploy !

Could you re-add this file to .gitignore?

mishaschwartz: @tlvu I've added a PR #314 . Is this what is breaking the Jenkins build as well do you think or is that unrelated?

tlvu: Completely un-related to Jenkins.  I merged https://github.com/bird-house/birdhouse-deploy/pull/308 yesterday and today I went onto our staging machine and see it still did not autodeploy that merge.  So I checked the logs and found out the reason.

PR #317, opened 2023-04-17
Config var `PAVICS_FQDN_PUBLIC` not usable in component `default.env` and external scripts

tlvu:
Currently, `PAVICS_FQDN_PUBLIC` is only usable in `.template` files, in `docker-compose-extra.yml` files and in component pre/post compose scripts because they are handled by `pavics-compose.sh`.

It was good enough but now with delayed eval feature, we can do better. `PAVICS_FQDN_PUBLIC` can be as accessible as the other `PAVICS_FQDN` var.

Both vars allow a host to have a different public and internal hostname. Some scripts, `certbotwrapper` for example, prefer the public hostname than the internal hostname when they are different because Let's Encrypt only knows about the public hostname.

With the pluggable nature of this stack, we can have many external scripts from many external repos reading the config vars and they can have the need to specifically access the public hostname.

Bonus, we now have a sample use of `DELAYED_EVAL` list, right in the main `default.env`.


birdhouse/default.env
```diff
@@ -15,7 +15,21 @@ export DATA_PERSIST_ROOT="/data"
...
+# Ensure DELAYED_EVAL is initially empty.
+# Only this base default.env is allowed to reset this variable.  All component
+# default.env should only append to DELAYED_EVAL and not reset it like this.
+DELAYED_EVAL=''
+
+# Append to DELAYED_EVAL list.
+export DELAYED_EVAL="
+  $DELAYED_EVAL
+  PAVICS_FQDN_PUBLIC
+"
```
fmigneault, line 32: I think a test would be in order here since it is not the "usual" `default.env` within a component/config sub-dir.

birdhouse/default.env
```diff
@@ -15,7 +15,21 @@ export DATA_PERSIST_ROOT="/data"
...
-export DELAYED_EVAL=''
+# Allow different public and internal hostname.
+# Default to PAVICS_FQDN, which must be set in env.local.
+# Must use single-quote for delayed eval.
+export PAVICS_FQDN_PUBLIC='${PAVICS_FQDN}'
+
+# Ensure DELAYED_EVAL is initially empty.
+# Only this base default.env is allowed to reset this variable.  All component
+# default.env should only append to DELAYED_EVAL and not reset it like this.
+DELAYED_EVAL=''
```
fmigneault, line 26: Is there a way to avoid this, like setting it empty within `pavics-compose` before sourcing anything else, to avoid instructions in the form "*don't do this except here*"?

mishaschwartz: I don't think setting it as initially empty is useful actually. $DELAYED_EVAL is initially unset so explicitly setting it to the empty string doesn't add anything. 

Also, we have a comment at the top of this file that describes how DELAYED_EVAL should be used. I don't think it is necessary to elaborate on it here.

mishaschwartz (approved):
Looks good! I don't have anything additional to add to @fmigneault's comments. Overall I think this is a good idea. So I'll approve it for now, pending the changes requested by @fmigneault.

birdhouse/default.env
tlvu: Well before Misha's refactoring into components, delayed eval was used directly in the main `default.env` !

But maybe I should get acquainted to add test to Misha testsuite.

birdhouse/default.env
tlvu: Maybe my "reset" wording is confusing you guys, sorry.  I actually want to "initialize before use".  Will remove this "reset" wording.

If you guys noticed, almost all other vars in any `default.env` are setting to some default value, but this `DELAYED_EVAL` var is meant to be appended so it has to be initialized somewhere.  

The main `default.env` is the best place to initialize since it's the first config file being read.  

The initialization can not be in `pavics-compose.sh` since it means other external scripts sourcing the config files will not have the proper initialization.

tlvu:
> I don't think setting it as initially empty is useful actually. $DELAYED_EVAL is initially unset so explicitly setting it to the empty string doesn't add anything.

See my "initialize before use" comment https://github.com/bird-house/birdhouse-deploy/pull/317#discussion_r1170199132

> Also, we have a comment at the top of this file that describes how DELAYED_EVAL should be used. I don't think it is necessary to elaborate on it here.

I'd rather be safe than sorry.  Personally I sometime go to fast and forget to read comments that are far from the code snippet that interest me.  So to defend against someone with the same bad habit as me, I'd rather provide directly good code they can copy/paste directly if they need to extend `DELAYED_EVAL`.

birdhouse/default.env
fmigneault: >The main `default.env` is the best place to initialize since it's the first config file being read.

I would agree if it was something meant to be overridden by users, like other configuration settings, but I think `DELAYED_EVAL` is more of a control variable used by `pavics-compose` than a setting. 

Can it be set in `read_default_env` instead, or anywhere that is sourced by everything before including all the other components? 
https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/read-configs.include.sh#L57-L63

tlvu: Ah that's a much better suggestion, will put the initialization in `read_default_env`.

tlvu:
@fmigneault relocated the initialization, added test.

fmigneault (approved):

PR #318, opened 2023-04-18
fix flaky WPS provider responses causing failure during their registration in weaver

fmigneault:
## Overview

Fix flaky WPS provider responses (i.e.: other WPS birds) causing failure during their registration in `weaver`.

In some cases, the WPS birds would not respond properly when starting the stack, either because they are still initiating or due to other temporary failures such as services being restarted until healthy. This fix introduces a retry mechanism to attempt WPS registration in `weaver` up to `WEAVER_WPS_PROVIDERS_RETRY_COUNT=5` times (1 initial attempt + 5 retries), and with `WEAVER_WPS_PROVIDERS_RETRY_AFTER=5` second intervals between each retry. If the maximum number of retries for any WPS provider or the `WEAVER_WPS_PROVIDERS_MAX_TIME` across all registrations are reached, the operation is aborted.


fmigneault:
run tests

fmigneault:
run tests

birdhouse/components/weaver/post-docker-compose-up
```diff
@@ -268,37 +283,53 @@ for prov in ${WEAVER_WPS_PROVIDERS}; do
...
+        body=$(echo "${resp}" | head -n -1)
+        if [ ${ret} -ne 0 ] || [ "${code}" -ne 201 ]; then
+            printf "\n%s\n" "${WARN}Failed registration of remote WPS provider [${prov}] on [${prov_url}]${retry_msg}."
+            printf "Error:\n%s\n" "${body}"
+            if [ ${retry} -gt ${total} ]; then
+              echo "${ERROR}Maximum retry attempts ${total} reached for WPS provider [${prov}]. Aborting."
+              reset_state
+              exit 23
+            fi
+            echo "${WARN}Will retry after ${WEAVER_WPS_PROVIDERS_RETRY_AFTER}s..."
```
mishaschwartz, line 325: For messages displayed to the user, I think it's better to actually use the word "seconds". As in:

```
Will retry after 5 seconds
```

instead of 

```
Will retry after 5s
```

birdhouse/components/weaver/post-docker-compose-up
```diff
@@ -81,26 +78,47 @@ else
...
+echo "${PREFIX}Will retry registration of each provider up to ${WEAVER_WPS_PROVIDERS_RETRY_COUNT} times"
+echo "${PREFIX}Will retry registration of each provider with ${WEAVER_WPS_PROVIDERS_RETRY_AFTER}s intervals"
 
 if [ -z "$WEAVER_CURL_IMAGE" ]; then
     WEAVER_CURL_IMAGE="curlimages/curl:7.87.0"
 fi
 
 # POSIX portable RNG if RANDOM does not exist on the current shell
-RANDOM_NUMBER=${RANDOM:-$(tr -dc 0-9 < /dev/urandom | head -c 5)}
+RANDOM_NUMBER=${RANDOM:-$(tr -dc 0-9 < /dev/urandom | (head -c 5 ; dd of=/dev/null))}
```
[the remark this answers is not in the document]
fmigneault: It didn't hang when running via the CI, but seems to do when executed only in the terminal. 
I'll try simply redirecting stderr to null instead. The original code worked fine, but only printed the broken pipe error.

birdhouse/components/weaver/post-docker-compose-up
fmigneault: I'll leave it as is because it aligns with all other outputs reporting seconds and keeps the message more succinct.
It's only for debugging anyway.

fmigneault:
run tests

tlvu (approved):
LGTM.  Please mention in the PR description and CHANGES that the fix to flaky WPS response is to introduce a retry mechanismn.  Max number of retries and retries duration are also introduced to avoid retrying forever.

fmigneault:
@tlvu 
see #321 replacing this PR.
Same content, but different base, so I don't integrate #284 changes yet.

PR #319, opened 2023-04-19
output COMPOSE_CONF_LIST in a readable format

fmigneault:
## Overview

Prints the activated compose file list line-by-line such that it can be more easily readable. 

## Changes

**Non-breaking changes**
- Prints the activated compose file list line-by-line such that it can be more easily readable. 

**Breaking changes**
- n/a

## Related Issue / Discussion

- n/a


tlvu (approved):
Great idea!  Can you just put a sample of "before" and "after" in the PR description and also in the CHANGES?

PR #321, opened 2023-04-19
Fix flaky WPS provider responses (i.e.: other WPS birds) causing failure during their registration in `weaver`

fmigneault:
## Overview

Fix flaky WPS provider responses (i.e.: other WPS birds) causing failure during their registration in `weaver`.

## Changes

**Non-breaking changes**
- Fix flaky WPS provider responses (i.e.: other WPS birds) causing failure during their registration in `weaver`.

  In some cases, the WPS birds would not respond properly when starting the stack, either because they are still
  initiating or due to other temporary failures such as services being restarted until healthy. This fix introduces 
  a retry mechanism to attempt WPS registration in `weaver` up to `WEAVER_WPS_PROVIDERS_RETRY_COUNT=5` times 
  (1 initial attempt + 5 retries), and with `WEAVER_WPS_PROVIDERS_RETRY_AFTER=5` second intervals between each retry.
  If the maximum number of retries for any WPS provider or the `WEAVER_WPS_PROVIDERS_MAX_TIME` across all registrations
  are reached, the operation is aborted.

**Breaking changes**
- n/a

## Related Issue / Discussion

- Close #318 (using `fix-canarieapi` as base branch) instead of `master` (this PR)


tlvu (approved):
+1

Issue #120, opened 2021-01-29
[Feature] Add a bird that allows for loose coupling between components while keeping them in synch


dbyrns:
This use case became obvious with the fact that a lot of components needs to be aware and configured when a new user is created.
Magpie handles those users, but we don't want it to be tighly-coupled with other components.
Magpie has also the ability to notify whoever registers a hook, but we don't want other components to be tighly-coupled to the Magpie hook neither.
Here comes Cowbird (a bird that lays in the nests of other birds), who will be able to register a hook in Magpie and do whatever is necessary for all the components in needs when it is notify.  
With this paradigm, we can do a lot of things : 

- When a user is created, workspace will be created in geoserver, thredds and jupyterhub
- When a user is deleted a cleanup will be performed
- When new data is added (via the ingestion service, a process output or a user notebook), a monitoring sub-system will be able to update components in needs.
- For ressources being served by more than one services, a permissions synchronizer will be able to keep them aligned across services.

Here we got a preliminary design : 
![Cowbird Components 1 2 0](https://user-images.githubusercontent.com/4202336/106329921-4e918f00-6250-11eb-99b3-202fde0c38cb.png)

dbyrns:
Hi @cehbrecht, we want to share with you our plan to add another bird in the PAVICS stack to help us synchronize states between components. Any comments would be welcomed.

fmigneault:
for reference: 
https://github.com/Ouranosinc/cowbird/pull/1

PR #323, opened 2023-04-20
Update cowbird config with permission webhook cases 

ChaamC:
PR to add support to the new cases added to `Cowbird`, relates to [cowbird PR #35](https://github.com/Ouranosinc/cowbird/pull/35).

## Changes

- Add Magpie webhook definitions for permission creation and deletion cases to be processed by Cowbird.

## Fixes

- Fix Cowbird's `sync_permissions` config which used invalid Magpie service types.
- Fixes https://github.com/bird-house/birdhouse-deploy/issues/120


birdhouse/components/cowbird/config/cowbird/config.yml.template
```diff
@@ -107,7 +107,7 @@ sync_permissions:
       - "geoserver_workspace : createStoredQuery <-> thredds_workspace : write"
   weaver_outputs:
     services:
-      weaver:
+      api:
```
fmigneault, line 110: This definition doesn't match the example https://github.com/Ouranosinc/cowbird/blob/master/config/config.example.yml
The example does not have the repeated
```yaml 
 - name: weaver
   type: service
```
for each resource to sync as defined below.

We can also see in https://github.com/Ouranosinc/cowbird/blob/master/config/config.example.yml#L58 that "service name" is used, which brings more confusion. It should be service types instead?

Is the example out of date?

ChaamC: You're right. It should indeed be "service type", not "service name".
I will add a commit here and to my other PR to review this and make it more explicit that it is a service type.

fmigneault (approved):

fmigneault:
@ChaamC You can either merge right away with bump versioning, or wait after https://github.com/Ouranosinc/cowbird/pull/35 and do it all in one. As you prefer.

birdhouse/components/cowbird/config/cowbird/config.yml.template
```diff
@@ -153,8 +153,7 @@ sync_permissions:
             type: route
           - name: "{outputID}"
             type: route
-      # see 'optional-components/secure-data-proxy' for more details on protected WPS-outputs
-      wps_outputs:
```
tlvu, line 157: Lost the key "wps_oputputs", expected?

tlvu (approved):
LGTM as I do not really know the inner working between cowbird and magpie.

birdhouse/components/cowbird/config/cowbird/config.yml.template
ChaamC: Yep, the `wps_outputs` key here and `weaver` key above are supposed to be Magpie service types, but they are not actual service types. Instead, all their contained values now use the key `api` which is a valid Magpie service type.

mishaschwartz: In fact, without this change, all cowbird hooks fail with:

```
cowbird.config.ConfigErrorInvalidServiceKey: Service `weaver` used in sync config is not valid since it was not found in Magpie services (['access', 'api', 'geoserver', 'geoserverwfs', 'geoserverwms', 'geoserverwps', 'ncwms', 'thredds', 'wfs', 'wps']).
```

so I think this change is necessary to ensure that cowbird works properly with this default config file.

If we're still waiting on https://github.com/Ouranosinc/cowbird/pull/35 to merge in this PR can we at least get the bug fix merged in the meantime. I've factored out that portion of this PR here: #336

ChaamC: The cowbird PR is still a work in progress, so I was still waiting to merge this PR.

We can add the necessary part from your PR for now, to avoid the error, and I will add the rest when the other PR is done.

CHANGES.md
```diff
@@ -63,8 +63,14 @@
...
   *  Add birdhouse community
 
+- Add Magpie webhook definitions for permission creation and deletion cases to be processed by Cowbird.
+- Add `USER_WORKSPACE_UID` and `USER_WORKSPACE_GID` env variables to manage ownership of the user workspaces used by
+  Cowbird, JupyterHub and others.
 - Licence: update copyright line with year and ownership
 
+## Fixes
+
+- Fix Cowbird's `sync_permissions` config which used invalid Magpie service types.
```
fmigneault, line 73: @ChaamC 
Will need to be moved up due to other releases since then.

CHANGES.md
```diff
@@ -63,8 +63,14 @@
   *  Add Misha to creators
   *  Add birdhouse community
 
+- Add Magpie webhook definitions for permission creation and deletion cases to be processed by Cowbird.
+- Add `USER_WORKSPACE_UID` and `USER_WORKSPACE_GID` env variables to manage ownership of the user workspaces used by
+  Cowbird, JupyterHub and others.
```
fmigneault, line 68: Need to add Magpie version update from 3.31.0 -> 3.34.0. 
See other entries in the changes for the format pointing as release tags.

CHANGES.md
```diff
@@ -174,15 +181,8 @@
...
   *  Add birdhouse community
 
-- Add Magpie webhook definitions for permission creation and deletion cases to be processed by Cowbird.
-- Add `USER_WORKSPACE_UID` and `USER_WORKSPACE_GID` env variables to manage ownership of the user workspaces used by
-  Cowbird, JupyterHub and others.
 - Licence: update copyright line with year and ownership
 
-## Fixes
-
-- Fix Cowbird's `sync_permissions` config which used invalid Magpie service types.
```
fmigneault: This was not transferred in the latest section?

fmigneault (commented):
Code-wise, everything looks fine. 
Need to get the test outputs posted on the PR to validate that everything works. 
Maybe latest run failed due to `cowbird==1.2.0` docker not yet released (not finished building)?

ChaamC:
I was able to download the image `cowbird==1.2.0` before pushing the version bump here, so the image should already be available. The e2e failure seemed to be occuring in other latest PRs and not only in this one.

fmigneault:
run tests

fmigneault:
🚀 🎉

PR #324, opened 2023-05-01
Globally added X-Robots-Tag: noindex HTTP response header

perronld:
## Overview

Search engines currently indexes dev/staging installations of Birdhouse. This may create SEO issues for organizations.

## Changes

- Add ``optional-components/x-robots-tags-header`` and ``X_ROBOTS_TAGS_HEADER`` variable to allow setting the desired
  header value server-wide.

**Breaking changes**
- X-Robots-Tag: noindex response header is globally added to nginx. This changes is breaking only if you wish to be indexed by search engines.


fmigneault (changes requested):
To make this non-breaking for everyone, it should be defined as an [optional-component](https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse/optional-components).
In this component, a nginx conf would be added and mounted in the `proxy` docker-compose service, such that in can be loaded by the `include /etc/nginx/conf.d/*.conf;` step. 
The instances that want to make use of this component will have to add it explicitly in `EXTRA_CONF_DIRS` of their `env.local` file.

perronld:
Before working further on this PR, lets see first if anyone actually wish to be indexed by public search engines.

huard:
Not sure I understand 100%, but I don't see why we'd want the staging and dev instances to be indexed.

fmigneault:
For staging and dev instances, I agree that indexing is most probably irrelevant. However, this definition can also be used for prod. So we must allow toggling the option if any indexing should be permitted for any use case.

perronld:
I think that the indexing is a problem even in prod.
In our case, Googlebot crawled Geoserver and Thredds (including wps_outputs).

fmigneault:
Maybe in the case of CRIM's instances, even prod indexing is not relevant because of our use cases, but other organizations could be taking advantage of it. For example, https://pavics.ouranos.ca/ entrypoint has a lot of documentation that indexes could facilitate search/matches. It could be the same eventually for U-Toronto that intends to swap the Jupyter Notebook interface by their own. For this reason, it should be an "opt-in" feature.

tlvu:
I think for Ouranos, we want the homepage of PAVICS to be indexed so we can be discovered.

Not sure I want all the data in Thredds and Geoserver to be indexed though.   And no indexing for staging and test servers of course.

Is there a way to exclude some path from being indexed?

@huard @tlogan2000 Any problems to have our data on Thredds and Geoserver being indexed by Google?

tlvu:
> In our case, Googlebot crawled Geoserver and Thredds (including wps_outputs).

Oh geez !  Which logs and what kind of log pattern?  I'd like to check if we've been crawled too.

tlogan2000:
> Any problems to have our data on Thredds and Geoserver being indexed by Google?

I think I would prefer only the homepage as you suggest but have no real idea how much trouble it is to avoid having indexing on geoserver / thredds.   Aany ballpark estimate in terms of amount of effort involved?

fmigneault:
Isn't there support for a `robots.txt` file at the root of the website as well? With this, you can normally have specific paths you want included or excluded for crawling, and by which bots.

huard:
Note the existence of https://developers.google.com/search/docs/appearance/structured-data/dataset

So one option would be to turn off THREDDS indexing, but add formatted metadata information to our landing page describing the datasets we want to be indexed.

fmigneault:
There's a lot of overlap between the JSON-LD Dataset definition and the STAC definitions. 
Maybe the JSON-LD Dataset could be generated by an output plugin with https://github.com/cedadev/stac-generator while the THREDDS contents are generated for STAC collections themselves. However, we will have to figure out how to inject the resulting JSON-LD definitions in the HTML `<head>` of each relevant dataset...

tlvu:
> So, for example, in the location block of the homepage, all we have to do is to add `add_header X-Dummy-Header asdf` to cancel the X-Robots-Tag that I added in the _http_ block.

The homepage is immediately at the root so I think overriding the root with `add_header X-Dummy-Header asdf` will not work.

However, `/thredds` and `/geoserver` are under their own location directives so something like this might work?  See examples from https://yoast.com/x-robots-tag/
```
location ~* \.(doc|pdf)$ {
    add_header  X-Robots-Tag "noindex, noarchive, nosnippet";
}
```

This means we should not forget to add this header for each new location directive we add to the proxy.  Not so ideal but not sure what are other solutions since the homepage is at the root.

Another solution is move the homepage to `/homepage` and the root will simply redirect to `/homepage` by default.

In all cases, I think this X-Robots-Tag should probably be optional and opt-in, unless we all agree that allow crawling of Thredds and Geoserver is a bad idea.

perronld:
Unfortunately, I don't know how the homepage is configured, but even if it's at the root, in should be within a `location / { ... }`  block

perronld:
I added the X-Robots-Tag header in the _http_ block, which is parent to all _location_ blocks, so if any _location_ block has the add_header directive, then if will cancel the one from the _http_ block.
So, the fix for the homepage would be to set the variable like this:
`export PROXY_ROOT_LOCATION="alias /data/homepage/; add_header X-Dummy-Header asdf;"`

tlvu:
> I added the X-Robots-Tag header in the _http_ block, which is parent to all _location_ blocks, so if any _location_ block has the add_header directive, then if will cancel the one from the _http_ block.
> So, the fix for the homepage would be to set the variable like this:
> `export PROXY_ROOT_LOCATION="alias /data/homepage/; add_header X-Dummy-Header asdf;"`

That seems to work !

@perronld Please document this opt-out for the homepage and describe this change in the CHANGES.md file (Usually some form of copy/paste of the PR description).

@mishaschwartz @eyvorchuk any concern from UofT and PCIC side, about this globally denying seach engine indexing of PAVICS instances?  Note an opt-out exists for the homepage.

Before:

```
$ curl --include --silent https://boreas.ouranos.ca/geoserver/ | head -20                                  
HTTP/1.1 302 
Server: nginx/1.23.4
Date: Wed, 03 May 2023 18:33:00 GMT
Content-Length: 0
Connection: keep-alive
X-Frame-Options: SAMEORIGIN
Set-Cookie: JSESSIONID=A34F9B545B03CBAC2282CCE2025C5EC4; Path=/geoserver; HttpOnly
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: Access-Control-Allow-Origin,Access-Control-Allow-Credentials
Location: /geoserver/index.html


$ curl --include --silent https://boreas.ouranos.ca/index.html | head -20
HTTP/1.1 200 OK
Server: nginx/1.23.4
Date: Wed, 03 May 2023 18:31:05 GMT
Content-Type: text/html
Content-Length: 1475
Last-Modified: Fri, 10 Mar 2023 17:17:47 GMT
Connection: keep-alive
ETag: "640b663b-5c3"
X-Robots-Tags: noindex
Accept-Ranges: bytes

<!doctype html>
<html>
  <head>
```

After:
```
$ curl --include --silent https://boreas.ouranos.ca/geoserver/ | head -20                                                                                            
HTTP/1.1 302 
Server: nginx/1.23.4
Date: Wed, 03 May 2023 18:29:23 GMT
Content-Length: 0
Connection: keep-alive
X-Frame-Options: SAMEORIGIN
Set-Cookie: JSESSIONID=74625D96DA69A90A4A3CBE082A834CE4; Path=/geoserver; HttpOnly
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: Access-Control-Allow-Origin,Access-Control-Allow-Credentials
Location: /geoserver/index.html
X-Robots-Tags: noindex

$ curl --include --silent https://boreas.ouranos.ca/index.html | head -20                                                                                            
HTTP/1.1 200 OK
Server: nginx/1.23.4
Date: Wed, 03 May 2023 18:29:42 GMT
Content-Type: text/html
Content-Length: 1475
Last-Modified: Fri, 10 Mar 2023 17:17:47 GMT
Connection: keep-alive
ETag: "640b663b-5c3"
X-Dummy-Header: asdf
Accept-Ranges: bytes

<!doctype html>
<html>
  <head>
```

tlvu:
@fmigneault does this Global X-Robots-Tag impact later implementation of https://github.com/bird-house/birdhouse-deploy/issues/325 ?  We want to advertize data on Thredds the various web crawlers/indexers?

fmigneault:
> if any _location_ block has the add_header directive, then if will cancel the one from the _http_ block.
> So, the fix for the homepage would be to set the variable like this:
`export PROXY_ROOT_LOCATION="alias /data/homepage/; add_header X-Dummy-Header asdf;"`

> does this Global X-Robots-Tag impact later implementation of https://github.com/bird-house/birdhouse-deploy/issues/325 ? We want to advertize data on Thredds to various web crawlers/indexers?

Each `location` block would need a specific variable, similar to `PROXY_ROOT_LOCATION`, to allow per-endpoint overrides.
I don't think this solution is quite the right one. The `X-Robots-Tags: noindex` is still present in the `/geoserver/` response, so crawlers are not available for any other endpoint other than the entrypoint. If we wanted to allow *also* `/thredds/`, there is no way to allow both `/` and `/thredds/`, and keep others like `/geoserver/` still with disallowed indexing.

Not sure why the example *before* case of `/geoserver/` did not have the `X-Robots-Tags: noindex`? Seems like this produces the opposite of what was expected (allow crawl only on the entrypoint, and not anywhere else).

I don't really like the solution to inject a random header to cancel another one. Seems like a very obscure workaround that most won't understand intuitively. Seems like it would be easier to only add `X-Robots-Tags: noindex` where we want them instead.

mishaschwartz:
> I don't really like the solution to inject a random header to cancel another one

What if we used the default `all` value to override subpaths instead of a random string like `asdf`

The all value is documented here: https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#directives

fmigneault:
> What if we used the default `all` value to override subpaths instead of a random string like `asdf`

That makes more sense since we will be working with the same `X-Robots-Tag` header.

tlvu:
> The all value is documented here: https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#directives

Should we be extra sure and add the "deny" there?  `noindex, nofollow, noarchive, nositelinkssearchbox, nosnippet, notranslate, noimageindex`

fmigneault:
I think `noindex, nofollow` are sufficient. 
The others seem to depend on the index to occur to be generated, or are irrelevant with the contents that we have.

fmigneault:
@perronld @huard @tlvu @mishaschwartz 
I pushed some commits to transfer the original change proposed into a separate optional-component. 

@perronld 
On our side, that would mean only adding the following toour daccs-config env.local for all staging/dev/prod instances. 
```shell
EXTRA_CONFS_DIR='
   <...>
   optional-components/x-robots-tags-header
'
```

mishaschwartz (commented):
I really like this idea. I just have one questions about the way its set up:



Is there a reason why this needs to be configured in an optional component? Why not just make it an environment variable in the proxy component? That seems like it would be simpler since this setting only changes a single line in the proxy configuration.

Or are you thinking that this could be expanded later to update additional header related settings that are not specifically for the proxy?

fmigneault:
Using only an environment variable would work as well.
However, the user would have to write it properly to include the full directive, e.g.: 
```shell
export X_ROBOTS_TAGS='add_header X-Robots-Tags "noindex, nofollow";'
```
I felt it was less error-prone to abstract that need for them to have the exact nginx syntax.
I also took advantage of the usual optional-component documentation approach to provide more details about the feature. 

For now, the component only overrides `X-Robots-Tags` globally, but we could have additional variables later on if need be for per-location overrides.

mishaschwartz (approved):

fmigneault:
@tlvu  Added in 497ded4

tlvu (approved):

fmigneault (approved):

PR #329, opened 2023-05-24
Add /version and /service endpoints

mishaschwartz:
## Overview

A new endpoint "/services" is added that provides a json string describing each of the user facing services currently enabled on the stack. This is a static string and serves a different purpose than the endpoints served by canarie-api (monitoring status). This endpoint is meant to be polled by the node registry scripts (in development) to provide information about what services are meant to be available without having to poll other endpoints directly.

A new endpoint "/version" is added that provides a string containing the current version number of the stack (e.g. "1.26.0"). This endpoint is meant to be polled by the node registry scripts (in development).

## Changes

**Non-breaking changes**
- Adds two new endpoints serving static content from the nginx proxy

**Breaking changes**
- None


tests/service-config-schema.json
```diff
@@ -0,0 +1,34 @@
+{
```
fmigneault, line 1: For all of the new `service-config.json.template` files, is it possible to provide this reference as `{"$schema": "<url>"}`?

Since the schema and configs are relevant for the node registry, it should be specified under the node registry source code itself instead of birdhouse-deploy.

birdhouse/config/proxy/conf.d/all-services.include.template
```diff
@@ -7,6 +7,16 @@
...
     }
 
+    location /services {
+        default_type application/json;
+        return 200 '${BIRDHOUSE_DEPLOY_SERVICES_JSON}';
+    }
+
+    location /version {
+        default_type text/plain;
+        return 200 '${BIRDHOUSE_VERSION}';
```
fmigneault, line 17: Why not return both the version and the release datetime in JSON?

birdhouse/scripts/get-services-json.include.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# default value in case of error or missing definitions
+export BIRDHOUSE_DEPLOY_SERVICES_JSON='{}'
+if [ -z "${ALL_CONF_DIRS}" ]; then
+  echo "No services in DEFAULT_CONF_DIRS and EXTRA_CONF_DIRS. SERVICES JSON list will be empty!"
+  return
+fi
+
+for adir in $(echo "$ALL_CONF_DIRS" | grep -v " ./core\| ./data"); do
```
fmigneault, line 10: Refers to the directories from the other PR.

tests/test_deployment.py
```diff
@@ -0,0 +1,56 @@
+import os
+import pytest
+import glob
+import json
+from string import Template
+
+import jsonschema
+
+COMPONENT_LOCATIONS = ("core", "data", "services", "extensions", "test-helpers")
```
fmigneault, line 9: Refers to the other PR directories.

tests/test_deployment.py
```diff
@@ -0,0 +1,56 @@
...
+        for path in component_paths:
+            service_config_file = os.path.join(path, "service-config.json.template")
+            if os.path.isfile(service_config_file):
+                with open(service_config_file) as f:
+                    service_config = json.loads(Template(f.read()).substitute(TEMPLATE_SUBSTITUTIONS))
+                config_name = service_config.get("name")
+                path_name = os.path.basename(path)
+                if config_name != path_name:
+                    invalid_names.append((config_name, path_name))
+        assert not invalid_names, "service names in service-config.json.template should match the directory name"
```
fmigneault, line 42: Provide the list of invalid names in the error message to help debug which one(s) are problematic.

birdhouse/components/weaver/service-config.json.template
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "weaver",
+    "url": "https://${PAVICS_FQDN_PUBLIC}/${WEAVER_MANAGER_NAME}",
+    "type": "ems",
```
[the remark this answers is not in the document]
mishaschwartz: The "type" field is intended as some values to sort/search on so I'd be fine to change this to a list

tests/service-config-schema.json
mishaschwartz: That's a good idea. Since the node registry source code is still in development why don't you make this an issue and we can revisit it in a later PR.

birdhouse/config/proxy/conf.d/all-services.include.template
mishaschwartz: Only the version is relevant for our use-case. If you have a use-case that also requires the release time we can add it in a different endpoint.

tests/test_deployment.py
mishaschwartz: pytest provides nicely formatted error messages that contain that information. The additional assert error message is just to provide some additional explanation and doesn't need to provide that information twice.

birdhouse/config/proxy/conf.d/all-services.include.template
fmigneault: I just though that we might as well provide it since it's already there. The node manager could also validate when was the last update.

tests/test_deployment.py
fmigneault: You're right. I forgot about that from static code analysis.

tests/service-config-schema.json
fmigneault: Yes. Sounds good. I let you create it.

CHANGES.md
```diff
@@ -17,6 +17,14 @@
 
 ## Changes
 
+- A new endpoint "/services" is added that provides a json string describing each of the user facing services currently 
+  enabled on the stack. This is a static string and serves a different purpose than the endpoints served by canarie-api
+  (monitoring status). This endpoint is meant to be polled by the node registry scripts (in development) to provide 
+  information about what services are meant to be available without having to poll other endpoints directly.
+
+- A new endpoint "/version" is added that provides a string containing the current version number of the stack
```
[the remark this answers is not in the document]
fmigneault: Just "typical" code formatting. Class names, methods, URL references, etc. should stand out more that normal text.

birdhouse/components/weaver/service-config.json.template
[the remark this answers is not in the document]
mishaschwartz: After some additional discussion, we proposed further restrictions to the type. See here: https://github.com/DACCS-Climate/DACCS-node-registry/pull/7

birdhouse/config/geoserver/service-config.json.template
```diff
@@ -1,7 +1,7 @@
 {
     "name": "geoserver",
     "url": "https://${PAVICS_FQDN_PUBLIC}/geoserver",
-    "type": ["geoserver"],
+    "type": ["data", "service-wms"],
```
fmigneault, line 4: Not only a WMS, but also a WFS. 
(technically, WPS also possible, but not used)

birdhouse/components/weaver/service-config.json.template
```diff
@@ -1,7 +1,7 @@
 {
     "name": "weaver",
     "url": "https://${PAVICS_FQDN_PUBLIC}/${WEAVER_MANAGER_NAME}",
-    "type": ["ogc-api"],
+    "type": ["service-ogcapi"],
```
fmigneault, line 4: Need to be specific about *OGC API - Processes*, as there are *Features*, *Maps*, etc.

birdhouse/config/thredds/service-config.json.template
```diff
@@ -1,7 +1,7 @@
 {
     "name": "thredds",
     "url": "https://${PAVICS_FQDN_PUBLIC}/thredds",
-    "type": ["thredds"],
+    "type": ["data"],
```
fmigneault, line 4: No `service-thredds`? Also, missing `catalog`? Based on the terms of the pattern in test schema.

mishaschwartz: > Also, missing catalog? 

That sort of depends on the intended use-case for a lot of these. Once stac is in place then we intend to use that as the user-facing catalog and thredds will just be used to serve data.

In fact a lot of these will depend on the specifics of the permissions in magpie. For example: geoserver can optionally have wms, wfs, wps depending on the permission settings enabled. 

So the question ends up being: 

- does type indicate all _possible_ use cases?
- does type indicate all _intended_ use cases?
- does type indicate all _enabled_ use cases? (this gets complicated as permissions for users may differ)

I'm interested in any opinions on this. But I'm also just happy to keep things simple for now and go with all _possible_ use cases if that simplifies things for now.

mishaschwartz: After a quick discussion with Alex who's writing code to consume this information, we think that indicating all _possible_ use cases is sufficient and easiest to describe and maintain.

I'll go for that for now.

mishaschwartz:
@matprov it looks like tests are failing to deploy because there is no room left on disk:

```
+ docker pull pavics/workflow-tests:230530-1
230530-1: Pulling from pavics/workflow-tests

....

failed to register layer: Error processing tar file(exit status 1): write /opt/conda/envs/birdy/lib/python3.9/site-packages/mamba_gator/navigator/static/bundle.js.map: no space left on device
```


from: https://daccs-jenkins.crim.ca/job/PAVICS-e2e-workflow-tests/job/master/1152/console

birdhouse/config/thredds/service-config.json.template
fmigneault: I don't think there is a big distinction between *intended*/*enabled* (as in the intention will drive what gets enabled or not), but I think this information is more relevant than *possible* use cases from a user points of view that will interact with the node. Basically, a user is interested in knowning what they can achieve with the node. I think specifying that a service *could* do something, but doesn't support it, is somewhat pointless for the users or developers interacting with that node. Listing all *possible* cases also somewhat poses a burden on the node provider, as in, it would be expected that they support all use cases, which is not something that is required nor that we expect.

mishaschwartz: While I agree in theory, I don't see how we can implement this in practice.

The difference between intended and enabled are:

- intended: this service is available for some users of the node
- enabled: this service is available to *you*, a specific user of the node

One option would be to tie this in to the magpie providers definitions for this node but right now not all services are necessarily using magpie/twitcher for authentication.

If we wait until all services are authenticated through magpie we could take that strategy. In the meantime though, we need some solution to move this PR forward.

- if we're comfortable with keeping the "possible" strategy for now: this PR is ready
- if we'd prefer the "intended" strategy: we need to wait until all service are protected by magpie/twitcher so I suggest we remove the "type" key for now and implement it later
- if we'd prefer the "enabled" strategy: we need to go back to the drawing board and rethink the entire mechanism that this endpoint uses to provide the information about each service so I suggest we remove the "type" key for now and implement it later

fmigneault: > I don't see how we can implement this in practice.

Aren't new nodes expected to be added by providers that want to register it? So they would be the ones to open a PR on the node-registry to add their instance. Only them should know exactly what they support. This could also include addionnal services and functionalities that are not part of the existing set in birhdouse-deploy.

I'm still failing to see any use case where a user browsing for node capabilities would even care about *possible* features. Only what they can achieve right away on a given node matters.

mishaschwartz: You're talking about the "intended" strategy here, and you're right about the details of how this is implemented. 

I'm saying that we can't implement the "enabled" strategy since users will not be logged in when searching for services.

Even with the "intended" strategy, this gets more complicated. For example, consider the following scenario:

- a provider registers geoserver in magpie with the following configuration:

   ```yaml
   providers:
      geoserver:
        url: http://localhost:1234
        type: geoserver
        configuration:
          wfs: true
          wms: false
          wps: false
   ```

- we'd expect that the "type" key would contain `["services-wfs", "data"]` when the static json string is created when the proxy container starts up.
- if later the provider changes it to `wms: true` then we'd expect the "type" key to change to include `"service-wms"` but this will only be picked up by changing the json template file and restarting the proxy container.
- we could make it more automatic by generating the "types" from the magpie configuration

mishaschwartz: That being said, I don't care that much about how this is implemented. We can adapt the front end to work with whatever data is provided by the services endpoint. 

@fmigneault please let me know what you would like to see here and we'll go with that so that we can move forward with this PR.

birdhouse/config/proxy/conf.d/all-services.include.template
mishaschwartz: @fmigneault is having the release datetime a requirement for you? We can adapt the front end to accommodate whatever data you want provided by this endpoint.

fmigneault: I find it to be a quicker indication of an "outdated" stack or not, without having to check GitHub when the version was realeased, since we can loose track of the version numbers across all services.

birdhouse/config/thredds/service-config.json.template
fmigneault: Maybe were are confusing the terms between each other. I don't really mind about the "name" of the strategy. 
The only thing important is that it should reflect what a user can achieve when interacting with the node.

I don't think relying on Magpie configuration is the right way to go.
In the case of the birdhouse-deploy configuration provided in this repository, we can parse which WFS/WMS/WPS or any other provider are activated. However, another DACCS node with overrides will not necessarily expose those custom configurations.

Therefore, I can see only 2 valid options about what we can do to report node capabilities:
1. Rely on the node managers to open a PR on the node-registry to provide updates to the reference repository about the modified capabilities of their services for their node.
2. Have some kind of capabilities endpoint, similar to `/components` that provides directly the contents what the node-registry is expecting (schema to be completed/validated). The `/conformance` endpoint is quote common now of OGC and STAC APIs. Probably a similar approach with normative URI references in a `conformsTo` list could be defined. 

Either way, if they don't provide the capabilities updates which either approach, it is sadly "just too bad" for their node.

mishaschwartz: > 1. Rely on the node managers to open a PR on the node-registry to provide updates to the reference repository about the modified capabilities of their services for their node.

This is a fine fallback if needed but I can see there being a lot of issues due to human error that I'd like to avoid if possible. 

>  2. Have some kind of capabilities endpoint, similar to /components that provides directly the contents what the node-registry is expecting (schema to be completed/validated).

This is what the `/services` endpoint is supposed to be. 

>  The /conformance endpoint is quote common now of OGC and STAC APIs. Probably a similar approach with normative URI references in a conformsTo list could be defined.

 I don't really understand how a conformance endpoint helps us achieve what we want here. I checked out the example in weaver (https://pavics.ouranos.ca/weaver/conformance) and environment canada (https://api.weather.gc.ca/conformance) and it is not clear to me how that is a solution to what we are trying to achieve here. Can you help me understand how adding a conformance endpoint helps us achieve our goal:

- We are trying to logically organize services into categories so that we can group them together for documentation purposes in a way that makes sense to the end user (paraphrased from here: https://github.com/DACCS-Climate/DACCS-node-registry/pull/7#issuecomment-1578693022)

One last thing I will say: being able to categorize services in this way is _not_ a requirement for the project. If we think that this sort of categorization is misleading since we can't define the categories well enough then we should _not_ include them at all. The json object would then only include "name", "url", "description", "documentation" as keys which would be entirely sufficient.

fmigneault: > I don't really understand how a conformance endpoint helps us achieve what we want here.

If using the birdhouse-deploy stack as is, it is possible to know "somewhat" which services types/categories apply. 
However, this is technically not guaranteed to be correct or sufficient information to use the services, and it is not really that clear which subset of capabilities are supported.

For example, I could toggle off the `wms` capabilities of my instance's `geoserver`, and then, default services types/categories returned by `geoserver` on `/services` would incorrectly report that my node supports `service-wms`, when it doesn't.
The https://github.com/bird-house/birdhouse-deploy/blob/service-version-endpoints/birdhouse/config/geoserver/service-config.json.template file therefore needs to be dynamically populated, which becomes quickly very complicated. However, this is not even the worst case.

The goal of the `/conformance` endpoint is to be more explicit about the exact capabilities supported by each service.
The one I know the most is for the case of `weaver`. The *OGC API - Processes* is technically sub-divided into:
- https://github.com/opengeospatial/ogcapi-processes/tree/master/core (Part 1: Core)
- https://github.com/opengeospatial/ogcapi-processes/tree/master/extensions/deploy_replace_undeploy (Part 2: DRU)
- https://github.com/opengeospatial/ogcapi-processes/tree/master/extensions/workflows (Part 3: Workflows)
- https://github.com/opengeospatial/ogcapi-processes/tree/master/extensions/quotation
  https://github.com/opengeospatial/ogcapi-processes/tree/master/extensions/billing (WIP: Quotation & Billing)

And each of those parts are themselves sub-divided in a lot of conformances options that can be optionally supported.
There are similar sets of conformance URLs for WFS, STAC, etc. Some of them even overlap, since STAC is a superset around *OGC API - Features*, which is the predecessor of WFS (and partially compatible with it). 

In summary, I think having the general categorization is a good idea. We should keep that. 
However, I wonder if we could take advantage of the widely used approach of `/conformance` instead of redefining our own structure, to basically list similar "capabilities" (although less verbose for the moment). If this `/conformance` was used, many services would not need to update anything as they already provide it, as you illustrated with Weaver and Env Canada. STAC API also provides this list, and its a requirement for all new *OGC API*s.

fmigneault: I have a better idea. Let's keep `/services`, but also allow a `links` section.
For https://github.com/bird-house/birdhouse-deploy/blob/service-version-endpoints/birdhouse/components/weaver/service-config.json.template, it would then be:
```json
{
  "name": "weaver",
  "type": [
    "service-ogcapi_processes"
  ],
  "description": "An OGC-API flavored Execution Management Service",
  "links": [
    { 
      "rel": "service",
      "type": "application/json",
      "href": "https://${PAVICS_FQDN_PUBLIC}/${WEAVER_MANAGER_NAME}/"
    },
    { 
      "rel": "service-desc",
      "type": "text/html",
      "href": "https://pavics-weaver.readthedocs.io/"
    },
    { 
      "rel": "service-doc",
      "type": "application/json",
      "href": "https://${PAVICS_FQDN_PUBLIC}/${WEAVER_MANAGER_NAME}/"
    },
    {
      "rel": "conformance",
      "type": "application/json",
      "href": "https://${PAVICS_FQDN_PUBLIC}/${WEAVER_MANAGER_NAME}/conformance/"
    }
  ]
}
```
This uses valid/recommended https://www.iana.org/assignments/link-relations/link-relations.xhtml names and is extensible.

fmigneault: Given that the list of `conformance` items could be identified from some provided link, I would worry less about the specific values of `type`. They would almost be the same as `keywords`, which are common in other APIs (but with a more restrictive set to your linking). 

In the case of GeoServer and the birds, that would be each of the WFS, WMS, etc. `GetCapabilities` endpoint that could return what they support (see links in https://pavics.ouranos.ca/geoserver/web/). If OGC APIs were added to GeoServer, they would each report their own `/conformance` endpoint as well (https://geoserver.geosolutionsgroup.com/edu/en/ogcapi/intro.html#openapi-based).

In THREDDS, the best I could find was: https://pavics.ouranos.ca/testthredds/catalog.xml
So at least, we have some form of listing of reference schemas and capapilities it supports.

I don't expect "conformance" to be a *requirement* for all services, just that it is available whenever possible.

dchandan: @fmigneault The point of this PR is simply to expose information that will be consumed by the node-registry, and some DACCS components that use the registry, such as the front-end. There are, so far, no other meaningful uses of the `/services` endpoint by anyone else, for anything else. Given this limited use case, it seems unnecessary to get into the weeds of what something should be called and the conformance standards that should be used. We could all simply agree to go with calling things whatever we want for now and revise things later.

In fact, I do not see why we cannot simply go with @mishaschwartz's proposal to expose all possible capabilities at this point and then in future, should the need arise for one reason or another, constrain the `types` field to intended or enabled capabilities. I see this as a more agile way of doing things and letting experience and use cases dictate how the behaviour of the endpoint evolves, rather than beating the issue to death in this PR before we have even gotten started to using the feature. 

I think the `links` format makes sense, and I think an amendment to this PR that includes that change should be sufficient to warrant merging this PR and moving forward.

fmigneault: @dchandan 

> There are, so far, no other meaningful uses of the `/services` endpoint by anyone else, for anything else. 

But should we limit ourselves to this narrow mindset because of this?
While designing a registry to centralize the node definitions, we should try to identify ahead of time possible issues, and emphasise on approaches that will favor reusing this information effectively. Otherwise, we end up in a similar situation as ESGF, were we have to patch up a metagrid definition after the fact to address the huge mess of capabilities from each node.

>  I do not see why we cannot [...] expose all possible capabilities at this point

This is exactly what I'm trying to achieve. I'm simply pointing out that saying for example `service-ogcapi_processes` is not sufficient to indicate what this service is capable of. It is much more complicated than that. 

I'm simply sharing experience I have gathered over time from working with OGC and other Geo-related standards. 
Instead of reinventing the wheel and making a custom schema, I propose alternative methods to represent the same data, but that just so happens to align with existing web best pratices used by many APIs.

If the goal is to allow nodes to interact with each other and be able to properly communicate their capabilities, they should be able to provide this information is the most descriptive manner possible. If the way to represent these references always changes, this will become a huge blocker for interroperability.

mishaschwartz: @fmigneault Sure that's fine, we'll use the links key as you suggest. 

Based on your comments I'm going to assume that your answer to the question I raised in this comment (https://github.com/bird-house/birdhouse-deploy/pull/329#discussion_r1221550632):

> Are you saying that you don't care what is in the type key if we change the links key to your suggestion?
If so, please let me know and I'll update the links key to your suggestion and I'll decide what goes in the type key.

is: "use the format described here (https://github.com/bird-house/birdhouse-deploy/pull/329#discussion_r1220478936) and leave the current values of the `type` key as is". 

If that is not what you meant, please let me know as soon as possible so that I can proceed to update this PR.

fmigneault: I would like to have the `ilnks` provided for reference, but you are free to use the `type` as well with whichever set of values that fills your needs for grouping the services in the node-registry.

mishaschwartz:
Unit tests are expected to fail until https://github.com/DACCS-Climate/DACCS-node-registry/pull/9 is merged.

mishaschwartz:
@fmigneault please have a look at the `links` values for the various services to let me know if they match what you were thinking. And feel free to suggest any additions (either here or in a later PR).

mishaschwartz:
E2E tests are failing because the JSON string is now too long for an nginx parameter.

TODO: move this content to a file and serve it from there (we should probably do the same to the /components endpoint as well)

tlvu (commented):
Quick look, LGTM.

Will deploy to my test VM to play with and report back.

Have you tried this under autodeploy?  Meaning have autodeploy invoke `./pavics-compose.sh` instead of calling `./pavics-compose.sh` manually ourselves.  I vaguely remember the `/components/` is not generated properly under autodeploy.

birdhouse/config/geoserver/service-config.json.template
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "geoserver",
+  "keywords": [
+    "data",
+    "service-wms",
+    "service-wfs",
+    "service-wps"
+  ],
+  "documentation": "https://docs.geoserver.org/",
```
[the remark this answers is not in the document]
mishaschwartz: Thanks for catching that. It should have been removed after one of the format re-writes but it was missed. I'll remove it.

mishaschwartz:
Yes the autodeploy works for both `service` and `version` endpoints. The `components` endpoint is empty after autodeploy:

```json
{"components": []}
```

which I didn't realize was an issue but that looks like it's specifically an issue with components and should be addressed in a different PR

tlvu (approved):
LGTM

tlvu (changes requested):
Houston, I have test failure

```
self = <test_read_configs_include.TestReadConfigs object at 0x7f882796a650>                                                                                                 
read_config_include_file = '/zstore/repos/birdhouse-deploy-master/birdhouse/read-configs.include.sh'                                                                        
                                                                                                                                                                            
    def test_delayed_eval_custom_value(self, read_config_include_file) -> None:                                                                                             
        """Test delayed eval when value is set in env.local"""                                                                                                              
        extra = {"PAVICS_FQDN": '"fqdn.example.com"',                                                                                                                       
                 "PAVICS_FQDN_PUBLIC": '"public.example.com"',                                                                                                              
                 "DATA_PERSIST_ROOT": '"/my-data-root"',  # indirectly change JUPYTERHUB_USER_DATA_DIR                                                                      
                 "GEOSERVER_DATA_DIR": '"/my-geoserver-data"',                                                                                                              
                 }                                                                                                                                                          
        proc = self.run_func(read_config_include_file, extra,                                                                                                               
                             'echo "$PAVICS_FQDN_PUBLIC - $JUPYTERHUB_USER_DATA_DIR - $GEOSERVER_DATA_DIR"')                                                                
        # If PAVICS_FQDN_PUBLIC is set in env.local, that value should be effective.                                                                                        
>       assert (split_and_strip(get_command_stdout(proc))[-1] ==                                                                                                            
                "public.example.com - /my-data-root/jupyterhub_user_data - /my-geoserver-data")                                                                             
E       AssertionError: assert 'public.examp...eoserver-data' == 'public.examp...eoserver-data'                                                                             
E         - public.example.com - /my-data-root/jupyterhub_user_data - /my-geoserver-data                                                                                    
E         ?                       ---    -----                                                                                                                              
E         + public.example.com - /data/jupyterhub_user_data - /my-geoserver-data                                                                                            
                                                                                                                                                                            
tests/test_read_configs_include.py:155: AssertionError                                                                                                                      
___________________________________________________________ TestCreateComposeConfList.test_compose_no_overrides ____________________________________________________________
                                                                                                                                                                            
self = <test_read_configs_include.TestCreateComposeConfList object at 0x7f882796b610>                                                                                       
read_config_include_file = '/zstore/repos/birdhouse-deploy-master/birdhouse/read-configs.include.sh'                                                                        
                                                                                                                                                                            
    def test_compose_no_overrides(self, read_config_include_file):                                                                                                          
        """Test that COMPOSE_CONF_LIST is set correctly when there are no overrides"""                                                                                      
        proc = self.run_func(                                                                                                                                               
            read_config_include_file, {"ALL_CONF_DIRS": "./config/finch ./config/raven"}, 'echo "$COMPOSE_CONF_LIST"'                                                       
        )                                                                                                                                                                   
>       assert split_and_strip(get_command_stdout(proc), split_on="-f") == [                                                                                                
            "docker-compose.yml",                                                                                                                                           
            "./config/finch/docker-compose-extra.yml",                                                                                                                      
            "./config/raven/docker-compose-extra.yml",                                                                                                                      
        ]                                                                                                                                                                   
E       AssertionError: assert ['docker-compose.yml'] == ['docker-comp...se-extra.yml']                                                                                     
E         Right contains 2 more items, first extra item: './config/finch/docker-compose-extra.yml'                                                                          
E         Use -v to get more diff  
```

Will try on a `master` to see if error already exist.

tlvu (approved):

birdhouse/README.rst
```diff
@@ -262,6 +262,11 @@ To run the tests:
     python3 -m pip install -r tests/requirements.txt
     pytest tests/
 
+Some tests require internet access (to access JSON schemas used to validate
+JSON structure). If you need to run tests offline, you can skip the tests that
+require internet access by setting the `DACCS_SKIP_ONLINE_TESTS` environment
+variable to a non-empty value.
```
fmigneault, line 268: Instead of using custom logic, it would be much better to take advantage of pytest markers and add something like `@pytest.mark.online` to the relevant tests. It is easier after to ignore using `pytest -k 'not online' tests`, which allows any logical combination of markers as needed. It will also automatically perform the skip, no need to handle it in try/expect and `pytest.fail` re-raise.

birdhouse/config/proxy/conf.d/all-services.include.template
```diff
@@ -4,17 +4,20 @@
...
     }
 
     location /services {
         default_type application/json;
-        return 200 '${BIRDHOUSE_DEPLOY_SERVICES_JSON}';
+        root /static;
+        try_files /services.json =404;
     }
 
     location /version {
```
fmigneault: Does these endpoints properly work by themselves? 
i.e.: calling `https://${PAVICS_FQDN_PUBLIC}/services` and `https://${PAVICS_FQDN_PUBLIC}/version` without trailing `/` does not cause a redirect to `/jupyter/hub`?

Asking because `/components` does cause the redirect to Jupyter if not defined explicitly as `/components/`, due to how the location is defined above. If it works, maybe `/components/` should be changed to `/components` instead, to align all the endpoints together.

birdhouse/config/thredds/service-config.json.template
```diff
@@ -1,7 +1,25 @@
...
+  "name": "thredds",
+  "keywords": [
+    "data",
+    "catalog"
+  ],
+  "description": "A server that provides catalog, metadata, and data access services for scientific data.",
+  "links": [
+    {
+      "rel": "service",
+      "type": "application/json",
```
fmigneault, line 11: `text/html`

birdhouse/scripts/get-services-json.include.sh
```diff
@@ -1,15 +1,13 @@
...
-  return
-fi
 
 for adir in ${ALL_CONF_DIRS}; do
   [ -f "${adir}/service-config.json" ] || continue
   SERVICES="${SERVICES}$([ -n "${SERVICES}" ] && echo ',') $(cat "${adir}/service-config.json")"
 done
 
+if [ -z "${SERVICES}" ]; then
+  echo "No services in DEFAULT_CONF_DIRS and EXTRA_CONF_DIRS. SERVICES JSON list will be empty!"
```
fmigneault, line 11: Could use the prefix as in https://github.com/bird-house/birdhouse-deploy/blob/22fc7c01a0279b2c98a9059b874dee5922bcac21/birdhouse/pavics-compose.sh#L74

fmigneault (commented):
I think it would be a good idea to include this to all `service-config.json.template` files:
```json
"$schema": "https://raw.githubusercontent.com/DACCS-Climate/DACCS-node-registry/main/node_registry.schema.json"
```

Using the `main` branch for now because there is no tagged version, but would be even better if some were made available to target specific revisions.

birdhouse/config/proxy/conf.d/all-services.include.template
mishaschwartz: Yeah I'd prefer to change it to `/components` too. It can be annoying if you forget that final `/`. I can do that in this PR

tlvu:
> Houston, I have test failure
> 
> ```
> self = <test_read_configs_include.TestReadConfigs object at 0x7f882796a650>                                                                                                 
> read_config_include_file = '/zstore/repos/birdhouse-deploy-master/birdhouse/read-configs.include.sh'                                                                        
>                                                                                                                                                                             
>     def test_delayed_eval_custom_value(self, read_config_include_file) -> None:                                                                                             
>         """Test delayed eval when value is set in env.local"""                                                                                                              
>         extra = {"PAVICS_FQDN": '"fqdn.example.com"',                                                                                                                       
>                  "PAVICS_FQDN_PUBLIC": '"public.example.com"',                                                                                                              
>                  "DATA_PERSIST_ROOT": '"/my-data-root"',  # indirectly change JUPYTERHUB_USER_DATA_DIR                                                                      
>                  "GEOSERVER_DATA_DIR": '"/my-geoserver-data"',                                                                                                              
>                  }                                                                                                                                                          
>         proc = self.run_func(read_config_include_file, extra,                                                                                                               
>                              'echo "$PAVICS_FQDN_PUBLIC - $JUPYTERHUB_USER_DATA_DIR - $GEOSERVER_DATA_DIR"')                                                                
>         # If PAVICS_FQDN_PUBLIC is set in env.local, that value should be effective.                                                                                        
> >       assert (split_and_strip(get_command_stdout(proc))[-1] ==                                                                                                            
>                 "public.example.com - /my-data-root/jupyterhub_user_data - /my-geoserver-data")                                                                             
> E       AssertionError: assert 'public.examp...eoserver-data' == 'public.examp...eoserver-data'                                                                             
> E         - public.example.com - /my-data-root/jupyterhub_user_data - /my-geoserver-data                                                                                    
> E         ?                       ---    -----                                                                                                                              
> E         + public.example.com - /data/jupyterhub_user_data - /my-geoserver-data                                                                                            
>                                                                                                                                                                             
> tests/test_read_configs_include.py:155: AssertionError                                                                                                                      
> ___________________________________________________________ TestCreateComposeConfList.test_compose_no_overrides ____________________________________________________________
>                                                                                                                                                                             
> self = <test_read_configs_include.TestCreateComposeConfList object at 0x7f882796b610>                                                                                       
> read_config_include_file = '/zstore/repos/birdhouse-deploy-master/birdhouse/read-configs.include.sh'                                                                        
>                                                                                                                                                                             
>     def test_compose_no_overrides(self, read_config_include_file):                                                                                                          
>         """Test that COMPOSE_CONF_LIST is set correctly when there are no overrides"""                                                                                      
>         proc = self.run_func(                                                                                                                                               
>             read_config_include_file, {"ALL_CONF_DIRS": "./config/finch ./config/raven"}, 'echo "$COMPOSE_CONF_LIST"'                                                       
>         )                                                                                                                                                                   
> >       assert split_and_strip(get_command_stdout(proc), split_on="-f") == [                                                                                                
>             "docker-compose.yml",                                                                                                                                           
>             "./config/finch/docker-compose-extra.yml",                                                                                                                      
>             "./config/raven/docker-compose-extra.yml",                                                                                                                      
>         ]                                                                                                                                                                   
> E       AssertionError: assert ['docker-compose.yml'] == ['docker-comp...se-extra.yml']                                                                                     
> E         Right contains 2 more items, first extra item: './config/finch/docker-compose-extra.yml'                                                                          
> E         Use -v to get more diff  
> ```
> 
> Will try on a `master` to see if error already exist.

The fix is in this commit 449c36a78ba20f8fa7bb420f0886ea74d00853b3 and 952be7e591d7d98fbcf60645ac59e8ec11eaf50e from PR https://github.com/bird-house/birdhouse-deploy/pull/345

mishaschwartz:
@tlvu 
> The fix is in this commit https://github.com/bird-house/birdhouse-deploy/commit/449c36a78ba20f8fa7bb420f0886ea74d00853b3 and https://github.com/bird-house/birdhouse-deploy/commit/952be7e591d7d98fbcf60645ac59e8ec11eaf50e from PR https://github.com/bird-house/birdhouse-deploy/pull/345

Thanks for catching that!

fmigneault (approved):

PR #330, opened 2023-05-24
Update labeling

fmigneault:
## Overview

Update CI labeling cases.

## Changes

**Non-breaking changes**
- CI changes only. No impact to the insfrastructure.

**Breaking changes**
- n/a

## Related Issue / Discussion

- n/a

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -59,3 +59,8 @@ feature/WPS:
   - birdhouse/**/weaver/**/*
   - birdhouse/**/wps_output-volume/**/*
   - birdhouse/**/*wps*/**/*
+
+feature/node-registry:
+  - birdhouse/**/service-config.json.template
+  - birdhouse/scripts/get-version.include.sh
+  - birdhouse/scripts/get-services-json.include.sh
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,5 +1,7 @@
 # label rules used by .github/workflows/label.yml
 
+# NOTE: all globs starting by a '*' must be quoted to parse properly
+
 # label 'ci/operations' for all automation-related steps and files
 #   Since this repository is in itself an automation process to deploy a server instance,
 #   we refer here to CI as the 'meta' configuration files for managing the code and integrations with the repository,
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -17,3 +17,33 @@ documentation:
...
+  - birdhouse/components/weaver/**/*
+  - birdhouse/**/weaver/**/*
+
+component/cowbird:
+  - birdhouse/components/cowbird/**/*
+  - birdhouse/**/cowbird/**/*
+
+component/magpie:
+  - birdhouse/config/magpie/**/*
+  - birdhouse/**/magpie/**/*
```
tlvu, line 31: In anticipation of the big move, are we allowed to proactively add the future path `birdhouse/components/magpie/**/*`?  Same comments for all other components.

Another question is `birdhouse/**/magpie/**/*` already include everything before so do we even need to specify `/config|components/magpie`?

.github/labeler.yml
```diff
@@ -17,3 +17,33 @@ documentation:
...
+
+component/twitcher:
+  - birdhouse/config/twitcher/**/*
+  - birdhouse/**/twitcher/**/*
+
+component/THREDDS:
+  - birdhouse/config/thredds/**/*
+  - birdhouse/**/thredds/**/*
+
+feature/WPS:
```
tlvu, line 41: Jupyterhub? Geoserver?

tlvu (approved):
LGTM

.github/labeler.yml
fmigneault: Yes. We can add them too, but it shouldn't be necessary as mentioned by the globs including everything.

mishaschwartz: Won't these globs match directories like `birdhouse/config/finch/config/magpie/` as well? Is that desirable here?

.github/labeler.yml
fmigneault: Added

.github/labeler.yml
fmigneault: Indeed. I think this is valid. It will apply both tags for magpie/wps, which will be relevant if a file in there was modified.

PR #311, opened 2023-04-10
Deprecated unused/unmaintained components

mishaschwartz:
## Overview

Move unused and unmaintained components to a separate `deprecated-components/` subdirectory and remove
them from the DEFAULT_CONF_DIRS list if required.

## Changes

**Non-breaking changes**

**Breaking changes**
- The following components are no longer present by default:
  - catalog
  - frontend
  - malleefowl
  - ncops
  - ncwms2
  - project-api
  - solr
  - phoenix

## Related Issue / Discussion

- Resolves #214
- Closes #287 
- Resolves #206
- Closes #291 
- Resolves #9
- Closes #292 
- Resolves #290

PR #331, opened 2023-05-29
Do not expose additional ports

### LINKED ISSUE #222: :bulb: [Feature] Protected ports by design

mishaschwartz:
## Overview

Docker compose no longer exposes any container ports outside the default network except for ports 80 and 443 from the proxy container. This ensures that ports that are not intended for external access are not exposed to the wider internet even if firewall rules are not set correctly.

Note that if the `monitoring` component is used then port 9100 will be exposed from the `node-exporter` container. This is because this container must be run on the host machine's network and unfortunately there is no known workaround that would not require this port to be exposed on the host machine.

## Changes

**Non-breaking changes**

- Changes all internal URL references to components

**Breaking changes**

- removes the lb_flyingpigeon endpoint because it requires direct access to an exposed port

## Related Issue / Discussion

- Resolves #222

## Additional Information

- This does not make any changes to the components that are deprecated in #311


birdhouse/components/monitoring/config/magpie/config.yml.template
```diff
@@ -0,0 +1,48 @@
...
+    c4i: false
+    type: api
+    sync_type: api
+  alertmanager:
+    url: http://alertmanager:9093
+    title: AlertManager
+    public: true
+    c4i: false
+    type: api
+    sync_type: api
```
fmigneault, line 22: Just curious if this works (UI-wise, everything responds/reacts correctly)?
Have you tried validating access to those endpoints?

birdhouse/config/catalog/config/magpie/providers.cfg.template
```diff
@@ -8,7 +8,7 @@ providers:
     sync_type: wps
 
   ${CATALOG_THREDDS_SERVICE}:
-    url: http://${PAVICS_FQDN}:8083/twitcher/ows/proxy/thredds
+    url: http://thredds:8080/twitcher/ows/proxy/thredds
```
fmigneault, line 11: Will need to check with @tlvu. 

There was an issue regarding THREDDS that needed the full `${PAVICS_FQDN_PUBLIC}${TWITCHER_PROTECTED_PATH}` to be defined as Magpie's URL so that Twitcher could redirect in the right location. Not sure if this impacts the catalog browsing/listing/access or not.

birdhouse/config/flyingpigeon/config/canarie-api/canarie_api_monitoring.py.template
```diff
@@ -24,15 +24,14 @@ SERVICES['flyingpigeon'] = {
         'releasenotes': 'https://github.com/bird-house/flyingpigeon/blob/master/CHANGES.rst',
         'support': 'https://github.com/bird-house/flyingpigeon/issues',
         'source': 'https://github.com/bird-house/flyingpigeon',
-        'tryme': 'http://${PAVICS_FQDN}:8093/wps?service=WPS&version=1.0.0&request=GetCapabilities',
+        'tryme': 'https://${PAVICS_FQDN}/flyingpigeon/wps?service=WPS&version=1.0.0&request=GetCapabilities',
```
fmigneault, line 27: `PAVICS_FQDN_PUBLIC` should be used.

birdhouse/config/hummingbird/config/canarie-api/canarie_api_monitoring.py.template
```diff
@@ -24,15 +24,14 @@ SERVICES['hummingbird'] = {
         'releasenotes': 'https://github.com/bird-house/hummingbird/blob/master/CHANGES.rst',
         'support': 'https://github.com/bird-house/hummingbird/issues',
         'source': 'https://github.com/bird-house/hummingbird',
-        'tryme': 'http://${PAVICS_FQDN}:8097/wps?service=WPS&version=1.0.0&request=GetCapabilities',
+        'tryme': 'http://${PAVICS_FQDN}${TWITCHER_PROTECTED_PATH}/hummingbird/wps?service=WPS&version=1.0.0&request=GetCapabilities',
```
fmigneault, line 27: Use HTTPS and `PAVICS_FQDN_PUBLIC`

birdhouse/config/geoserver/config/proxy/conf.extra-service.d/geoserver.conf.template
```diff
@@ -1,5 +1,5 @@
     location /geoserver/ {
-        proxy_pass http://${PAVICS_FQDN}:8087;
+        proxy_pass http://geoserver:8080/geoserver/;
```
fmigneault, line 2: That bypasses the whole Magpie/Twitcher auth. It should be adjusted to use the protected proxy location while we are modifying this.

birdhouse/config/jupyterhub/config/proxy/conf.extra-service.d/jupyterhub.conf.template
```diff
@@ -1,5 +1,5 @@
     location /jupyter/ {
-        proxy_pass http://${PAVICS_FQDN}:8800/jupyter/;
+        proxy_pass http://jupyterhub:8000/jupyter/;
```
fmigneault, line 2: Bypasses the Magpie/Twitcher auth. 
Is JupyterHub sufficient by itself to form the Magpie login through the UI?

birdhouse/config/portainer/config/proxy/conf.extra-service.d/portainer.conf.template
```diff
@@ -1,5 +1,5 @@
     location /portainer/ {
-        proxy_pass http://${PAVICS_FQDN}:9000/;
+        proxy_pass http://portainer:9000/;
```
fmigneault, line 2: Bypasses Magpie/Twitcher auth.

birdhouse/config/proxy/default.env
```diff
@@ -8,7 +8,7 @@ export PROXY_READ_TIMEOUT_VALUE="240s"
 
 # Content of "location /" in file config/proxy/conf.d/all-services.include.template
 # Useful to have a custom homepage.
-export PROXY_ROOT_LOCATION="return 302 https://\$host/jupyter/hub/login;"
+export PROXY_ROOT_LOCATION="return 302 https://\$host/magpie;"
```
fmigneault, line 11: This doesn't seem related to the other changes.

fmigneault (changes requested):
Some CanarieAPI configs that were modified should be ported to https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/optional-components/canarie-api-full-monitoring/config/canarie-api/canarie_api_full_monitoring.py.template if not already defined in the list of public endpoints.

birdhouse/components/monitoring/config/magpie/config.yml.template
mishaschwartz: As of yesterday, no ...I was still working on it but everything should be working now. I'm not too familiar with what these pages should necessarily look like so if @tlvu can take a look and let me know if anything looks off, I'd appreciate it.

birdhouse/config/catalog/config/magpie/providers.cfg.template
mishaschwartz: I actually didn't even mean to make this change since catalog is being deprecated. It must have slipped in from the cherry-pick from the other branch.

I'm ok to either revert this change or leave it as is (since catalog is deprecated)

birdhouse/config/geoserver/config/proxy/conf.extra-service.d/geoserver.conf.template
mishaschwartz: Yes I didn't make any changes to which services were behind twitcher in this PR. As we discussed in #328 that feels like a change that should happen in a different PR.

mishaschwartz: Note that the monitoring routes were an exception to that rule since putting those behind twitcher was the only way I could think to not expose the monitoring ports and still protect those routes.

birdhouse/config/jupyterhub/config/proxy/conf.extra-service.d/jupyterhub.conf.template
mishaschwartz: Yes jupyter can handle its own authentication by communicating with magpie directly. I'd love to figure out how to put jupyterhub behind twitcher eventually but I spent a few days trying to figure it out and didn't come up with a good solution so I put it back on the to-do list for later. 

(also, I didn't make any changes to which services are behind twitcher in this PR, see my comment above)

birdhouse/config/portainer/config/proxy/conf.extra-service.d/portainer.conf.template
mishaschwartz: Yes.... portainer does manage its own authorization so this is not necessarily insecure but we could make it more secure by putting it behind twitcher as well.  I'd be open to double-protecting portainer since its such a powerful tool.

(also, I didn't make any changes to which services are behind twitcher in this PR, see my comment above)

birdhouse/config/catalog/config/magpie/providers.cfg.template
fmigneault: If it works with `thredds:8080`, then it is better to change it. 
It is guaranteed to fail with `${PAVICS_FQDN}:8083` after changes are applied.

birdhouse/config/geoserver/config/proxy/conf.extra-service.d/geoserver.conf.template
fmigneault: Yes. It's ok if done through a subsequent PR. I just want to make sure we don't forget about it.

fmigneault: https://github.com/bird-house/birdhouse-deploy/issues/333

birdhouse/config/jupyterhub/config/proxy/conf.extra-service.d/jupyterhub.conf.template
fmigneault: To Do: https://github.com/bird-house/birdhouse-deploy/issues/334

birdhouse/config/portainer/config/proxy/conf.extra-service.d/portainer.conf.template
fmigneault: To Do: https://github.com/bird-house/birdhouse-deploy/issues/335

mishaschwartz:
@tlvu @fmigneault can I merge this one?

birdhouse/components/README.rst
```diff
@@ -153,17 +153,14 @@ Prometheus stack is used:
 Usage
 -----
 
-- Grafana to view metric graphs: http://PAVICS_FQDN:3001/d/pf6xQMWGz/docker-and-system-monitoring
-- Prometheus alert rules: http://PAVICS_FQDN:9090/rules
```
tlvu, line 157: I did not add the prometheus stack to the proxy initially so it is not exposed to the internet.  Only someone over VPN can access it because prometheus and alertmanager do not have any authentication.  So I would rather this stays as-is.

birdhouse/components/monitoring/config/magpie/config.yml.template
```diff
@@ -0,0 +1,48 @@
...
+
+permissions:
+  - service: grafana
+    permission: read
+    group: administrators
+    action: create
+  - service: grafana
+    permission: write
+    group: administrators
```
tlvu, line 31: Oh you're adding permissions here, interesting, let's see how this would work.

birdhouse/components/README.rst
```diff
@@ -153,17 +153,14 @@ Prometheus stack is used:
...
-- AlertManager to manage alerts: http://PAVICS_FQDN:9093
-
-The paths above are purposely not behind the proxy to not expose them publicly,
-assuming only ports 80 and 443 are publicly exposed on the internet.  All other
-ports are not exposed.
-
-Only Grafana has authentication, Prometheus alert rules and AlertManager have
-no authentication at all so had they been behind the proxy, anyone will be
-able to access them.
+- Grafana to view metric graphs: http://PAVICS_FQDN/grafana/d/pf6xQMWGz/docker-and-system-monitoring
```
tlvu, line 156: So this should be httpS and not http I guess

birdhouse/config/jupyterhub/config/proxy/conf.extra-service.d/jupyterhub.conf.template
tlvu: > Bypasses the Magpie/Twitcher auth.

@fmigneault FYI Jupyter has never been behind Twitcher since day one, it has its own authentication.

tlvu (commented):
Just took a quick look.  Did deploy this PR to a test server and run Jenkins and spot check on it?

birdhouse/components/monitoring/config/proxy/conf.extra-service.d/monitoring.conf.template
```diff
@@ -0,0 +1,15 @@
+
+    location /grafana {
+        proxy_pass https://${PAVICS_FQDN_PUBLIC}${TWITCHER_PROTECTED_PATH}/grafana;
```
tlvu, line 3: We are not using the Verify trick?

tlvu:
Tried on my VM, got a weird repeated `NS_ERROR_CONNECTION_REFUSED`.

Screenshot from 2023-06-30 17-50-18

Tried Thredds and same error is seen, although it did not break loading of Thredds.
Screenshot from 2023-06-30 17-49-36

This could be my VM too weak.  Will try to increase its CPU and RAM.  Nonetheless, this is suspicious.

fmigneault:
> `NS_ERROR_CONNECTION_REFUSED`

In the case of Grafana, something is definitely not working, since it is the page itself that fails. 

However, the error is `NS_ERROR_CORRUPTED_CONTENT` for the stylesheet of THREDDS, and I can see this same error on master. I think this is simply Twitcher blocking `css` files, causing an unexpected 0-length body.

birdhouse/components/monitoring/config/proxy/conf.extra-service.d/monitoring.conf.template
mishaschwartz: No, but we could. I don't see any reason not to and maybe that will help with the issues below (I'll look into that today)

tlvu:
Just to clarify, when I am not logged into Magpie, navigating to `/grafana/` does result in the expected unauthorized behavior.

```
$ curl --include --silent https://lvupavicsmaster.ouranos.ca/grafana/
HTTP/1.1 401 Unauthorized
Server: nginx/1.23.4
Date: Tue, 04 Jul 2023 18:25:44 GMT
Content-Type: text/xml; charset=UTF-8
Content-Length: 471
Connection: keep-alive
X-Content-Type-Options: nosniff

<?xml version="1.0" encoding="utf-8"?>
<ExceptionReport version="1.0.0"
    xmlns="http://www.opengis.net/ows/1.1"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd">
    <Exception exceptionCode="NoApplicableCode" locator="AccessForbidden">
        <ExceptionText>Access to service is forbidden.</ExceptionText>
    </Exception>
```

When I logged into Magpie as the `admin` user, then I get the `NS_ERROR_CONNECTION_REFUSED` above in comment https://github.com/bird-house/birdhouse-deploy/pull/331#issuecomment-1615228216

Maybe a clue about this problem: when accessing `/grafana` (without the ending slash `/`), I get no response at all !

```
$ curl --include --silent https://lvupavicsmaster.ouranos.ca/grafana
#  No response at all
```

So hopefully this is simply a redirection or context-root mapping issue?

mishaschwartz:
@tlvu 

I think that the twitcher proxy was messing things up so I'm using the verify path instead and everything seems to be working for me. Can you have another look and let me know if it works for you as well?

birdhouse/components/monitoring/config/magpie/config.yml.template
```diff
@@ -1,20 +1,26 @@
...
     title: Prometheus
     public: true
     c4i: false
     type: api
     sync_type: api
   alertmanager:
-    url: http://alertmanager:9093
+    # below URL is only used to fill in the required location in Magpie
+    # actual auth validation is performed with Twitcher 'verify' endpoint without accessing this proxied URL
+    url: http://proxy:80
```
fmigneault: Although the `/twitcher/ows/verify` endpoint approach is used, there should not be any need to remove the actual endpoint where it could be reached if we desired to use the `/twitcher/ows//proxy` endpoint, or opened the port, for debugging or similar.

This is what is currently done for other services. 
For example, `flyingpigeon` uses `http://${PAVICS_FQDN}:8093/wps` although `http` and `8093` are not directly exposed, but it can be done by docker compose overrides.

mishaschwartz: @fmigneault just to clarify, you're suggesting we change the `url: http://proxy:80` back to what it was before so that it's easier to change back to the proxy path if needed?

I'm happy to do that, but I'm a little concerned that this will encourage people to use the "proxy" method instead of the "verify" method which (as we saw previously) makes these components inaccessible from the UI.

tlvu:
> @tlvu
> 
> I think that the twitcher proxy was messing things up so I'm using the verify path instead and everything seems to be working for me. Can you have another look and let me know if it works for you as well?

@mishaschwartz Will try to re-test your fix next week.

birdhouse/components/monitoring/config/magpie/config.yml.template
fmigneault: Yes. Thanks.
I'm not worried about them using this approach. We can always recommend the verify method if a PR is submitted. It is a relatively easy switch.

tlvu:
Sorry for the late report, too busy lately.

So I found another problem with the monitoring components: in the template expanded config `components/monitoring/config/proxy/conf.extra-service.d/monitoring.conf` the variable `TWITCHER_VERIFY_PATH` did not resolve.

I have a feeling we might be using this one more and more often to avoid performance penalty.

So should we have central way to make this variable resolve for every components that depend on it, or just copy the same thing as

https://github.com/bird-house/birdhouse-deploy/blob/2b344d3383d8c0cd3f4cc4a0bf4266e3bae06a47/birdhouse/optional-components/secure-data-proxy/default.env#L21-L28

I did not have this `secure-data-proxy` enabled, that's how I found this.

If Twitcher is going to be on all deployments, maybe that snippet should be moved to twitcher/default.env ?

mishaschwartz:
@tlvu 

> If Twitcher is going to be on all deployments, maybe that snippet should be moved to twitcher/default.env ?

Good catch and I'm sorry I thought I'd done this already. It may have been in a different PR and I'm confused. I'll update it now because I agree that the variable should be defined in twitcher/default.env

fmigneault:
Should we add `TWITCHER_PROTECTED_PATH=/twitcher/ows/proxy` to the Twitcher's `default.env` as well?

fmigneault (approved):

tlvu:
> @tlvu are you ok if we merge this one?

Let me try it again today.  I had some trouble logging into Grafana last time but did not have time to investigate.

tlvu:
@mishaschwartz We have a problem.  All the Grafana graphs are broken in this PR.  I have ensured they worked properly before updating to this PR.

Screenshot from 2023-08-08 19-04-18

I honestly do not want to delay this PR longer since we have way too many opened already.

How about you undo only the part about Grafana, keeping Prometheus and Alertmanager, those are working fine.  The rest is fine too.

Grafana happens to already have its own authentication so that's not unsecure, which is the spirit of this PR.

tlvu:
I managed to find a fix for the Grafana dashboard but now I am losing stats for 4 of the graphs.  It's as if Prometheus is unable to access node-exporter which make no sense since it is able to still get some stats from it.

I still think we might want to rollback the changes for all the monitoring components so we can merge this PR.  I do not have time to look more into those broken 4 broken graphs (Load, Unused Disk Space, Available Memory, Disk I/O) currently.

Screenshot from 2023-08-08 23-39-12

mishaschwartz:
@tlvu this looks like a problem caused by the fact that the `node-exporter` container is on the host network:

```yml
  node-exporter:
    image: quay.io/prometheus/node-exporter:v1.0.0
    container_name: node-exporter
    volumes:
      - /:/host:ro,rslave
    network_mode: "host"
    pid: "host"
    command: --path.rootfs=/host
    restart: always
```
which makes it inaccessible to other containers (in this case it needs to be visible to `prometheus`). I'll make a fix for this

tlvu:
> this looks like a problem caused by the fact that the `node-exporter` container is on the host network:

Then this is even more weird because if `prometheus` is unable to access `node-exporter` completely, why only 4 graphs lose data?  All the graphs in the top 2 rows should all lose data !!!

I vaguely remember `node-exporter` needs to use the host networking because it needs to monitor the entire machine.  Maybe a newer version has different way without using host networking.

mishaschwartz:
> Then this is even more weird because if prometheus is unable to access node-exporter completely, why only 4 graphs lose data? All the graphs in the top 2 rows should all lose data !!!

The data from cadvisor was always working, just the data from node-exporter was not accessible to prometheus.

See the fix here: https://github.com/bird-house/birdhouse-deploy/pull/331/commits/d4ee23f83cded42c92218f99e85d38f5b834f439

Unfortunately it means that this one port is available outside of the docker network (9100) but its a requirement of node-exporter and there's nothing we can do about that if we want to use it.

mishaschwartz:
@tlvu Here is what my test machine's grafana looks like now:

image

tlvu:
@mishaschwartz thanks for finding this.  I confirm your fix worked on my side as well.

Screenshot from 2023-08-09 17-05-15

CHANGES.md
```diff
@@ -16,6 +16,12 @@
 ------------------------------------------------------------------------------------------------------------------
 
 ## Changes
+- Do not expose additional ports:
+  - Docker compose no longer exposes any container ports outside the default network except for ports 80 and 443 from 
+    the proxy container. This ensures that ports that are not intended for external access are not exposed to the wider 
+    internet even if firewall rules are not set correctly.
```
tlvu, line 22: Add a note that `node-exporter` bind on host networking so its port is also exposed.

tlvu (approved):
All good for me, just a little update needed for the changelog.

Merge away when ready.

tlvu:
Forgot to say, please bump minor and not patch on this one since anyone relying on those open ports internally will have to change their habit/scripts.

mishaschwartz:
@tlvu I don't want to merge this yet because it will mess up the order unless the version is correctly bumped from #297.
@matprov can we please resolve that soon? If you don't have time let me know and I will do it.

mishaschwartz:
@matprov no worries. All done #364

PR #336, opened 2023-06-02
Fix Cowbird's `sync_permissions` config which used invalid Magpie service types

mishaschwartz:
## Overview

In the cowbird `config.yml.template` file: the `wps_outputs` and `weaver` keys are supposed to be Magpie service types, not magpie service names. This causes all cowbird hooks to fail with:

```
cowbird.config.ConfigErrorInvalidServiceKey: Service `weaver` used in sync config is not valid since it was not found in Magpie services (['access', 'api', 'geoserver', 'geoserverwfs', 'geoserverwms', 'geoserverwps', 'ncwms', 'thredds', 'wfs', 'wps']).
```

This fix is already implemented in #323 but since that is waiting on https://github.com/Ouranosinc/cowbird/pull/35 to get pulled in, this PR factors out the bugfix that isn't dependent on the cowbird PR. 

This allows us to fix this configuration bug in the meantime

## Changes

**Non-breaking changes**

None (bug fix)

**Breaking changes**

None


ChaamC (approved):

fmigneault (approved):

mishaschwartz:
@tlvu  can I merge this one?

tlvu (approved):

PR #341, opened 2023-06-12
Prototype Thredds verify twitcher

tlvu:
Prototype, not to be merged.  PR just for discussion.

I was not able to get the "verify" to work when I keep the old Thredds context-root `/twitcher/ows/proxy/thedds` to avoid breaking all existing notebooks.

@mishaschwartz could you check which config I overlooked.  I think verify was working initially when the context-root as simply `/thredds`.

However I was able to completely  bypass Twitcher while keeping  `/twitcher/ows/proxy/thedds` as context-root.  Search for all changes containing the word `bypass`.


mishaschwartz (commented):
@tlvu I would go with a different strategy actually. I'd keep the changes up to commit 7706c5d21608e5799af021d0528d93c8d5168653 and then add the following line to the nginx configuration (in a separate optional component):

```
rewrite ^${TWITCHER_PROTECTED_PATH}/thredds/(.*) /thredds/$1 permanent;
```

This will ensure that the existing paths in the notebooks will still work with a lot fewer complex changes to the code.

mishaschwartz:
@tlvu 

> rewrite ^${TWITCHER_PROTECTED_PATH}/thredds/(.*) /thredds/$1 permanent;

I've tested this out and can confirm that the redirect works as intended

Issue #342, opened 2023-06-14
:bug: [BUG]: `/components` endpoint is empty using auto-deploy


fmigneault:
## Details

The `/components` endpoint is empty after auto-deploy:

```json
{"components": []}
```

This does not happen when using `pavics-compose.sh`.

_Originally posted by @mishaschwartz in https://github.com/bird-house/birdhouse-deploy/issues/329#issuecomment-1591259721_

mishaschwartz:
It looks like the issue is that `grep` in the `pavics/docker-compose-git:docker-18.09.7-compose-1.25.1` image does not support searching for multiple patterns that are separated by newlines (see: https://github.com/bird-house/birdhouse-deploy/blob/1.26.4/birdhouse/scripts/get-components-json.include.sh#L58-L59).

The fix would be to replace `BIRDHOUSE_DEPLOY_COMPONENTS_LIST_KNOWN` which currently looks like:

```
components/cowbird
components/monitoring
components/scheduler
components/weaver
config/canarie-api
... etc.
```

with 

```
-e components/cowbird -e components/monitoring -e components/scheduler -e components/weaver -e config/canarie-api
```

which should work with all versions of `grep`. 

I can make this change shortly.

PR #343, opened 2023-06-14
Autodeploy: optionally fix file permissions

mishaschwartz:
## Overview

The autodeploy mechanism creates new files owned by root. If this is not desired then users have to manually update the file ownership after each autodeployment. This adds an option to change the ownership of all files to a specific user after each autodeployment. 

For example, if the code in this repo is currently owned by a user named `birduser` with uid 1002, then by setting `export AUTODEPLOY_CODE_OWNERSHIP="1002:1002"` in `env.local`, all files and folders in this repo will continue to be owned by `birduser` after each autodeployment. 

## Changes

**Non-breaking changes**

adds an optional environment variable to env.local

**Breaking changes**

None


tlvu (approved):
LGTM thanks !

tlvu:
Humm, funny this permission problem is the root cause since `birdhouse/deployment/fix-write-perm` was meant to specifically fix this kind of permission problem.  Or maybe I did not fully understand the usage of `setfacl`.  Thanks for catching this.

tlvu:
I think this fixes https://github.com/bird-house/birdhouse-deploy/issues/342

fmigneault (approved):

mishaschwartz:
> I think this fixes https://github.com/bird-house/birdhouse-deploy/issues/342

I just double checked and unfortunately it does not fix that issue

PR #344, opened 2023-06-14
Components endpoint now returns valid json

mishaschwartz:
## Overview

The JSON string reported by the `/components/` path was not valid JSON due to a misconfigured regular expression used to generate the content. The issue was that integers were not being properly parsed by the regular expression meaning that paths that contained integers other than 0 were not recognized as valid paths.

## Changes

**Non-breaking changes**

None: bugfix

**Breaking changes**

None

## Related Issue / Discussion

- Resolves #339

PR #347, opened 2023-06-15
`components/` endpoint displays intended information after auto-deploy

### LINKED ISSUE #342: :bug: [BUG]: `/components` endpoint is empty using auto-deploy

mishaschwartz:
## Overview

Previously, the script that generates the content for the `components/` endpoint was using a feature of `grep` that is not supported by all versions of `grep`.
  
This meant that this script running in the auto-deployment docker container was not able to properly parse the running components using `grep`. 

This fixes the issue by making the script compliant with all versions of `grep`.

## Changes

**Non-breaking changes**

None: bugfix

**Breaking changes**

None

## Related Issue / Discussion

- Resolves #342

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,16 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Fixes
+- `components/` endpoint displays intended information after auto-deploy
+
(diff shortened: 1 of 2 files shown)


mishaschwartz:
@fmigneault that issue is fixed in #344

tlvu (commented):
LGTM, thanks for looking into this annoying behavior.

tlvu (approved):

PR #348, opened 2023-06-16
Geoserver: protect web interface and ows routes behind magpie/twitcher

### LINKED ISSUE #333: :bulb: [Feature] Protect GeoServer
## Description

Ensure GeoServer is accessed following Magpie/Twticher authentication/authorization. 

## References

- https://github.com/bird-house/birdhouse-deploy/pull/331#discussion_r1210538488
- [DAC-454](https://crim-ca.atlassian.net/browse/DAC-454) *Move geoserver behind Twitcher proxy*
- [DAC-472](https://crim-ca.atlassian.net/browse/DAC-472) *Update Cowbird's Geoserver service post-Magpie service integration*

mishaschwartz:
## Overview

Geoserver: protect web interface and ows routes behind magpie/twitcher
  
Updates Magpie version to [3.35.0](https://github.com/Ouranosinc/Magpie/tree/3.35.0) in order to take advantage of updated Geoserver Service.

The `geoserverwms` Magpie service is now deprecated. If a deployment is currently using this service, it is highly recommended that the permissions are transferred from the deprecated `geoserverwms` service to the `geoserver` service.

The `/geoserver` endpoint is now protected by default. If a deployment currently assumes open access to Geoserver and would like to keep the same permissions after upgrading to this version, please update the permissions for the `geoserver` service in Magpie to allow the `anonymous` group access.

A `Magpie` service named `geoserver` with type `wfs` exists already and must be manually deleted before the new `Magpie` service created here can take effect.

The `optional-components/all-public-access` component provides full access to the `geoserver` service for the `anonymous` group in Magpie. Please note that this includes some permissions that will allow anonymous users to perform destructive operations. Because of this, please remember that enabling the `optional-components/all-public-access` component is not recommended in a production environment.

Introduces the `GEOSERVER_SKIP_AUTH` environment variable. If set to `True`, then requests to the geoserver endpoint  will not be authorized through twitcher/magpie at all. This is not recommended at all. However, it will slightly improve performance when accessing geoserver endpoints.

## Changes

**Non-breaking changes**

In order to provide public access to geoserver by default now, the `all-public-access` optional component must be enabled

**Breaking changes**

The current `wfs` Magpie service named `geoserver` must be deleted before the change here can take effect.

## Related Issue / Discussion

- Resolves #333


fmigneault (changes requested):
I don't think this is the right approach. 
Although this looks to work, it is very hackish and does not handle all use cases.

For example, the WPS endpoint can be accessed in multiple ways: 
`https://pavics.ouranos.ca/geoserver/ows?request=GetCapabilities&service=wps`
`https://pavics.ouranos.ca/geoserver/wps?request=GetCapabilities`
Same for all other W*\[X]*S variants:
`https://pavics.ouranos.ca/geoserver/wms?request=GetCapabilities`
`https://pavics.ouranos.ca/geoserver/ows?request=GetCapabilities&service=wms`

This does not include also the upcoming OGC API endpoints: 
`/geoserver/ogc/features/...`
`/geoserver/ogc/maps/...`
etc.

Using different services causes a lot of permission duplication across the endpoints, which will cause access errors. Instead the Magpie implementation should be updated to handle the `/web/` endpoint accordingly: 

https://github.com/Ouranosinc/Magpie/blob/f9b00c6142372aff96fd0edf0537c8b383fd5ee9/magpie/services.py#L1475-L1488

mishaschwartz:
@fmigneault 

What if we flipped the nginx conf for now:

```
    location /geoserver/ {
        proxy_pass http://${PAVICS_FQDN}${TWITCHER_PROTECTED_PATH}/geoserver-ows/;
    }

    location /geoserver/web/ {
        proxy_pass http://${PAVICS_FQDN}${TWITCHER_PROTECTED_PATH}/geoserver-web/;
```

That way we can accommodate the `wms/`, `wps/`, `ogc/` routes all together and the `web/` routes will be the exception.

Then when magpie catches up we can put it back under one magpie service definition

fmigneault:
> Is this a change you already have proposed for Magpie or should I add an issue for it?

You can add it. It was not already planned. It should be fairly easy to extend the existing class using `ServiceAPI`.

birdhouse/config/geoserver/config/proxy/conf.extra-service.d/geoserver.conf.template
```diff
@@ -1,5 +1,16 @@
     location /geoserver/ {
-        proxy_pass http://${PAVICS_FQDN}:8087;
+        proxy_pass http://${PAVICS_FQDN}${TWITCHER_PROTECTED_PATH}/geoserver-web/;
```
tlvu, line 2: Would it be easy to have a toogle to disable this?  Meaning a toogle to restore the original proxy config to not go through Twitcher?

fmigneault:
Maggie https://github.com/Ouranosinc/Magpie/releases/tag/3.35.0 is released with the relevant functionality.

CHANGES.md
```diff
@@ -15,7 +15,12 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Changes
+- Geoserver: protect web interface and ows routes behind magpie/twitcher
+ 
+  Updates Magpie version to 3.35.0 in order to take advantage of the updated Geoserver Service.
```
fmigneault, line 21: Use the tag link `[3.35.0](https://github.com/Ouranosinc/Magpie/tree/3.35.0)`

CHANGES.md
```diff
@@ -15,7 +15,12 @@
...
+
+  See https://github.com/bird-house/birdhouse-deploy/issues/333 for details.
```
fmigneault, line 23: Should be a bit more verbose about the "potentially breaking" changes:

- `geoserverwms` Magpie service deprecated (could need manual permission updates to be transferred to `geoserver` if used by some external configuration)
- `geoserver` having new secured access by default instead of using https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse/optional-components/test-geoserver-secured-access 
  - breaks instances that assumed fully-open access to `/geoserver` endpoint and its OWS requests
  - makes `optional-components/test-geoserver-secured-access` deprecated and `geoserver-secured` endpoint and service deprecated as well
- warning about `all-public-access` that will now apply additional permissions for geoserver as well
  (somewhat of an edge-case, but someone that was using `geoserver-sercured` + `all-public-access` technically had a "more secured" geoserver that what these changes provide, see https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/optional-components/test-geoserver-secured-access/config/magpie/permissions.cfg.template).
  Since the new permission set allows "writable" or "destructive" operations on layers, this could be an issue.

birdhouse/optional-components/all-public-access/config/geoserver/permissions.cfg
```diff
@@ -0,0 +1,79 @@
...
+   group: anonymous
+   action: create
+ - service: geoserver
+   permission: getfeatureinfo
+   group: anonymous
+   action: create
+ - service: geoserver
+   permission: describeprocess
+   group: anonymous
+   action: create
```
fmigneault, line 47: This is not relevant if `wps: false`. Same as other comment.

birdhouse/optional-components/all-public-access/config/geoserver/permissions.cfg
```diff
@@ -0,0 +1,79 @@
...
+ - service: geoserver
+   permission: getmap
+   group: anonymous
+   action: create
+ - service: geoserver
+   permission: execute
+   group: anonymous
+   action: create
```
fmigneault, line 31: This is not relevant if `wps: false`.
I don't think it would break, so can be left there with a comment regarding `wps` option for reference.

birdhouse/optional-components/all-public-access/config/geoserver/permissions.cfg
```diff
@@ -0,0 +1,79 @@
+permissions:
+ - service: geoserver
+   resource: /
+   type: route
```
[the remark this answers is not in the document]
mishaschwartz: No problem. I don't think I understood how to set this properly with the new geoserver. Do you mind pointing me to the documentation or an example please

CHANGES.md
mishaschwartz: Ok I can add some of these as warnings.

> makes optional-components/test-geoserver-secured-access deprecated and geoserver-secured endpoint and service deprecated as well

If these are now deprecated, I could move them to the deprecated-components directory as part of this PR. What would you suggest @fmigneault

fmigneault: Yes, that would be good. Thank you.

birdhouse/config/geoserver/config/proxy/conf.extra-service.d/geoserver.conf.template
```diff
@@ -1,5 +1,5 @@
     location /geoserver/ {
-        proxy_pass http://geoserver:8080/geoserver/;
+        proxy_pass http://${PAVICS_FQDN}${TWITCHER_PROTECTED_PATH}/geoserver/;
```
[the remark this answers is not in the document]
tlvu: Opps sorry, I mixed up the "verify/protected" path performance problem with JupyterHub and this one.

Maybe Geoserver won't be affected by the performance problem but just to prepare in advance, can this behavior be toggled off?  Meaning the redirect in the front proxy be default behind Twitcher but make it configurable.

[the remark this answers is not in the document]
tlvu: Should this `proxy_pass` directive use `https` since it is using `PAVICS_FQDN`?

tlvu:
By the way, any changes related to Geoserver is not automatically tested because the notebooks is hardcoding the production Geoserver, see https://github.com/Ouranosinc/pavics-sdi/issues/183

fmigneault:
We should un-deprecate `test-geoserver-secured-access` from [`22d006c` (#348)](https://github.com/bird-house/birdhouse-deploy/pull/348/commits/22d006cbfea41fe47eed2a8fc6fc6180a056a917) to reuse it instead only for test-data and test-permission setup for https://github.com/Ouranosinc/pavics-sdi/issues/183, https://github.com/bird-house/birdhouse-deploy/pull/381 and https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/125.

birdhouse/optional-components/all-public-access/config/geoserver/permissions.cfg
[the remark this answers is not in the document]
fmigneault: The `resource` field is needed mostly when applying a permission on a child resource under the service. Since the permission is applied directly on the service such that it applies for all its children, the `resource` shouldn't be needed. 
The config might be completely valid, I just don't remember if Magpie code handled the `/` correctly to resolve as the `service`. If it does, it's good to leave the config as is.

fmigneault (commented):
Same comments as @tlvu regarding the `proxy_pass`. The rest looks good.

birdhouse/config/geoserver/config/proxy/conf.extra-service.d/geoserver.conf.template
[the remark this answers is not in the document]
mishaschwartz: I've put the check behind the verify endpoint and introduced the `GEOSERVER_SKIP_AUTH` setting to skip checking twitcher entirely if needed. 

I've added comments in a few places to make it clear that we do not recommend using `GEOSERVER_SKIP_AUTH` since it opens up geoserver to full public access without any checks.

birdhouse/config/geoserver/config/proxy/conf.extra-service.d/geoserver.conf.template
```diff
@@ -14,6 +14,7 @@
 
     location = /secure-geoserver-auth {
         internal;
+        ${GEOSERVER_SKIP_AUTH_PROXY_INCLUDE}
```
fmigneault, line 17: Can you add a space after this and a comment before the following `proxy_pass` that highlights that this part is skipped or nor according to `GEOSERVER_SKIP_AUTH` value? This can help figure out what is happening if looking at the resulting file after template variables were applied.

fmigneault (approved):
LGTM. Thanks!

birdhouse/config/geoserver/default.env
```diff
@@ -26,9 +26,14 @@ export GEOSERVER_COMMUNITY_EXTENSIONS="geopkg-plugin"
 # Must use single-quote for delayed eval.
 export GEOSERVER_DATA_DIR='${DATA_PERSIST_ROOT}/geoserver'
 
+# If set, requests to the geoserver endpoint will not be authorized through twitcher/magpie
+export GEOSERVER_SKIP_AUTH=False
+export GEOSERVER_SKIP_AUTH_PROXY_INCLUDE='$([ x"${GEOSERVER_SKIP_AUTH}" = x"True" ] && echo "return 200;")'
```
tlvu, line 31: @mishaschwartz Thanks for adding `GEOSERVER_SKIP_AUTH` this PR looks good for me as well.  

Can you eventually add a similar switch for THREDDS.  Currently we are applying a custom hack on our fork of this repo for production to disable Twitcher in front of our Thredds.  Performance increased between 3 and 4x.

@fmigneault is there a DB upgrade?  If no upgrade I won't need to deploy this PR to test and can approuve straight.

tlvu:
> We need to coordinate merging this PR because of:
> 
> > A Magpie service named `geoserver` with type `wfs` exists already and must be manually deleted before the new Magpie service created here can take effect.
> 
> Unless there is a configuration option that I don't know about to delete/rename an existing magpie service that I don't know about @fmigneault ?

OH!  That would not be a seemless upgrade !  @fmigneault  Could the DB upgrade be made backward compatible?

mishaschwartz:
One option would be to just name the new `geoserver` service something different. The old services would still be around, they just would be redundant.

birdhouse/config/geoserver/default.env
fmigneault: No DB upgrade in Magpie. The existing `geoserser` type is extended.

fmigneault:
@mishaschwartz @tlvu 

> OH! That would not be a seemless upgrade ! @fmigneault Could the DB upgrade be made backward compatible?

No. There is no DB upgrade involved. It is 2 distinct service types that remain available. Magpie shouldn't really perform changes automatically on the data itself, as this would assume that a specific service *entry* name exists and must be modified.  DB upgrade is for handling schema modifications only.

> One option would be to just name the new geoserver service something different. The old services would still be around, they just would be redundant.

This would require another path as well, since the name is used in `/twitcher/ows/proxy/{service_name}`. 
Any scripts targeting `/twitcher/ows/proxy/geoserver` will not work anymore.

> Unless there is a configuration option that I don't know about to delete/rename an existing magpie service that I don't know about @fmigneault ?

No. This is something that must be done manually in the database.
Magpie will recreate the service from scratch if booted directly with a new type. This will also destroy any child resource/permission under the service that was deleted.

Note that what causes the "recreate from scratch" is the change of `url: http://proxy:80` in the new `geoserver` definition. 
If only the `type` and `configuration` where modified, the service would be updated and would preserve the existing hierarchy. 
See code: https://github.com/Ouranosinc/Magpie/blob/1d602eedbe4047fd95c4d2684b5731f4d0600977/magpie/register.py#L490-L500
 
To preserve the whole resource/permission hierarchy, it is probably more effective to rename `services.type = geoserver` field in the DB for the existing service `geoserver` (type: `wfs`). This new `configuration` will be updated automatically on startup when parsing the config files and detecting the change (assuming `url` was not changed). Otherwise, `configuration` must be added as well.  

Then, if some resource/permissions were set under for `geoserverwms`, its children resources would need an UDPATE of the `parent_id` field to point at `resource_id` of `geoserver` directly. 

To preserve any hierarchy that could already exist, I think it is safer to do the manual edit of `url` to the expected `http://proxy:80` while editing the DB before starting Magpie.

mishaschwartz:
Let me summarize the two problems:

1. There exists a service `geoserver` (type = wfs), when we try to create a service `geoserver` (type = geoserver) Magpie complains that a resource with that name already exists and _does not create it at all_
2. Permissions that are currently set for groups and users for the `geoserver` (type = wfs) and `geoserverwms` services are not translated over to the new `geoserver` service

If I understand what you're saying @fmigneault the solutions to these problems are:

1. manually change the url of `geoserver` (type = wfs) to `http://proxy:80` before applying the changes in this PR
    OR
    manually change the type of `geoserver` (type = wfs) to type = geoserver, before applying the changes in this PR
    OR 
    both?
2. manually change the parent id of resources and permissions from `geoserverwms` to the new `geoserver` (type = geoserver), after applying the changes in this PR

How do you recommend we achieve this? Should we provide a script that can be run in the magpie-postgres docker container to run these updates?

mishaschwartz:
Here's the relevant section of the Magpie logs:

```
[2023-10-03 16:46:11,894] WARNING    [MainThread][magpie.register] Services configuration are empty.
[2023-10-03 16:46:11,925] INFO       [MainThread][magpie.register] Service URL already properly set [http://cowbird:7000] (cowbird)
[2023-10-03 16:46:11,927] INFO       [MainThread][magpie.register] Service URL already properly set [http://finch:5000/wps] (finch)
[2023-10-03 16:46:11,929] INFO       [MainThread][magpie.register] Service URL update [http://geoserver:8080/geoserver] => [http://proxy:80] (geoserver)
[2023-10-03 16:46:11,933] INFO       [MainThread][magpie.register] Service URL already properly set [http://hummingbird:8080/wps] (hummingbird)
[2023-10-03 16:46:11,934] INFO       [MainThread][magpie.register] Service URL already properly set [http://raven:9099/wps] (raven)
[2023-10-03 16:46:11,936] INFO       [MainThread][magpie.register] Service URL already properly set [http://proxy:80] (secure-data-proxy)
[2023-10-03 16:46:11,944] INFO       [MainThread][magpie.register] Service URL already properly set [http://stac:8000] (stac)
[2023-10-03 16:46:11,946] INFO       [MainThread][magpie.register] Service URL already properly set [http://geoserver:8080/geoserver] (geoserver-secured)
[2023-10-03 16:46:11,947] INFO       [MainThread][magpie.register] Service URL already properly set [http://thredds:8080/twitcher/ows/proxy/thredds] (thredds)
[2023-10-03 16:46:11,950] INFO       [MainThread][magpie.register] All services processed.
```

The URL for the geoserver (type = wfs) is updated but the actual "type" is not updated. Since Magpie thinks that it has processed the changes for this provider, it doesn't create a new service with type = geoserver or update to current service to type = geoserver.

I think that we're missing a line like:

```
svc.type = svc_type
```

here: https://github.com/Ouranosinc/Magpie/blob/1d602eedbe4047fd95c4d2684b5731f4d0600977/magpie/register.py#L496-L500

fmigneault:
@mishaschwartz 
Indeed. It's missing `svc.type`. I misread `svc.sync_type` and had in mind the type was updated as well.

mishaschwartz:
>  Indeed. It's missing svc.type. I misread svc.sync_type and had in mind the type was updated as well.

Ok cool. Should we update that on the Magpie side then? This PR can wait until Magpie has been updated

fmigneault:
@mishaschwartz @tlvu https://github.com/Ouranosinc/Magpie/pull/590

fmigneault:
https://github.com/Ouranosinc/Magpie/tree/3.36.0 pushed. The docker image should be ready in ~10min.

fmigneault:
@mishaschwartz 
> With this update, we still need to manually do steps 2 and 3 from here first right? [#348 (comment)](https://github.com/bird-house/birdhouse-deploy/pull/348#issuecomment-1745310963)

yes

tlvu (approved):
LGTM, as long as the new Magpie has no DB upgrade.  We have a very old DB in production so we've had many db upgrade problem in the past.

mishaschwartz:
@tlvu 

If you have permissions in magpie set for the `geoserverwms` service then you'll need to update the database (see https://github.com/bird-house/birdhouse-deploy/pull/348#issuecomment-1755977142 and https://github.com/bird-house/birdhouse-deploy/pull/348#issuecomment-1745310963 for details). Note that this does not require an update to the database schema, you may just need to move some settings around which can be done with the SQL commands that @fmigneault mentions in the comment above or manually through the UI.

CHANGES.md
```diff
@@ -17,6 +17,37 @@
 
 [//]: # (list changes here, using '-' for each new entry, remove this when items are added)
 
+[1.37.0](https://github.com/bird-house/birdhouse-deploy/tree/1.37.0) (2023-11-01)
+------------------------------------------------------------------------------------------------------------------
+
+## Changes
+- Geoserver: protect web interface and ows routes behind magpie/twitcher
+ 
+  Updates Magpie version to [3.35.0](https://github.com/Ouranosinc/Magpie/tree/3.35.0) in order to take advantage of
```
tlvu, line 26: Woops, the actual Magpie in this PR is 3.36 and not 3.35 !

tlvu:
@fmigneault @mishaschwartz 

I test deploy this against our Magpie production DB, got the following warning, just want to ensure these are normal and expected:
```
[2024-04-12 03:31:36,349] WARNING    [MainThread][magpie.register] Service URL update [http://geoserver:8080/geoserver] => [http://proxy:80] (geoserver)
[2024-04-12 03:31:36,350] WARNING    [MainThread][magpie.register] Service type update [wfs] => [geoserver] (geoserver). If children resources/permissions
 are not compatible, this could break the instance.  
```

fmigneault:
@tlvu 
Yes. This is intentional. The request to GeoServer should now use Twitcher's `/ows/verify` as a pre-auth request instead of `/ows/proxy` request going through Twitcher, which was causing the memory/latency issues.
See: 
https://github.com/bird-house/birdhouse-deploy/blob/ee4087f111e3b487e1610d5d641de68f581f3845/birdhouse/config/geoserver/config/proxy/conf.extra-service.d/geoserver.conf.template#L2-L4

PR #356, opened 2023-07-13
Make bind-mount locations more flexible

### LINKED ISSUE #355: :bulb: [Feature] Make bind-mount locations configurable
## Description

Some bind mind locations are configurable so the user can choose where on disk the data should be located (for example: a user can set the `JUPYTERHUB_USER_DATA_DIR` variable). 

Others are partially configurable in that users can choose the name of a subdirectory of the directory specified by the `DATA_PERSIST_ROOT` variable (for example: a user can set the `USER_WORKSPACES` environment variable)

Others must be located under the directory specified by `DATA_PERSIST_ROOT` (for example: `"${DATA_PERSIST_ROOT}/magpie_persist"`)

It would be better to allow more/all such directories to be mounted from a customizable location on disk. This would allow the user to choose the best configuration for their own set-up. 

For example, they may choose to store data that will be regularly modified to a disk with high write speeds whereas they may choose to store data that will rarely be updated to a different disk entirely.

## Concerned Organizations

@fmigneault @tlvu

mishaschwartz:
## Overview

Previously, most bind mount locations on the host machine were subdirectories of the folder specified by the `DATA_PERSIST_ROOT` environment variable (`/data` by default). This change allows the user to set custom locations for the following additional variables, so that they don't need to be all under the same common directory.

  - `LOGROTATE_DATA_DIR` (default: `${DATA_PERSIST_ROOT}/logrotate`)
  - `MONGODB_DATA_DIR` (default: `${DATA_PERSIST_ROOT}/mongodb_persist`)
  - `POSTGRES_DATA_DIR` (default `${DATA_PERSIST_ROOT}/frontend_persist`)
  - `USER_WORKSPACES` (default `${DATA_PERSIST_ROOT}/user_workspaces`)
  - `WEAVER_MONGODB_DATA_DIR` (default `${DATA_PERSIST_ROOT}/mongodb_weaver_persist`)
  - `WEAVER_WPS_PRIVATE_DIR` (default `${DATA_PERSIST_ROOT}/wps_private`)

## Changes

**Non-breaking changes**
- Adds additional configuration options but all defaults remain the same as before

**Breaking changes**

None

## Related Issue / Discussion

- Resolves #355 

## Additional Information

- [ ] Things to do...


CHANGES.md
```diff
@@ -15,7 +15,19 @@
...
+- Make bind-mount locations more flexible
+
+  Previously, most bind mount locations on the host machine were subdirectories of the folder specified by the 
+  `DATA_PERSIST_ROOT` environment variable (`/data` by default). This change allows the user to set custom locations
+  for the following additional variables, so that they don't need to be all under the same common directory.
+
+  - `LOGROTATE_DATA_DIR` (default: `${DATA_PERSIST_ROOT}/logrotate`)
+  - `MONGODB_DATA_DIR` (default: `${DATA_PERSIST_ROOT}/mongodb_persist`)
+  - `POSTGRES_DATA_DIR` (default `${DATA_PERSIST_ROOT}/frontend_persist`)
+  - `USER_WORKSPACES` (default `${DATA_PERSIST_ROOT}/user_workspaces`)
```
dchandan, line 28: Suggestion: maybe this can be named `USER_WORKSPACES_DIR` to be consistent with other `*_DIR` names?

dchandan (approved):

CHANGES.md
mishaschwartz: this variable is already in use so I kept it this way for backwards compatibility

mishaschwartz:
@ChaamC do you have time to chat about whether this change will affect your work with cowbird?

ChaamC:
@mishaschwartz 
In my current work on Cowbird (briefly mentioned by @fmigneault on [issue#355](https://github.com/bird-house/birdhouse-deploy/issues/355#issuecomment-1635005749)), we need to be able to create hardlinks between the folder of the wpsoutputs data and the user workspaces folder. They both need to be in a common parent directory so that the parent directory can be mounted as a volume to generate hardlinks correctly.
This is working well if both are in `/data` (or `${DATA_PERSIST_ROOT}`. 

I think it would be better not to have those variables (`${USER_WORKSPACES}` and an eventual `${WPS_OUTPUTS_DIR}`) as customizable, to ensure they do not break those new Cowbird functionalities.

Also, using a customizable wps outputs directory could also break weaver.

As for the other variables concerned by this PR, I haven't worked with them, but I don't think they will have an impact on Cowbird right now.

From my point of view, there's just the problem of customizing the `USER_WORKSPACES` variable which could impact Cowbird.

birdhouse/components/cowbird/docker-compose-extra.yml
```diff
@@ -36,7 +36,7 @@ services:
       - ./components/cowbird/config/cowbird/cowbird.ini:/opt/local/src/cowbird/config/cowbird.ini
       # even if not running tasks here, they must be registered to send them off to the right place!
       - ./components/cowbird/config/cowbird/celeryconfig.py:/opt/local/src/cowbird/config/celeryconfig.py
-      - "${DATA_PERSIST_ROOT}/${USER_WORKSPACES}:/${USER_WORKSPACES}"
+      - "${USER_WORKSPACES}:/user_workspaces"
```
fmigneault, line 39: As mentioned by @ChaamC, this could easily break Cowbird.
Something tricky about these volumes is that Cowbird creates dynamic hardlinks between different locations that must resolve properly once mounted. However, even though Cowbird sees the mounted location `/user_workspaces` inside its docker container, the hardlinks it must create must resolve properly on the host side, such that once this link is mounted elsewhere such as in Jupyter user-specific dockers, the file still resolves correctly with a reduced/nested directory of the specific user-workspace. 

@ChaamC and I have had a lot of discussions and did many exploration tests to try to resolve this already complicated problem. I would prefer not to introduce customizable paths here, to avoid Cowbird having to deal with complicated inverse-volume-mapping resolution to create the links correctly.

Something that could be considered is to have some kind of `DATA_PERSIST_SHARED_DIR` (instead of  `DATA_PERSIST_ROOT`) where stuff like user-workspaces and wps-outputs that must share locations between services are all nested under. However, the `${DATA_PERSIST_SHARED_DIR}/user_workspaces` and `${DATA_PERSIST_SHARED_DIR}/wps_outputs` would be hardcoded to ensure they work properly.

birdhouse/components/weaver/default.env
```diff
@@ -96,7 +96,13 @@ export WEAVER_WPS_PROVIDERS_RETRY_COUNT=5
 # control interval time between retries (duration in seconds, counts toward maximum timeout)
 export WEAVER_WPS_PROVIDERS_RETRY_AFTER=5
 
+export WEAVER_MONGODB_DATA_DIR='${DATA_PERSIST_ROOT}/mongodb_weaver_persist'
+
+export WEAVER_WPS_PRIVATE_DIR='${DATA_PERSIST_ROOT}/wps_private'
```
fmigneault, line 101: This one might need to be revisited. 
Might be deprecated considering the switch to `wpsoutputs` with nested user-context managed by Cowbird.

fmigneault (commented):
Except the `USER_WORKSPACES`, all others should be ok to allow overrides.

ChaamC:
@mishaschwartz 
I think we can procede with `DATA_PERSIST_SHARED_DIR`, as it was suggested by @fmigneault.
For Cowbird, I need `user_workspaces` and `wpsoutputs` to have a common parent directory for valid hardlinks, so the solution should be compatible with these upcoming changes.

birdhouse/components/weaver/default.env
mishaschwartz: @fmigneault are you suggesting that this shouldn't be included in this PR or just that this might be deprecated some time in the future?

mishaschwartz:
@ChaamC I've made the changes suggested (I ended up naming the variable `DATA_PERSIST_SHARED_ROOT` to be analogous with `DATA_PERSIST_ROOT`)

Please let me know if the solution is compatible with the cowbird changes or if I've missed anything

ChaamC (changes requested):
Good job!
Minor comments, but looks good on my side. I tried it with my upcoming Cowbird changes and was able to make it work without trouble. FYI, I have just opened a PR with my upcoming changes [#360](https://github.com/bird-house/birdhouse-deploy/pull/360). I will eventually update it with your changes, as I cannot merge it for the moment anyway.

It would be good to get @fmigneault's approval too, especially for the weaver part that he is more familiar with. Unfortunately he is in vacation, I think for 2 weeks.

birdhouse/default.env
```diff
@@ -11,6 +11,11 @@ export BASH_IMAGE="bash:5.1.4"
 # Root directory under which all data persistence should be nested under
 export DATA_PERSIST_ROOT="/data"
 
+# shellcheck disable=SC2016
+# Root directory for all files that are persisted on disk and may contain links (ie. the files
+# are "shared" between subdirectories). This means that the subdirectory structure is fixed.
```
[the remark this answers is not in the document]
mishaschwartz: I think I just meant the files in this directory should not be moved once created so that the links don't break. But I agree that the wording is not great. I'll update it.

birdhouse/components/weaver/default.env
fmigneault: Yes. I think it will be deprecated eventually.

mishaschwartz:
Should now include all changes from #360

birdhouse/default.env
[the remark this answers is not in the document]
fmigneault: On Weaver front, it should be valid as well since WPS outputs dir remains mounted the same way on both host/docker side.

fmigneault (commented):
I do not recall why, but `WEAVER_WPS_OUTPUTS_DIR` (which depends on `WPS_OUTPUTS_DIR`, `DATA_PERSIST_ROOT` and now the alternative `DATA_PERSIST_SHARED_ROOT`) is defined as `DELAYED_EVAL` variable.

Similarly, `COWBIRD_MONGODB_DATA_DIR` is in `DELAYED_EVAL`. 

Does that mean that all those variable dependencies should also be marked as `DELAYED_EVAL`?

Since the integration of Cowbird to manage private user workspace vs public ones under a common (shared) data persistence dir, and Weaver using the same approach for WPS outputs using the Magpie hook + `X-WPS-Output-Context` header, I suggest getting rid of `WEAVER_WPS_PRIVATE_DIR` altogether.

mishaschwartz:
> Does that mean that all those variable dependencies should also be marked as DELAYED_EVAL?

Yes, and they all should be already. If I've missed one let me know.

>  I suggest getting rid of WEAVER_WPS_PRIVATE_DIR altogether.

Ok, done

ChaamC (approved):

fmigneault (approved):

birdhouse/components/weaver/default.env
```diff
@@ -96,7 +96,12 @@ export WEAVER_WPS_PROVIDERS_RETRY_COUNT=5
...
 export WEAVER_WPS_PROVIDERS_RETRY_AFTER=5
 
+export WEAVER_MONGODB_DATA_DIR='${DATA_PERSIST_ROOT}/mongodb_weaver_persist'
+
+
 export DELAYED_EVAL="
   $DELAYED_EVAL
   WEAVER_WPS_OUTPUTS_DIR
+  WEAVER_MONGODB_DATA_DIR
+  WEAVER_WPS_PRIVATE_DATA_DIR
```
tlvu, line 106: Left-over not deleted `WEAVER_WPS_PRIVATE_DATA_DIR` ?

birdhouse/components/weaver/docker-compose-extra.yml
```diff
@@ -97,7 +96,7 @@ services:
     networks:
       - weaver-mongodb
     volumes:
-      - ${DATA_PERSIST_ROOT}/mongodb_weaver_persist:/data/db
+      - ${WEAVER_MONGODB_DATA_DIR}/mongodb_weaver_persist:/data/db
```
tlvu, line 99: Should this be `${WEAVER_MONGODB_DATA_DIR}:/data/db` instead?

tlvu (commented):
Sorry for the late review, I wasn't tagged as reviewer on this PR initially.

Spotted a wrong volume-mount path.

Issue #334, opened 2023-06-01
:bulb: [Feature]


fmigneault:
## Description

Ensure JupyterHub is accessed *behind* Magpie/Twticher authentication/authorization. 
Currently, it uses Magpie/Twticher after the fact to login the user, but the initial request and subsequent ones are not protected by default. Therefore, any user can still reach the entrypoints (though they cannot login).

## References

- https://github.com/bird-house/birdhouse-deploy/pull/331#discussion_r1210541774

tlogan2000:
I supposed my quick thought would be to agree with @huard but certain changes need to be made to the magpie page in my opinion ... Currently magpie signin  is very much aligned to an admin user (options for user, service perms etc.) visible even if unaccessable after login : see https://pavics.ouranos.ca/magpie/.    The basic user would have to see something like 'Please sign in here'  then a  'take me to the Jupyterlab ' type navigation.  Possibily an account info / change password option

huard:
Is it possible to rig this so that if I go to pavics.ouranos.ca/jupyter, once the user signs-in in Magpie, it goes directly back to Jupyter (ie you're not stuck in magpie) ?

Also, could we "re-brand" the magpie sign-in page ? I'm concerned if people see Magpie and its logo, they'll think they left the DACCS node.

mishaschwartz:
@huard

> Is it possible to rig this so that if I go to pavics.ouranos.ca/jupyter, once the user signs-in in Magpie, it goes directly back to Jupyter (ie you're not stuck in magpie) ?

Yes that should be possible

> Also, could we "re-brand" the magpie sign-in page ? I'm concerned if people see Magpie and its logo, they'll think they left the DACCS node.

Another option is to leave magpie alone and create a separate "DACCS branded" sign in page that makes a call to the magpie api (see: https://pavics-magpie.readthedocs.io/en/latest/api.html#tag/Session%2Fpaths%2F~1signin%2Fpost). That way we don't have to force magpie to be something it isn't.

huard:
I like the API call idea. Good for me. 
@tlvu Thoughts ?

fmigneault:
I also agree with the option 2 (keep the jupyterhub sign in page as the default but also have the jupyterhub sign in trigger the magpie sign in). I think it is much easier to use JupyterHub as the entrypoint, because even if we login on Magpie first, we would not have the JupyterHub session ID and session cookie... 

When JupyterHub does the login, it performs some Magpie login using this:
[...]

PR #358, opened 2023-07-14
Protect jupyterhub behind twitcher authentication

mishaschwartz:
## Overview

Also sets magpie cookies whenever a user logs in through jupyterhub so that they are automatically logged in through magpie as well.

## Changes

**Non-breaking changes**
- adds jupyterhub as a provider in magpie so that admin users can set api permissions in magpie for jupyterhub

**Breaking changes**

## Related Issue / Discussion

- implements step 1 from this comment: https://github.com/bird-house/birdhouse-deploy/issues/334#issuecomment-1591178151

## Additional Information

- [ ] Move authenticator changes to the github.com/Ouranosinc/jupyterhub project
- [ ] Add custom logout handler so that logout in jupyterhub also logs out of magpie


mishaschwartz:
@tlvu I've moved the jupyterhub changes to https://github.com/Ouranosinc/jupyterhub/pull/17 > I'm not a member of that repo so I can't directly request a review. So can you (or the relevant person at Ouranos) please review that PR. 

When that one is pulled in, I will update this one to reflect the changes.

tlvu:
Tried this PR, getting 502 Bad Gateway when trying to load the Jupyter login page.

So the user have to first go to `/magpie`, then navigate back to `/jupyter`?  This is a bit awkward.

```
$ curl https://lvupavicsmaster.ouranos.ca/jupyter/hub/login --include --silent
HTTP/1.1 502 Bad Gateway
Server: nginx/1.23.4
Date: Mon, 28 Aug 2023 19:14:46 GMT
Content-Type: text/html
Content-Length: 497
Connection: keep-alive
ETag: "64230162-1f1"

<!DOCTYPE html>
<html>
<head>
<title>Error</title>
<style>
html { color-scheme: light dark; }
body { width: 35em; margin: 0 auto;
font-family: Tahoma, Verdana, Arial, sans-serif; }
</style>
</head>
<body>
<h1>An error occurred.</h1>
<p>Sorry, the page you are looking for is currently unavailable.<br/>
Please try again later.</p>
<p>If you are the system administrator of this resource then you should check
the error log for details.</p>
<p><em>Faithfully yours, nginx.</em></p>
</body>
</html>
```

tlvu:
Tried to first login to `/magpie` then navigate to `/jupyter` but still unable to load the login page.  It seems the proxy is unable to connect to jupyterhub.

docker logs proxy:
2023/08/28 19:33:40 [error] 17#17: *615 connect() failed (111: Connection refused) while connecting to upstream, client: 10.10.10.11, server: localhost, request: "GET /jupyter/hub/login HTTP/1.1", upstream: "http://172.22.0.33:8000/jupyter/hub/login", host: "lvupavicsmaster.ouranos.ca"

mishaschwartz:
@tlvu 

> Tried this PR, getting 502 Bad Gateway when trying to load the Jupyter login page.

This is because of the upgrade of jupyterhub versions (we were using one that was really really out of date, I don't think that we ever updated the version in this repo when we updated the github.com/Ouranosinc/jupyterhub project). 

You may need to update the jupyterhub database for running instances (as you already noticed): 

```
./pavics-compose.sh run jupyterhub jupyterhub upgrade-db
```

I will add a comment 

> So the user have to first go to /magpie, then navigate back to /jupyter? This is a bit awkward.

No, they just need to log in on jupyter and it will log them in to magpie as well. If you're seeing something different please let me know with some more details but that is not what I'm seeing when I test this PR. Once I upgrade the database, the PR works as intended for me. Did you also restart proxy after you upgraded the database?

> Tried to first login to /magpie then navigate to /jupyter but still unable to load the login page. It seems the proxy is unable to connect to jupyterhub.

If you log in to jupyterhub, it will log you in to magpie, not the other way around. Please see the full discussion here (https://github.com/bird-house/birdhouse-deploy/issues/334). This PR specifically only implements step one described in this comment (https://github.com/bird-house/birdhouse-deploy/issues/334#issuecomment-1591178151)

tlvu:
> You may need to update the jupyterhub database for running instances (as you already noticed):

I did.  It even created a backup of the old file.

```
# ls /var/lib/docker/volumes/jupyterhub_data_persistence/_data/
jupyterhub_cookie_secret  jupyterhub.sqlite  jupyterhub.sqlite.2023-08-28-192536
```

> Did you also restart proxy after you upgraded the database?

I even did a full `./pavics-compose.sh down && ./pavics-compose.sh up -d` just to be sure.

I'd restore the old DB and perform the upgrade again, maybe it failed !

tlvu:
So indeed it was the DB upgrade that did not worked the first time, not sure why.

By the way, we should add `--rm` to `./pavics-compose.sh run --rm jupyterhub jupyterhub upgrade-db` to avoid leaving stopped container behind.

Also, I just got this problem twice

Screenshot from 2023-08-29 23-09-22

Will return my VM to `master` branch state to see if I can reproduce.  If not, then probably something with verify trick might trigger this.  If still reproducible, then probably my VM is too weak again and I have to increase its resources.  I try to replicate the exact same config from prod, less all the data and lower cronjob frequency.

This is the `climex.ipynb` notebook, one of the heavy ones in our collection of tutorial notebooks that I often use when I need to benchmark something in Jupyter.

tlvu:
After many increase in RAM and CPU for my VM, I am finally able to run the `climex.ipynb`.  Then I which to the branch of this PR and again I am unable to run the notebook.

I notice my RAM was completely maxed out with lots of IO Wait.
Screenshot from 2023-08-30 18-26-43

So I do not think we should go ahead with this PR before understanding the performance penalty of the verify trick.

The Jupyterhub already has a login page so I don't think we are losing any security aspect.  We are only losing additional integration with Magpie.

mishaschwartz:
I'm surprised the verify method is so IO heavy. 

> The Jupyterhub already has a login page so I don't think we are losing any security aspect. We are only losing additional integration with Magpie.

Agreed, we could still check the magpie permissions at login time only though which would get us 90% of the way there. We'd only be missing out on the fine-grained magpie permissions for jupyterhub but I can't imagine a good use case for those anyway. I'll make an update to this PR with that proposed change shortly.

> So I do not think we should go ahead with this PR before understanding the performance penalty of the verify trick.

This is only part of the goal of this PR. The main thing that this PR solves is the ability to log in to magpie when we log in to jupyterhub in order to have a consistent login experience across the platform. We can still do this without the need to run everything through twitcher.

I'll make a few changes to avoid the twitcher verify at every request.

tlvu:
> I'm surprised the verify method is so IO heavy.

We have had many many performance problem in the past with the Twitcher/Magpie combo.  There is a reason why there is a `stress-test.ipynb` in the Jenkins test suite.  I probably should open an issue to group together all the various past and current issues about Twitcher/Magpie performance.

That said, in the current PR, there is also another variable, the JupyterHub update.  Maybe the issue is due to both JupyterHub and Twitcher/Magie?  Do you think the new Magpie integration can work in the existing JupyterHub version?

fmigneault:
> I'm surprised the verify method is so IO heavy.

Maybe it is not the verify method itself that is heavy, but rather that Jupyter could be sending a lot of requests for loading images, CSS, etc.? 
It could be possible to filter the `/jupyter` proxy redirect to limit verify checks only for certain subsets of paths/content-types. 
Maybe that could help reduce the load?

fmigneault:
As I suspected, Jupyter does an initial 40-50 requests to load its interface, and then a subsequent 3-5 request every second or so to monitor active sessions, kernel, terminal and metrics. All of those would be redirected to the auth verify endpoint. Therefore, the Twitcher gunicorn workers (

fmigneault (approved):

mishaschwartz:
What if we only protected the initial `/hub/login` route with Magpie. That way we still allow admins to restrict access to specific users (by not giving them access to the jupyterhub login page) but all other requests go to the jupyterhub proxy directly instead, so we don't have all of those additional requests going through magpie once they've logged in. 

We would lose the ability to restrict access to other jupyterhub routes through Magpie but I'm not sure if there's really a good use-case for that? As long as an admin has the ability to pick which users have access to jupyter through Magpie, that's good enough for me.

fmigneault:
> What if we only protected the initial `/hub/login` route with Magpie. 

Wouldn't that mean that `/jupyterhub/` would basically be exposed? If I manually set the necessary sessionId/cookie/Authorization headers in my request, I could directly target that endpoint, bypassing Magpie/Twitcher auth. The token wouldn't even need to be valid, just that it is there to fulfill the requirement that redirects to login, because it would not be validated against Magpie. 

If I misunderstood, please detail more the solution you have in mind.

> That way we still allow admins to restrict access to specific users (by not giving them access to the jupyterhub login page) 

Just to make sure, but that should be possible with the current change. 
Just need to set (eg: `group:users, allow` and `group:anonymous, deny`) on that login endpoint.

mishaschwartz:
> Wouldn't that mean that /jupyterhub/ would basically be exposed? If I manually set the necessary sessionId/cookie/Authorization headers in my request, I could directly target the /jupyter/hub/login route, bypassing Magpie/Twitcher auth. The token wouldn't even need to be valid, just that it is there to fulfill the requirement to skip redirect to login, because it would not be validated against Magpie.

You're right about that. What I think I had in mind was to protect the login request that jupyterhub makes to Magpie. But I don't really know how I'd go about doing that since that's just a `/signin` request which has to be publicly available.

Ok what about this... what if when we log in to jupyterhub it tries to login to magpie and then makes a request to `/ows/verify/jupyterhub` and doesn't log in the jupyterhub user if it gets a 403. But the nginx proxy doesn't interact with Magpie at all.
Basically, this means that the jupyterhub login is responsible for checking whether a user has permission to log in according to Magpie. 

Is this getting too complicated?

> Just to make sure, but that should be possible with the current change.

Yes, it's possible with the current change. I just want to make sure it stays.

fmigneault:
> what if when we log in to jupyterhub it tries to login to magpie and then makes a request to `/ows/verify/jupyterhub` and doesn't log in the jupyterhub user if it gets a 403

Maybe that could work, but we would have to evaluate if a user could be able to handcraft a request that would bypass the login step. Right now, I'm able to call the `/jupyter/` endpoint by skipping the login step if my browser happens to have some specific Cookie set with a `jupyterhub-session-id`, even though I am not logged in Magpie. 

Also, we would have to consider how the user experience feels.
For example, after login in Magpie/Jupyter (if they work together), I can easily switch to any other service endpoint of the platform without having to do another login in Magpie.

mishaschwartz:
> Maybe that could work, but we would have to evaluate if a user could be able to handcraft a request that would bypass the login step. Right now, I'm able to call the /jupyter/ endpoint by skipping the login step if my browser happens to have some specific Cookie set with a jupyterhub-session-id, even though I am not logged in Magpie.

Yes that makes sense. Jupyterhub stores a session cookie separate from Magpie. If a user logs in through jupyterhub it will set both cookies but jupyterhub really only cares about its own cookie for its own authentication purposes.

I don't think this edge case is a problem. A user can manipulate their cookies to skip the jupyterhub login step but that doesn't buy them any additional access and I don't think that will be something that a user is likely to do by accident.

> Also, we would have to consider how the user experience feels.
> For example, after login in Magpie/Jupyter (if they work together), I can easily switch to any other service endpoint of the platform without having to do another login in Magpie.

Yes, good idea. That's definitely possible with the current changes proposed here. Is there a change that you're suggesting?

fmigneault:
> I don't think this edge case is a problem. A user can manipulate their cookies to skip the jupyterhub login step but that doesn't buy them any additional access and I don't think that will be something that a user is likely to do by accident.

Well, given that this allows them to spawn a docker with jupyter instance running, if the authentication proxy portion is removed and that the instance must restrict this access to certain users, then it becomes a problem. 

> Yes, good idea. That's definitely possible with the current changes proposed here. Is there a change that you're suggesting?

No. I like this capability. I was only referring to the other brainstorming approach that would skip the Magpie/Twitcher proxy. If that method has a side effect to lose this capability, it makes it even less desirable.

tlvu (changes requested):
Set status to "request change" for the performance problem.

mishaschwartz:
> Well, given that this allows them to spawn a docker with jupyter instance running, if the authentication proxy portion is removed and that the instance must restrict this access to certain users, then it becomes a problem.

I don't understand what you think is happening that would allow the user to spawn a docker container without logging in to jupyterhub.

Right now, all that we're doing is logging in to Magpie when we log in to jupyterhub. Jupyterhub still enforces all of its own authentication and authorization. We're just suggesting to add another authorization layer on top of that with Magpie. We're adding more authorization layers, not removing any.

fmigneault:
> I don't understand what you think is happening that would allow the user to spawn a docker container without logging in to jupyterhub.

I have managed to skip the login portion of JupyterHub simply by setting some jupyter session id and some cookies in the right combination in my browser. Just like you mentionned before, Jupyterhub only checks for its own auth cookies. Right now, because Magpie is also needed to perform an actual login when spawning the docker for the user, it protects on that front even if jupyterhub cookies were bypassed. 

My comment was referring to the brainstorm discussion where potentially removing that Magpie login validation layer could then open access to Jupyterhub when doing this hack. 
However, if the Magpie login step remains present, whether all Jupyterhub requests are behind the proxy or only the login, I don't see any immediate issue.

mishaschwartz:
@tlvu 

I've added a new proposed feature the the MagpieAuthenticator that checks if the user that is logging in has access to jupyterhub. Please see details here: https://github.com/Ouranosinc/jupyterhub/pull/19

This will allow us to check whether a user is allowed to *log in* to jupyterhub but will no longer check at every request (by querying twitcher's ows/verify endpoint). I believe that this is a good compromise that will allow admins some control over who is allowed to access jupyterhub while simultaneously not affecting performance.

 @fmigneault 

This only checks for authorization at the login step so this does not address the issue you raise here:

> I have managed to skip the login portion of JupyterHub simply by setting some jupyter session id and some cookies in the right combination in my browser.

But I'm assuming that the only way to get that session id and cookie is to have previously gone through the login step, am I right?

So this solution still works for our purposes, because _at some point_ you need to log-in in order to acquire the access you need.

If I'm not understanding what you are saying please let me know.

mishaschwartz:
Also ...

DO NOT MERGE THIS PR UNTIL https://github.com/Ouranosinc/jupyterhub/pull/19 is merged and the pavics/jupyterhub version has been updated here!

fmigneault:
> But I'm assuming that the only way to get that session id and cookie is to have previously gone through the login step, am I right?
> 
> So this solution still works for our purposes, because _at some point_ you need to log-in in order to acquire the access you need.

Somewhat. 
What you would need to do to test it is to perform a Magpie login (this is just to succeed the current setup login and make the operation quicker, otherwise it is troublesome to handcraft the tokens manually). Then, after the Jupyter login was completed and you land on its notebook interface, copy the session cookies that Jupyterhub sets in the browser. 
After that, logout everywhere (or use another browser), and re-login in Magpie using another user that shouldn't be allowed access to Jupyterhub with your new feature. If you can then gain access to Jupyterhub by dumping the copied session cookies in the browser, then the feature can be bypassed. If not, then the proposed method is probably a good candidate.

mishaschwartz:
@fmigneault 

Ok I understand what you mean. You're saying that if an attacker can steal a session cookie, they can access jupyterhub even if the magpie permissions have changed. Yes, this is currently possible, and has been possible all along.

This is an outstanding issue already and has been raised here: https://github.com/Ouranosinc/jupyterhub/issues/2

The best solution is to implement `Authenticator.refresh_user(user)` as discussed in that issue. The other option is to protect jupyterhub paths behind `twitcher/ows/verify` at the nginx proxy level (but this causes performance issues as @tlvu points out).

I can also address https://github.com/Ouranosinc/jupyterhub/issues/2 in this PR as well but I feel like that should probably be its own PR.

I'll also point out that if an attacker is able to gain access to a valid jupyterhub session cookie they will almost certainly have access to a valid magpie session cookie as well.

fmigneault:
@mishaschwartz 
It can be done in iterative PRs for the refresh user part. 

My concern is mainly that there is nothing at the moment that strongly ties the Magpie login with the Jupyterhub login. 
Someone could be logged in as "user1" from Magpie's point of view while being "user2" in Jupyterhub depending on own cookies were handled. Anything that can improve (even partially) remedy this situation would be good.

mishaschwartz:
> My concern is mainly that there is nothing at the moment that strongly ties the Magpie login with the Jupyterhub login.

I get it. It's a good concern to have. I think that the the `refresh_user` feature will help with that. I can bump that up the priority list.

tlvu:
> This will allow us to check whether a user is allowed to _log in_ to jupyterhub but will no longer check at every request (by querying twitcher's ows/verify endpoint). I believe that this is a good compromise that will allow admins some control over who is allowed to access jupyterhub while simultaneously not affecting performance.

Super!  I think this should fix the performance problem.  Does it still set Magpie cookie as before?  And the reverse scenario, someone already logged into Magpie will not need to login again for JupyterHub?

> Someone could be logged in as "user1" from Magpie's point of view while being "user2" in Jupyterhub depending on own cookies were handled.

I am not sure I follow this one.  Right now, before this PR, this situation can not happen?  How can this situation happen now with this PR?

mishaschwartz:
> Does it still set Magpie cookie as before? And the reverse scenario, someone already logged into Magpie will not need to login again for JupyterHub?

Yes it still sets the Magpie cookie.
No, even if you've logged in to Magpie, you still need to log in to jupyterhub. We decided that jupyterhub should be considered the only login screen that users should interact with.

 > I am not sure I follow this one. Right now, before this PR, this situation can not happen? How can this situation happen now with this PR?

It could happen before this PR as well. A user could log in to juptyerhub with one set of credentials and then go over to magpie and log in with another. This PR makes the situation slightly better because if a user logs in to Magpie first as user1 and then to jupyterhub as user2, they will be logged into magpie as user2 at the same time.

fmigneault:
@tlvu 
> I am not sure I follow this one. Right now, before this PR, this situation can not happen? How can this situation happen now with this PR?

Look! (using current `master`). If cookies are set correctly, users are not enforced between Jupyterhub and the rest of the platform. 

image

mishaschwartz:
> DO NOT MERGE THIS PR UNTIL https://github.com/Ouranosinc/jupyterhub/pull/19 is merged and the pavics/jupyterhub version has been updated here!

This has now been updated. This PR is now ok to merge

tlvu:
I will redo the performance test and report back.

mishaschwartz:
> I will redo the performance test and report back.

@tlvu have you had a chance to do this? I'd like to get this PR merged in soon if possible

tlvu:
> > I will redo the performance test and report back.
> 
> @tlvu have you had a chance to do this? I'd like to get this PR merged in soon if possible

Opps, sorry, this one felt through the cracks with too many interruptions.   Will try to re-test this week for you.

tlvu:
@mishaschwartz 

In the spirit of quickly merging this PR and to future-proof any unforeseen future problems in production because production is always full of surprises and we can never anticipate everything, can you add a toogle to disable this Jupyterhub behind Twitcher?  The same kind of toogle as in PR https://github.com/bird-house/birdhouse-deploy/pull/348.

This way, if something happen in production, we can toogle this off immediately to avoid breaking users.  Then we have time for further investigations.

The toogle would allow the previous behavior before this PR.

mishaschwartz:
@tlvu 

I've added a toggle here: https://github.com/bird-house/birdhouse-deploy/pull/358/commits/1e6b98b7882e38bd606dc5770d32ebc4c86aabfb

I have not added this variable to the `env.local.example` file because I believe that it shouldn't be advertised to most users since I really don't recommend disabling this functionality unless absolutely necessary.

However, you know about it now so Ouranos can always disable it if necessary.

tlvu:
> @tlvu
> 
> I've added a toggle here: [1e6b98b](https://github.com/bird-house/birdhouse-deploy/commit/1e6b98b7882e38bd606dc5770d32ebc4c86aabfb)

@mishaschwartz great thanks.  I do not see a change to the proxy to make the verify path always return 200 (http success), like the other Geoserver behind Twitcher PR https://github.com/bird-house/birdhouse-deploy/pull/348

I pulled out the full diff of this PR and there is no proxy change to add the verify path either.  

Could you quickly remind me how this one perform the access control differently than other Geoserver behind Twitcher PR?  How come it does not need a verify path added to the proxy?

Because I won't deploy this PR for manual testing this time, I just want to make sure the deactivation works as intended the day I actually need it.

> 
> I have not added this variable to the `env.local.example` file because I believe that it shouldn't be advertised to most users since I really don't recommend disabling this functionality unless absolutely necessary.

That's perfect.  Not every settings in `default.env` are exposed to `env.local.example`.

mishaschwartz:
@tlvu 

> Could you quickly remind me how this one perform the access control differently than other Geoserver behind Twitcher PR? How come it does not need a verify path added to the proxy?

That's right. In this case, the proxy does not do any verification for us. The MagpieAuthenticator is the one that checks whether a user is authorized or not by querying twitcher's verify endpoint directly.

If the proxy does the check then every request to jupyterhub is checked which causes a bottleneck as you discovered (https://github.com/bird-house/birdhouse-deploy/pull/358#issuecomment-1700294695 https://github.com/bird-house/birdhouse-deploy/pull/358#issuecomment-1701243838)

We implemented a solution so that access is checked when the user logs in to jupyterhub (https://github.com/bird-house/birdhouse-deploy/pull/358#issuecomment-1735844742) and we will add another feature in the future to check periodically as well (https://github.com/bird-house/birdhouse-deploy/pull/358#issuecomment-1737374614)

PR #359, opened 2023-07-20
Flexible locations for data served by thredds

dchandan:
## Overview

Currently, the following limitations are present in the birdhouse stack with respect to where data is located, how that is documented and whether or not one has the flexibility to change either of those:

1. The locations of data on both the host and on the container are hardcoded in the docker-compose file:

>      - ${DATA_PERSIST_ROOT}/datasets:/pavics-data
>      - ${DATA_PERSIST_ROOT}/ncml:/pavics-ncml

2. The names that are used for data locations are not very clear. Specifically:

     2.1. The names of top level `datasetScan` directories 'Birdhouse' and 'Datasets' is not descriptive (and also hardcoded).

     2.2. 'Datasets' is supposed to be the main data serving location containing _any_ catalog data but is associated with filesystem locations containing the label 'ncml' which wrongly suggests that the directory is only supposed to serve ncml catalog files.
     2.3 It is unclear what role the 'Birdhouse' folder is supposed to play in a general deployment (it appears that on PAVICS it contains raw data or data that is otherwise not available through the catalog).
    2.4 'pavics' is used in the names of data locations which is something we have to move away from for a general DACCS deployment.

## Changes

**Non-breaking changes**
In recognition of the present PAVICS distinction between the 'Datasets' and 'Birdhouse' directories, the new changes continue to expose two data locations on the THREDDS server. The first location, defined on the host by the new environment variable  `STAC_DATA_LOCATION_FOR_THREDDS_ON_HOST` is meant to replace 'Datasets' as the **location containing catalog data** and the variable name is chosen to reflect the upcoming changes where STAC will be used to as the catalog. The other location `MISC_DATA_LOCATION_FOR_THREDDS_ON_HOST` is meant to replace 'Birdhouse' as **the place for all non catalog data**.

**Breaking changes**
There should be no breaking changes. The defaults of all newly added environment variables are those that are currently being used. A new deployment will just need to override these variables in `env.local` files.

## Related Issue / Discussion

- Related to https://github.com/bird-house/birdhouse-deploy/pull/356


birdhouse/default.env
```diff
@@ -10,6 +10,9 @@ export BASH_IMAGE="bash:5.1.4"
 
 # Root directory under which all data persistence should be nested under
 export DATA_PERSIST_ROOT="/data"
+export STAC_DATA_LOCATION_FOR_THREDDS_ON_HOST=${DATA_PERSIST_ROOT}/ncml  # this default is for backward compatibility
```
mishaschwartz, line 13: Also they should be added to the `DELAYED_EVAL` variable for the same reason (if `DATA_PERSIST_ROOT` is updated)

dchandan:
Thanks for the review @mishaschwartz! I'll incorporate these into a revised submission.

birdhouse/config/thredds/catalog.xml.template
```diff
@@ -31,7 +31,7 @@
 
     </datasetScan>
 
-    <datasetScan name="Datasets" ID="datasets" path="datasets" location="/pavics-ncml">
+    <datasetScan name="${STAC_DATA_LOCATION_NAME}" ID="${STAC_DATA_LOCATION_NAME_lower}" path="${STAC_DATA_LOCATION_NAME_lower}" location="${STAC_DATA_LOCATION_FOR_THREDDS_ON_CONTAINER}">
```
fmigneault, line 34: This variable name assumes `components/stac` will be present, but nothing guarantees nor even require this component to be loaded. Avoid mixing `STAC_` prefixed variables that should be reserved only to this component.

However, this part of the XML could be injected entirely using an environment variable (or an extra XML if THREDDS allows load others?), when STAC+THREDDS are enabled.  That would be an XML defined somewhere in `components/stac/config/thredds`.

birdhouse/config/thredds/default.env
```diff
@@ -6,6 +6,15 @@ export THREDDS_DOCKER=pavics/thredds-docker
 export THREDDS_IMAGE="${THREDDS_DOCKER}:${THREDDS_VERSION}"
 export THREDDS_ORGANIZATION="Birdhouse"
 
+# All default values for data locations are set to ensure backward compatibility
+export STAC_DATA_LOCATION_FOR_THREDDS_ON_CONTAINER="/pavics-ncml"
```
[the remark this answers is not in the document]
fmigneault: `CATALOG_` might bring confusion with `catalog` service (although deprecated, other similar variables can still be defined).
If the intent is to have a generic "dataset" shared between many services, `BIRDHOUSE_DATA_...` or simply `DATA_...` seems reasonable.

birdhouse/config/thredds/docker-compose-extra.yml
```diff
@@ -21,8 +20,8 @@ services:
       PAVICS_FQDN_PUBLIC: $PAVICS_FQDN_PUBLIC
     volumes:
       - thredds_persistence:/usr/local/tomcat/content/thredds
-      - ${DATA_PERSIST_ROOT}/datasets:/pavics-data
-      - ${DATA_PERSIST_ROOT}/ncml:/pavics-ncml
+      - ${MISC_DATA_LOCATION_FOR_THREDDS_ON_HOST}:${MISC_DATA_LOCATION_FOR_THREDDS_ON_CONTAINER}
+      - ${STAC_DATA_LOCATION_FOR_THREDDS_ON_HOST}:${STAC_DATA_LOCATION_FOR_THREDDS_ON_CONTAINER}
```
fmigneault, line 24: Since it could map to another location (depending on what the user sets), it can potentially cause other locations to be incorrectly mapped because they don't align anymore.

For example: 
https://github.com/bird-house/birdhouse-deploy/blob/2b344d3383d8c0cd3f4cc4a0bf4266e3bae06a47/birdhouse/config/wps_outputs-volume/config/proxy/docker-compose-extra.yml#L6

https://github.com/bird-house/birdhouse-deploy/blob/2b344d3383d8c0cd3f4cc4a0bf4266e3bae06a47/birdhouse/components/weaver/default.env#L85

Considering that `wps_outputs` volume is configured accordingly such that they will appear next to `${DATA_PERSIST_ROOT}/datasets` so that they are all listed under
`https://${HOST}/twitcher/ows/proxy/thredds/catalog/birdhouse/catalog.html`.

fmigneault (commented):
Although `${DATA_PERSIST_ROOT}/datasets:/pavics-data` and `${DATA_PERSIST_ROOT}/ncml:/pavics-ncml` are both "hardcoded", it is possible to set `DATA_PERSIST_ROOT` to whichever value you desire. This could be a directory with various mount points. Therefore, technically, anything nested under `DATA_PERSIST_ROOT` really isn't that much hardcoded depending on how the drive is mapped. The reason of the shared `DATA_PERSIST_ROOT` is that, once this is defined, all relative locations can be properly mapped between the various services that must use similar volume mounts. 

I'm not against having separate variables to help facilitate setup of these directories to separate locations, with defaults reverting to the current locations. However, anything interacting with items mapped by `DATA_PERSIST_ROOT` must take into consideration the multiple inter-service complexities detailed in https://github.com/bird-house/birdhouse-deploy/issues/355#issuecomment-1635005749

I think that simply offering variables to edit anything as of the latest update does not sufficiently warns the developer about what they might inadvertently break, including but not limited to, leaking private user-workspace files publicly.

birdhouse/config/thredds/docker-compose-extra.yml
tlvu: I think I also agree with you @fmigneault on this one but is there a possible bug with the value of `WEAVER_WPS_OUTPUTS_DIR`?  

If `${DATA_PERSIST_ROOT}/datasets` is `/pavics-data`, then shouldn't `WEAVER_WPS_OUTPUTS_DIR` be `${DATA_PERSIST_ROOT}/datasets/wps_outputs/weaver` (missing `datasets`)?

If it is indeed the case, then `WEAVER_WPS_OUTPUTS_DIR='${MISC_DATA_LOCATION_FOR_THREDDS_ON_HOST}/wps_outputs/weaver'`

tlvu (changes requested):
I agree with most feedbacks from Misha and Francis here.

A few historical explanations on why things are that way:



    The names that are used for data locations are not very clear. Specifically:

    2.1. The names of top level datasetScan directories 'Birdhouse' and 'Datasets' is not descriptive (and also hardcoded).

In 2019 when I joined Ouranos, DACCS did not started and at that time I believe we wanted to be part of the Birdhouse community, and we still want it now.

During all those years many tutorial notebooks have the string `birdhouse` burned in the Thredds URL and those notebooks most probably been duplicated by many of our users.  Changing that string will be a big pain for all our existing rather large userbase. 

    2.2. 'Datasets' is supposed to be the main data serving location containing any catalog data but is associated with filesystem locations containing the label 'ncml' which wrongly suggests that the directory is only supposed to serve ncml catalog files.

'Datasets' was not meant to be the main data serving location, the 'Birdhouse' location was.  But under 'Birdhouse' we only have raw `.nc` files.

Then we discover the NCML feature of Thredds that allows aggregating multiples `.nc` files into one single file so one single URL for our users to deal with.

NCML also allows for renaming the data collections name and hierarchy, for changing attributes on the fly without changing the raw data, which would break all shasums of all the `.nc` files, which would make it very hard for us to perform integrity check on those raw `.nc` files, especially when these `.nc` are downloaded from another organization.  We want to preserve the shasum and hierarchy on disk to match the source organization we downloaded from.

So we wanted to have a separated new top-level for those more user-friendly NCML with nicer hierarchy and since the name 'Datasets' was not taken yet, we used it.  But on the disk is it `/pavics-ncml` because it should only contain `.ncml` files.

Since all the NCML was much more user-friendly, it is used more widely so it appears to be the main collection but that was not the original intention.  No harm if the usage has evolved.

    2.3 It is unclear what role the 'Birdhouse' folder is supposed to play in a general deployment (it appears that on PAVICS it contains raw data or data that is otherwise not available through the catalog).

The Catalog only crawl the NCML collections because that collection has nicer name and hierarchy and proper attribute fixes.

    2.4 'pavics' is used in the names of data locations which is something we have to move away from for a general DACCS deployment.

Agreed.  DACCS came after PAVICS project so unfortunately PAVICS is hardcoded everywhere.

birdhouse/config/thredds/default.env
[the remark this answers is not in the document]
tlvu: I would like to propose to rename `MISC_DATA_LOCATION_NAME` ==> `RAW_DATA_LOCATION_NAME`, `MISC_DATA_LOCATION_FOR_THREDDS_ON_HOST` ==> `RAW_DATA_LOCATION_FOR_THREDDS_ON_HOST` and `MISC_DATA_LOCATION_FOR_THREDDS_ON_CONTAINER` ==> `RAW_DATA_LOCATION_FOR_THREDDS_ON_CONTAINER`.

`STAC_DATA_LOCATION_NAME` ==> `DATA_LOCATION_NAME` and similar for all its friends.

Avoid using `STAC` because this is the Thredds component, it should not know which other component(s) is/are crawling its data.

birdhouse/config/thredds/docker-compose-extra.yml
tlvu: `WEAVER_WPS_OUTPUTS_DIR='${MISC_DATA_LOCATION_FOR_THREDDS_ON_HOST:-${DATA_PERSIST_ROOT}/datasets}/wps_outputs/weaver'` to avoid Weaver component having direct dependency on Thredds component.

tlvu:
Please update `CHANGES.md` and follow the release procedure when merging this PR as documented in https://github.com/bird-house/birdhouse-deploy/blob/2b344d3383d8c0cd3f4cc4a0bf4266e3bae06a47/birdhouse/README.rst#release-procedure

birdhouse/config/thredds/catalog.xml.template
fmigneault: Exactly. Thanks for providing the example I had in mind.

birdhouse/config/thredds/docker-compose-extra.yml
fmigneault: The expectation is that the (partial) structure is as follows:
```
/data
    /datasets
    /wps_outputs
        /<bird>
    /magpie_persists
    [...]
```

I can see that there is indeed a `/data/datasets/wps_outputs` that gets created (but empty) due to the `${DATA_PERSIST_ROOT}/datasets:/pavics-data` volume mount. However, I'm not quite sure how the expected `/data/wps_outputs/<bird>` locations get mapped in THREDDS, but they do appear in the right place: 

- https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/catalog/birdhouse/wps_outputs/catalog.html
- https://hirondelle.crim.ca/twitcher/ows/proxy/thredds/catalog/birdhouse/wps_outputs/catalog.html

They are populated with the appropriate outputs from the WPS services.

And https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/catalog/datasets/catalog.html does not have `wps_outputs` somehow?

tlvu: > I'm not quite sure how the expected `/data/wps_outputs/<bird>` locations get mapped in THREDDS, but they do appear in the right place:

Thredds get it from here https://github.com/bird-house/birdhouse-deploy/blob/2b344d3383d8c0cd3f4cc4a0bf4266e3bae06a47/birdhouse/config/thredds/docker-compose-extra.yml#L26

Note the content of folder `wps_outputs/` is from a data-volume `wps_outputs` and not from the disk volume-mount `/data/wps_outputs`.  

That `/data/wps_outputs` path from `WEAVER_WPS_OUTPUTS_DIR` hopefully is the path inside the container, like the birds, and not the path on disk.  If path on disk, it is probably wrong.

We have to be careful when we reference a path, be specific if it is path on the disk of the host or path inside the container.

Each bird also volume-mount the same `wps_outputs` data-volume so that's why everything "works together": https://github.com/bird-house/birdhouse-deploy/blob/2b344d3383d8c0cd3f4cc4a0bf4266e3bae06a47/birdhouse/config/finch/config/wps_outputs-volume/docker-compose-extra.yml#L6


> And https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/catalog/datasets/catalog.html does not have `wps_outputs` somehow?

Because not needed.  We only need one single location for `wps_outputs` on Thredds, for all the WPS services (birds).

birdhouse/config/thredds/default.env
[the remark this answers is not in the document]
mishaschwartz: @tlvu thank you for the naming suggestions. I agree that those are all good names and that I'm happy to use those names.

birdhouse/config/thredds/docker-compose-extra.yml
mishaschwartz: Just to clarify this situation:

- `wps_outputs` is a volume mounted to each of the birds' containers and to the proxy container at `pavics-data/wps_outputs`
- `${WEAVER_WPS_OUTPUTS_DIR}` is a location on host that is mounted to the weaver, weaver-worker, and proxy container at `pavics-data/wps_outputs/weaver`

`${WEAVER_WPS_OUTPUTS_DIR}` needs to be a bind mount, not a volume, because it needs to be mounted in the container at the same location as on the host so that it can be mounted to the weaver-worker containers by a process running in the weaver container.

@fmigneault @tlvu is that correct?

mishaschwartz:
@tlvu thank you for the historical explanations. 

> During all those years many tutorial notebooks have the string birdhouse burned in the Thredds URL and those notebooks most probably been duplicated by many of our users. Changing that string will be a big pain for all our existing rather large userbase.

This is a good reason to keep the defaults as they are for now and PAVICS should keep these defaults for as long as you want. This doesn't mean that we shouldn't provide this flexibility for other DACCS nodes that don't have the historical user-base that you have already.

> So we wanted to have a separated new top-level for those more user-friendly NCML with nicer hierarchy and since the name 'Datasets' was not taken yet, we used it. But on the disk is it /pavics-ncml because it should only contain .ncml files.

This is a great use-case and we can/should even suggest this to other people managing data on their DACCS node. However, it shouldn't be the only option which is why I support renaming this directory to something that doesn't contain "ncml"

birdhouse/config/thredds/catalog.xml.template
dchandan: > This variable name assumes `components/stac` will be present, but nothing guarantees nor even require this component to be loaded.

That's a good point and the solutions suggested could work. If I recall there is a current consensus (or emerging consensus?) that STAC should be a required component of a minimum DACCS (Marble?) node even if no data is served at the node. If that were to be the case, then would the existing structure suffice?

tlvu:
> > This is a good reason to keep the defaults as they are for now and PAVICS should keep these defaults for as long as you want. This doesn't mean that we shouldn't provide this flexibility for other DACCS nodes that don't have the historical user-base that you have already.
> 
> Yes, fully agree. A possible option comes to mind if PAVICS also chooses to move to a newer configuration: redirect links to older locations to newer locations at the proxy layer. @mishaschwartz @tlvu thoughts on this?

Just to be clear, I am not saying I want the default to stay that way.  In fact, I am up for the default values to change so new nodes won't have this legacy naming.

However, for existing deployments, we should be able to keep the old values to be considerate to existing users.  So as long as there is an easy way to override all the new names and URL paths, I am okay.

So maybe we take this opportunity to think forward about all the public facing names and URL path because starting the right way is much better than having to add layers of backward compat (I like your redirect links suggestion Deepak, hopefully a mean to transition to newer names).

Instead of `/twitcher/ows/proxy/finch`, can we have just `/finch` or `/wps/finch`?  Same remark for all other birds.

Instead of `/twitcher/ows/proxy/thredds`, just `/thredds`?

Because once you published a URL path, if you change those paths, you'll have to ensure redirects to not break your users.  WPS responses sometime contain the public URL back in the response.  In the redirect case, this probably won't match.  Same for OpenDap protocol, I think some scenario contain the public URL in the response.

My point is it might be tricky to rename these URL paths after the fact.  Better start right.  For exiting deployment leave a way to easily return to the old config.

birdhouse/config/thredds/docker-compose-extra.yml
tlvu: I'll let @fmigneault reply for Weaver, I am not fully familiar with it.

huard (commented):
Agree with the proposed name changes. 

I would not bake STAC into names, especially since we don't yet have an operational instance using it and confirmation this is our long term catalog solution. 

Essentially, what PAVICS has at the moment is 
- raw data, not catalogued because too messy and subject to change. I see this as "internal" data, like private functions in a code that you're free to use, but with the knowledge it could break at the next release. 
- clean aggregated data, catalogued, changes are communicated to users.

tlvu:
> Your comment on thinking about all public facing names and URLs is very good. I very much do like the suggestions about the short urls to individual birds and threads. I don't think we, as a DACCS group, have had a serious discussion about finalizing this. Perhaps this is something for which it is worth opening an issue?

Yes.  Let's make other changes not Thredds related outside of this PR that should only be relevant to Thredds naming.

birdhouse/config/thredds/docker-compose-extra.yml
dchandan: I've now incorporated all the other comments. Any thoughts on what we need to do about this?

birdhouse/config/thredds/default.env
```diff
@@ -19,3 +37,13 @@ OPTIONAL_VARS="
...
 "
+
+export DELAYED_EVAL="
+  $DELAYED_EVAL
+  DATA_LOCATION_NAME
+  DATA_LOCATION_FOR_THREDDS_ON_CONTAINER
+  DATA_LOCATION_FOR_THREDDS_ON_HOST
+  RAW_DATA_LOCATION_NAME
+  RAW_DATA_LOCATION_FOR_THREDDS_ON_CONTAINER
+  RAW_DATA_LOCATION_FOR_THREDDS_ON_HOST
```
tlvu, line 48: Missing all the `_lower` vars in `DELAYED_EVAL`. 

Basically all variables that depend on other vars need to be in `DELAYED_EVAL`.

So `DATA_LOCATION_NAME` and `DATA_LOCATION_FOR_THREDDS_ON_CONTAINER`, and their `RAW_` counterparts, do not need to be in `DELAYED_EVAL` since they do not depend on any other variables.

tlvu:
@dchandan Next time please just do a regular merge with merging with master, avoid doing a rebase then a force push like this.

Anyone that pulled from your PR will be confused and git won't be able to resolve any conflicts properly, automatically. I am not saying this happened.  

Just saying that any history editing (rebase, force push, commit amend) is fine as long as you have not submitted a PR yet.  Once you have a submitted PR, anyone can pull your branch for verification and expected to get your follow up fixes or updates with a simple `git pull`.  They should not be have to deal with any messy merge or `git reset --hard` to get back to your "new" branch.

dchandan:
@tlvu Sorry about that! Still pretty new to this... I appreciate the explanation, this is very helpful :)

tlvu:
> Sorry about that! Still pretty new to this... I appreciate the explanation, this is very helpful :)

No worries, I did not pull your branch to test yet so did not get into that issue.

If you ever want to read more about problems with history editing (read destruction of old commits), you can see the lengthy discussion here https://github.com/Ouranosinc/xclim/issues/373 about rebase merge and squash merge.  Both merge methods will destroy old commits.  I've disabled both merge methods in this repo.

dchandan:
Are we able to pull this in?

tlvu:
> Are we able to pull this in?

I think unfinished discussion about `WEAVER_WPS_OUTPUTS_DIR` ?

dchandan:
Yes, I was hoping to finish that discussion and get this pulled in, but I learned this afternoon that Francis is away ATM. No rush, I was just checking if this is something I can check off.

fmigneault:
I'm back, but I think @ChaamC will be better suited to validate if there are any potential issues regarding how WPS outputs directories are mounted and synched between the various services, including pending https://github.com/Ouranosinc/cowbird/pull/40 feature. He should be back next week to answer.

birdhouse/config/thredds/docker-compose-extra.yml
fmigneault: As long as the mapping paths are the same on both sides of `:` for Weaver, it doesn't matter where the directory is actually located. 

Other than that, the paths are only "conveniently" named in such a way that they match `<data-root>/wps_outputs/<bird>` format to align with other services that do not have this "same path" requirement as Weaver does. All those paths need to align to allow Cowbird to perform permission synchronization between WPS-outputs file access, THREDDS, Jupyter workspaces, etc.

If all those conditions are met, there shouldn't be any issue with how paths/env-vars are used to define mounts. 
However, I would like to avoid the possibility for inexperienced users that are not aware of all these conditions and implications to inadvertently break functionalities simply because they changed some base directory path.

dchandan: What is the right `wps_output` dir? Is it: (a) `${DATA_PERSIST_ROOT}/wps_outputs` or (b) `${DATA_PERSIST_ROOT}/datasets/wps_outputs`? If it is (a), and from the responses it sounds like it is (a), then there shouldn't be an issue with the change proposed because they don't affect the folders `${DATA_PERSIST_ROOT}/wps_outputs` and `${DATA_PERSIST_ROOT}/user_workspaces`.

fmigneault: Yes, I believe it should be (a).

dchandan:
Based on the discussion of the only item that was outstanding (https://github.com/bird-house/birdhouse-deploy/pull/359#discussion_r1321528204), it seems we can pull this PR.

birdhouse/config/thredds/default.env
```diff
@@ -6,10 +6,28 @@ export THREDDS_DOCKER=pavics/thredds-docker
 export THREDDS_IMAGE="${THREDDS_DOCKER}:${THREDDS_VERSION}"
 export THREDDS_ORGANIZATION="Birdhouse"
 
+# All default values for data locations are set to ensure backward compatibility
```
mishaschwartz, line 9: We have an unofficial convention that I like: start environment variables with the component name. That way it is really easy to determine which component each variable applies to. 

With that in mind, I'd recommend renaming some of these to:

```sh
export THREDDS_DATA_LOCATION_ON_CONTAINER='/pavics-ncml'
export THREDDS_RAW_DATA_LOCATION_ON_CONTAINER='/pavics-data'
export THREDDS_DATA_LOCATION_ON_HOST='${DATA_PERSIST_ROOT}/ncml'  # this default is for backward compatibility 
export THREDDS_RAW_DATA_LOCATION_ON_HOST='${DATA_PERSIST_ROOT}/datasets'  # this default is for backward compatibility
export THREDDS_DATA_LOCATION_NAME='Datasets'  # this default is for backward compatibility 
export THREDDS_RAW_DATA_LOCATION_NAME='Birdhouse'  # this default is for backward compatibility 
export THREDDS_DATA_LOCATION_NAME_lower=$(echo "$DATA_LOCATION_NAME" | tr '[:upper:]' '[:lower:]')
export THREDDS_RAW_DATA_LOCATION_NAME_lower=$(echo "$RAW_DATA_LOCATION_NAME" | tr '[:upper:]' '[:lower:]')
```

birdhouse/config/thredds/default.env
```diff
@@ -6,10 +6,28 @@ export THREDDS_DOCKER=pavics/thredds-docker
...
+export DATA_LOCATION_FOR_THREDDS_ON_CONTAINER='/pavics-ncml'
+export RAW_DATA_LOCATION_FOR_THREDDS_ON_CONTAINER='/pavics-data'
+export DATA_LOCATION_FOR_THREDDS_ON_HOST='${DATA_PERSIST_ROOT}/ncml'  # this default is for backward compatibility 
+export RAW_DATA_LOCATION_FOR_THREDDS_ON_HOST='${DATA_PERSIST_ROOT}/datasets'  # this default is for backward compatibility
+export DATA_LOCATION_NAME='Datasets'  # this default is for backward compatibility 
+export RAW_DATA_LOCATION_NAME='Birdhouse'  # this default is for backward compatibility 
+export DATA_LOCATION_NAME_lower=$(echo "$DATA_LOCATION_NAME" | tr '[:upper:]' '[:lower:]')
```
mishaschwartz, line 16: This should be single quoted so that it's not evaluated too soon (since it's one of the DELAYED_EVAL variables now)

And the line below as well

mishaschwartz (changes requested):

birdhouse/config/thredds/default.env
```diff
@@ -6,10 +6,28 @@ export THREDDS_DOCKER=pavics/thredds-docker
...
 
+# All default values for data locations are set to ensure backward compatibility
+export THREDDS_DATA_LOCATION_ON_CONTAINER='/pavics-ncml'
+export THREDDS_RAW_DATA_LOCATION_ON_CONTAINER='/pavics-data'
+export THREDDS_DATA_LOCATION_ON_HOST='${DATA_PERSIST_ROOT}/ncml'  # this default is for backward compatibility 
+export THREDDS_RAW_DATA_LOCATION_ON_HOST='${DATA_PERSIST_ROOT}/datasets'  # this default is for backward compatibility
+export THREDDS_DATA_LOCATION_NAME='Datasets'  # this default is for backward compatibility 
+export THREDDS_RAW_DATA_LOCATION_NAME='Birdhouse'  # this default is for backward compatibility 
+export THREDDS_DATA_LOCATION_NAME_lower='$(echo "$THREDDS_DATA_LOCATION_NAME" | tr "[:upper:]" "[:lower:]")'
+export THREDDS_RAW_DATA_LOCATION_NAME_lower='$(echo "$THREDDS_RAW_DATA_LOCATION_NAME" | tr '[:upper:]' '[:lower:]')'
```
fmigneault, line 17: Even if the variables are for storing a lowercase representation, the names themselves should be kept all uppercase. 

I would like to know also why is lowercase enforced? I don't see any technical reason for this to be needed and it adds a lot of unnecessary variables that mostly duplicates configuration and `VARS`/`DELAYED_EVAL` references.

The only non-uppercase variant I find is the `name` property in the config XML. I believe this could be any-case with the same value as whichever value is set for these variables.

fmigneault, line 17: From the point of view of a new node manager, it is not clear what consists of a "`RAW_DATA"` vs plain `"DATA"`. Naming could be improved to provide better guidance.

From my understanding, the original 2 locations refer to something like:
- `birdhouse`: data shared amongst services of the server, such as user-workspaces data, test data, dynamically generated outputs from processing, etc.
- `datasets`: external data sources provided by the server, usually as read-only and static, provided for reference/research

fmigneault (changes requested):

birdhouse/config/thredds/default.env
dchandan: > Even if the variables are for storing a lowercase representation, the names themselves should be kept all uppercase.

Fair.

> I would like to know also why is lowercase enforced? I don't see any technical reason for this to be needed and it adds a lot of unnecessary variables that mostly duplicates configuration and `VARS`/`DELAYED_EVAL` references.

These lower case variables are needed for maintaining full backward compatibility.

dchandan: > From the point of view of a new node manager, it is not clear what consists of a "`RAW_DATA"` vs plain `"DATA"`. Naming could be improved to provide better guidance.

This was already discussed earlier in the PR and this naming convention was seen as the best path forward. I'm hesitant to re-visit it at this point. Regardless of the choice of name, we will have to document this properly for a new node admin, so that will be an appropriate venue to make the purpose of the two locations clear.

fmigneault: I think the current names only swapped non-descriptive names (birdhouse, datasets) by new ones. 
These names do not improve item 2. of the PR description. 
If the admin has to find the comment that describes the variables (which are not present at the moment), it is already an indication that they are improperly named.

fmigneault: I just realized, `THREDDS_DATA_LOCATION_NAME_lower` is misleading. 
If overridden, nothing actually enforces it to be lowercase... 
It would be better to simply have `THREDDS_DATA_LOCATION_DIR` or similar. 
And the default would simply be `THREDDS_DATA_LOCATION_DIR=datasets`.
Idem for "`RAW_DATA`".

mishaschwartz: > I just realized, THREDDS_DATA_LOCATION_NAME_lower is misleading.
> If overridden, nothing actually enforces it to be lowercase...

Typically we have handled cases like this by not including variables like `THREDDS_DATA_LOCATION_NAME_lower` in the env.local.example file which indicates to the user that they are not meant to be overridden and are for internal use only.

mishaschwartz: > If the admin has to find the comment that describes the variables (which are not present at the moment), it is already an indication that they are improperly named.

I disagree. It is impossible to fully describe the purpose of the variable in a name. This is why we use documentation to properly describe the variable's purpose. 

> From the point of view of a new node manager, it is not clear what consists of a "RAW_DATA" vs plain "DATA". Naming could be improved to provide better guidance.

Ok, what would you suggest @fmigneault?

dchandan: It is not ideal to have a "lower case" variable name from many perspectives, I agree with that. The issue is that to maintain full backward compatibility with what already exists in `config/thredds/catalog.xml.template` (as shown below) one need both naming case.

> `<datasetScan name="Birdhouse" ID="birdhouse" path="birdhouse" location="${THREDDS_RAW_DATA_LOCATION_ON_CONTAINER}">`
> `<datasetScan name="Datasets" ID="datasets" path="datasets" location="${THREDDS_DATA_LOCATION_ON_CONTAINER}">`

dchandan: I think Misha's suggestion works here.

dchandan: The new names reflect the way in which data is currently stored on the thredds server on PAVICS, specifically, one folder for catalog data (now defined by the new variable `THREDDS_DATA_LOCATION_NAME`) and one for data that is not meant to be part of the official catalog  (now indicated by the variable name `THREDDS_RAW_DATA_LOCATION_NAME`). It seems very sensible to me given the dual needs of supporting legacy decisions while bringing much needed clarity and flexibility.

fmigneault: IMO `datasets` and `birdhouse` made sense, because one is for storing static/external "datasets", while the other is for operational needs of the services under "birdhouse"-deploy. Maybe there was some misuses in the past in path names such as using `/pavics-data` or regarding were some data was placed for convenience instead of where they should actually be, but the original intention of these names were right.

mishaschwartz: I agree that they make sense for developers who are organizing the code but these names are visible to users who are browsing the thredds catalog. A user doesn't know what a "birdhouse" is or how that relates to their data. 

I think we're getting off track here though since the default values "birdhouse" and "datasets" aren't changed by this PR, this PR just provides the option to set them to something different if a node administrator wants.

fmigneault: If I decide that I want my instance to use "CRIM" instead of "birdhouse", I could set `THREDDS_DATA_LOCATION_NAME=CRIM`. The current `THREDDS_DATA_LOCATION_NAME_lower` would then convert that to `crim`, so both directories and URLs will use `crim`. I have no choice to override `THREDDS_DATA_LOCATION_NAME_lower=CRIM` as well to obtain `/twitcher/ows/proxy/thredds/fileserver/CRIM/...`. 

`THREDDS_DATA_LOCATION_NAME_lower=CRIM` feels odd, and make uses of an "undocumented" `env.local.example` variable. So instead, I say we document them explicitly such that `default.env` and `env.local.example` both have: 
```env
# name of the directory used in THREDDS Catalog interface
export THREDDS_DATA_LOCATION_NAME=Birdhouse
# name of the directory in file system paths and URLs
export THREDDS_DATA_LOCATION_DIR=birdhouse
```

fmigneault: > I agree that they make sense for developers who are organizing the code but these names are visible to users who are browsing the thredds catalog. A user doesn't know what a "birdhouse" is or how that relates to their data.

Yes, this is why these new variables allow to override the *values* to whichever name that feels right for your instance, such that what users will see is relevant on that server. The *name* of the variables themselves though should be meaningful for the devs, not the users that will not see those variable names anyway. It is irrelevant that a user knows about eg `THREDDS_BIRDHOUSE_DATA_LOCATION_NAME`/`THREDDS_DATASET_DATA_LOCATION_NAME` or not. The code base is "birdhouse-deploy", so I don't see what would be more relevant than "birdhouse" for a dev to indicate data directories used by its services. This would make it the same default name as the docker default network and prefix used in  named volumes automatically created by docker-compose.

Of course, the defaults *values* need to remain "birdhouse" and "datasets" respectively for backward compatibility. I'm only takling about the *name* of the new variables.

mishaschwartz: If you want to control the name of the URL then sure, that makes sense. 

But just to be clear, in your example `THREDDS_DATA_LOCATION_DIR` is only used to set the URL path, not the name of the directory on the file system. That is currently set by the `THREDDS_DATA_LOCATION_ON_CONTAINER` variable: 

https://github.com/bird-house/birdhouse-deploy/pull/359/files#diff-4bd22da1236dbeeb19113dd6f1cec1bb094d578e8e4ae796309bc9b5c17f6958R34

If we're going to do it this way I'd suggest the more descriptive name `THREDDS_DATA_URL_PATH` 

We can change `THREDDS_DATA_LOCATION_ON_CONTAINER` to `THREDDS_DATA_LOCATION_DIR` as well

dchandan: I also agree that providing this flexibility makes sense. Let's make the following changes:

`THREDDS_DATA_LOCATION_NAME_lower -> THREDDS_DATA_URL_PATH`
`THREDDS_RAW_DATA_LOCATION_NAME_lower ->  THREDDS_RAW_DATA_URL_PATH`

I'd like to keep `THREDDS_DATA_LOCATION_ON_CONTAINER` as is to make it a clear counterpart of the variable `THREDDS_RAW_DATA_LOCATION_ON_HOST`. There are too many directories and volume mounts involved in the birdhouse stack and I think a nomenclature like this helps make sense of them while reducing the likelihood of errors.

dchandan: The problem is the "Birdhouse" location, for historical reasons/misuse, is currently used for storing data used by services under birdhouse-deploy as well as for storing and exposing raw data that are not part of (the currently used) catalog, i.e. THREDDS. 

Ideally, there should be three directories: One strictly for storing catalog data, one strictly for data that for whatever reason one doesn't want to store in the catalog at the moment, and one for internal use by birdhouse.

dchandan: I am not opposed to this three directory structure. In fact, I'd prefer it. Then, we can have variables like: `THREDDS_BIRDHOUSE_DATA_LOCATION_NAME`, `THREDDS_RAW_DATA_LOCATION_NAME` and `THREDDS_DATA_LOCATION_NAME`.

fmigneault: I'm not sure to undertand the use case of "data that for whatever reason one doesn't want to store in the catalog". 
If that data is not a dataset to be stored in the catalog, nor used by a service, what is it used for? I don't think the birdhouse-deploy config should have an unused variable. A service could however define more variables for additional data locations, such as what is done for `jupyterhub_user_data`, `magpie_persist`, etc., but we would add them as needed later on.

For the names, I would prefer `THREDDS_DATASET_LOCATION_NAME` or `THREDDS_CATALOG_LOCATION_NAME` (to make a parallel to your mention "storing *catalog* data"), but not simply `RAW_DATA` or `DATA`. These are too generic and not meaningful to disambiguate them, which will lead to more data locations being mixed/misused in the future.

fmigneault: It is good to have the explicit distinction of `..._URL_PATH` and `..._LOCATION_ON_HOST` to offer the flexibility of each part. It is important to allow control over the exposed URL and the host path. 

I don't think the `..._LOCATION_ON_CONTAINER` is really that much important, since the purpose of THREDDS's `catalog.xml` is to remap any local (in-docker) path to some URL, so where the data actually resides in the container is irrelevant as long as it align with the configured volume path.

tlvu: > `THREDDS_DATA_LOCATION_NAME_lower -> THREDDS_DATA_URL_PATH`
> `THREDDS_RAW_DATA_LOCATION_NAME_lower -> THREDDS_RAW_DATA_URL_PATH`

Agreed, nicer name.  And they don't have to be delayed eval anymore right?   Documentation should be clear so the user can decide what to do with them (need lower case or not).

> I'd like to keep `THREDDS_DATA_LOCATION_ON_CONTAINER` as is to make it a clear counterpart of the variable `THREDDS_RAW_DATA_LOCATION_ON_HOST`.

Agreed too.


> I don't think the `..._LOCATION_ON_CONTAINER` is really that much important, since the purpose of THREDDS's `catalog.xml` is to remap any local (in-docker) path to some URL, so where the data actually resides in the container is irrelevant as long as it align with the configured volume path.

It is important if you are the admin of the host and you `docker exec` into the `thredds` container or you manage the files on disk (ex: to sync new files).  In this case, you do want to know the path in the container vs path on the real disk.

Another user that want to know the real path is the ncml developer.  All path referenced in an `.ncml` are location on container, example https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/fileServer/datasets/reanalyses/day_ERA5_NAM.ncml, which correspond to this file https://github.com/Ouranosinc/pavics-vdb/blob/master/1-Datasets/reanalyses/day_ERA5_NAM.ncml (`.ncml` files in this repo are synced to our Thredds automatically)

So for the same `.nc` or `.ncml` file, the end users will care for the public URL on Thredds but the admin and developer will care for the actual location on disk and in the container.

tlvu: From my comment https://github.com/bird-house/birdhouse-deploy/pull/359#pullrequestreview-1540391553 explaining the historical reason for the naming of the 2 collections "Birdhouse" and "Datasets", those 2 names are, let's say, not the greatest choice.

So I think I'd rather not have "birdhouse" again in the variable name `THREDDS_BIRDHOUSE_DATA_LOCATION_NAME`.

The current "Birdhouse" catalog on our Thredds contains:
* `.nc` files that will be aggregated by an `.ncml` file in the "Datasets" catalog
* `.nc` files that we simply host for other orgs or other projects, ex: eccc/forecasts/geps/, that might or might not have a corresponding nicer `.ncml` aggregation in the "Datasets" catalog
* wps_outputs
* testdata

So it's pretty much a dumping ground for anything.  That "Birdhouse" catalog could be called "Root", "Misc" or "Raw" and it would still fit.

When we discovered all the features `.ncml` offers, we decided to create a brand new "Datasets" catalog to make it really stand out.

So from my perspective, `RAW_DATA` and `DATA` flavors for those 2 starting catalogs are fine for now.  Just document they are just 2 catalogs, nothing special, the admin will decide what goes in there.  It could be just testdata, just wps_outputs, not a mixed bag of everything like "Birdhouse" is right now.

What would really help future node admin is the ability to add new catalogs as they need, so they are not stuck with 2 like us and one of the 2 end up being a dumping ground for anything.

So I suggest we add a placeholder variable that can be set in `env.local` to dynamically add extra xml snippets to add any number of new catalogs.  This will cover Deepak 3 catalogs suggestions and any other special needs for any future node admin.  Nothing says 3 catalogs is enough or 5 or 10.  Each node admin will decide how many catalogs they want.  We start with 2 just to show that it is possible to have more than 1 catalog.

Example of placeholder type of variable: https://github.com/bird-house/birdhouse-deploy/blob/33042e6b0b1cc46ca0ec18bffca9d74b4d385f53/birdhouse/config/jupyterhub/jupyterhub_config.py.template#L190 

Sample use of that placeholder var in `env.local`: https://github.com/bird-house/birdhouse-deploy/blob/33042e6b0b1cc46ca0ec18bffca9d74b4d385f53/birdhouse/env.local.example#L338-L350

dchandan: Yes, there is absolutely no need to constrain ourselves to a pre-determined number of catalogs. This sounds like a very good way to proceed. For now, on PAVICS, the `RAW_DATA` and `DATA` catalogs would map directly to the existing PAVICS setup and for UofT the `RAW_DATA` would contain, among other things, (i) temporary data that we don't want to host long-term and so we don't want to put in the main catalog, (ii) data that needs to be quickly made available to users for their analysis while we work on properly ingesting it into the main catalog.

dchandan: > I don't think the `..._LOCATION_ON_CONTAINER` is really that much important, since the purpose of THREDDS's `catalog.xml` is to remap any local (in-docker) path to some URL, so where the data actually resides in the container is irrelevant as long as it align with the configured volume path.

Exactly what @tlvu said. Also, in the absence of defining it in a variable we are just hardcoding it in the deployment everywhere it is needed, and that is not good.

tlvu: To not constraint any data type in any catalog, how about renaming from `RAW_DATA` and `DATA` flavors to simply `CATALOG_1` and `CATALOG_2`.   Given the generic variable name, the default values will look funny.  Just encourage users to set different names for them in the documentation.

fmigneault: @tlvu 
>   Just document they are just 2 catalogs, nothing special, the admin will decide what goes in there. It could be just testdata, just wps_outputs, not a mixed bag of everything like "Birdhouse" is right now.

I strongly disagree with this.

The plateform itself, especially with Cowbird, tries to establish "shared" locations between services, and synchronize their permissions. 
It is already hard enough to make everything work with the various interfaces of all these services, I would like at least to have 1 directory that is "platform-managed". So maybe `PLATFORM_DATA` or `SERVICE_DATA` could work as well instead of `BIRDHOUSE_DATA`. Basically, I would like the name to make it stand out for admins as "*you should not host your data here, services could wipe it*", and avoid this mixed uses that were listed. It over-complicates how Cowbird needs to work around different locations and filesystem structures without breaking permissions. 

For the other data locations, whether there is 1, 2 or more, and whether it is called "datasets" or anything else doesn't matter that much for me, because it is for data-dumping whatever the node admin wants to share.

> So I suggest we add a placeholder variable that can be set in env.local to dynamically add extra xml snippets to add any number of new catalogs. 

That would be a good way to help node admins add more dataset locations as needed without completely overriding the current catalog XML.

fmigneault: If that can make the life easier for users/dev/maintainers, then that is fine. 👍

mishaschwartz: Ok, I'm going to try to summarize where I think we're currently at with this discussion:

1. We need one location for data that is created/managed by components like cowbird and weaver  
    - Let's use `SERVICE` in the variable names for this location (I don't have a strong preference but I think we just need to pick one)
2. We can have any other number of locations that can be added by adding extra xml snippets
     - note that we will also need a mechanism to specify which locations on the host should be mounted to the thredds container but we can probably do that by setting variables in env.local as well

Any objections to this strategy?

tlvu: Agreed with 1 of the 2 location called `SERVICE` to denote its shared service nature.  Mention that in the documentation.

fmigneault: Sounds good. We can have 1 `SERVICE` and 1 `DATASET` to make 2 locations quickly avaiable (without need of custom/complicated XML override), and N extentable locations to the admin's discretion using a 3rd variable.

dchandan: If we are deciding that one of the two locations will be called `SERVICE` then we will need the ability to have two datasets right from the start. So, I will have to include the XML extension ability in this PR as well. Give me some time to figure that out.

tlvu: @dchandan The ability to have extra catalogs is simply a variable in `env.local` that will be template expanded into the Thredds config file.  Should be really simple, unless there is template expansion problem.

For the 2 existing catalogs, hardcode them to make things simple.  Do not make the 2nd catalog configurable. 

A new org deploying PAVICS, will simply put all its data into the second `DATASET` catalog and not touch the `SERVICE` catalog.  In the `SERVICE` catalog, the only item is `wps_outputs` that are shared by all the birds and Cowbird.

fmigneault: >  the only item is `wps_outputs`

And `testdata` as well IMO, but I agree with @tlvu for the rest.

dchandan: > unless there is template expansion problem.

Yeah, that is where I expect any potential problems to arise. Hopefully, not.

fmigneault: If the variable uses `'` quotes, there shouldn't be issues to inject XML as is, since it would be dropped in without shell expansions, but that would not allow for it to nest other template variables easily. Otherwise, `"` are possible but they must deal with any necessary escapes since `"` are used extensively in XML. 

Alternatively, including another XML file could be possible if THREDDS's XML parser allows it (not sure if it does?). 
```xml
<!DOCTYPE doc [
<!ENTITY otherFile SYSTEM "otherFile.xml">
]>
&otherFile;
```
Assuming this is allowed, the admin could then volume mount that XML, which could be generated from a corresponding `otherFile.xml.template`, and not have to worry about any escapes.

tlvu:
Just so it's clear to the user, please give an example of adding a new Thredds catalog via the placeholder variable in `env.local.example`.  For example of what kind of escape is required when using the placeholder variable.

birdhouse/config/thredds/default.env
dchandan: Done. I hope [this implementation](https://github.com/bird-house/birdhouse-deploy/pull/359/commits/8c5f78c27dfba47c18ba7b34918cb1df2e013923) makes sense.

birdhouse/env.local.example
```diff
@@ -409,6 +409,28 @@ export GEOSERVER_ADMIN_PASSWORD=geoserverpass
...
+export THREDDS_ADDITIONAL_CATALOG=""
+#export THREDDS_ADDITIONAL_CATALOG="
+#    <datasetScan name="dataset_location_name" ID="dataset_url_path" path="dataset_url_path" location="dataset_location_on_container">
+#
+#      <metadata inherited="true">
+#        <serviceName>all</serviceName>
+#      </metadata>
+#
+#      <filter>
+#        <include wildcard="*.nc" />
```
tlvu, line 423: If you actually uncomment his example code, does it work?  No need to escape all the double quotes?

dchandan: Oh right... I haven't tested. I think just replacing it with single quote is fine right?

tlvu: Yes I think single quote would do the trick.  Easy, just uncomment that sample override to test.

dchandan: Fix all problems. Tested to check the feature works.

birdhouse/config/thredds/default.env
```diff
@@ -5,11 +5,30 @@ export THREDDS_VERSION=4.6.18-unidata-2022-01
 export THREDDS_DOCKER=pavics/thredds-docker
 export THREDDS_IMAGE="${THREDDS_DOCKER}:${THREDDS_VERSION}"
 export THREDDS_ORGANIZATION="Birdhouse"
+export THREDDS_ADDITIONAL_CATALOG=""
+
+# All default values for data locations are set to ensure backward compatibility
+export THREDDS_DATASET_LOCATION_ON_CONTAINER='/pavics-ncml'
+export THREDDS_SERVICE_DATA_LOCATION_ON_CONTAINER='/pavics-data'
```
tlvu, line 12: Might as well say `/pavics-ncml` and `/pavics-data` are for backward compatibilty because their defaut value are also "funny".

birdhouse/config/thredds/docker-compose-extra.yml
```diff
@@ -19,8 +18,8 @@ services:
       PAVICS_FQDN_PUBLIC: $PAVICS_FQDN_PUBLIC
     volumes:
       - thredds_persistence:/usr/local/tomcat/content/thredds
-      - ${DATA_PERSIST_ROOT}/datasets:/pavics-data
-      - ${DATA_PERSIST_ROOT}/ncml:/pavics-ncml
+      - ${THREDDS_SERVICE_DATA_LOCATION_ON_HOST}:${THREDDS_SERVICE_DATA_LOCATION_ON_CONTAINER}
+      - ${THREDDS_DATASET_LOCATION_ON_HOST}:${THREDDS_DATASET_LOCATION_ON_CONTAINER}
       - wps_outputs:/pavics-data/wps_outputs
```
tlvu, line 23: `/pavics-data` here to replace by the proper var.  Do a full repo search for the string `/pavics-data` and `/pavics-ncml` to be sure.

tlvu (approved):
Looks good now.  Some old hardcode did not replaced by the new vars.

Please test all hardcode has been replaced please override all the new vars in `env.local` and `./pavics-compose.sh down` and `./pavics-compose.sh up -d` see if a WPS service can still produce wps outputs that can be downloaded and Thredds can start with all the new catalog names and url path.

birdhouse/config/thredds/docker-compose-extra.yml
dchandan: Good catch! Thanks!

birdhouse/components/weaver/config/proxy/docker-compose-extra.yml
```diff
@@ -7,6 +7,6 @@ services:
       - ./components/weaver/config/proxy/conf.extra-service.d:/etc/nginx/conf.extra-service.d/weaver:ro
       # because of mounting path naming restrictions (see note in 'worker' definition),
       # we must add the custom path on top of named 'wps_outputs' volume of other birds for the proxy to expose results
-      - ${WEAVER_WPS_OUTPUTS_DIR}:/pavics-data/wps_outputs/weaver:ro
```
fmigneault, line 10: Not sure what `THREDDS` has to do here for `weaver`/`proxy` service combination regarding WPS outputs.

Same comment for all other locations edited in https://github.com/bird-house/birdhouse-deploy/pull/359/commits/91af29eff88393f5a7d4c1778c8e5d90a8e8615d that are not directly related to `thredds` service.

fmigneault (changes requested):

birdhouse/components/weaver/config/proxy/docker-compose-extra.yml
```diff
@@ -7,6 +7,6 @@ services:
...
+      - ${WEAVER_WPS_OUTPUTS_DIR}:${THREDDS_SERVICE_DATA_LOCATION_ON_CONTAINER}/wps_outputs/weaver:ro
```
fmigneault, line 10: Not sure what `THREDDS` has to do here for `weaver`/`proxy` service combination regarding WPS outputs.

Same comment for all other locations edited in [`91af29e` (#359)](https://github.com/bird-house/birdhouse-deploy/pull/359/commits/91af29eff88393f5a7d4c1778c8e5d90a8e8615d) that are not directly related to `thredds` service.

fmigneault (changes requested):

birdhouse/components/weaver/config/proxy/docker-compose-extra.yml
dchandan: Aren't we putting the WPS outputs under "service data" location that is then served by THREDDS as a separate catalog? (this seems to be the case now).

fmigneault: Either we rename them, or we simply make the location in-docker a hardcoded `/data/wps_outputs/<bird>` for these cases. 
In the case of THREDDS, I understand that the in-docker location could matter because it allows different configurations for managing multiple datasets, but in WPS birds, there can only be 1 location for WPS outputs. It is really irrelevant where they are mounted, and it forces platform admins to learn how each bird behaves (which is not consistent). Making flexible path editable in docker for these cases can lead to more configuration errors than it seems.

dchandan: > It is really irrelevant where they are mounted, and it forces platform admins to learn how each bird behaves (which is not consistent).

But WPS outputs directory and user workspaces directory now need to be on the same mount point right? (for cowbird to synchronize WPS outputs to user workspaces) In this case, it seems to me that moving the variables out of THREDDS to another location and renaming them would be the way to go.

dchandan: Sorry, my bad. We are talking about location on container.

fmigneault: Weaver needs exact mounting on both sides when running its child processes. Once the results are generated, the same WPS outputs on the server can be mounted in other services any way they want. However, we align the paths in a similar fashion to what THREDDS will do, such that the request URL looks somewhat similar. 

Cowbird and Jupyter should mount WPS-outputs and user-worspaces in a similar fashion to be resolved correctly from the point of view of the user directory while using the notebook. 

Other cases (so far), can mount however they desire as they do not interact with other data locations.

dchandan: I am still quite confused, I admit the operations of weaver are quite opaque to me. At [first](https://github.com/bird-house/birdhouse-deploy/pull/359#discussion_r1358917967) it read to me like the issue is that variables from the THREDDS component are controlling the working of the weaver component and that was a sacrilege. Now, it sounds to me that you are saying the changes are will break something. Could you please clarify which is it?

fmigneault: For the case above where my comment was placed, `${WEAVER_WPS_OUTPUTS_DIR}:${THREDDS_SERVICE_DATA_LOCATION_ON_CONTAINER}/wps_outputs/weaver:ro` is applied under the `proxy` service. This will not break `weaver` service, because it only defines how to mount the WPS outputs generated by Weaver at some point, for access through the `/wpsoutputs` HTTPS alias. 

Under `weaver` service however, it must be mounted exactly as `${WEAVER_WPS_OUTPUTS_DIR}:${WEAVER_WPS_OUTPUTS_DIR}`, or it will break how it runs its processes.

In a similar fashion, I know that some birds like `finch` mount WPS outputs only as `wpsoutputs:/data/wpsoutputs` because it automatically nests its outputs under a sub-dir named `finch` (contrary to Weaver which must indicate the final `/weaver`), while others like `hummingbird` use `wps_outputs:/opt/birdhouse/var/lib/pywps/outputs` due to configuration particularities. Their respectives volumes/dirs are mounted in `proxy` such that they "align" natually as: 
```yaml
${PAVICS_FQDN_PUBLIC}/wpsoutputs
    /weaver
    /finch
    /hummingbird
    ...
```

The takeaway is that each service could have implementation-specific requirements, and trying to fit a single variable in each case is probably not feasable. In some cases, it can outright break the service. For these cases, I prefer to leave hardcoded mount locations, and remove the chances of a node admin "breaking" them unconsciously by changing configurable values. This is particularly important because "broken" paths do not necessarily raise an error, but could simply place outputs in wrong locations, and that can be hard to diagnose considering each service-specific logic/code-base.

dchandan: Thanks for the clarification. I've made some reversions, do these look good? My only remaining concern now is that I prefer not to use the term `pavics-data`, but we can think about that after we've settled that nothing is breaking with these latest changes.

fmigneault: Looks good.

dchandan: Great! Done. Can you approve the PR?

tlvu: @dchandan I started to be confused by all the changes.  Can you confirm if you change all the default values to something else, the different pieces of the stack still work together?  

https://github.com/bird-house/birdhouse-deploy/blob/91af29eff88393f5a7d4c1778c8e5d90a8e8615d/birdhouse/config/thredds/default.env#L11-L18 

Meaning
* we can browse Thredds catalog from a web browser
* we can add/remove permissions to Thredds catalogs in Magpie
* a WPS service can create a wps_output that is accessible via the proxy and Thredds
* whatever other workflows that Weaver and Cowbird supports that I am not very familiar with that @fmigneault  can fill in

fmigneault: To test all of this, you can use this notebook: 
https://github.com/Ouranosinc/pavics-sdi/blob/master/docs/source/notebook-components/weaver_example.ipynb

It will test Weaver, which relays a call to hummingbird WPS at the same time, using a NetCDF from the THREDDS catalog, and the resulting WPS output file should be available through the `/wpsoutput` endpoint.

fmigneault:
run tests

birdhouse/components/weaver/config/proxy/docker-compose-extra.yml
fmigneault: Uisng the default (unset) values for the new variables, operations seems to behave correctly.
https://github.com/bird-house/birdhouse-deploy/pull/359#issuecomment-1771124315

A test with overrides to validate everything would be good to make sure alternative setups work as intended.

dchandan: Thanks for running the tests. Should I still check WPS outputs using the notebook you linked above or is that check run as part of tests you initiated?

tlvu: > Thanks for running the tests. Should I still check WPS outputs using the notebook you linked above or is that check run as part of tests you initiated?

@dchandan FYI the automated tests use the default values.  To test that all the overrides works (which also indirectly test whether we forgot any replacement of hardcode to variable), that has to be done manually.

mishaschwartz (changes requested):
- There's also a reference to `${DATA_PERSIST_ROOT}/datasets` in `birdhouse/scripts/bootstrap-testdata` that should be updated to `${DATA_PERSIST_ROOT}/${THREDDS_SERVICE_DATA_LOCATION_ON_HOST}`

birdhouse/components/weaver/config/proxy/docker-compose-extra.yml
mishaschwartz: @dchandan @tlvu @fmigneault 

I've run the test notebook (https://github.com/Ouranosinc/pavics-sdi/blob/master/docs/source/notebook-components/weaver_example.ipynb) with non-default values for all the variables and the notebook ran without any issues. 

The only thing I had change in the notebook was the `WEAVER_TEST_DEFAULT_FILE` variable so that it could use the updated URL (since I changed `THREDDS_SERVICE_DATA_URL_PATH` from the default `birdhouse` for the test)

PR #360, opened 2023-07-25
Mount public wpsoutputs data

ChaamC:
## Overview

This PR addresses the need for users to have access to public wps outputs data. The latest version of Cowbird monitors the wps outputs data directory (which contains public data and user restricted data) and isolates its public data via hardlinks to another directory. This other directory is then mounted on JupyterLab instances to give data access to the users.

[cowbird#40](https://github.com/Ouranosinc/cowbird/pull/40) must first be merged before merging this PR.

### Non-breaking changes
- Add public wps outputs directory to Cowbird and add corresponding volume mount to JupyterHub.
- Add `WPS_OUTPUTS_DIR` env variable to manage the location of the wps outputs data.

### Breaking changes
- Update `cowbird` service from [1.2.0](https://github.com/Ouranosinc/cowbird/tree/1.2.0) to latest version
- Require `MongoDB==5.0` Docker image for Cowbird's database.

Because of the new `MongoDB==5.0` database requirement for Cowbird that uses (potentially) distinct version from other 
birds, a separate Docker image is employed only for Cowbird. If some processes, jobs, or other Cowbird-related data 
was already defined on one of your server instances, manual transfer between the generic 
`${DATA_PERSIST_ROOT}/mongodb_persist` to new  `${DATA_PERSIST_ROOT}/mongodb_cowbird_persist` directory must be 
accomplished. The data in the new directory should then be migrated to the new version following the same procedure as
described for Weaver in 
[Database Migration](https://pavics-weaver.readthedocs.io/en/latest/installation.html?#database-migration).

## Related Issue / Discussion

- Related to Jira task [DAC-570](https://crim-ca.atlassian.net/browse/DAC-570) and PR [cowbird#40](https://github.com/Ouranosinc/cowbird/pull/40)

- [x] Update Cowbird's version after Cowbird's PR is merged


birdhouse/components/cowbird/docker-compose-extra.yml
```diff
@@ -65,6 +69,20 @@ services:
...
+      - "${DATA_PERSIST_ROOT}/${USER_WORKSPACES}:${DATA_PERSIST_ROOT}/${USER_WORKSPACES}"
     restart: always
     logging: *default-logging
+
+  # Dedicated database for Cowbird, since other 'mongodb' image does not employ the same version.
+  cowbird-mongodb:
+    image: mongo:${COWBIRD_MONGODB_VERSION}
+    container_name: ${COWBIRD_MONGODB_SERVICE}
+    networks:
+      - "${COWBIRD_MONGODB_SERVICE}"
```
fmigneault, line 81: Since the name is hardcoded in `networks` below, it is better to hardcode it here as well, otherwise they could mismatch if the variable is overridden.

Idem in `cowbird` and `cowbird-worker` services.

CHANGES.md
```diff
@@ -41,6 +41,11 @@
   [CMIP Dataset Ingestion Workflows](https://github.com/cedadev/stac-generator-example/tree/master/conf).
 
 - Adds ``optional-components/stac-public-access`` to give public access to the STAC catalog.
+- Add public wps outputs directory to Cowbird and add corresponding volume mount to JupyterHub.
+- Update `cowbird` service from [1.2.0](https://github.com/Ouranosinc/cowbird/tree/1.2.0)
+  to [2.0.0](https://github.com/Ouranosinc/cowbird/tree/2.0.0)
```
fmigneault, line 46: Add end period.

CHANGES.md
```diff
@@ -41,6 +41,11 @@
   [CMIP Dataset Ingestion Workflows](https://github.com/cedadev/stac-generator-example/tree/master/conf).
 
 - Adds ``optional-components/stac-public-access`` to give public access to the STAC catalog.
+- Add public wps outputs directory to Cowbird and add corresponding volume mount to JupyterHub.
```
fmigneault, line 44: wps -> WPS

CHANGES.md
```diff
@@ -41,6 +41,11 @@
   [CMIP Dataset Ingestion Workflows](https://github.com/cedadev/stac-generator-example/tree/master/conf).
 
 - Adds ``optional-components/stac-public-access`` to give public access to the STAC catalog.
```
fmigneault, line 43: Changes must be rebased on top with latest versions.

birdhouse/components/cowbird/docker-compose-extra.yml
```diff
@@ -23,7 +23,11 @@ services:
       # root user
       COWBIRD_FILESYSTEM_ADMIN_UID: 0
       COWBIRD_FILESYSTEM_ADMIN_GID: 0
-      WORKSPACE_DIR: /${USER_WORKSPACES}
+      # Note that WPS_OUTPUTS_DIR and WORKSPACE_DIR must both point to paths from the same volume.
+      # This is to allow the creation of hardlinks between the wpsoutputs and the user workspace.
+      WPS_OUTPUTS_DIR: ${WPS_OUTPUTS_DIR}
+      WPS_OUTPUTS_PUBLIC_SUBDIR: ${WPS_OUTPUTS_PUBLIC_SUBDIR}
+      WORKSPACE_DIR: ${DATA_PERSIST_ROOT}/${USER_WORKSPACES}
```
fmigneault, line 30: Are the variables themselves actually used in Cowbird (through `os.getenv` or similar)? 
It seems they should already be included in the [cowbird/config.yml.template](https://github.com/bird-house/birdhouse-deploy/pull/360/files#diff-0752f5d9727dae924cef084a00c6b65c26945b4a08dd681bb0d6cf702b68d33e), but maybe I'm mistaken? 
Not a big deal though if they are provided again here.

fmigneault (changes requested):

CHANGES.md
ChaamC: I'll move those changes when the other PR is merged and after doing a new pull from master to get latest changes.

birdhouse/components/cowbird/docker-compose-extra.yml
ChaamC: The variables are indeed also included in the `config.yml.template` file, but when that template file gets converted to `config.yml` file, the env variables are not all resolved. I notice for example that the `WORKSPACE_DIR` and `WPS_OUTPUTS_DIR` keep their unresolved format `${VAR_NAME}`. By including the environment variables here in the `docker-compose`, the variables can get resolved later in Cowbird if necessary. 

When loading the config in Cowbird, it attempts to resolved remaining unresolved env vars found in the config 
 using `os.path.expandvars()`.

CHANGES.md
```diff
@@ -41,6 +41,11 @@
...
+- Add public wps outputs directory to Cowbird and add corresponding volume mount to JupyterHub.
+- Update `cowbird` service from [1.2.0](https://github.com/Ouranosinc/cowbird/tree/1.2.0)
+  to [2.0.0](https://github.com/Ouranosinc/cowbird/tree/2.0.0)
+- Require `MongoDB==5.0` Docker image for Cowbird's database.
```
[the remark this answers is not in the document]
ChaamC: I updated the changelog, I took most of the description that was already used in the precent similar change applied to Weaver.

.readthedocs.yml
```diff
@@ -19,14 +19,15 @@ sphinx:
...
     - requirements: docs/requirements.txt
 
 #conda:
 #  environment: environment-docs.yml
 
 build:
-  image: stable
+  os: "ubuntu-22.04"
+  tools:
+    python: "3.6"
```
fmigneault, line 33: Unless something breaks during build, 3.8 at least, since this is already EOL for a while.
(could also be a separate PR)

ChaamC: Version updated to 3.10, readthedocs seems to build successfully.

fmigneault (approved):

.readthedocs.yml
fmigneault: Very weird that even latest commit built RTD without problem with the options that you found were causing errors: 
https://github.com/bird-house/birdhouse-deploy/commit/ce5c14df65ebc0dbfc50e097dee0d2b50a0007f5

But your latest one is also good:
[`90611a7` (#360)](https://github.com/bird-house/birdhouse-deploy/pull/360/commits/90611a7984e5e1ef59c3562d77818e69af55e40f)

Funky stuff happening recently on RTD it seems...

ChaamC: The reason the latest build on master could be successful was because RTD is deprecating the option progressively, doing temporary deprecation enforcements until in a few weeks where the option becomes fully deprecated (see [blog here](https://blog.readthedocs.com/use-build-os-config/)). I just happened to build it on Sept. 18th where I got the deprecation error.

birdhouse/components/cowbird/docker-compose-extra.yml
```diff
@@ -23,20 +23,25 @@ services:
...
     depends_on:
-      - "${COWBIRD_MONGODB_SERVICE}"
+      - cowbird-mongodb
     volumes:
       - ./components/cowbird/config/cowbird/config.yml:/opt/local/src/cowbird/config/cowbird.yml
       - ./components/cowbird/config/cowbird/cowbird.ini:/opt/local/src/cowbird/config/cowbird.ini
       # even if not running tasks here, they must be registered to send them off to the right place!
       - ./components/cowbird/config/cowbird/celeryconfig.py:/opt/local/src/cowbird/config/celeryconfig.py
-      - "${DATA_PERSIST_ROOT}/${USER_WORKSPACES}:/${USER_WORKSPACES}"
+      - "${DATA_PERSIST_ROOT}:${DATA_PERSIST_ROOT}"
```
tlvu, line 44: Woo, this has write-access to the entire `$DATA_PERSIST_ROOT` !!!  Isn't this too much access?

I am guessing it only needs access to `JUPYTERHUB_USER_DATA_DIR="$DATA_PERSIST_ROOT/jupyterhub_user_data"` and Thredds wps_output dir?

birdhouse/default.env
```diff
@@ -52,3 +52,5 @@ export DEFAULT_CONF_DIRS='
 
 export USER_WORKSPACE_UID=1000
 export USER_WORKSPACE_GID=1000
+
+export WPS_OUTPUTS_DIR="${DATA_PERSIST_ROOT}/wps_outputs"
```
tlvu, line 56: Use single-quote for delayed eval and add to `DELAY_EVAL` list since it depends on another var.

tlvu (commented):
Sorry for the late reply.  This is a kind of PR I needed to deploy to my test VM for testing but I just don't have the bandwidth to work on multiple PR anymore.  The requirement for a VM has increased quite a bit and I am not able to run multiple VMs on my workstation simultaneously anymore.

Anyways, without a real deploy I spot a few changes that I think is worth considering fixing.

birdhouse/components/cowbird/docker-compose-extra.yml
mishaschwartz: @ChaamC correct me if I'm wrong... but the reason for this is that both of those directories need to be mounted in the same volume for hardlinks to work: 

@tlvu In #356 we introduce the [`DATA_PERSIST_SHARED_ROOT`](https://github.com/bird-house/birdhouse-deploy/pull/356/files#diff-752de5e561aca0a197c4f5dc44af11b0926716d3f953a1f9fe378670c37d6b10R32) variable here instead. This gives us the ability to mount _only the required subdirectories_ as as single volume.

tlvu: Yes I am a bit behind in most things cowbird.  Understood that for hard links to work, the src and dest of the link should be on the same filesystem.

So if we need to expose Thredds wps_outputs and possibly Geoserver workspace in the Jupyter env workspace, then we need to only volume-mount those 3, in the cowbird container, not the entire `/data`.

If `DATA_PERSIST_SHARED_ROOT` is the same as `DATA_PERSIST_ROOT` by default, then it comes back to the same as exposing the entire `/data` to cowbird writable.

Basically would something like the following works:
```yaml
services:
  cowbird:
    volumes:
      - ./components/cowbird/config/cowbird/config.yml:/opt/local/src/cowbird/config/cowbird.yml
      # even if not running tasks here, they must be registered to send them off to the right place!
      - ./components/cowbird/config/cowbird/celeryconfig.py:/opt/local/src/cowbird/config/celeryconfig.py
      - "${DATA_PERSIST_SHARED_ROOT}/jupyter_user_data:${DATA_PERSIST_SHARED_ROOT}/jupyter_user_data"
      - "${DATA_PERSIST_SHARED_ROOT}/datasets/wps_outputs:${DATA_PERSIST_SHARED_ROOT}/datasets/wps_outputs"
      - "${DATA_PERSIST_SHARED_ROOT}/geoserver/workspaces:${DATA_PERSIST_SHARED_ROOT}/geoserver/workspaces"
```

So the paths inside and outside of the cowbird container is exactly the same, giving it the impression it has access to the entire `/data` but it's not true.

By the way, I am not sure `${DATA_PERSIST_SHARED_ROOT}/datasets/wps_outputs` actually works since I think the actual wps_outputs is in a data-volume https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/config/wps_outputs-volume/docker-compose-extra.yml

Usage of that wps_outputs data-volume:
```
birdhouse/config/raven/config/wps_outputs-volume/docker-compose-extra.yml
6:      - wps_outputs:/data/wpsoutputs

birdhouse/config/thredds/docker-compose-extra.yml
24:      - wps_outputs:/pavics-data/wps_outputs
```

Anyways, many inter-connected pieces so not simple to wrap my head around.

mishaschwartz: > If `DATA_PERSIST_SHARED_ROOT` is the same as `DATA_PERSIST_ROOT` by default, then it comes back to the same as exposing the entire `/data` to cowbird writable.

Yes it would. But... I would like to see `DATA_PERSIST_SHARED_ROOT` not be the same as `DATA_PERSIST_ROOT`. The only reason it is like that by default is to maintain backwards compatibility. 

> Basically would something like the following works:

No, unfortunately that wouldn't work. They actually need to be mounted at the same mount-point. The relative location matters for symlinks, but a shared mount-point matters for hard-links.

> By the way, I am not sure `${DATA_PERSIST_SHARED_ROOT}/datasets/wps_outputs` actually works since I think the actual wps_outputs is in a data-volume

I don't have enough knowledge to comment on this. Sorry

fmigneault:
@ChaamC Can you take a look at the items following @tlvu review?

birdhouse/components/cowbird/docker-compose-extra.yml
ChaamC: > @ChaamC correct me if I'm wrong... but the reason for this is that both of those directories need to be mounted in the same volume for hardlinks to work:

Yes, exactly. If we want to share the wps_outputs data to the user, we need to use hardlinks, and cowbird requires the src and destination of the hardlinks to be in the same volume/file partition, or it will trigger a `Cross-device link`.

I needed to mount the full data directory to be able to make hardlinks between files from the `/data/wps_outputs` and `/data/user_workspaces` directories.

We can update the volume mount to use the upcoming new variable `DATA_PERSIST_SHARED_ROOT` as long as all files used by cowbird for this PR's feature are contained in this directory.

birdhouse/default.env
ChaamC: Seems to be fixed in a new [commit](https://github.com/bird-house/birdhouse-deploy/pull/356/commits/601749ea29663e69038542903f48f919a0ba48ba) from PR [#356](https://github.com/bird-house/birdhouse-deploy/pull/356), so I won't add it here for now, unless we decide to change things about the data volume concerns.

tlvu: Yes, ignore this comment, I saw the fix in PR https://github.com/bird-house/birdhouse-deploy/pull/356 which is already merged.

birdhouse/components/cowbird/docker-compose-extra.yml
tlvu: Just to double check, for existing deployments, if we want `DATA_PERSIST_SHARED_ROOT=/data_shared_mount` different than `DATA_PERSIST_ROOT=/data`, we have to move 
  * /data/wps_outputs => /data_shared_mount/wps_outputs
  * /data/user_workspaces => /data_shared_mount/user_workspaces
  * anything else needed for the migration?
  
I think this migration should be documented because anyone starting with the current default values and do not want Cowbird to have access to the entire `/data` will have to perform this manual migration.

Or we set a different value immediately in `default.env` and in `env.local.example` we document the backward-compatible value `DATA_PERSIST_SHARED_ROOT=/data` and warn this gives Cowbird too much write-access.  I prefer different default value now to avoid migration later.  Cowbird is not in the default enable list right now so this should not break too many existing deployments.

Also, I still do not understand why we are referring to `/data/wps_outputs` when currently wps_outputs is a data-volume as mentioned in my comment https://github.com/bird-house/birdhouse-deploy/pull/360#discussion_r1336344103 above.  

This means different volume-mount between `wps_outputs` dir and `user_workpaces` dir and hardlink probably do not work.

Does Cowbird fallback to a regular copy when hardlink do not work?  Does it log somewhere when hardlink do not work?

Is there a notebook or test script that tests Cowbird in the PAVICS stack I can try?

Also question for the future as I heard maybe Cowbird will manage Geoserver workspace too.  Does this means we will need to migrate Geoserver data dir under `DATA_PERSIST_SHARED_ROOT` later?

ChaamC: > anything else needed for the migration?

That's it, at least concerning this PR's changes.
Although, I'm thinking about something else to watch out for. Not sure if this variable's path could be affected, but changing the [`JUPYTERHUB_USER_DATA_DIR`](https://github.com/bird-house/birdhouse-deploy/blob/547cc085c52924a21d853eeb1149790c2cb7577d/birdhouse/config/jupyterhub/default.env#L22) variable's path could break the symlink that is used in the cowbird workspace to link to the `JUPYTERHUB_USER_DATA_DIR` directory. So, maybe something to watch out for. The user workspaces could eventually be resynchable if we decide to change the jupyterhub path, but the resync is not yet implemented in cowbird for this symlink, so for now it would need a manual recreation of symlinks in each user workspace if that variable is changed.

If we know we will eventually do the migration, I don't see any reason not to do it now. I think it would be okay to change it now if you prefer.

> Also, I still do not understand why we are referring to /data/wps_outputs when currently wps_outputs is a data-volume as mentioned in my comment https://github.com/bird-house/birdhouse-deploy/pull/360#discussion_r1336344103 above.
> This means different volume-mount between wps_outputs dir and user_workpaces dir and hardlink probably do not work.

I am not sure to understand the problem here. Do you mean the fact that the birds who will produce wps_outputs data only mount the wps_outputs volume (and not the user workspace volume)? Any bird that produces wps_outputs only needs access to the wps_outputs volume. Only cowbird requires the full access to the data directory to allow the creation of hardlinks between the wps_outputs and user workspace directories. After that, when the hardlink is created, the volume mount does not require the full data mount to have the hardlink accessible. It is really only for the moment of creation, in cowbird that we require the full mount of the data directory.

> Does Cowbird fallback to a regular copy when hardlink do not work? Does it log somewhere when hardlink do not work?

If the hardlink fails, the failure is only logged in Cowbird's logs. We want to avoid having 2 independant copies that could diverge. We did implement the resync function for the wps_outputs data, so if we need, we can call that API endpoint on cowbird, and cowbird will regenerate any missing hardlinks that could have resulted from a previous failure.

> Is there a notebook or test script that tests Cowbird in the PAVICS stack I can try?

Nope, not at the moment, although I am checking to create a notebook to verify the different parts of the user workspace (notebook symlink, wps_outputs hardlinks, geoserver data, etc.), but I am not sure yet if it is achievable.

> Also question for the future as I heard maybe Cowbird will manage Geoserver workspace too. Does this means we will need to migrate Geoserver data dir under DATA_PERSIST_SHARED_ROOT later?

Cowbird currently already manages geoserver's file permissions on the user workspace. Since the data is only found in the user workspace directly, it doesn't require any symlink/hardlinks, as data is already accessible to the user,it shouldn't be a problem. 
Unless there is external geoserver data to be made accessible to the user that I am not aware of yet?

PR #361, opened 2023-07-31
Make local development possible with self-signed SSL certificate

mishaschwartz:
## Overview

- Ensure that containers in the stack do not require access to the host machine in order to operate properly
- This means that a developer can deploy a local version of the stack without requiring that the deployment be exposed publicly on the internet.

## Changes

**Non-breaking changes**

- the stack will now work with self signed certificates for localhost

**Breaking changes**

- None for production

## Additional Information

- [ ] currently monitoring capabilities (eg. canarie-api) don't work in local development because they depend on having access to routes only visible to the host machine.


birdhouse/README.rst
```diff
@@ -114,6 +114,37 @@ Disk: at least 100 TB, depending how much data is hosted on Thredds and Geoserve
...
+
+.. code-block::
+
+    openssl req -x509 -out localhost.crt -keyout localhost.key \
+      -newkey rsa:2048 -nodes -sha256 \
+      -subj '/CN=localhost' -extensions EXT -config <( \
+       printf "[dn]\nCN=localhost\n[req]\ndistinguished_name = dn\n[EXT]\nsubjectAltName=DNS:localhost\nkeyUsage=digitalSignature\nextendedKeyUsage=serverAuth")
+
+    # https://ubuntu.com/server/docs/security-trust-store
+    sudo cp localhost.crt /usr/local/share/ca-certificates/
```
tlvu, line 136: Oh I don't have this part in my Vagrant steps.  If you use Vagrant could you also update https://github.com/bird-house/birdhouse-deploy/blob/2b344d3383d8c0cd3f4cc4a0bf4266e3bae06a47/birdhouse/vagrant-utils/configure-pavics.sh#L65-L68

If not, it's okay, I can next time I try to rebuild my Vagrant box from scratch.  I do this about once a year.

birdhouse/README.rst
```diff
@@ -114,6 +114,37 @@ Disk: at least 100 TB, depending how much data is hosted on Thredds and Geoserve
...
+
+    cat localhost.crt localhost.key > cert.pem
+
+    rm localhost.crt localhost.key
+
+Then set the following following variables in the env.local file:
+
+* `SSL_CERTIFICATE=/absolute/path/to/cert.pem`
+* `PAVICS_FQDN=localhost`
+
```
tlvu, line 146: VERIFY_SSL=false?

tlvu (commented):
Sorry for commenting on a draft PR.  I usually only comment on ready to review PR because I understand draft PR are either not ready or for testing or prototyping only.

Just some minor points.

birdhouse/docker-compose.yml
```diff
@@ -1,3 +1,7 @@
 version: "3.4"
 
 services: {} # no services loaded by default
+
+networks:
+  default:
+    name: ${BIRDHOUSE_DEFAULT_NETWORK}
```
tlvu, line 7: Isn't this the default network name?  Do we still need to specify it?  But I guess no harm.

tlvu (approved):
I think this is good enough for local development.

Maybe add a note in the README that this setup do not work if someone use Jenkins to hit the PAVICS host that use self-signed SSL cert but for manual testing and manually accepting the mismatch self-signed cert and hostname, it should work.

birdhouse/docker-compose.yml
mishaschwartz: Yes, the default value for this is the same as the default that docker-compose specifies. But, it's nice to be explicit about the name so that we don't have hard-coded values in the code with the assumption that the network name will never change.

birdhouse/README.rst
```diff
@@ -114,6 +114,37 @@ Disk: at least 100 TB, depending how much data is hosted on Thredds and Geoserve
...
+Then add the certificate you just generated to the trust store as described `here <https://ubuntu.com/server/docs/security-trust-store>`_ .
+Finally, combine the generated key and certificate files together to use as the `SSL_CERTIFICATE` for this stack.
+
+The following should work to generate a self-signed ssl certificate on ubuntu 20.04+:
+
+.. code-block::
+
+    openssl req -x509 -out localhost.crt -keyout localhost.key \
+      -newkey rsa:2048 -nodes -sha256 \
+      -subj '/CN=localhost' -extensions EXT -config <( \
```
fmigneault, line 132: Shouldn't this include a `-days N` as well to avoid it being expired after a few days?

birdhouse/README.rst
fmigneault: I don't know how the trust store works with the .crt, but on my side, I only did the concat of the files and it works. What is the purpose exactly of this step?

birdhouse/README.rst
```diff
@@ -114,6 +114,37 @@ Disk: at least 100 TB, depending how much data is hosted on Thredds and Geoserve
...
+       printf "[dn]\nCN=localhost\n[req]\ndistinguished_name = dn\n[EXT]\nsubjectAltName=DNS:localhost\nkeyUsage=digitalSignature\nextendedKeyUsage=serverAuth")
+
+    # https://ubuntu.com/server/docs/security-trust-store
+    sudo cp localhost.crt /usr/local/share/ca-certificates/
+
+    cat localhost.crt localhost.key > cert.pem
```
fmigneault, line 138: I think the generated `cert.pem` file needs to be `chmod 0600`, otherwise `nginx` complains in some situations about invalid permissions.

CHANGES.md
```diff
@@ -15,7 +15,12 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Changes
+- Make local development possible with self-signed SSL certificate
+  - Ensure that containers in the stack do not require access to the host machine in order to operate properly
+  - This means that a developer can deploy a local version of the stack without requiring that the deployment
```
fmigneault, line 21: No separate items for a sentence that resumes a thought from the previous one.

birdhouse/components/weaver/post-docker-compose-up
```diff
@@ -246,7 +246,8 @@ for prov in ${WEAVER_WPS_PROVIDERS}; do
     if [ -z "${prov}" ]; then
         continue
     fi
-    prov_url="https://${PAVICS_FQDN_PUBLIC}${TWITCHER_PROTECTED_PATH}/${prov}"
+    prov_url="${prov#*:}"
+    prov="${prov%%:*}"
     prov_cap="${prov_url}?service=WPS&request=GetCapabilities"
```
fmigneault, line 251: Just commenting here, but applies for all `default.env` of each WPS bird provider as well.

The public URL is intentionally employed such that permissions applied on those respective birds are enforced even when sending the execution request through Weaver. Using the internal URL bypasses this security check, which could allow a user to execute protected WPS using Weaver as a backdoor.

birdhouse/config/twitcher/twitcher.ini.template
```diff
@@ -65,7 +65,7 @@ twitcher.workdir =
 twitcher.prefix =
 
 # magpie (for twitcher.adapter)
-magpie.url = https://${PAVICS_FQDN}/magpie
+magpie.url = http://magpie:2001
```
fmigneault, line 68: This is not used only for performing requests. 
It is used for reporting back to the user the access point they should use or be redirected to when interacting with Magpie.
It is also used to set exposed hostname references in authorization headers/cookies. 
This change will break some logins. 

A separate config parameter could be considered to handle internal requests, but for now this is not possible.

fmigneault (changes requested):
I like the addition of the SSL self-signed doc. Good for someone that didn't know how to do it.

However, I don't understand why the rest of the changes are required?
Using self-signed certificates already worked. One only had to use the local IP of their machine instead of the generic `localhost`. No need for internet access.

birdhouse/README.rst
mishaschwartz: the nginx docker container runs all workers as root so there is no permission limits enforced

birdhouse/components/weaver/post-docker-compose-up
mishaschwartz: I thought cowbird synced the permissions so that this wasn't necessary

birdhouse/config/twitcher/twitcher.ini.template
mishaschwartz: That's unfortunate. I think this limitation is enough to kill this PR for now until this can be resolved.

birdhouse/components/weaver/post-docker-compose-up
fmigneault: For the moment, Cowbird only syncs permissions of endpoints to access job outputs for the same user that pushed the job. 
The permissions between Weaver and the "remote" WPS is not synced.

Either way, I still prefer using the public URLs because they are displayed in `/weaver/providers` and `/weaver/providers/{name}` responses.

birdhouse/README.rst
fmigneault: I don't think it is a problem in the sense of accessing/reading the file. It is nginx or the underlying SSL libraries themselves that flag the permissions as invalid, as it is prone to security problems. Running as root is valid, but it should be only root that can read it (no groups/others), and for security, it should not be modifiable from within the container, so `600` + mount with `:ro`.

PR #363, opened 2023-08-09
Updating contribution

Nazim-crim:
## Overview
Updating contribution relating changes made in daccs-e2e-jenkins

## Changes

birdhouse_skip_ci: false


CONTRIBUTING.rst
```diff
@@ -37,6 +37,10 @@ Policy rules
 
 3. Contributions will trigger a test suite that must successfully pass before being merged (or integrated).
 
+    * The test suite can be run using a different DACCS config with ``daccs_configs_branch: branch_name`` in the PR description.
```
mishaschwartz, line 40: Personally, I think these variables should be named `birdhouse_configs_branch` and `birdhouse_skip_ci`.

DACCS is the name of the organization that is developing Marble which uses software from Birdhouse. 

Mixing these names around can get confusing very quickly. I think that we should always use "birdhouse" to refer to elements that are specific to this repository.

mishaschwartz (changes requested):

mishaschwartz:
@matprov @Nazim-crim Thank you for developing this! This is a great feature

CONTRIBUTING.rst
mishaschwartz: @matprov  I think mentioning daccs at all is going to lead to confusion, but if it can't be avoided then it can't be helped. I would just add an explanatory note somewhere why the name "daccs" is used here.

CONTRIBUTING.rst
```diff
@@ -37,6 +37,10 @@ Policy rules
...
+    * It is possible to skip the test suite if the latest commit contains either ``[skip ci]``, ``[ci skip]`` or ``[no ci]``.
+      To globally skip the test suite regardless of the commit message use ``daccs_skip_ci: true`` in the PR description.
+
 4. Contributions must be reviewed by every willing organizations
    (Default reviewers are `@tlvu`_ for `Ouranos`_ and `@MatProv`_ for `CRIM`_).
```
tlvu, line 45: Since we are at it, should we add Misha as default reviewer for UofT for this repo and update CRIM default reviewer as well?

Nazim-crim: @mishaschwartz Can you check if you want to add something else under Contributions

fmigneault (approved):

tlvu (approved):
+1

mishaschwartz (approved):

PR #372, opened 2023-08-18
Renaming DACCS to Marble in service-config.json.template files

dchandan:
## Overview

Small change to the location of schema describing node services following the renaming of DACCS to Marble.

## Changes

**Non-breaking changes**
- Changed https://github.com/DACCS-Climate/DACCS-node-registry to https://github.com/DACCS-Climate/Marble-node-registry in all `service-config.json.template` files.

**Breaking changes**
None


tests/test_deployment.py
```diff
@@ -1,10 +1,10 @@
-import os
-import pytest
 import glob
 import json
+import os
```
tlvu, line 3: I am surprised the import is really missing?

tlvu (approved):
LGTM but I think @mishaschwartz knows more about this rename.

tests/test_deployment.py
dchandan: No it's not, its just the IDE sorting the imports.

mishaschwartz (approved):
Looks good. ~This makes me realize that we need a servic-config.json.template file for stac as well. Though we should do that in a different PR.~

EDIT: @dchandan you're on it already! (#373) Thank you!

fmigneault (approved):

PR #373, opened 2023-08-18
Adding service information for the STAC catalog

dchandan:
## Overview

STAC is a new service to the birdhouse-stack and it should be made visible at the `/services` endpoint. Currently this is missing.

## Changes

**Non-breaking changes**
- Adding a `service-config.json.template` file to `birdhouse/components/stac`.

**Breaking changes**
N/A


birdhouse/components/stac-browser/service-config.json.template
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "https://raw.githubusercontent.com/DACCS-Climate/Marble-node-registry/main/node_registry.schema.json#service",
+  "name": "stac-browser",
+  "keywords": [
+    "catalog"
+  ],
+  "description": "A web interface to browse the STAC catalog of the Marble Network.",
```
mishaschwartz, line 7: Don't mention Marble here (see comment below)

mishaschwartz (changes requested):

CHANGES.md
```diff
@@ -15,7 +15,9 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## CHANGES
+
+- Adding a description for the STAC service that will be served at the `/services` endpoint
```
fmigneault, line 20: Need sync with latest commits.

fmigneault (commented):
I agree with other comments from @mishaschwartz

birdhouse/components/stac/service-config.json.template
```diff
@@ -0,0 +1,20 @@
...
+  "links": [
+    {
+      "rel": "service",
+      "type": "text/html",
+      "href": "https://${PAVICS_FQDN_PUBLIC}/stac/"
+    },
+    {
+      "rel": "service-doc",
+      "type": "text/html",
+      "href": "https://stacspec.org/en"
```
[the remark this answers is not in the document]
mishaschwartz: > https://raw.githubusercontent.com/radiantearth/stac-api-spec/main/core/openapi.yaml ?

Wouldn't that be better for a `service-desc` link than a `service-doc` link?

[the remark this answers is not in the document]
fmigneault: Yes. 
Just need to be careful about which links are provided. "STAC Spec" refers to the contents of a STAC document, while "STAC-API Spec" are for the service.

Even then, it's somewhat complicated because there are multiple OpenAPI for core, item-seach and ogc-api-features extensions.

Anyway, I think for the JSON structure of the API, the following would be the best we can do:

```json
{
  "rel": "service-doc",
  "type": "text/html",
  "href": "https://github.com/radiantearth/stac-api-spec"
},
{
  "rel": "service-desc",
  "type": "application/yaml",
  "href": "https://raw.githubusercontent.com/radiantearth/stac-api-spec/main/core/openapi.yaml"
}
```

dchandan:
I think all suggestions and requests are addressed now.

birdhouse/components/stac/service-config.json.template
```diff
@@ -0,0 +1,25 @@
...
+    },
+    {
+      "rel": "service-doc",
+      "type": "text/html",
+      "href": "https://github.com/radiantearth/stac-api-spec"
+    },
+    {
+      "rel": "service-desc",
+      "type": "application/yaml",
+      "href": "https://raw.githubusercontent.com/radiantearth/stac-api-spec/main/core/openapi.yaml"
```
fmigneault, line 22: I think it would be better actually to refer to a pinned version (release/tag). 
https://github.com/radiantearth/stac-api-spec/tree/v1.0.0 should be good. 
We should probably make this a configurable env-var with the corresponding app tag for the docker image. 
Right now, `crim-ca/stac-app:main` is used as well: 
https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/components/stac/docker-compose-extra.yml#L13

dchandan: I agree about the pinned version. I didn't quite understand your second comment..

fmigneault (approved):

tlvu (approved):
LGTM

birdhouse/components/stac/service-config.json.template
mishaschwartz: @fmigneault can this be addressed in a different PR? It's a good point but probably should be addressed later when we actually start using a version other than `crim-ca/stac-app:main`

fmigneault: Yes. I agree to move it to another PR while addressing the reference image.

mishaschwartz (approved):

mishaschwartz (changes requested):
Needs update to .gitignore and info in `stac-browser/service-config.json` doesn't actually get added to the json string.

(I can work on this)

EDIT: fixed

mishaschwartz (approved):

mishaschwartz:
@fmigneault @tlvu @dchandan  I had to relax the tests to permit a service config file to allow the name of _any_ service in the file to be the same as the component folder since we have to now support multiple services per component (the `stac` component offers two services: stac and stac-browser). 

The other option is to treat both stac and stac-browser as the same service and include the url for one of them in the `links` key for the other. For example

```
{
  "name": "stac",
  ...
  "links": [
      {"rel": "via",
       "type": "application/json",
       "href": "https://${PAVICS_FQDN_PUBLIC}/stac/"},
       ...
   ]
}
```

I chose `via` as the rel value since it "Identifies a resource that is the source of the information in the link's context." (see [here])(https://www.iana.org/assignments/link-relations/link-relations.xhtml) but I'm open to other suggestions.

What is your preferred strategy for this? I'm leaning towards allowing multiple services per component because it allows us more flexibility and I'm having a hard time seeing a downside to this.

fmigneault:
@mishaschwartz  
I like the idea of linking the services between each other. The rel `alternate` could also be considered.

fmigneault (commented):
Waiting on follow-up on https://github.com/bird-house/birdhouse-deploy/pull/373#issuecomment-1739263443 and https://github.com/bird-house/birdhouse-deploy/pull/373#issuecomment-1739811266.

mishaschwartz:
@fmigneault 

Just to clarify:

>  I like the idea of linking the services between each other. The rel alternate could also be considered.

Do you recommend two services (stac and stac-browser) with links referring to each other, or do you recommend one service (stac-browser) with a link to the stac endpoint?

tlvu (approved):
LGTM

fmigneault:
> Do you recommend two services (stac and stac-browser) with links referring to each other, or do you recommend one service (stac-browser) with a link to the stac endpoint?

Two services with links referring to each other since they are related, but still two distinct entities.

Each service in this case has different references: `/stac/` pointing at STAC-API spec and crimca/stac-app implementation;  while STAC-browser using the stac-utils/pgstac implementation. 

For the test validation portion, matching at least 1 service by name seems reasonable. Cross-links could also be validated. 
Service configs unmatched by name must still pass schema validation though.

PR #378, opened 2023-09-06
Adding new jupyterlab image 

Nazim-crim:
## Overview
Add a jupyterlab image to be launch from jupyterhub containing mlflow, this image can be found under https://github.com/bird-house/pavics-jupyter-base/tree/mlflow-proxy 

## Changes
**Non-breaking changes**
- Changes jupyterhub version from 1.4.0-20210506 to 4.0.2-20230816
- Add an image to the list of images that can be launched from JupyterHub which will be used to start an instance of MLflow.

## Related Issue / Discussion

- Resolves [DAC-573](https://crim-ca.atlassian.net/browse/DAC-573) 

Notes
jupyter lab google drive extension is not supported with this image


birdhouse/config/jupyterhub/default.env
```diff
@@ -5,7 +5,7 @@
 # are applied and must be added to the list of DELAYED_EVAL.
 
 export JUPYTERHUB_DOCKER=pavics/jupyterhub
-export JUPYTERHUB_VERSION=1.4.0-20210506
+export JUPYTERHUB_VERSION=3.0.0-20220926
```
fmigneault, line 8: Just to be sure, are the existing images that used the older JupyterHub version still able to run using this newer version?

fmigneault (commented):
Need to add the details from the PR description to the CHANGES.rst file.
Otherwise, looks good.

birdhouse/config/jupyterhub/default.env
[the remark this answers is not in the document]
Nazim-crim: Last time I had spawning issues with newer versions, but this not seems to be the case anymore. We can use the new one

Nazim-crim: Yes jupyterLab works out of the box with JupyterHub 1.0 +

CHANGES.md
```diff
@@ -22,6 +22,10 @@
 
 ## Changes
 
+- Changes `JUPYTERHUB_VERSION` from `1.4.0-20210506` to `4.0.2-20230816`
```
fmigneault, line 25: Add the note from your PR description as well: 
"jupyter lab google drive extension is not supported with this image"

Nazim-crim:
run tests

birdhouse/config/jupyterhub/docker-compose-extra.yml
```diff
@@ -40,6 +40,8 @@ services:
     # the Hub and the rest of the stack are on network 'default'
       - default
       - jupyterhub_network
+    # ensure the deployment does not fail when upgrading Jupyterhub version
+    command: bash -c "jupyterhub upgrade-db && jupyterhub"
```
fmigneault, line 44: @tlvu 
I would like to have your opinion whether auto upgrade-db seems reasonable? Or do we purposely want this operation to be manual to allow some backups as needed beforehand? Is it possible that this corrupts the database?

tlvu: I've tried it before and it does make a backup before performing the upgrade so if a corruption happens, it's reversible.  Maybe this backup only happens because we used the file based DB, if it had been Postgres, this auto backup won't happen.

I just wonder why the Jupyterhub team did not want auto updating the DB in the first place.

Anyways, for our specific setup, it's probably fine.

birdhouse/env.local.example
```diff
@@ -242,7 +242,8 @@ export GEOSERVER_ADMIN_PASSWORD=geoserverpass
 # Jupyter single-user server images
 #export DOCKER_NOTEBOOK_IMAGES="pavics/workflow-tests:210216 \
 #                               pavics/crim-jupyter-eo:0.3.0 \
-#                               pavics/crim-jupyter-nlp:0.4.0"
+#                               pavics/crim-jupyter-nlp:0.4.0 \
+#                               birdhouse/pavics-jupyter-base:mlflow-proxy"
```
tlvu, line 246: Just curious, why is this the base image?   I thought on CRIM's side, all specific image are child image of the base image and the base image stays generic?

CHANGES.md
```diff
@@ -31,6 +31,11 @@
...
+
+- Add an image to the list of images that can be launched from JupyterHub which will be used to start an instance of MLflow.
```
tlvu, line 36: This changelog is funny.  I think it has not been merged with latest of `master`.

birdhouse/config/jupyterhub/default.env
```diff
@@ -5,7 +5,7 @@
 # are applied and must be added to the list of DELAYED_EVAL.
 
 export JUPYTERHUB_DOCKER=pavics/jupyterhub
-export JUPYTERHUB_VERSION=1.4.0-20210506
+export JUPYTERHUB_VERSION=4.0.2-20230816
```
tlvu, line 8: Just curious, this upgrade is needed or just nice to have?  Maybe explain the reason for the upgrade of the JupyterHub in the changelog as well.

birdhouse/env.local.example
fmigneault: It is different in this case because the introduction of the MLflow extension breaks the Google Drive extension, since the Jupyter version is bumped, and gdrive is not compatible with latest changes (Jupyter v1 or v2 only, project archived and not updated since 2020). 

In the long run, we will probably bump the other images as well to realign them all with the same base image at some point, but this allows temporarily to keep the older Jupyter v1 images and this updated v4 one at the same time.

fmigneault (approved):
Ok my side, but check remaining comments from Long.

birdhouse/env.local.example
tlvu: Not sure I follow here.  JupyterHub upgrade from v1 to v4 is different from this new Jupyter image for mlflow-proxy.

My original question that I expected a child mlflow image from the base image, not a direct mlflow flavor from the base image.

I was also curious if both changes have to happen in lock step.

Nazim-crim: The base image use `jupyterlab==2.2.9` which does not respect the requirements for the mlflow-proxy plugin. But upgrading the base image and all the dependencies would cause compatibility issues for the two other child images.(Which would take more time to investigate and resolve)

The safest and most efficient choice is to have a spin off of the base image. If there is a need to update nlp and eo images, we would just need to update the env.local.example. 

JupyterHub upgrade from v1 to v4 is needed to run this new image. Since there is a compatibilty issue with `jupyter-server 2.7.3` and Jupyterhub v1. (when using `Spawner.disable_user_config = True` )

CHANGES.md
Nazim-crim: The merge was done before the version bump on master. I'll move the change to the unreleased section.

tlvu (approved):
LGTM

PR #382, opened 2023-09-19
Adding redirect to /stac/stac when using twitcher

Nazim-crim:
## Overview

Fix redirect when behind twitcher to automatically redirect to /stac/stac 

## Related Issue / Discussion

- Resolves [DAC-580](https://crim-ca.atlassian.net/browse/DAC-580)

*Notes*

- For permissions on magpie to be correctly applied on the stac ressource a child /stac is needed.
- When using the current proxy .../stac redirect correctly to .../twitcher/ows/proxy/stac/stac but using twitcher/ows/proxy/stac/ does not.
- This change make sure that api calls correctly redirect to twitcher/ows/proxy/stac/stac without having a recursive call to itself if /stac/stac is already in the url.


mishaschwartz:
@Nazim-crim do you mind summarizing the issue in [DAC-580](https://crim-ca.atlassian.net/browse/DAC-580) here. I don't have access to that URL and I'm not sure what issue this is resolving. Thanks

fmigneault (commented):
@Nazim-crim 
I believe you also found out that the Magpie service/resources needed the nested `stac/stac` as well to secure access properly?

To make sure one does not create resources directly under the `stac` service, I think it would be wise to apply the permissions in https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/optional-components/stac-public-access/config/magpie/config.yml.template#L11-L19 directly on a new  `stac` resource under the `stac` service.

Nazim-crim:
@mishaschwartz it's in the pr description

mishaschwartz:
> @mishaschwartz it's in the pr description

Yes, I don't have permission to view CRIM's atlassian pages

mishaschwartz:
Wait, I'm also confused why there's no `stac/config/magpie` directory. At the very least, the Magpie providers definition should be there or else the current nginx definition won't work (unless we also enable `stac-public-access`).

Sorry, I should have noticed that earlier.

mishaschwartz:
And also the current dependency structure will mean that stac will always be enabled if `all-public-access` is enabled. 

- all-public-access > stac-public-access > stac

That seems confusing as well.

There's some really weird things going on with the way this is set up. Am I looking at this wrong?

Nazim-crim:
I think we should remove the stac-public-access and in the default.env of all-public-access and just have the definition of providers and permissions in /stac/config/magpie . Would that make more sense or is there another reason to have it separate.  @mishaschwartz @fmigneault

Nazim-crim:
@mishaschwartz We can incorporate theses changes in this pr. Do you want me to pull your branch or are you gonna push it from your side here. Also just to make sure that there's a `resource: /stac` under both permission.

tlvu (approved):
LGTM.

Approuve the spirit of the PR.  Will let Francis have the final say since he knows more about Magpie config.

fmigneault:
Updated to take into account changes in https://github.com/bird-house/birdhouse-deploy/pull/383.
Will merge both PRs together before bump.

PR #383, opened 2023-09-20
move STAC service Magpie definition under its component

fmigneault:
## Overview

Move STAC service definitions under `components/stac` configuration.

## Changes

**Non-breaking changes**
- Move initial ``stac`` service Magpie definition under its component configuration.
  - Before this change, ``optional-components/stac-public-access`` was mandatory since the ``stac`` service under
    Magpie was not created otherwise, leading to "*service not found*" error when requesting the ``/stac`` endpoint.
  - Ensure that the first ``stac`` resource under ``stac`` service in Magpie is created by default.
    Without this resource being defined initially, it is very easy to forget creating it, which would not take into
    account the required ``/stac/stac`` request path to properly resolve the real endpoints where STAC API is served.
- Remove `optional-components/stac-public-access` dependency under `optional-components/all-public-access`
  to avoid indirectly enforcing `components/stac` when `optional-components/all-public-access` is enabled.
  Users that desire using `optional-components/stac-public-access` will have to add it explicitly to the list
  of `EXTRA_CONF_DIRS`.

- Rename `optional-components/stac-public-access/config/magpie/config.yml.template` to
  `optional-components/stac-public-access/config/magpie/permissions.cfg` in order to align
  with permissions-specific contents as accomplished with other components.

**Breaking changes**
- n/a

## Related Issue / Discussion

- Relates to #382


Nazim-crim (approved):
Okay so we can merge this seperately from the fix with the redirect

mishaschwartz (changes requested):
I think we should also completely remove stac-public-access and put those permission changes under all-public-access. (see my concern here: https://github.com/bird-house/birdhouse-deploy/pull/382#issuecomment-1727880587)

See proposed changes from here (https://github.com/bird-house/birdhouse-deploy/compare/stac-permission-fixes)

fmigneault:
@mishaschwartz 
I believe `stac-public-access` should not be removed. 
Instead, its dependency in `all-public-access` should be removed. 
This is to avoid `stac` being enforced by `all-public-access`, while still providing a quick method to make it public as needed.

mishaschwartz:
>  Instead, its dependency in all-public-access should be removed.

Ok I'm fine with that compromise. Can you remove it from the `default.conf` file then?

tlvu (commented):
LGTM as well, pending the comment from Misha.

birdhouse/components/stac/config/magpie/config.yml.template
mishaschwartz, line 1: Can we rename this to `config.yml` since there isn't anything in the file that needs to be "templated". 
Then we can also remove `config/magpie/config.yml` from the .gitignore file

tlvu (approved):

mishaschwartz (changes requested):

mishaschwartz (approved):

Issue #246, opened 2022-05-24
Productize Notos: autodeploy keeps recreating the containers even if unchanged


tlvu:
Guessing it is because the docker and/or docker-compose version used by autodeploy container and the docker and/or docker-compose version installed are too different.

fmigneault:
> Another solution is for `pavics-compose.sh` to also run inside a docker image with `docker-compose` (the autodeploy docker image) so then any `docker-compose` update will be 100% transparent because the `docker-compose` on the host is not used anymore.

Not sure if this is an actual fix. If running in a docker, the docker engine it uses must be compatible to the `docker-compose` running inside it. It only moves the problem elsewhere, and can greatly complicate the definition of networks definition since they would now have to be bridged between the host and that internal-docker network

tlvu:
Relate to https://github.com/bird-house/birdhouse-deploy/pull/384

PR #384, opened 2023-09-21
Docker compose version missing in ``stac/config/magpie/`` compose file

mishaschwartz:
## Overview

The `version:` key was not set in the `stac/config/magpie/docker-compose-extra.yml` file which caused `docker-compose` to report a version mismatch and fail to start.

## Changes

**Non-breaking changes**
- Bug fix

**Breaking changes**
None

## Additional Information

birdhouse_skip_ci: false

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,11 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Fixes
+
+- Docker compose version missing in ``stac/config/magpie/`` compose file
+  - The ``version:`` key was not set in the ``stac/config/magpie/docker-compose-extra.yml`` file which caused
+    ``docker-compose`` to report a version mismatch and fail to start.
 
 [1.31.3](https://github.com/bird-house/birdhouse-deploy/tree/1.31.3) (2023-09-21)
 ------------------------------------------------------------------------------------------------------------------
--- a/birdhouse/components/stac/config/magpie/docker-compose-extra.yml
+++ b/birdhouse/components/stac/config/magpie/docker-compose-extra.yml
@@ -1,3 +1,5 @@
+version: "3.4"
+
 services:
   magpie:
     volumes:


tlvu (approved):
This one is easy :D

fmigneault (approved):
This is something I intentionally left out, because as soon as any version mismatches anywhere, combining them fails. So it is somewhat redundant to have it duplicated in each file. We could have it defined only once in https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/docker-compose.yml. 

Not against adding it explicitly, though. Just wanted to share why it was missing.

tlvu:
> Not against adding it explicitly, though. Just wanted to share why it was missing.

I agree the `version` key is optional.  We should remove them all.  However, when doing this, all external repos should be done at the same time.  And I think we should do it eventually but communications has to very clear with all partners because this is a breaking change.

For now, if we do not include the `version` key, the entire stack will not start and we have to ensure all changes are not breaking change, unless it is scheduled and communicated very clearly in advance.

mishaschwartz:
@fmigneault 

> This is something I intentionally left out

Which version of docker compose are you using though? Because older ones (like the version suggested in the vagrant installation scripts) will complain about the mismatch.

I'm very very happy to update the docker compose version though instead

tlvu:
> Which version of docker compose are you using though? Because older ones (like the version suggested in the vagrant installation scripts) will complain about the mismatch.

The physical host has the same `docker-compose` version as what vagrant installs, which is also the same version as the autodeploy container.

https://github.com/bird-house/birdhouse-deploy/blob/2045cc5083dc96a64cae41ef2e72caf8f54eeeb7/birdhouse/vagrant-utils/install-docker.sh#L102-L105

The reason to keep all the versions in-sync is documented in the comment above.  When there is a mismatch, the containers will re-create themselves all the time, even for a simple `./pavics-compose.sh up -d` that should do nothing because no change to any configs.

> 
> I'm very very happy to update the docker compose version though instead

Updating the `docker-compose` version means all the physical hosts of all existing deployments have to do it at the same time.

I was thinking of a solution less disruptive documented here https://github.com/bird-house/birdhouse-deploy/issues/246

All comments about `docker-compose` update should probably stay in that issue https://github.com/bird-house/birdhouse-deploy/issues/246 to keep everything organized.

fmigneault:
@mishaschwartz 
> Which version of docker compose are you using though? Because older ones (like the version suggested in the vagrant installation scripts) will complain about the mismatch.

I'm currently running `v2.17.2`, but was running `1.2x` before. Both complained the same that it could not merge YAMLs with 
different `version` values.

PR #386, opened 2023-09-28
STAC hooks

Nazim-crim:
## Overview
Add hooks for STAC requests to dynamically create Magpie resources.

## Changes

**Non-breaking changes**

- Add a Magpie Webhook to create the Magpie resources corresponding to the STAC-API path elements when a `STAC-API` 
  `POST /collections/{collection_id}` or `POST /collections/{collection_id}/items/{item_id}` request is accomplished.
  - When creating the STAC `Item`, the `source` entry in `links` corresponding to a `THREDDS` file on the same instance 
    is used to define the Magpie `resource_display_name` corresponding to a file to be mapped later on
    (eg: a NetCDF `birdhouse/test-data/tc_Anon[...].nc`).
  - Checking same instance source path is necessary because `STAC` could refer to external assets, and we do not want to 
    inject Magpie resource that are not part of the active instance where the hook is running.

## Related Issue / Discussion
relates to [DAC-578](https://crim-ca.atlassian.net/browse/DAC-578)


Nazim-crim:
run test

Nazim-crim:
run tests

birdhouse/components/stac/config/magpie/stac_hooks.py
```diff
@@ -0,0 +1,144 @@
...
+    Extract THREDD path from a STAC links
+    """
+    display_name = None
+    for link in links:
+        if link['rel'] == 'source':
+            # Example of title `thredds:birdhouse/CMIP6` -> `birdhouse/CMIP6`
+            display_name = link['title'].split(':')[1]
+            break
+
+    return display_name
```
fmigneault, line 144: There should be a check here that raises directly if `display_name` was not resolved, to avoid all following computations that will end up in an error regardless when reaching `ru.create_resource`.

birdhouse/components/stac/config/twitcher/docker-compose-extra.yml
```diff
@@ -0,0 +1,10 @@
+version: "3.4"
+
+services:
+  # extend twitcher with MagpieAdapter hooks employed for stac proxied requests
```
fmigneault, line 4: Use "STAC" in comments and function docstrings.

birdhouse/components/stac/config/twitcher/docker-compose-extra.yml
```diff
@@ -0,0 +1,10 @@
...
+  twitcher:
+    volumes:
+      # NOTE: MagpieAdapter hooks are defined within Magpie config, but it is actually Twitcher proxy that runs them
+      # target mount location depends on main docker-compose 'MAGPIE_PROVIDERS_CONFIG_PATH' environment variable
+      - ./components/stac/config/magpie/config.yml:/opt/birdhouse/src/magpie/config/stac-config.yml:ro
+      - ./components/stac/config/magpie/stac_hooks.py:/opt/birdhouse/src/magpie/hooks/stac_hooks.py:ro
```
fmigneault, line 10: Missing a newline at the end.

birdhouse/components/stac/config/twitcher/docker-compose-extra.yml
```diff
@@ -0,0 +1,10 @@
+version: "3.4"
+
+services:
+  # extend twitcher with MagpieAdapter hooks employed for stac proxied requests
+  twitcher:
+    volumes:
+      # NOTE: MagpieAdapter hooks are defined within Magpie config, but it is actually Twitcher proxy that runs them
+      # target mount location depends on main docker-compose 'MAGPIE_PROVIDERS_CONFIG_PATH' environment variable
+      - ./components/stac/config/magpie/config.yml:/opt/birdhouse/src/magpie/config/stac-config.yml:ro
```
fmigneault, line 9: I realise that this comment came from Weaver example, but it is somewhat confusing looking back at it. 
`MAGPIE_PROVIDERS_CONFIG_PATH` is used in other places with a different value (ie: defined in https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/config/magpie/default.env#L28 and everywhere else that it is used). 

Can you adjust the comment (here and in Weaver's corresponding config) to have a reference to `birdhouse/config/twitcher/docker-compose-extra.yml` where this "other" `MAGPIE_PROVIDERS_CONFIG_PATH` mentioned here actually refers to?

birdhouse/components/stac/config/magpie/stac_hooks.py
```diff
@@ -0,0 +1,144 @@
...
+            collection_res_id = collection_res.json["resource"]["resource_id"]
+
+        # create resource /stac/stac/collections/<collection_id>
+        # In try catch since Magpie resource are persistent but stac-db could be wiped
+        try:
+            collection = ru.create_resource(collection_id, display_name, Route.resource_type_name, collection_res_id, db_session=session)
+        except Exception as exc:
+            LOGGER.warning("Failed creation of the collection %s %s", collection_id, str(exc), exc_info=exc)
+
+        session.commit()
```
fmigneault, line 70: Session commit should probably be avoided here if the creation failed. 
Calling some abort or rollback method?
And then return the response directly.

fmigneault (changes requested):

tlvu (commented):
I approuve the spirit of this PR.  Francis knows more about Magpie so I defer the actual approuval to him.

birdhouse/components/stac/config/twitcher/docker-compose-extra.yml
tlvu: Yes I find this confusing too.  I think what Francis meant is to reference the location where the variable `MAGPIE_PROVIDERS_CONFIG_PATH` is actually defined, here https://github.com/bird-house/birdhouse-deploy/blob/547cc085c52924a21d853eeb1149790c2cb7577d/birdhouse/config/twitcher/docker-compose-extra.yml#L21

So we know which definition is actually "in effect".

Just curious, why do we have 2 flavors of the same variable?  Because one is the Magpie container and the other is Twitcher container?

CHANGES.md
```diff
@@ -15,7 +15,15 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Changes
+
+- Add a Magpie Webhook to create the Magpie resources corresponding to the STAC-API path elements when a `STAC-API` 
+  `POST /collections/{collection_id}` or `POST /collections/{collection_id}/items/{item_id}` request is accomplished.
```
mishaschwartz, line 21: Do we also need a hook when a collection or item is deleted? (this is the sort of thing I mean: https://api.stacspec.org/v1.0.0-rc.1/ogcapi-features/extensions/transaction/#tag/Transaction/operation/deleteFeature)

birdhouse/components/stac/config/magpie/stac_hooks.py
```diff
@@ -0,0 +1,144 @@
...
+            LOGGER.warning("Failed creation of the item %s %s", display_name, str(exc), exc_info=exc)
+        session.commit()   
+   
+    except Exception as exc:
+        LOGGER.error("Unexpected error while creating the item %s %s", display_name, str(exc), exc_info=exc)
+    
+    return response
+
+def extract_display_name(links):
+    # type: (JSON array) -> string
```
mishaschwartz, line 133: Shouldn't this be: `# type: List[Dict] -> Optional[str]`

birdhouse/components/stac/config/magpie/config.yml.template
```diff
@@ -7,6 +7,15 @@ providers:
     c4i: false
     type: api
     sync_type: api
+    hooks:
+      - type: response
+        path: "/stac/collections/?"
+        method: POST
+        target: /opt/birdhouse/src/magpie/hooks/stac_hooks.py:create_collection_resource
+      - type: response
+        path: "/stac/collections/[a-z-A-Z-0-9]+/items/?"
```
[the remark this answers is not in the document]
Nazim-crim: The id on the url is not exactly an UUID see on : `https://hirondelle.crim.ca/stac/collections/0798aa197d54eb4332767a5a4077fb0f` , also on the branch `arch-changes` of the populator the collection title is used as an id ex: "CMIP6". I think we should either match any random string with numbers ``[a-zA-Z0-9]+` or make a change in the populator to force the use of an UUID when creating the collection. What do you think ? @fmigneault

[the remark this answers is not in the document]
fmigneault: Oh! My bad. I assumed it was a typical UUID based on previous examples I saw.
Go ahead with `[a-zA-Z0-9-]+` (careful, you example was missing the last `-`).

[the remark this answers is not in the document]
fmigneault: Actually, it is possible to POST something like `my_id`? If so, maybe `[\\w_-]+` would work better?

birdhouse/components/stac/config/twitcher/docker-compose-extra.yml
fmigneault: Indeed. This is what I meant.

The 2 flavors is because one is inside Magpie, and the other in Twitcher, but both are loaded by Magpie code (`MagpieAdapter` specifically in the case of Twticher). 
Inside Twitcher, the directory structure is not that important because other files are not used.
The only thing to respect is the multi-config loading strategy, as described in the comment just above that variable definition.

birdhouse/components/stac/config/magpie/stac_hooks.py
```diff
@@ -0,0 +1,144 @@
...
+from magpie.api.requests import get_service_matchdict_checked
+from magpie.models import Route, Service
+from magpie.utils import get_logger
+from magpie.db import get_session_from_other
+
+if TYPE_CHECKING:
+    from pyramid.request import Request
+    from pyramid.response import Response
+
+LOGGER = get_logger("magpie.stac",level=logging.DEBUG)
```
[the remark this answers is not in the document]
Nazim-crim: So it should just be `get_logger("magpie.stac")` and let it set by itself the log level ?

birdhouse/components/stac/config/magpie/config.yml.template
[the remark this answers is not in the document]
Nazim-crim: It could be possible. We can use `[\\w-]+` since `_` is found with `\w`

CHANGES.md
Nazim-crim: I dont think it would be something to do atm. The `STAC` items are generated from files found on `thredds`. There shouln't be a request to delete an item sent to `STAC` if the file is still on `thredds`. AFAIK there is no mechanism in place to notify `STAC` when an item or directory is deleted on `thredds` / if it's something we should implement in the futur. What do you think? @fmigneault

birdhouse/components/stac/config/magpie/stac_hooks.py
Nazim-crim: yes you are right thanks

CHANGES.md
fmigneault: I could be done in spirit of cleanup in Magpie, but if the resource is actually deleted in STAC, only a 404 would be returned instead of 403. So not high priority IMO.

birdhouse/components/stac/config/magpie/stac_hooks.py
fmigneault: `# type: (List[Dict[str, str]]) -> Optional[str]` 

But I would prefer `-> str` with raise error as per https://github.com/bird-house/birdhouse-deploy/pull/386#discussion_r1341407674

birdhouse/components/stac/config/magpie/stac_hooks.py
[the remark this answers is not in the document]
fmigneault: yes

mishaschwartz (approved):
Ok! LGTM (pending suggested changes from @fmigneault)

CHANGES.md
```diff
@@ -17,13 +17,13 @@
...
 ## Changes
 
-- Add a Magpie Webhook to create the Magpie resources corresponding to the STAC-API path elements when a `STAC-API` 
+- Add a Magpie Webhook to create the Magpie resources corresponding to the STAC-API path elements when a `STAC-API`
   `POST /collections/{collection_id}` or `POST /collections/{collection_id}/items/{item_id}` request is accomplished.
-  - When creating the STAC `Item`, the Asset URL corresponding to a `THREDDS` file on the same instance is used to 
-    define the Magpie `resource_display_name` corresponding to a file to be mapped later on
+  - When creating the STAC `Item`, the `source` entry in `links` corresponding to a `THREDDS` file on the same instance
+    is used to define the Magpie `resource_display_name` corresponding to a file to be mapped later on
     (eg: a NetCDF birdhouse/test-data/tc_Anon[...].nc).
```
fmigneault, line 24: missing ``` ` ```  quotes around the path

birdhouse/components/stac/config/magpie/stac_hooks.py
```diff
@@ -81,64 +64,84 @@ def create_item_resource(response):
...
     display_name = None
     for link in links:
-        if link['rel'] == 'source':
+        if link["rel"] == "source":
             # Example of title `thredds:birdhouse/CMIP6` -> `birdhouse/CMIP6`
-            display_name = link['title'].split(':')[1]
+            display_name = link["title"].split(":")[1]
             break
+    if display_name is None:
+        raise Exception("The display name was not extracted properly")
```
fmigneault, line 106: Better to check onky `if display_name`, this way we reject empty value as well. 
Can raise `ValueError` which is more appropriate than the generic `Exception`.

birdhouse/components/stac/config/magpie/stac_hooks.py
```diff
@@ -81,64 +64,84 @@ def create_item_resource(response):
...
+    for childs in nodes.values():
+        # Find the resource in the current children.
+        if childs["node"].resource_name == resource:
+            children = childs["children"]
+            # Since the resource exists, we can use its id to create the next resource.
+            parent_id = childs["node"].resource_id
+            next_depth = current_depth + 1
+            return create_resource_tree(resource_tree, children, next_depth, parent_id, session, display_name)
+            create_resource = False
+            break
```
fmigneault, line 132: The two lines after `return` will never execute. 
The `return` on the same line as `created_resource_tree`, although valid, is misleading with nothing (`None`) returned.

birdhouse/components/stac/config/magpie/stac_hooks.py
```diff
@@ -81,64 +64,84 @@ def create_item_resource(response):
...
+        # Creating the last resource in the tree, we need to use the display_name.
+        if current_depth == len(tree) - 1:
+            ru.create_resource(resource, display_name, Route.resource_type_name, parent_id, db_session=session)
+            return
+        else:
+            # Creating the resource somewhere in the middle of the tree before using its id.
+            node = ru.create_resource(resource, None, Route.resource_type_name, parent_id, db_session=session)
+            parent_id = node.json["resource"]["resource_id"]
+            next_depth = current_depth + 1
+            create_resource_tree(resource_tree, nodes, next_depth, parent_id, session, display_name)
```
fmigneault, line 145: Since following resources must all be created, `{}` instead of `nodes` would more quickly skip the for loop.

birdhouse/components/stac/config/magpie/stac_hooks.py
Nazim-crim: Did you mean ` if not display_name` raise or `if display_name` return and then raise the error ?

birdhouse/components/stac/config/magpie/stac_hooks.py
Nazim-crim: Oh wow, it took me some time to see it even after your comment. I get it now, nice catch, thanks you !

birdhouse/components/stac/config/magpie/stac_hooks.py
fmigneault: Yes. My bad. `if not display_name`.

dchandan:
A STAC catalog can contain 100s of thousands of items. UofT node has already hit that number of catalog items (in THREDDS). What are the performance implications of creating these many resources in Magpie?

fmigneault:
> A STAC catalog can contain 100s of thousands of items. UofT node has already hit that number of catalog items (in THREDDS). What are the performance implications of creating these many resources in Magpie?

The creation itself and their retrieval when accessing the resources/permissions via Twitcher should be good since they target a specific item in the DB.

However, this is a valid point when `ru.get_resource_children` is called. That could get gradually slower because children resources are looked up by name one by one. A better implementation would perform a search directly, yielding either 0/1 match, since the target `resource_name` and its expected `parent_id` are known.

Nazim-crim:
> A better implementation would perform a search directly, yielding either 0/1 match, since the target `resource_name` and its expected `parent_id` are known.

This other implementation would ask at least 5 calls to db for each post of item `stac/collections/{collection_id}/items/{item_id}` but would indeed reduce the processing time for searching the right children. I haven't seen in `Magpie` a function to perform a search directly with a `resource_name` and `parent_id`. Do you know where it is and If there is such function do you want me to make the changes in this PR ?
@fmigneault

fmigneault:
@Nazim-crim 
There is no pre-made function in Magpie for querying a specific resource other than `ResourceService.by_resource_id`. 
Even then, this would not help you find `resource.resource_name == '<search>' WHERE resource.parent_id == <previous-resource-found>`. 

If you wanted to reduce the number of db queries, you could make a very specific filter for this case, but I think it would be resonable to start with only a generic filter as above and run it a few times in a row for each deeper resource to retrieve. 

If you want to try out some form of recursive approach, you can take some inspiration from the installed package `ziggurat_foundations.models.services.resource_tree_postgres:ResourceTreeServicePostgreSQL.from_resource_deeper`, but you don't need to waste a lot of time on this if you can't figure it out easily.

Nazim-crim:
run tests

birdhouse/components/stac/config/magpie/stac_hooks.py
```diff
@@ -0,0 +1,140 @@
...
+    body = request.json
+    item_id = body["id"]
+    try:
+        display_name = extract_display_name(body["links"])
+    except Exception as exc:
+        LOGGER.error("Error when extracting display_name from links %s %s", body["links"], str(exc), exc_info=exc)
+        return response
+
+    # Get the <collection_id> from url -> /collections/{collection_id}/items
+    collection_id = re.search(r'(?<=collections\/).*?(?=\/items)', request.url).group()
```
fmigneault, line 74: The regex should be improved to use accepted characters (`[0-9a-zA-Z_.-]+`) instead of `.*?`.
Currently, some invalid regexes are mached such as `/collections//items` or `/collections/   /items`, which will cause an error when reaching the DB queries.

birdhouse/components/stac/config/magpie/stac_hooks.py
```diff
@@ -0,0 +1,140 @@
...
+def extract_display_name(links):
+    # type: (List[Dict[str, str]]) -> str
+    """
+    Extract THREDD path from a STAC links
+    """
+    display_name = None
+    for link in links:
+        if link["rel"] == "source":
+            # Example of title `thredds:birdhouse/CMIP6` -> `birdhouse/CMIP6`
+            display_name = link["title"].split(":")[1]
```
fmigneault, line 101: Shouldn't the service part be kept?
Otherwise, how will Cowbird know under which service the associated resource is located under?

Nazim-crim: In the cowbird config? If we leave "thredds:" in the display_name, won't we need to split it in cowbird before being able to map the directory path to a stac item ?

fmigneault: You should be able to define a regex in the cowbird path segments. 
Provided that the feature that allows extracting `resource_display_name` instead of `resource_name` is added, it should be possible to have `thredds:{var}`. Maybe some adjustments must be added in https://github.com/Ouranosinc/cowbird/blob/aad19bac737bc0ff60bb3c2cf977eeb9383c9f85/cowbird/permissions_synchronizer.py for `_generate_regex_from_segments` and `_find_matching_res`. 
The `thredds:` prefix would be important in order to support multiple distinct THREDDS servers.

Nazim-crim:
run tests

Nazim-crim:
run tests

Nazim-crim:
run tests

birdhouse/components/stac/config/magpie/stac_hooks.py
```diff
@@ -0,0 +1,140 @@
...
+    body = request.json
+    item_id = body["id"]
+    try:
+        display_name = extract_display_name(body["links"])
+    except Exception as exc:
+        LOGGER.error("Error when extracting display_name from links %s %s", body["links"], str(exc), exc_info=exc)
+        return response
+
+    # Get the <collection_id> from url -> /collections/{collection_id}/items
+    collection_id = re.search(r'(?<=collections\/)[0-9a-zA-Z_.-]+?(?=\/items)', request.url).group()
```
mishaschwartz, line 74: no need to escape a `/` character

mishaschwartz, line 74: If there is no match found then `re.search` will return `None`, and you'll get an attribute error since `None.group` is not a method. Is this what you want? Is there a case where this URL could be incorrectly formatted?

Maybe it would be better to parse this in a safer way using [urllib.parse](https://docs.python.org/3/library/urllib.parse.html)

birdhouse/components/stac/config/magpie/stac_hooks.py
```diff
@@ -0,0 +1,140 @@
...
+        if current_depth == len(tree) - 1:
+            ru.create_resource(resource_name, display_name, Route.resource_type_name, parent_id, db_session=session)
+        else:
+            # Creating the resource somewhere in the middle of the tree before using its id.
+            node = ru.create_resource(resource_name, None, Route.resource_type_name, parent_id, db_session=session)
+            parent_id = node.json["resource"]["resource_id"]
+            next_depth = current_depth + 1
+            create_resource_tree(resource_tree, next_depth, parent_id, session, display_name)
+
+    return
```
mishaschwartz, line 140: unnecessary explicit return

mishaschwartz (changes requested):

birdhouse/components/stac/config/magpie/stac_hooks.py
Nazim-crim: Since the hook is triggered by a post made to the `STAC-api` using the endpoint `.../collections/{collection_id}/items`. There should always be a match.

fmigneault (approved):

tlvu (approved):
I approuve the spirit of this PR. Francis knows more about Magpie so I defer the actual approuval to him.

birdhouse/components/stac/config/magpie/stac_hooks.py
Nazim-crim: @mishaschwartz Do you think the rest is good to be merged ?

mishaschwartz: Thanks! Sorry I was off for a few weeks. I'll approve it all now

mishaschwartz (approved):

PR #388, opened 2023-10-04
allow users to request public Weaver WPS outputs

fmigneault:
## Overview

Allow users to submit a Weaver job requesting to store outputs to the public location instead of their user-workspace.

## Changes

**Non-breaking changes**
- Users-submitted jobs are not enforced to their user-workspace location if specifically requesting `public` location.

**Breaking changes**
- n/a

## Related Issue / Discussion

- relates to https://github.com/crim-ca/weaver/pull/569
- resolves [DAC-190](https://crim-ca.atlassian.net/browse/DAC-190)

birdhouse_skip_ci: false


birdhouse/components/weaver/config/magpie/weaver_hooks.py
```diff
@@ -30,7 +30,21 @@
...
+#  The output hierarchy would be:
+#
+#       ${WEAVER_WPS_OUTPUTS_DIR}/
+#           users/
+#               <user-id>/
+#                   <job-id>/
+#           ${PUBLIC_USER_CONTEXT}/
+#               <job-id>/
+#
+# NOTE: This value should align with Cowbird's setting 'PUBLIC_WORKSPACE_WPS_OUTPUTS_SUBDIR'.
```
mishaschwartz, line 46: Instead of adding this note. Why don't we actually use the `PUBLIC_WORKSPACE_WPS_OUTPUTS_SUBDIR` variable to set this constant?

If we make this a `.template` file then we can just insert the correct value to ensure that it aligns.

birdhouse/components/weaver/config/magpie/weaver_hooks.py
```diff
@@ -30,7 +30,21 @@
 
     from magpie.adapter import HookContext
 
-LOGGER = get_logger("birdhouse-weaver-hooks")
+LOGGER = get_logger("magpie.weaver-hooks")
```
mishaschwartz, line 33: Is this related to the rest of the changes?

birdhouse/components/weaver/config/magpie/weaver_hooks.py
```diff
@@ -50,15 +64,18 @@ def add_x_wps_output_context(request):
     # if explicitly provided, ensure it is permitted (admin allow any, otherwise self-user reference only)
     if header is not None:
```
mishaschwartz, line 65: I feel like the logic of these `if` statements is unnecessarily complex. Can we simplify it to:

```py
if request.user is None:
    header = PUBLIC_USER_CONTEXT
elif header is None or (header != PUBLIC_USER_CONTEXT and not is_admin(request)):
    header = "users/" + str(request.user.id)
```

With the appropriate comments added in of course

mishaschwartz (changes requested):
Overall this looks good. 

I'm a bit confused as to how this relates to https://github.com/crim-ca/weaver/pull/569 though.

Does this PR depend on the weaver PR to be merged in first? If so, does this PR need to include a change to the weaver version once that one is pulled in?

fmigneault:
> I'm a bit confused as to how this relates to [crim-ca/weaver#569](https://github.com/crim-ca/weaver/pull/569) though.
> 
> Does this PR depend on the weaver PR to be merged in first? If so, does this PR need to include a change to the weaver version once that one is pulled in?

It does not depend on it, but that PR adds a usefull CLI & Python Client option that aligns with the same header handled here. 
The header feature was already available in Weaver, no need to update it.

birdhouse/components/weaver/config/magpie/weaver_hooks.py
fmigneault: Good idea. 

I think it would be wise also to mask away the nesting logic of `PUBLIC_WORKSPACE_WPS_OUTPUTS_SUBPATH` here.
Technically, the original path was `public` only, but Cowbird adjusted recently to be `public/wps_outputs` in prevision that other "public" stuff might be used. 
I think we should "help" the user such that if `public` or `${PUBLIC_WORKSPACE_WPS_OUTPUTS_SUBPATH}` is passed, both will redirect to `${PUBLIC_WORKSPACE_WPS_OUTPUTS_SUBPATH}`. This way, they can just request "make my output public" and not worry about instance-specific configuration.

Sounds good?

birdhouse/components/weaver/config/magpie/weaver_hooks.py
fmigneault: I agree for moving `if request.user is None:` to avoid repeting it, but the other `elif` is harder to read, and less diff when adding new combinations. Better to leave it sepated.

birdhouse/components/weaver/config/magpie/weaver_hooks.py
mishaschwartz: Yes that's a great idea!

fmigneault:
For info, I updated Weaver while at it, but this is still not a requirement for this feature. It has just been a while since it got updated, and a lot of bugfixes have been applied since.

birdhouse/components/weaver/config/magpie/weaver_hooks.py.template
```diff
@@ -47,20 +64,16 @@ def add_x_wps_output_context(request):
...
-            header = "users/" + str(request.user.id)
+    # if no user logged in, ignore specified value to avoid writing in someone else's user-workspace
+    # if explicitly requested to make their job output public right away, whether logged in or not,
+    # use the public location instead of specific user-workspace
+    if request.user is None or header in PUBLIC_USER_CONTEXTS:
+        header = PUBLIC_WORKSPACE_WPS_OUTPUTS_SUBDIR
+    # if explicitly provided, ensure it is permitted, and if not provided, default to current user
+    # admin-level user is allowed to write anywhere (up to them to use a sensible location)
+    # otherwise, enforce the current user-workspace even if provided explicitly
+    if not header or not is_admin(request):
```
mishaschwartz, line 75: If `request.user is None` then `is_admin(request)` is always going to return False right?

So, in this case wouldn't the next line raise an error since `request.user` is `None` and so `request.user.id` raises an AttributeError?

If we make this line an `elif` instead of a new `if` it should work though since the case where `request.user is None` is handled on line 70.

mishaschwartz (changes requested):

birdhouse/components/weaver/config/magpie/weaver_hooks.py.template
fmigneault: Good catch. I guess that's why I did the `if request.user is None:` on its own in the original implementation.

fmigneault: I'll just revert to something separate and stop wasting time on refactoring these `if` combinations.

mishaschwartz (approved):

birdhouse/components/weaver/config/magpie/weaver_hooks.py.template
mishaschwartz: I like the code better now. It's much easier to read :+1:

PR #389, opened 2023-10-11
add `optional-components/jupyterhub-stop-idle`

fmigneault:
## Overview

Jupyter instances that have been started by users are simply left in a running status indefinitely if not manually logged of.
This causes dockers `jupyter-<user>` to remain active, using resources unnecessarily. 

This feature adds a new component that helps quickly detecting idle jupyter servers, and stopping them after a given inactivity timeout. The logic behind it depends on what JupyterHub reports on its activity API endpoint for the given user.

To have the utility available, the following requirements must be applied: https://github.com/Ouranosinc/jupyterhub/pull/21
To test quickly, simply build the Docker on this branch and override `JUPYTERHUB_DOCKER` and `JUPYTER_VERSION` accordingly, and enable `optional-components/jupyterhub-stop-idle` in `EXTRA_CONF_DIRS`. Setting `JUPYTERHUB_STOP_IDLE_TIMEOUT` can be set to adjust the timeout interval. 

## Changes

**Non-breaking changes**
- Adds `optional-components/jupyterhub-stop-idle` allowing culling of idle jupyter servers of users.

**Breaking changes**
- n/a

## Related Issue / Discussion

- Depends on https://github.com/Ouranosinc/jupyterhub/pull/21

## To Do

- [ ] update changelog
- [ ] update readme about new variables and how to enable the component

### Still investigating

The timeout frequency is properly applied and docker logs show that activity status checks are performed, but servers are still considerd in a "running" status for some reason, even if nothing is running on them, nor any tab refreshing them being open. 
I will look more into that, but I want to open this PR right away to gather early feedback.

### DIFF
--- a/birdhouse/config/jupyterhub/default.env
+++ b/birdhouse/config/jupyterhub/default.env
@@ -43,6 +43,15 @@ export JUPYTER_LOGIN_BANNER_BOTTOM_SECTION=""
 # server for the change to take effect.
 export JUPYTERHUB_README=""
 
+# Additional configuration to be provided by services or components managed by the stack.
+# Any component extending this configuration using this variable *MUST* include it back to allow further
+# components to extend it as well. Also, they *MUST* extend the variable with each new definitions applied on
+# a new line (without indents) to ensure that Python configurations will be executable without syntax error.
+# This variable differs from 'JUPYTERHUB_CONFIG_OVERRIDE' that does not require users to include it back to make
+# its definition extensible by other components. However, 'JUPYTERHUB_CONFIG_OVERRIDE' will still be applied
+# after 'JUPYTERHUB_CONFIG_EXTENDED' to ensure it can override anything as needed after all components were resolved.
+export JUPYTERHUB_CONFIG_EXTENDED=""
+
 # Allow for adding new config or override existing config in
 # config/jupyterhub/jupyterhub_config.py.template.
 export JUPYTERHUB_CONFIG_OVERRIDE=""
@@ -65,6 +74,7 @@ OPTIONAL_VARS="
   \$JUPYTER_LOGIN_BANNER_TOP_SECTION
   \$JUPYTER_LOGIN_BANNER_BOTTOM_SECTION
   \$JUPYTER_LOGIN_TERMS_URL
+  \$JUPYTERHUB_CONFIG_EXTENDED
   \$JUPYTERHUB_CONFIG_OVERRIDE
   \$JUPYTERHUB_DOCKER
   \$JUPYTERHUB_VERSION
--- a/birdhouse/config/jupyterhub/jupyterhub_config.py.template
+++ b/birdhouse/config/jupyterhub/jupyterhub_config.py.template
@@ -187,4 +187,5 @@ blocked_users = {'authtest', '${CATALOG_USERNAME}', 'anonymous'}
 c.Authenticator.blacklist = blocked_users  # v0.9+
 c.Authenticator.blocked_users = blocked_users  # v1.2+
 
+${JUPYTERHUB_CONFIG_EXTENDED}    # noqa
 ${JUPYTERHUB_CONFIG_OVERRIDE}    # noqa
--- a/birdhouse/optional-components/jupyterhub-stop-idle/default.env
+++ b/birdhouse/optional-components/jupyterhub-stop-idle/default.env
@@ -1,0 +1,74 @@
+#!/bin/sh
+
+# All env in this default.env can be overridden by env.local.
+
+# All env in this default.env must NOT depend on any other env.  If they do, they
+# must use single quotes to avoid early expansion before overrides in env.local
+# are applied and must be added to the list of DELAYED_EVAL.
+
+# idle timeout of a user jupyterhub server to be shut down automatically (default: 1 day)
+export JUPYTERHUB_STOP_IDLE_TIMEOUT=${JUPYTERHUB_STOP_IDLE_TIMEOUT:-86400}
+
+# maximum age before stoping user servers (includiung active/running ones, regardless of last activity status)
+# if not specified, this will not be defined by the utility
+# (i.e.: users that provide activity status before timeout could run their server indefinitely)
+export JUPYTERHUB_STOP_IDLE_MAX_AGE=
+
+# for details about the configuration, see https://github.com/jupyterhub/jupyterhub-idle-culler
+# WARNING: be careful about indents here, Python code execution depends on it
+export JUPYTERHUB_STOP_IDLE_CONFIG="
+import sys
+
+jupyterhub_stop_idle_timeout = ${JUPYTERHUB_STOP_IDLE_TIMEOUT}
+c.JupyterHub.load_roles = [
+    {
+        'name': 'jupyterhub-idle-culler-role',
+        'scopes': [
+            'list:users',
+            'read:users:activity',
+            'read:servers',
+            'delete:servers',
+        ],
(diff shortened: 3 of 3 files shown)


birdhouse/optional-components/jupyterhub-stop-idle/default.env
```diff
@@ -0,0 +1,74 @@
...
+# however, jupyter has by default 5min intervals, which is usually much lower than 'JUPYTERHUB_STOP_IDLE_TIMEOUT'
+c.JupyterHub.last_activity_interval = int(min(jupyterhub_stop_idle_timeout / 2, 300))
+"
+
+export JUPYTERHUB_CONFIG_EXTENDED="
+${JUPYTERHUB_CONFIG_EXTENDED}
+${JUPYTERHUB_STOP_IDLE_CONFIG}
+"
+
+#export DELAYED_EVAL="
```
mishaschwartz, line 60: Remove commented code please

birdhouse/optional-components/jupyterhub-stop-idle/default.env
```diff
@@ -0,0 +1,74 @@
...
+# All env in this default.env can be overridden by env.local.
+
+# All env in this default.env must NOT depend on any other env.  If they do, they
+# must use single quotes to avoid early expansion before overrides in env.local
+# are applied and must be added to the list of DELAYED_EVAL.
+
+# idle timeout of a user jupyterhub server to be shut down automatically (default: 1 day)
+export JUPYTERHUB_STOP_IDLE_TIMEOUT=${JUPYTERHUB_STOP_IDLE_TIMEOUT:-86400}
+
+# maximum age before stoping user servers (includiung active/running ones, regardless of last activity status)
```
mishaschwartz, line 12: Is this maximum age in seconds? Pleas specify the unit of time.

mishaschwartz (approved):
Looks good. Just a few documentation and clean-up comments below:

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -187,4 +187,5 @@ blocked_users = {'authtest', '${CATALOG_USERNAME}', 'anonymous'}
 c.Authenticator.blacklist = blocked_users  # v0.9+
 c.Authenticator.blocked_users = blocked_users  # v1.2+
 
+${JUPYTERHUB_CONFIG_EXTENDED}    # noqa
```
tlvu, line 190: Why not re-use the existing `JUPYTERHUB_CONFIG_OVERRIDE`?  

If you want to keep both, please leave 2 lines between the 2 variables so that when both are actually in use, they do not look mixed up.

birdhouse/config/jupyterhub/default.env
```diff
@@ -43,6 +43,15 @@ export JUPYTER_LOGIN_BANNER_BOTTOM_SECTION=""
 # server for the change to take effect.
 export JUPYTERHUB_README=""
 
+# Additional configuration to be provided by services or components managed by the stack.
+# Any component extending this configuration using this variable *MUST* include it back to allow further
+# components to extend it as well. Also, they *MUST* extend the variable with each new definitions applied on
+# a new line (without indents) to ensure that Python configurations will be executable without syntax error.
+# This variable differs from 'JUPYTERHUB_CONFIG_OVERRIDE' that does not require users to include it back to make
```
tlvu, line 50: What do you mean by "include it back"?

tlvu (changes requested):
Missing `optional-components/README.rst` and `CHANGES.md` update.

Also an existing idle culling mechanism was already possible via `env.local`, see https://github.com/bird-house/birdhouse-deploy/blob/2da7a69c33bd797b1c8b31e34c08e2c467e6d1e7/birdhouse/env.local.example#L338-L350 

Not clear to me how better or different this external idle culling differ but I wonder if this is some changes between Jupyterhub v1 vs v4.  I would say to keep this external idle culler in case the built-in idle culler do not work anymore in v4.

Please in the `optional-components/README.rst`, mention the existing  internal idle culler documented in `env.local.example`, to give users an option if one of the 2 works better or just plain do not work anymore.

Has this newer culler been tested by lowering the timeout to very short amount?

tlvu:
The README of the external jupyterhub-idle-culler also mention about the JupyterHub can already cull idle jupyter itself without explaining why would someone need to use this external version https://github.com/jupyterhub/jupyterhub-idle-culler/blob/737dfa155b809453e6e795dd9de42d6a926fd4a0/README.md?plain=1#L263-L274 

This is rather confusing to me.

@mishaschwartz what do you think about keeping 2 different culling options?  Would this confuse more the users?

mishaschwartz:
> @mishaschwartz what do you think about keeping 2 different culling options? Would this confuse more the users?

Yeah, I'd rather keep one version or the other.  It will definitely confuse users. 

Can we test out if the old version still works and if not, remove the comment from env.local. 
If it does still work then we should use that one instead _unless_ there is some major advantage to the `jupyterhub-idle-culler` method. Is there @fmigneault ?

birdhouse/config/jupyterhub/jupyterhub_config.py.template
fmigneault: I tried doing that, but given that some instances already use `export JUPYTERHUB_CONFIG_OVERRIDE=...` in their `env.local`, which is sourced when runninig `pavics-compose.sh`, any overrides applied along the way get ignored. 

Because of how `JUPYTERHUB_CONFIG_EXTENDED` gets defined using extra newlines, they are already separated by multiple extra lines. I could add more though, or even add comments in the template to separate them explicitly.

birdhouse/config/jupyterhub/default.env
fmigneault: Doing something in the form: 
```shell
export JUPYTERHUB_CONFIG_EXTENDED="
${JUPYTERHUB_CONFIG_EXTENDED}
<extra python here>
"
```

fmigneault:
> Can we test out if the old version still works and if not, remove the comment from env.local.

Yes. I would like to test existing approach, and if it works, only keep that one. 
However, I would like to make it more easily/directly configurable by some optional variable ratger than asking of users to add it to their `env.local` via `JUPYTERHUB_CONFIG_OVERRIDE`. 
I did not even realize this code existed before being pointed out by @tlvu

tlvu:
> However, I would like to make it more easily/directly configurable by some optional variable ratger than asking of users to add it to their `env.local` via `JUPYTERHUB_CONFIG_OVERRIDE`. 

Yes, true, anything in production will need this to avoid long running jupyter containers wasting ram and cpu.

Then I would suggest moving that snippet of code into the default jupyterhub config and just make those timeout limits config variables.

> I did not even realize this code existed before being pointed out by @tlvu

Being a production server with many users, we faced this issue https://github.com/bird-house/birdhouse-deploy/issues/67 and although this culling might not fixed the root cause, it probably helped.

fmigneault:
@tlvu @mishaschwartz 
I found this thread that seems to explain properly the differences: https://discourse.jupyter.org/t/which-is-the-correct-way-to-cull-idle-kernels-and-notebook/8123/16

From my reading, it would seem the internal culler is sufficient for our use case.
However, I found in the comments that there is also a terminal culler that could keep the session/server active even if the kernel is idle, if only a terminal is keep open. (https://github.com/jupyter-server/jupyter_server/pull/438)

So the config might need some extra parameters, but I would stick with only the internal parameters if those work by themselves.

PR #390, opened 2023-10-12
make jupyterhub idle server more easily configurable

fmigneault:
## Overview

Add new variables to easily configure idle jupyter user instances.

## Changes

**Non-breaking changes**
- Jupyterhub configurable idle server culling.
  - Add optional variables `JUPYTER_IDLE_SERVER_CULL_TIMEOUT`, `JUPYTER_IDLE_KERNEL_CULL_TIMEOUT` and
    `JUPYTER_IDLE_KERNEL_CULL_INTERVAL` that allows fined-grained configuration of user-kernel and server-wide
    docker image culling when their activity status reached a certain idle timeout threshold.
  - Enable idle kernel culling by default with a timeout of 1 day, and user server culling with timeout of 3 days.
  - Avoids the need for custom `JUPYTERHUB_CONFIG_OVERRIDE` specifically for idle server culling.
    If similar argument parameters should be defined using an older `JUPYTERHUB_CONFIG_OVERRIDE` definition,
    the new configuration strategy can be skipped by setting `JUPYTER_IDLE_KERNEL_CULL_TIMEOUT=0`.

**Breaking changes**
- n/a

## Related Issue / Discussion

- Closes https://github.com/bird-house/birdhouse-deploy/pull/389 (replaces)
- Closes https://github.com/Ouranosinc/jupyterhub/pull/21 (not required anymore)


fmigneault:
@tlvu @mishaschwartz 

I've tested it using some low values (eg: `JUPYTER_IDLE_KERNEL_CULL_TIMEOUT=10`). 
At some point, my `jupyter-<user>` image got shut down, though it took longer than anticipated. 
Some docs indicate that shutdown could take longer, considering the timeout only triggers the request to shutdown.
Either way, at least, it shut down at some point, without the need of the other strategy. 

I've opted for a default 1day timeout to promote servers using defaults/minimally-overridden configs to have cleanup enabled with some reasonable time, so user can experiment for some time without being impacted by shutdown, while stopping automatiically when they are clearly not using it anymore.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -187,4 +187,37 @@ blocked_users = {'authtest', '${CATALOG_USERNAME}', 'anonymous'}
...
+# Timeout (in seconds) after which individual user kernels/terminals are considered idle and ready to be culled.
+jupyter_idle_kernel_cull_timeout = int("${JUPYTER_IDLE_KERNEL_CULL_TIMEOUT}" or 0)
+# Interval (in seconds) on which to check for idle kernels exceeding the cull timeout value.
+jupyter_idle_kernel_cull_interval = int("${JUPYTER_IDLE_KERNEL_CULL_INTERVAL}" or 0)
+if jupyter_idle_kernel_cull_timeout:
+    if not jupyter_idle_kernel_cull_interval or jupyter_idle_kernel_cull_interval > jupyter_idle_kernel_cull_timeout:
+        jupyter_idle_kernel_cull_interval = jupyter_idle_kernel_cull_timeout / 2
+    c.Spawner.args.extend([
+        '--MappingKernelManager.cull_idle_timeout={}'.format(jupyter_idle_kernel_cull_timeout),
+        '--MappingKernelManager.cull_interval={}'.format(jupyter_idle_kernel_cull_interval),
```
tlvu, line 208: Funny I didn't have to manually set the `cull_interval` to haft of the `cull_idle_timeout` and it was working fine.  What is the default value of the 2 `cull_interval` if not set?

tlvu (approved):
LGTM.  Please update `env.local.example` with the 3 new vars and their documentations and delete the corresponding sample for `JUPYTERHUB_CONFIG_OVERRIDE`.

tlvu:
> I've tested it using some low values (eg: `JUPYTER_IDLE_KERNEL_CULL_TIMEOUT=10`).
> At some point, my `jupyter-<user>` image got shut down, though it took longer than anticipated.

I assume you also set `JUPYTER_IDLE_SERVER_CULL_TIMEOUT=10`?  So with those 2 configs, after 10 sec, the notebook kernel got killed, then 10 sec later the server is killed.

The server cull timeout starts only when all kernels are down.

For us at Ouranos, we set the server timeout to 4 days and the kernel timeout to 1 day.   Together, they account for 5 days of complete inactivity before the server is actually gone, so any 4 days long weekend is taken care of.

tlvu:
> > I've tested it using some low values (eg: `JUPYTER_IDLE_KERNEL_CULL_TIMEOUT=10`).
> > At some point, my `jupyter-<user>` image got shut down, though it took longer than anticipated.
> 
> I assume you also set `JUPYTER_IDLE_SERVER_CULL_TIMEOUT=10`? So with those 2 configs, after 10 sec, the notebook kernel got killed, then 10 sec later the server is killed.
> 
> The server cull timeout starts only when all kernels are down.
> 
> For us at Ouranos, we set the server timeout to 4 days and the kernel timeout to 1 day. Together, they account for 5 days of complete inactivity before the server is actually gone, so any 4 days long weekend is taken care of.

Given this new config, if  `JUPYTER_IDLE_SERVER_CULL_TIMEOUT` is not set or is zero (the default), the server would not be killed at all.

I think we should put some value to avoid the server running forever.  Maybe a bigger value like a week would be reasonable?

birdhouse/config/jupyterhub/jupyterhub_config.py.template
mishaschwartz: [300 seconds](https://jupyter-server.readthedocs.io/en/latest/api/jupyter_server.services.kernels.html#jupyter_server.services.kernels.kernelmanager.MappingKernelManager.cull_interval_default) I believe

mishaschwartz (approved):
Looks good to me.

I agree with this though: 

> Given this new config, if JUPYTER_IDLE_SERVER_CULL_TIMEOUT is not set or is zero (the default), the server would not be killed at all.
> I think we should put some value to avoid the server running forever. Maybe a bigger value like a week would be reasonable?

So I'd also like to see that default changed before we merge this

birdhouse/config/jupyterhub/jupyterhub_config.py.template
fmigneault: Indeed, there is the default of 300 from jupyter-server, but with my proposed changes at L205, the value is automatically halved from the timeout if not specified. I had to do this, otherwise the interval between updates took at least 300s regardless of my timeout=10 test.

fmigneault:
> I assume you also set `JUPYTER_IDLE_SERVER_CULL_TIMEOUT=10`? So with those 2 configs, after 10 sec, the notebook kernel got killed, then 10 sec later the server is killed.

Indeed. Sorry for the confusion. Both variables were set at 10s.

tlvu (approved):
LGTM.  Maybe set the server timeout to 4 days instead of 3 days so the guy coming back from a 4 days long weekend have a chance resume his work?

fmigneault:
> LGTM. Maybe set the server timeout to 4 days instead of 3 days so the guy coming back from a 4 days long weekend have a chance resume his work?

I'd consider that an edge case. Also, they should be able to resume work from a notebook saved in the user-workspace that would be re-mounted on server restart.

mishaschwartz (approved):

tlvu:
> I'd consider that an edge case. Also, they should be able to resume work from a notebook saved in the user-workspace that would be re-mounted on server restart.

Yes it's an edge case.

It's not about preserving the notebooks, it's about preserving any custom installs the user had made without properly recording them in a `requirements.txt` or `environment.yml`.

It's fine, 3 or 4, the node admin with adjust if they receive complains over long weekend.

PR #391, opened 2023-10-17
Jupyterhub cull interval setting must be an integer

mishaschwartz:
## Overview

Previously, the default `jupyter_idle_kernel_cull_interval` setting is calculated by dividing the `jupyter_idle_kernel_cull_timeout` setting by 2 using float division. This meant that the result was a float instead of the expected integer value. This caused and error when the jupyterlab server spawned. In order to fix this, the value is cast to an integer after division.

## Changes

**Non-breaking changes**
- Bug fix

**Breaking changes**
None

## Additional Information

The error message encountered if the value is not cast to an integer is as follows:

```
Bad config encountered during initialization: The 'cull_interval' trait of an AsyncMappingKernelManager instance expected an int, not the str '43200.0'.
```

This indicates that the value is read in as a string and checked whether it is an integer or not (possibly with python's `str.isnumeric()` method).

birdhouse_skip_ci: false


birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -203,7 +203,7 @@ jupyter_idle_kernel_cull_timeout = int("${JUPYTER_IDLE_KERNEL_CULL_TIMEOUT}" or
 jupyter_idle_kernel_cull_interval = int("${JUPYTER_IDLE_KERNEL_CULL_INTERVAL}" or 0)
 if jupyter_idle_kernel_cull_timeout:
     if not jupyter_idle_kernel_cull_interval or jupyter_idle_kernel_cull_interval > jupyter_idle_kernel_cull_timeout:
-        jupyter_idle_kernel_cull_interval = jupyter_idle_kernel_cull_timeout / 2
+        jupyter_idle_kernel_cull_interval = int(jupyter_idle_kernel_cull_timeout / 2)
```
fmigneault, line 206: Just in case, I would add a `max(1, int(jupyter_idle_kernel_cull_timeout / 2))` to avoid some division rounded down to 0.

fmigneault (approved):

tlvu (approved):
Thanks for catching this !

PR #396, opened 2023-10-25
limit Magpie usernames

### LINKED ISSUE #393: :bug: [BUG]: magpie allows special characters in usernames that mess up jupyterhub
## Summary

Jupyterhub creates jupyterlab docker containers named `jupyter-{username}` where `{username}` is the user name of the logged in user.

Magpie permits usernames with `.` in them but jupyterhub will substitute the `.` with `-2e` when creating the jupyterlab container. This means that a user with the name `example.user` will have a container named `jupyter-example-2euser`. 

This causes problems since we assume that these names will align.

## Details

- Magpie restricts usernames according to this regex: `^[A-Za-z0-9]+(?:[\s_\-\.][A-Za-z0-9]+)*$`
    - https://github.com/Ouranosinc/Magpie/blob/3.36.0/magpie/api/management/user/user_utils.py#L868-L882
    - https://github.com/Ouranosinc/Magpie/blob/3.36.0/magpie/api/exception.py#L48
- docker does not restrict container names (it allows `.` in container names)
- jupyterhub escapes everything except lowercase ascii and digits
    - https://github.com/jupyterhub/dockerspawner/blob/12.1.0/dockerspawner/dockerspawner.py#L625

When spawning a new jupyterlab container, it fails with the following traceback:

```
  File "/opt/conda/envs/birdy/bin/jupyterhub-singleuser", line 10, in <module>
    sys.exit(main())
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/jupyterhub/singleuser/app.py", line 68, in main
    return SingleUserNotebookApp.launch_instance()
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/jupyter_core/application.py", line 277, in launch_instance
    return super().launch_instance(argv=argv, **kwargs)
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/traitlets/config/application.py", line 1040, in launch_instance
    app.initialize(argv)
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/jupyterhub/singleuser/mixins.py", line 971, in initialize
    result = super().initialize(*args, **kwargs)
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/jupyterhub/singleuser/mixins.py", line 642, in initialize
    super().initialize(argv)
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/traitlets/config/application.py", line 113, in inner
    return method(app, *args, **kwargs)
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/jupyter_server/serverapp.py", line 2461, in initialize
    self.init_configurables()
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/jupyter_server/serverapp.py", line 1864, in init_configurables
    connection_dir=self.runtime_dir,
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/traitlets/traitlets.py", line 703, in __get__
    return self.get(obj, cls)
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/traitlets/traitlets.py", line 659, in get
    default = obj.trait_defaults(self.name)
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/traitlets/traitlets.py", line 1871, in trait_defaults
    return self._get_trait_default_generator(names[0])(self)
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/jupyter_core/application.py", line 107, in _runtime_dir_default
    ensure_dir_exists(rd, mode=0o700)
  File "/opt/conda/envs/birdy/lib/python3.9/site-packages/jupyter_core/utils/__init__.py", line 25, in ensure_dir_exists
    os.makedirs(path, mode=mode)
  File "/opt/conda/envs/birdy/lib/python3.9/os.py", line 215, in makedirs
    makedirs(head, exist_ok=exist_ok)
  File "/opt/conda/envs/birdy/lib/python3.9/os.py", line 215, in makedirs
    makedirs(head, exist_ok=exist_ok)
  File "/opt/conda/envs/birdy/lib/python3.9/os.py", line 215, in makedirs
    makedirs(head, exist_ok=exist_ok)
  [Previous line repeated 1 more time]
  File "/opt/conda/envs/birdy/lib/python3.9/os.py", line 225, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/notebook_dir/writable-workspace/.home'
```

I'm not entirely sure why the issue is a permission denied error but the cause is obviously due to an unexpected mismatch between the expected username and the username according to the container.

One possible solution would be to further restrict the valid usernames in magpie, either permanently or only when jupyterhub is enabled.

### To Reproduce

1. start birdhouse deploy with the magpie and jupyterhub components enabled
2. create a user with a `.` in their name in magpie (eg: `user.name`)
3. log in to jupyterhub as that user and spawn a new jupyterlab server
4. inspect the jupyterhub logs for the error message: `docker logs -f jupyter-user.name`

## Environment

| Server/Platform URL   | daccs.cs.toronto.edu
| Version Tag/Commit    | 1.35.0
| Related issues/PR     | 
| Related components    | jupyterhub, magpie
| Custom configuration  | 
| docker version |  Docker version 24.0.2, build cb74dfc

mishaschwartz:
## Overview

When Jupyterhub spawns a new jupyterlab container, it escapes any non-ascii, non-digit character in the username. This results in a username that may not match the expected username (as defined by Magpie). This mismatch results in the container failing to spawn since expected volumes cannot be mounter to the jupyterlab container.

This fixes the issue by ensuring that juptyerhub does not convert the username that is receives from Magpie.

## Changes

**Non-breaking changes**
- limits usernames in Magpie

**Breaking changes**
- If any users currently exist with invalid usernames, an admin needs to update their user name in Magpie

## Related Issue / Discussion

- Resolves #393

## Additional Information

birdhouse_skip_ci: false


birdhouse/config/jupyterhub/config/magpie/docker-compose-extra.yml
```diff
@@ -0,0 +1,5 @@
+version: "3.4"
+services:
+  magpie:
+    environment:
+      MAGPIE_USER_NAME_EXTRA_REGEX: "^[a-z0-9]+$$"
```
tlvu, line 5: Just curious, instead of this env var could we have done this in the `magpie.ini` file?  I have a slight preference for all configs to be grouped together in one file for ease of tracing them.

Is the `magpie.ini` route not possible or there are advantages to use env var?

tlvu (approved):
LGTM but consider my suggestion to group all configs together, if possible.

birdhouse/config/jupyterhub/config/magpie/docker-compose-extra.yml
mishaschwartz: The reason that I put this here is that we only need to enforce this limitation if both the magpie and jupyterhub components are enabled. 
Yes, this will be 99% of the use-cases but it's important to keep the dependencies clear in the way we structure things in this repo.

mishaschwartz: Another way to think about this: This follows the "pluggable component architecture" that we've been trying to create since #214

tlvu: Oh I see !  Never thought about that one, having Magpie without JupyterHub do not need this config.  You are right.

fmigneault (commented):
Sorry to bring this up again, but just curious to see if all these Magpie changes were actually necessary. Maybe the solution was taken too hastily?

Instead of forcing a subset of escaped/lowercase characters from JupyterHub to Mapgie, it seems we could simply update the references to use `{raw_username}` instead of `{username}` and keep the large character set of Magpie usernames. Magpie names are restricted to `^[A-Za-z0-9]+(?:[\s_\-\.][A-Za-z0-9]+)*$`, which should all be valid characters for the file system.

https://github.com/jupyterhub/dockerspawner/blob/a6bf72e764dfcfdf4d1dbb1c5a84bedd183505e2/dockerspawner/dockerspawner.py#L355-L357

fmigneault:
And if `{raw_username}` doesn't work, we can also provide our own escape function: 
https://github.com/jupyterhub/dockerspawner/blob/a6bf72e764dfcfdf4d1dbb1c5a84bedd183505e2/dockerspawner/dockerspawner.py#L600-L618

mishaschwartz:
@fmigneault 

As @tlvu mentions here (https://github.com/bird-house/birdhouse-deploy/issues/393#issuecomment-1769089001) having a different username in Magpie as in Jupyterhub adds an additional burden on admins who are trying to manage the system.

Also, having mismatched names may cause confusion for users who log in with one username and see a different username displayed in jupyterlab.

Finally, any strategy that requires us to escape usernames means we have to be very very careful about conflicts. 

For example, in Magpie, we can theoretically have two users: `example.user` and `example-2euser`. If the escape function converts `.` to `-2e` (as it does now) then these two users will be able to access each other's jupyterlab container. 

I'm not saying we can't construct an escape function that will be foolproof, I just think that this adds a ton of extra complexity that we have to worry about. 

In my opinion it is simpler for both admins, developers, and users to simply restrict the usernames in Magpie

fmigneault:
@mishaschwartz 
But the links I referenced here are exactly *to avoid escaping*, in order to remove the problematic feature Jupyter applies by default causing modified user names. Therefore, user names from Magpie would be used as is. Users would not see an escaped variations of their name in Jupyter, it would be the same one they used for signin up in Magpie.

mishaschwartz:
@fmigneault

You're right, my comments were only responding to your suggestion to override escape_name. However:

- just using {raw_username} in the name_template does not fix the problem since the Authenticator also normalizes the username (https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html#normalize-usernames) by converting it to lowercase
- if we also update the Authenticator.normalize_username function to not convert the name to lowercase, we still have the problem that Magpie allows spaces in usernames and docker containers cannot be created with spaces.
- to avoid docker containers created with spaces, we can override escape_name instead of using {raw_username} in name_template (which I don't think we should do... see above).

I really think that the best and simplest option is to just restrict user names

fmigneault:
@mishaschwartz 
And override could simply be: 
```python
class CustomDockerSpawner(DockerSpawner): 
    @property
    def escaped_name(self):
        return self.user.name
```

Since the authenticator is also a custom implementation, it is a very quick to override `normalize_username` the same way.

> Magpie allows spaces in usernames

Good point.
I'm not against limiting spaces in the configured regex to avoid issues with docker and paths that would not be properly quoted. 
However, this is still lower impact than blocking `_`/`-`/uppercase which are more common in usernames.

tlvu:
Just to recap, we are still discussing about this PR simply to have `_`/`-`/uppercase in the username?

I don't think it is worth it.  No users have ever complained to us in production that they did not like the username we gave them and need to have `_`/`-`/uppercase in the username.

I would rather have the regular `{username}` over `{raw_username}` with the extra username validation regex.  It just feels simpler and more safe.  If JupyterHub take the effort to have a "safe" `{username}`, I think probably there is a reason.

fmigneault:
We do have some users that have `.`, `_`, `-`. 
Even some control users in birdhouse used these characeters, eg;  `admin-catalog`.
These users are not actively using Jupyter, hence why nothing "broke" for them, but restricting their usernames because of Jupyter seems like patching the issue the wrong way around. 

Recaping everything, I would suggest that `{username}` are always used (as already done), but override the 2 functionsthat do the escape (for volume and docker names respectively) to simply not do them. There is no need for extra validation regex, since that is already enforced by Magpie user registration. 
In Magpie, @mishaschwartz's feature would be used to set `^[A-Za-z0-9]+(?:[_\-\.][A-Za-z0-9]+)*$` such that only the spaces are removed from Magpie's default regex.

mishaschwartz:
> We do have some users that have ., _, -.
> Even some control users in birdhouse used these characeters, eg; admin-catalog.

@fmigneault This would have been useful to know earlier in the discussion so that I could have understood _why_ you were objecting so strongly to this change.

With regards to the admin-catalog user: this user is only added if the (**deprecrated**) catalog component is enabled and it is added to jupyterhub's blacklist anyway. I'm not inclined to limit this change in order to support a deprecated component.

Are there other control users that you are concerned about that are not deprecated? 

>  If JupyterHub take the effort to have a "safe" {username}, I think probably there is a reason.

This is giving me pause too @tlvu . I don't want to override these functions and then create other conflicts later on. However, the fact that jupyter provides public configuration options to override these settings makes me feel a little bit better about it. I would be willing to consider overriding these functions after thorough testing.

tlvu:
> * If there are non-deprecated control users that contain special character that we need to support,

The `admin-catalog` user will never login to Jupyter.  If any other non-deprecated control users that ever need to login to Jupyter, we can also just rename them since they are totally under our control.  So far, I think, probably, none of the control users need to login to Jupyter yet.   Otherwise, we would already see problems in production.

This is just to say, unless the effort to test the override functions is simple and quick, then yes go for it since it's nice to have `-` and `_` in username.  Otherwise I don't think it's worth the time.

Note that we can always defer overriding the functions later since it's easy to relax existing restriction.  It's much harder to tighten restrictions since it implies manually fixing all the relaxed usernames that become illegal with harder restriction.

birdhouse/config/jupyterhub/config/magpie/docker-compose-extra.yml
tlvu: Hey, I just think of a counter argument in favor of `magpie.ini` since we are discussing about control users already violating the extra regex.

Even if Magpie can initially be activated without Jupyterhub, we better ensure all the users created are "JupyterHub forward compatible".  Otherwise the poor node admin will be in for a surprise the day they activate JupyterHub.

That is to say, it's more future proof to have a stricter regex, then relax it if needed, than the other way around.

fmigneault (commented):
Looks good. 
Will have to see how that works with https://github.com/Ouranosinc/Magpie/pull/596.
I propose we update this PR with the Magpie version that has the fix at the same time.

birdhouse/config/jupyterhub/config/magpie/docker-compose-extra.yml
```diff
@@ -1,5 +1,7 @@
 version: "3.4"
 services:
   magpie:
+    environment:
+      MAGPIE_USER_NAME_EXTRA_REGEX: ^\S+$$
```
tlvu, line 5: I am completely confused here.  So `\S` will "matches any non-whitespace character; this is equivalent to the class [^ \t\n\r\f\v]." (definition from https://docs.python.org/3/howto/regex.html#simple-patterns).

So we want the usernames to only contain non-writespace char ???

Other than non-whitespace, we probably would want to avoid other kind of illegal char like `/`, `*`, `#` ...  Unless this is already handled by Magpie built-in username validation?

Or I might have completely mis-understood how the new `MAGPIE_USER_NAME_EXTRA_REGEX` works.

CHANGES.md
```diff
@@ -15,7 +15,17 @@
...
+## Changes
+
+- Limit usernames in Magpie to match restrictions by Jupyterhub's Dockerspawner
+
+  When Jupyterhub spawns a new jupyterlab container, it escapes any non-ascii, non-digit character in the username as
+  well as converting any uppercase character to lowercase. This results in a username that may not match the expected
+  username (as defined by Magpie). This mismatch results in the container failing to spawn since expected volumes
+  cannot be mounter to the jupyterlab container.
+
+  This fixes the issue by limiting Magpie usernames to contain only lowercase ascii characters and digits. This ensures
```
tlvu, line 27: Are we still "limiting Magpie usernames to contain only lowercase ascii characters and digits".  This should be updated to match the extra regex.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -25,6 +25,15 @@ c.JupyterHub.db_url = '/persist/jupyterhub.sqlite'
...
 
 class CustomDockerSpawner(DockerSpawner):
+    @property
+    def escaped_name(self):
+        """
+        Return the username without escaping. This ensures that mounted directories on the
+        host machine are discovered properly since we expect the username to match the username
+        set by Magpie.
+        """
+        return self.user.name
```
tlvu, line 35: Oh I am not sure I want to override `escape_name()`.  This will make `{username}` for old existing username that potentially can have invalid char actually use the raw username with invalid char.

`MAGPIE_USER_NAME_EXTRA_REGEX` take effect on new username only.  Existing ones, if they are invalid, are still invalid.

tlvu (changes requested):
The new regex surprised me.

birdhouse/config/jupyterhub/config/magpie/docker-compose-extra.yml
tlvu: @mishaschwartz Not sure if you've seen my previous comment above but I think it's better to have the new extra regex in `magpie.ini` so all users created are forward-compatible if JupyterHub is enabled later.

birdhouse/config/jupyterhub/config/magpie/docker-compose-extra.yml
tlvu: Opps sorry, my brain is dead past midnight, yes we want non-whitespace char.  

The regex is right but I'd like something more conservative, like an explicit allow list: `"^[a-zA-Z0-9_\-]+$$"`.  This explicit list will protect us from any other illegal char that we ever forgot to think about.  I'd rather be safe than sorry.

mishaschwartz: This regex is used in combination with the one already enforced by Magpie by default `^[A-Za-z0-9]+(?:[\s_\-\.][A-Za-z0-9]+)*$` (see https://github.com/bird-house/birdhouse-deploy/pull/396#pullrequestreview-1698302553).

The username must match both that default regex as well as `^\S$`. The default regex protects us from unexpected characters, the second regex protects us from whitespace.

I'd rather not duplicate most of the default regex here since the more complex we make it, the more prone to errors we are if we change it in the future.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
mishaschwartz: Any user who would have that issue will not be able to spawn a jupyterlab server right now and you would know about it already.

If we don't override `escape_name` then we have to limit the usernames to lowercase ascii and digits. I believe that @fmigneault is still not in favour of that strategy though.

birdhouse/config/jupyterhub/config/magpie/docker-compose-extra.yml
fmigneault: If we are going to override the regex either way, I agree with @tlvu that it is better to have it explicitly, even if they overlap partially. It is easier to analyze one location than having to go back into Magpie code to validate the full set. Even if we do make a mistake, the regex on Magpie side will still double-check it.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
fmigneault: The Jupyter usernames were always bounded by Magpie usernames at least. The only mismatch this escape caused was about lower/upper case (which won't even be a thing anymore with an upcoming Magpie PR), and `.-_` characters.

fmigneault (approved):
Assuming the fix of lower/upper-case user name handling is done in Magpie, this looks fine to me.

birdhouse/config/jupyterhub/config/magpie/docker-compose-extra.yml
```diff
@@ -1,7 +1,5 @@
 version: "3.4"
 services:
   magpie:
-    environment:
-      MAGPIE_USER_NAME_EXTRA_REGEX: ^\S+$$
```
tlvu: I assume you migrate this change to `magpie.ini` to ensure forward compat if later JupyterHub is enabled?  Missing corresponding change in `magpie.ini`.  Please use an explicit regex `"^[a-zA-Z0-9_\-\.]+$$"` which would allow upper case, `_-.` and nothing else.

This is to ensure even if Magpie validation regex changes later, we are independent and as @fmigneault said, this is to simplify the life of other node admin, avoiding them to lookup the validation regex on Magpie side.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
tlvu: > Any user who would have that issue will not be able to spawn a jupyterlab server right now and you would know about it already.

Agreed.  For existing user that never log into Jupyter before, if they generate an error, hopefully they will contact us and we'll create them a new username that is "valid".

Now that I think about it though, isn't overriding `escape_name()` to return `self.user.name` the same as simply using `{raw_username}`?  Why overriding when we can simply use `{raw_username}`?

tlvu (approved):
Please add back the extra regex in `magpie.ini` in explicit form.

For `{raw_username}` vs overriding `escape_name()` this is just a my reflection, don't have to change it if no new advantages.

Otherwise, the rest looks good, thanks.

birdhouse/config/jupyterhub/config/magpie/docker-compose-extra.yml
tlvu: When adding back the extra regex, please add a comment that Magpie already has a built-in regex too, just to raise awareness for those not in the know.

I guess now that Magpie do not allow uppercase, the extra regex becomes `"[a-z0-9_\-\.]+$$"`.

mishaschwartz: @tlvu 

With the new changes, we do not need to use the `MAGPIE_USER_NAME_EXTRA_REGEX` feature at all. The current default checks that Magpie does are sufficient (without the need to do an extra check), see https://github.com/Ouranosinc/Magpie/pull/597 for details

mishaschwartz:
> Now that I think about it though, isn't overriding escape_name() to return self.user.name the same as simply using {raw_username}? Why overriding when we can simply use {raw_username}?

This only works in the places where jupyterhub allows you to specify {raw_username} in a template. We want to ensure that this applied in all cases (not just when mounting user workspaces to the container)

birdhouse/config/jupyterhub/config/magpie/docker-compose-extra.yml
tlvu: @mishaschwartz 

I see, thanks.  In this case, just to help other future node admin not in the know, can you document in `magpie.ini` the exact permanent link to the magie regex, this one currently https://github.com/Ouranosinc/Magpie/blob/595602d2cebae94223b952f0cf04a1caa64f6546/magpie/api/management/user/user_utils.py#L66. 

So this is the regex we expect at this moment.  If in the future this regex changes on Magpie side, and somehow break us, we will know what was the original expected working regex.

PR #397, opened 2023-11-02
optional-components/all-public-access: remove GeoServer route permissions

fmigneault:
## Overview

The `type: route` is invalid in that case, because the permission is applied on the service itself (not a child `route` resource). The type should be `service`, but can be omitted when no `resource` is specified, as in the case for all other permissions.

## Changes

**Non-breaking changes**
- Remove erroneous `type: route` for a Magpie permission set directly on the service.

**Breaking changes**
- n/a

## Related Issue / Discussion

- Introduced by https://github.com/bird-house/birdhouse-deploy/pull/348


mishaschwartz (approved):
Is this related to our discussion here? https://github.com/bird-house/birdhouse-deploy/pull/348/files/c4f8d68ca6e7ed5daf735786511dad6856468332#r1338846201

I guess I misunderstood your directions

tlvu (approved):
One line change, the easiest PR approval ever, but did you forget CHANGES.md ? :D 

It can never be one line :(

Approve the spirit since you know Magpie much better than me.

tlvu:
By the way, I changed the PR title to clearly indicate this applies to the optional-components/all-public-access only.  

My first impression with the original PR title is whoah, are we locking down further GeoServer by default?  But then I realized it's not by default.

fmigneault:
> Is this related to our discussion here?

Yes.

PR #399, opened 2023-11-08
Update `DEFAULT_CONF_DIRS` to the minimal components required to deploy the stack

### LINKED ISSUE #357: Document how to employ a subset of components
# Summary

The intent of this issue is to document the *recommended approach* for future users, such that we reduce the potential combinations of unexpected overrides they might attempt, and they could as us to maintain support. 

# Description

Using the `EXTRA_CONF_DIRS`, it is possible to apply additional configurations on top of `DEFAULT_CONF_DIRS`. The activated components are then the super-set of `DEFAULT_CONF_DIRS | EXTRA_CONF_DIRS`. However, this still forces the user to have, *at least*, the `DEFAULT_CONF_DIRS` set of components enabled. Technically, it would be perfectly possible to override `DEFAULT_CONF_DIRS` to start the instance with an even smaller subset than the proposed default services.

The recommended approach to do so should be better documented in: 
https://github.com/bird-house/birdhouse-deploy/blob/2b344d3383d8c0cd3f4cc4a0bf4266e3bae06a47/birdhouse/env.local.example#L45-L47

Also, there might be a need to document a "minimal set" of dependencies (i.e.: what must absolutely be defined for the instance to start without error). Notably, `proxy` (to have `nginx`) and "some service" that overrides the root location would be needed. Maybe more?
https://github.com/bird-house/birdhouse-deploy/blob/2b344d3383d8c0cd3f4cc4a0bf4266e3bae06a47/birdhouse/env.local.example#L216-L219

This "minimal set" could be simply documented, and it would be up to the new node maintainers to keep them in `DEFAULT_CONF_DIRS`, or we could be more proactive and make another `MINIMAL_CONF_DIRS` variable (or some other method...?).

mishaschwartz:
## Overview

Changes `DEFAULT_CONF_DIRS` to refer exclusively to the proxy, magpie, twitcher, stac, and cowbird components. Also moves all components that were previously under the `birdhouse/config` directory to the `birdhouse/components` directory. This removes the arbitrary distinction between these groups of components that didn't have any functional or logical reason.

Because this change updates the default components, this is not backwards compatible unless the following changes are made to the local environment file (`birdhouse/env.local` by default):

  - add any components no longer in the `DEFAULT_CONF_DIRS` list to the `EXTRA_CONF_DIRS` list.
    For example, to keep the jupyterhub component enabled, add `./components/jupyterhub` to the `EXTRA_CONF_DIRS` list.

## Changes

**Non-breaking changes**

- changes `PROXY_ROOT_LOCATION` to refer to the magpie login page by default but to the jupyterhub login page when jupyterhub is also enabled.

**Breaking changes**

- moves all components under `./config` to `./components`
- sets `DEFAULT_CONF_DIRS` to refer exclusively to the proxy, magpie, twitcher, stac, and cowbird

## Related Issue / Discussion

- Resolves #357 

## Additional Information

birdhouse_skip_ci: false


birdhouse/.gitignore
```diff
@@ -0,0 +1 @@
+./config
```
fmigneault, line 1: Since we are purposely moving components, should we consider leaving this un-ignored? 
Could help detect an invalid definition that refers to the old path.

birdhouse/default.env
```diff
@@ -44,17 +44,11 @@ or a custom web interface.
...
-  ./config/portainer
-  ./config/magpie
-  ./config/twitcher
-  ./config/jupyterhub
+  ./components/proxy
+  ./components/magpie
+  ./components/twitcher
+  ./components/stac
+  ./components/cowbird
 '
```
fmigneault, line 52: Given this change, I propose we use `2.0.0` for the new version.

birdhouse/README.rst
fmigneault, line 1: Since many components are not enabled by default anymore, and are not distinguished from previous components/config location, it would be important to include details in this file about what each component accomplishes, such that users can have a better idea of what each one does in the ecosystem. Most of the details can be extracted from the corresponding canarie-api configs for each service/component.

birdhouse/components/proxy/default.env
```diff
@@ -8,7 +8,7 @@ export PROXY_READ_TIMEOUT_VALUE="240s"
 
 # Content of "location /" in file config/proxy/conf.d/all-services.include.template
 # Useful to have a custom homepage.
-export PROXY_ROOT_LOCATION="return 302 https://\$host/jupyter/hub/login;"
+export PROXY_ROOT_LOCATION="return 302 https://\$host/magpie;"
```
fmigneault, line 11: I saw that the same variable is set by default when enabling `jupyterhub`. 
Should we have a better strategy to define the desired root location redirect? 
It seems that which one will be applied depends on a combination of dependency resolution order + file-system listing order. I'm not sure if this is replicable, unless one explicitly overrides the variable in their `env.local`.

birdhouse/default.env
```diff
@@ -44,17 +44,11 @@ or a custom web interface.
...
-  ./config/thredds
-  ./config/portainer
-  ./config/magpie
-  ./config/twitcher
-  ./config/jupyterhub
+  ./components/proxy
+  ./components/magpie
+  ./components/twitcher
+  ./components/stac
+  ./components/cowbird
```
fmigneault, line 51: No difference technically, but can you put cowbird directly after magpie/twitcher given how closely related they are feature wise?

birdhouse/components/cowbird/default.env
```diff
@@ -68,5 +68,5 @@ DELAYED_EVAL="
 # (include this for Cowbird<2.0.0)
 #COMPONENT_DEPENDENCIES="
 #  $COMPONENT_DEPENDENCIES
-#  ./config/mongodb
+#  ./components/mongodb
 #"
```
fmigneault, line 72: Technically, `component/magpie` should be in there, otherwise Cowbird doesn't really have anything to interact with.

birdhouse/pavics-compose.sh
```diff
@@ -129,7 +129,7 @@ do
 
     # run postgres post-startup setup script
     # Note: this must run before the post-docker-compose-up scripts since some may expect postgres databases to exist
-    postgres_id=$(PROXY_SECURE_PORT=443 HOSTNAME=${PAVICS_FQDN} docker-compose ${COMPOSE_CONF_LIST} ps -q postgres)
+    postgres_id=$(PROXY_SECURE_PORT=443 HOSTNAME=${PAVICS_FQDN} docker-compose ${COMPOSE_CONF_LIST} ps -q postgres 2> /dev/null)
```
fmigneault, line 132: Just curious. Which error/output is getting muted here?

birdhouse/scripts/get-components-json.include.sh
```diff
@@ -35,7 +35,7 @@ cd "${BIRDHOUSE_DEPLOY_COMPONENTS_ROOT}" || true  # ignore error for now, empty
 
 # note: no quotes in 'ls' on purpose to expand glob patterns
 BIRDHOUSE_DEPLOY_COMPONENTS_LIST_KNOWN="$( \
-  ls -d1 ./*components/*/ ./config/*/ 2>/dev/null \
+  ls -d1 ./*components/*/ ./components/*/ 2>/dev/null \
```
fmigneault, line 38: Seems redundant to have component twice. Should be globed by the first one.

tests/pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+markers =
+    online: marks test as requiring access to the internet (to run offline: '-m "not online"'
```
fmigneault, line 3: Missing closing `)`.
The end message is somewhat misleading. It will not "run offline", but skip online tests entirely.

CHANGES.md
```diff
@@ -15,7 +15,21 @@
...
+  Also moves all components that were previously under the `birdhouse/config` directory to the `birdhouse/components`
+  directory. This removes the arbitrary distinction between these groups of components that didn't have any functional
+  or logical reason.
+
+  Because this change updates the default components, this is not backwards compatible unless the following changes are
+  made to the local environment file (`birdhouse/env.local` by default):
+
+  - add any components no longer in the `DEFAULT_CONF_DIRS` list to the `EXTRA_CONF_DIRS` list.
+  
+    For example, to keep the jupyterhub component enabled, add `./components/jupyterhub` to the `EXTRA_CONF_DIRS` list.
```
fmigneault, line 32: Given that `./components/jupyterhub` is not enabled by default anymore, but `./components/stac` is, it might be worth mentioning that adding `PROXY_ROOT_LOCATION` explicitly could be required depending on which components are enabled, and which one the node manager desires to put as default redirect (if any).

fmigneault (commented):
Since the `EXTRA_CONF_DIRS` was preemptively set with both `config` and `component` locations for all items explicitly, and that the CI test suite runs everything without major problem, I am confident the refactor must be good. 

However, I would like to hold off this PR for some time, as I am aware that @ChaamC, @Nazim-crim and @ahandan are all working on finalizing some relatively complex implementations that interact with multiple components simultaneously. I would prefer to avoid merge conflict burden on top of their changes that should be interacted in a near future.

birdhouse/.gitignore
mishaschwartz: I think we have to ignore it so that the autoupdate code works. Right @tlvu ?

birdhouse/default.env
mishaschwartz: I agree

birdhouse/README.rst
mishaschwartz: That makes sense. Do you think that belongs here or as a separate PR to update documentation.

birdhouse/components/proxy/default.env
mishaschwartz: > It seems that which one will be applied depends on a combination of dependency resolution order + file-system listing order

If jupyterhub is enabled it will always override the magpie settings in `default.env` since magpie is enabled in `DEFAULT_CONF_DIRS` which will always be read before `EXTRA_CONF_DIRS` (where jupyterhub is enabled). Because jupyterhub's `default.env` will be read second, it will override the variable set in magpie's `default.env`

fmigneault: Ok. Makes sense. I would add a comment that explicit this assumption, and indicate that this is only as last resort default if nothing else was enabled. Realistically, Magpie should probably not be the entrypoint, but we need "something" to fill the variable.

birdhouse/README.rst
fmigneault: Since we are aiming for a 2.0, I would include it all before bumping, but it can be done in a separate PR if that feels cleaner.

birdhouse/.gitignore
fmigneault: Since it would be a major update that could remove components if misconfigured, I guess it that it would feel safer to purposely break the autoupdate to have node admins make sure they set up their `EXTRA_CONF_DIRS` properly and clean up leftover configs if applicable before updating. However, if others don't feel this could be an issue, I'm fine with having the gitignore also. I'm not personally using autoupdate.

birdhouse/pavics-compose.sh
mishaschwartz: The error says that there is no container named postgres running (none of the default components require postgres now)

birdhouse/.gitignore
mishaschwartz: I'm not either so let's let @tlvu decide this one

birdhouse/pavics-compose.sh
fmigneault: I see! I suspected that was the case for a while.

birdhouse/README.rst
mishaschwartz: I've added it as an issue here: https://github.com/bird-house/birdhouse-deploy/issues/400 so that we don't forget about it. 

Since we're waiting on a few other PRs anyway before we merge this PR, we can wait a little to do the documentation updates.

birdhouse/.gitignore
tlvu: I'd rather prefer not breaking the autodeploy.  We had already the plan to do this: add the new paths in advance to `env.local` so it is already forward compatible when this PR merge so no manual interventions required.

However I would also ping PCIC in this PR so they are aware of this major change.

tlvu:
@eyvorchuk this PR will not enable my default many of the currently components enabled by default.

I think most of them are not used by PCIC so you guys will probably like this PR a lot as it will save CPU and RAM on your side. However I do know that PCIC uses Thredds and this PR will not enable Thredds by default.

To ensure forward compatibility when this PR will be merged, please add in advance Thredds to `EXTRA_CONF_DIRS` in `env.local` file:
```sh
EXTRA_CONF_DIRS="
  ./components/thredds
  /your/other/external/components
  (...)
"
```

The var `EXTRA_CONF_DIRS` allows  not existing dir to be specified so you can do this right now to avoid forgetting about it.

If there are other components in use by PCIC please also add them back explicitely.

I highly suggest you to actually test deploy this PR to catch all errors in advance.

Please approve this PR once you have made the change to your `env.local` and optionally test deploy this PR.

eyvorchuk:
@tlvu Thanks for letting me know about this. I just took note of what containers are running in our stack, so I'll add them to this list and test these changes.

birdhouse/components/cowbird/default.env
tlvu: Should we even keep this since "this dependency is only required if the mongo instance is the one provided in config/mongodb (include this for Cowbird<2.0.0)"

Is cowbird still <2.0.0?  Else this comment block not needed anymore?

birdhouse/components/geoserver/.gitignore
```diff
@@ -6,4 +6,3 @@ service-config.json
 geoserver_canarie_api_monitoring.py
 geoserver-magpie-provider.cfg
 config/proxy/canarie_api_monitoring.py
-config/magpie/providers.cfg
```
tlvu, line 9: Geoserver do not add to magpie providers anymore?  But should we still keep this old ignore to avoid breaking autodeploy?

birdhouse/components/proxy/default.env
tlvu: Add this comment about magpie as landing page is not the most user-friendly location to the var  `PROXY_ROOT_LOCATION` in `env.local.example` as well.

birdhouse/env.local.example
```diff
@@ -91,10 +102,6 @@ export GEOSERVER_ADMIN_PASSWORD=geoserverpass
 #    ./optional-components/wps-healthchecks
 #    ./optional-components/test-weaver
 #    ./optional-components/test-geoserver-secured-access
-#    ./components/scheduler
-#    ./components/monitoring
-#    ./components/weaver
-#    ./components/cowbird
```
tlvu, line 97: Oh the optional-components were before the components on purposes, to give the components higher precedence since the later item in the list can override configs of all previous items.

But now thinking about this, this means all the optional-components were already able to override all the default components.

I guess we can keep it like this but it feels a bit weird all the optional stuffs have higher override precedence than the "good" stuff.

birdhouse/components/cowbird/default.env
fmigneault: I wouldn't worry about `cowbird<2.0.0`. No reason to go back. Only add the relevant current dependencies.

tlvu: That's what I thought, no need to go back to older cowbird so no need to keep this comment block about adding mongodb as dependencies.

birdhouse/env.local.example
fmigneault: I'm not sure, it is as simple as that. @mishaschwartz can confirm, since he worked on it the most.
The "inspection" order follows the list, but each indicated component will then consider all their listed dependencies to load them as well before the component itself. So, components not listed or listed later could appear before the "currently inspected" component if they were marked as a dependency. 

To get a guaranteed override, one would need to set the variable value to override using single quotes and add the variable to `DELAYED_EVAL`, or define another custom component placed last in `EXTRA_CONF_DIRS` that overrides what is needed explicitly.

eyvorchuk (approved):
I just tested this PR on the PCIC stack and everything worked well. I took note of what `EXTRA_CONF_DIRS` will look like with this setup so that I can use that once this gets merged.

tlvu:
> I took note of what `EXTRA_CONF_DIRS` will look like with this setup so that I can use that once this gets merged.

@eyvorchuk You can add all changes to `EXTRA_CONF_DIRS` to your `env.local` immediately to avoid forgetting or having to time your change with the merge of this PR.

Like I said, `EXTRA_CONF_DIRS` will accept non-existing dirs.  This is the mechanism to allow for a smooth transition when components name changes.

eyvorchuk:
Ah, right, ok. I just added the changes to our `env.local`.

mishaschwartz:
@fmigneault @tlvu 

Now that https://github.com/bird-house/birdhouse-deploy/pull/386 and https://github.com/bird-house/birdhouse-deploy/pull/359 have been merged in to master. Is everyone ok if I proceed with updating this PR and getting it ready to merge in as well?

fmigneault:
@mishaschwartz 
Fine with me. Just need to resolve the conflicts to make sure I didn't forget a component reference for the CI.

birdhouse/components/geoserver/.gitignore
mishaschwartz: We need to remove this so that future changes to the providers.cfg file won't be ignored. It is no longer generated from a template file, it is part of the source code now.

birdhouse/env.local.example
mishaschwartz: > I guess we can keep it like this but it feels a bit weird all the optional stuffs have higher override precedence than the "good" stuff.

A lot of the optional components are not actually components, they're additional settings for existing components so we _want_ them to override the existing components.

> The "inspection" order follows the list, but each indicated component will then consider all their listed dependencies to load them as well before the component itself

Yeah that's a good summary. There's a detailed description of how this is enforced here: https://github.com/bird-house/birdhouse-deploy/blob/1.42.1/CHANGES.md?plain=1#L876-L962

The general recommendation is: load the optional components after the regular components since any overrides are usually desired.

tlvu (approved):
LGTM.  I assume an autodeploy test has been performed with the old list of components enabled in `env.local`?

I will pre-prepare our various `env.local` files in advance so once this PR is merged, our staging system will pick this up automatically.

fmigneault:
@tlvu 
The CI runs with this enabled: 
```shell
export EXTRA_CONF_DIRS="
    ./config/proxy
    ./config/magpie
    ./config/twitcher
    ./config/canarie-api
    ./config/geoserver
    ./config/finch
    ./config/raven
    ./config/hummingbird
    ./config/thredds
    ./config/portainer
    ./config/jupyterhub
    ./components/proxy
    ./components/magpie
    ./components/twitcher
    ./components/canarie-api
    ./components/geoserver
    ./components/finch
    ./components/raven
    ./components/hummingbird
    ./components/thredds
    ./components/portainer
    ./components/jupyterhub
    ./components/monitoring
    ./components/weaver
    ./components/cowbird
    ./components/stac
    ./optional-components/all-public-access
    ./optional-components/secure-data-proxy
    ./optional-components/stac-data-proxy
    ./optional-components/stac-populator
    ./optional-components/database-external-ports
    ./optional-components/test-weaver
    ./optional-components/test-cowbird-jupyter-access
    ./optional-components/test-geoserver-secured-access
    ./optional-components/x-robots-tag-header
"
```

fmigneault (approved):

fmigneault:
@mishaschwartz 
> basic documentation for the components

Great addition. More than "basic". Thanks !

Future updates should probably reorganize them slightly so they are sorted similarly to the components directory. 
A TOC would also be useful since there are quite a lot in there now.

PR #402, opened 2023-11-09
Jupyterhub: periodically check whether the logged-in user still have permission to access

mishaschwartz:
## Overview

By setting the `JUPYTERHUB_CRYPT_KEY` environment variable in the `env.local` file, jupyterhub will store user's authentication information (session cookie) in the database. This allows jupyterhub to periodically check whether the user still has permission to access jupyterhub (the session cookie is not expired and the permission have not changed).

The minimum duration between checks can be set with the `JUPYTERHUB_AUTHENTICATOR_REFRESH_AGE` variable which is an integer (in seconds).

Note that users who are already logged in to jupyterhub will need to log out and log in for these changes to take effect.

To forcibly log out all users currently logged in to jupyterhub you can run the following command to force the recreation of the cookie secret:

  ```shell
  docker exec jupyterhub rm /persist/jupyterhub_cookie_secret && docker restart jupyterhub
  ```

First discussed here: https://github.com/bird-house/birdhouse-deploy/pull/358#issuecomment-1737374614

## Changes

**Non-breaking changes**
- Adds two new environment variables to configure additional jupyterhub authentication
- New jupyterhub version pavics/jupyterhub:4.0.2-20231024

**Breaking changes**

## Related Issue / Discussion

Related to #334

- [x] Note that this PR requires https://github.com/Ouranosinc/jupyterhub/pull/23 to be merged in first and the jupyterhub version updated to match.

## Additional Information

birdhouse_skip_ci: false


birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -19,6 +19,11 @@ c.MagpieAuthenticator.magpie_url = "http://magpie:2001"
 c.MagpieAuthenticator.public_fqdn = "${PAVICS_FQDN_PUBLIC}"
 c.MagpieAuthenticator.authorization_url = "${JUPYTERHUB_AUTHENTICATOR_AUTHORIZATION_URL}"
 
+if os.getenv("JUPYTERHUB_CRYPT_KEY"):
+    c.MagpieAuthenticator.enable_auth_state = True
+    c.MagpieAuthenticator.refresh_pre_spawn = True
+    c.MagpieAuthenticator.auth_refresh_age = int("${JUPYTERHUB_AUTHENTICATOR_REFRESH_AGE}")
```
fmigneault, line 25: Might need some alignment with `MAGPIE_COOKIE_EXPIRE`.

birdhouse/config/jupyterhub/default.env
```diff
@@ -64,6 +64,15 @@ export JUPYTERHUB_CONFIG_OVERRIDE=""
...
 export JUPYTERHUB_AUTHENTICATOR_AUTHORIZATION_URL='http://twitcher:8000/ows/verify/jupyterhub'
 
+# 32 byte hex-encoded key used to encrypt a user's authentication state in the juptyerhub database.
+# If set, jupyterhub will periodically check if the user still has permission to access jupyterhub (according to Magpie)
+export JUPYTERHUB_CRYPT_KEY=
+
+# Jupyterhub will check if the current logged in user still has permission to access jupyterhub (according to Magpie)
+# if their authentication information is older that this value (in seconds). This value is only applied if
+# JUPYTERHUB_CRYPT_KEY is set.
+export JUPYTERHUB_AUTHENTICATOR_REFRESH_AGE=300
```
fmigneault, line 74: It looks like Jupyter checks its own session much more frequently (every second or so).
This doesn't need to be as often, but I think it could still be slightly decreased, like every minute.

birdhouse/config/jupyterhub/default.env
```diff
@@ -64,6 +64,15 @@ export JUPYTERHUB_CONFIG_OVERRIDE=""
 # recommended as it may permit unauthorized users from accessing jupyterhub.
 export JUPYTERHUB_AUTHENTICATOR_AUTHORIZATION_URL='http://twitcher:8000/ows/verify/jupyterhub'
 
+# 32 byte hex-encoded key used to encrypt a user's authentication state in the juptyerhub database.
+# If set, jupyterhub will periodically check if the user still has permission to access jupyterhub (according to Magpie)
+export JUPYTERHUB_CRYPT_KEY=
```
fmigneault, line 69: What happens if the value is changed between restarts, causing a different state to be obtained?
Does Jupyter perform a new check with Magpie to verify if an updated value works, and update the state transparently with the new encryption key?
In other words, once this is set, must it remain the same, or we can automatically regenerate it each time?

birdhouse/config/jupyterhub/default.env
```diff
@@ -5,7 +5,7 @@
 # are applied and must be added to the list of DELAYED_EVAL.
 
 export JUPYTERHUB_DOCKER=pavics/jupyterhub
-export JUPYTERHUB_VERSION=4.0.2-20231002
+export JUPYTERHUB_VERSION=4.0.2-20231024
```
tlvu, line 8: I am guessing the `JUPYTERHUB_VERSION` will change again after https://github.com/Ouranosinc/jupyterhub/pull/23 is merged an a new build is triggered on DockerHub?

tlvu (approved):
Approuve the spirit of this PR.

Please test for upgrade scenario of existing production deployments where multiples users are already logged in, already have their Jupyter session opened and potentially running some calculations.

Do they have to re-login or this is completely transparent to them?  If they have to re-login, please note that in the CHANGES.md and I will have to notify our users before rolling this change out.

birdhouse/config/jupyterhub/default.env
mishaschwartz: Yeah, I'll update it shortly.

birdhouse/config/jupyterhub/default.env
mishaschwartz: This is the default that jupyterhub uses (https://jupyterhub.readthedocs.io/en/latest/reference/api/auth.html#jupyterhub.auth.Authenticator.auth_refresh_age)

I'm happy to shorten it to one minute. I'll update that shortly

birdhouse/config/jupyterhub/jupyterhub_config.py.template
fmigneault: In case the value is set higher here than the actual cookie age. It would give the false feedback that the cookie/session is still valid until the next refresh happens, while Magpie would actually log the user out. The refresh rate should happen at least a few times per cookie-age/refresh cycle to provide a relevant status.

mishaschwartz:
@tlvu 

> Do they have to re-login or this is completely transparent to them? If they have to re-login, please note that in the CHANGES.md and I will have to notify our users before rolling this change out.

Yes, it is transparent to them. This feature will only start applying once the user has logged out and logged back in again.

If a user is already logged in, they will not have any authentication information added to the database yet, so the check that was added here (https://github.com/Ouranosinc/jupyterhub/pull/23) will allow them to remain logged in.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
mishaschwartz: Ok, I understand.

Since the default MAGPIE_COOKIE_EXPIRE value is 24 hours and this checks every minute now. That should be more that sufficient to align the values

tlvu:
> If a user is already logged in, they will not have any authentication information added to the database yet, so the check that was added here ([Ouranosinc/jupyterhub#23](https://github.com/Ouranosinc/jupyterhub/pull/23)) will allow them to remain logged in.

@mishaschwartz 

Very nice to ensure transparent upgrade.

Please add a note to the CHANGES.md that each existing Jupyter user have to logout and re-login for this new change to take effect.

This is to ensure node admin is not falsely thinking the security enhancement is effective simply with the new deployment.

For us, we probably will kill all user sessions somewhere during Xmas holidays so everyone is forced to re-login so we are sure this will take effect.

One way I found to force everyone to re-login again is to `./pavics-compose.sh stop jupyterhub`, then `rm /var/lib/docker/volumes/jupyterhub_data_persistence/_data/jupyterhub.sqlite`, then `./pavics-compose.sh start jupyterhub`.  If you have another way to force everyone to re-login, please let me know.  This way of going behind JupyterHub back is not very elegant.

Please also document in CHANGES.md one way to force everyone to re-login.

fmigneault (approved):

tlvu:
> @tlvu
> 
> Ok I've updated the CHANGES.md file with the relevant information. Please have a look to see if that makes sense

Good me for me, thanks.

tlvu (approved):

PR #403, opened 2023-11-15
add `optional-components/stac-data-proxy`

fmigneault:
## Overview

Provide a way to host local data that STAC API can refer to for use/download.

Currently, any STAC Asset that is referenced within responses by STAC-API Collections/Items must either be already hosted by another service of the stack (eg: CMIP6 netCDF in THREDDS), or point at some other external resource not on the server. 

Instead of having a custom config and mount point for each node, this optional component defines a standard way to define it.

## Changes

**Non-breaking changes**

- `optional-components/stac-data-proxy`: add a new feature to allow hosting of local STAC assets.

  The new component defines variables `STAC_DATA_PROXY_DIR_PATH` (default `${DATA_PERSIST_ROOT}/stac-data`) and
  `STAC_DATA_PROXY_URL_PATH` (default `/data/stac`) that are aliased (mapped) under `nginx` to provide a URL
  where locally hosted STAC assets can be downloaded from. This allows a server node to be a proper data provider,
  where its STAC-API can return Catalog, Collection and Item definitions that points at these local assets available
  through the `STAC_DATA_PROXY_URL_PATH` endpoint.

  When enabled, this component can be combined with `optional-components/secure-data-proxy` to allow per-resource
  access control of the contents under `STAC_DATA_PROXY_DIR_PATH` by setting relevant Magpie permissions under service
  `secure-data-proxy` for children resources that correspond to `STAC_DATA_PROXY_URL_PATH`. Otherwise, the path and
  all of its contents are publicly available, in the same fashion that WPS outputs are managed without
  `optional-components/secure-data-proxy`. 

  More details are provided in https://github.com/bird-house/birdhouse-deploy/blob/stac-data-proxy/birdhouse/optional-components/README.rst#provide-a-proxy-for-local-stac-asset-hosting

**Breaking changes**
- n/a

## Related Issue / Discussion

- Relates to https://github.com/crim-ca/stac-populator/pull/31
- Relates to contents in https://github.com/ai-extensions/stac-data-loader/tree/main/data/EuroSAT/stac
- Relates to https://github.com/ai-extensions/stac-data-loader/blob/main/notebooks/stac_eurosat.ipynb

STAC metadata generated from above notebook (see subset for example), will be able to use a location such as `https://${PAVICS_FQDN_PUBLIC}${STAC_DATA_PROXY_URL_PATH}/EuroSAT/...` instead of the temporary raw-GitHub content URLs. The STAC populator (with `DirectoryLoading` implementation), will be able to push the STAC Collection/Items toward that instances. The STAC Assets that they refer to will be placed under `${STAC_DATA_PROXY_DIR_PATH}/EuroSAT` to make them accessible externally.


fmigneault:
@ahandan @ChaamC @francisPLT @Nazim-crim 
Feel free to review or not. Just giving you the pointer as FYI for future geo projects.

fmigneault:
run tests

fmigneault:
@tlvu 
Friendly reminder regarding this PR. 

If everything looks good, that will help me synchronize merges with other STAC-related PRs to avoid conflicts.
- https://github.com/bird-house/birdhouse-deploy/pull/405
- #386

dchandan:
I am not able to see how this feature meshes with the functionality for controlling the mount points of THREDDS data, and support for multiple THREDDS catalog that was recently provided in #359. The variables `STAC_DATA_PROXY_DIR_PATH` and `STAC_DATA_PROXY_URL_PATH` appear to be doing something similar to what is done by `THREDDS_DATASET_LOCATION_ON_HOST`, `THREDDS_SERVICE_DATA_LOCATION_ON_HOST`, `THREDDS_DATASET_URL_PATH` and `THREDDS_SERVICE_DATA_URL_PATH` variables from #359. From my limited understanding so far, I get the feeling that this is meant to support access to data that is not stored on THREDDS? Could you provide some use cases that will further illustrate the use case?

fmigneault:
@dchandan  
Indeed. It is for other data not hosted through THREDDS. 

An example would be the GeoTiff, GeoJSON and PNG asset files under https://github.com/ai-extensions/stac-data-loader/tree/main/data/EuroSAT that are referenced by the STAC Collection/Items. At the moment, it is already possible to push those STAC definitions (eg: https://hirondelle.crim.ca/stac/collections/EuroSAT-subset-train), but the assets they refer to must make use of an external host source (temporarily using GitHub raw content endpoint for now, which is not optional). If the node is a data provider, it should have a way to host its own data without necessarily relying on an external hosting service.

dchandan:
I see. But then why not put that data on THREDDS on the node itself? In your example, it would mean putting the asset files from EuroSAT onto hirondelle's THREDDS catalog.

dchandan:
Yeah, I can see that. Although any data can be served through THREDDS, it does do better with climate-date related formats. But then this opens up a more broader question of if (and I am in favour of doing this) and how we should provide a more comprehensive file server based on nginx. If this essentially and completely achieves that, then I think we should rename it to something more general rather than `stac-data-proxy`.

tlvu (approved):
LGTM.  I have not test deployed this.  Please ensure it does not break the autodeploy.

See other minor comments.

dchandan:
Maybe something like `birdhouse-fileserver-proxy`?

tlvu:
> we should provide a more comprehensive file server based on nginx. If this PR essentially and completely achieves that, then I think we should rename it to something more general rather than `stac-data-proxy`.

> Maybe something like `birdhouse-fileserver-proxy`?

I sort of agree with @dchandan here.  Is this new URL and location on disk meant only for STAC data?  From the implementation, it looks pretty generic so it could be used for all sort of static data, not just STAC data.  

So, in the future, if we need for "some other kind of data" we won't need to again change the config, or to go through PRs review again.

It's like the new ability to add any new numbers of Thedds collections, there is no need to modify the code anymore.  So more flexibility and less "paperwork time" in PR reviews.

This stack will hopefully be deployed by many different organisations with very different needs, we should care for flexibility to help meet their needs.

fmigneault:
@dchandan @tlvu 

>  how we should provide a more comprehensive file server based on nginx. If this PR essentially and completely achieves that, then I think we should rename it to something more general rather than stac-data-proxy.

In a way, there is nothing more that needs to be done than adding a custom `alias` to the nginx configs to have as many mount points as desired. Even using secured accessed does not need any additional service in Magpie (as this PR shows, `secure-data-proxy` is simply reused).

The issue with combining everything under a common location/directory/component is that it can limit configuration mounts of the server. For now, only a simple Nginx `alias` is defined, but it could get much more complicated than that to control specific min/max contents-length/type and HTTP methods, and would I prefer to have a dedicated nginx configuration for a controlled use case than many variables for each use case. In the future, I would prefer to avoid the kind of defaults as below because of 1 service's use case / limitations. It is better to leave the option to mount each service's data individually for maximum flexibility. 

https://github.com/bird-house/birdhouse-deploy/blob/cc8dc1f7c1e62a8b8d7892f1c1714363fd002cd3/birdhouse/config/proxy/nginx.conf.template#L27-L34

Another reason that I named the component `stac-data-proxy` is due to the importance of STAC compared to other generic data endpoints (assuming `stac` is enabled by default following @mishaschwartz PR). For the moment, permissions are somewhat generic, but I foresee they could become much more STAC-specific in follow-up PRs.

birdhouse/optional-components/stac-data-proxy/config/magpie/permissions.cfg.template
```diff
@@ -0,0 +1,12 @@
+# NOTE:
+# Assume 'secure-data-proxy' would exist if needed.
```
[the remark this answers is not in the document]
fmigneault: I do not enforce it because `stac-data-proxy` can be used in 2 modes: 
- fully-open (similar to wps-outputs), in which case `secure-data-proxy` is irrelevant (since not enabled)
- secured access, when `secure-data-proxy` is enabled, and therefore created by that component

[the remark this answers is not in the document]
fmigneault: All documented here: 
https://github.com/bird-house/birdhouse-deploy/blob/cc8dc1f7c1e62a8b8d7892f1c1714363fd002cd3/birdhouse/optional-components/README.rst#populate-stac-catalog-with-sample-data

Important: look at the file in raw mode, GitHub somehow does not render the `list-table` RST directive :(

birdhouse/optional-components/README.rst
```diff
@@ -326,6 +326,51 @@ To enable this optional-component:
...
+- Edit ``env.local`` (a copy of `env.local.example`_)
+- Add ``./optional-components/stac-data-proxy`` to ``EXTRA_CONF_DIRS``.
+- Optionally, add any other relevant components to control access as desired (see below).
+
+When using this component, access to the endpoint defined by ``STAC_DATA_PROXY_URL_PATH``, and therefore all
+corresponding files contained under mapped ``STAC_DATA_PROXY_DIR_PATH`` will depend on how this
+feature is combined with ``./optional-components/stac-public-access`` and ``./optional-components/secure-data-proxy``.
+Following are the possible combinations and obtained behaviors:
+
+.. list-table::
```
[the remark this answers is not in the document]
fmigneault: Just noticed that two. I don't know why because that one is: 
https://github.com/bird-house/birdhouse-deploy/blob/master/README.rst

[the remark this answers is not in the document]
fmigneault: Figured it out. See https://github.com/bird-house/birdhouse-deploy/pull/403/commits/812e42ddfd0b073c1afc628c5134ff911eae1a50?short_path=d1bd7c6#diff-d1bd7c60b458436b881e8172ca76ae01f0e5a868e9815b1b9faaa0089aec9871

birdhouse/optional-components/stac-data-proxy/config/magpie/permissions.cfg.template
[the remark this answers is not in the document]
mishaschwartz: > Should I place this config under birdhouse/optional-components/stac-data-proxy/config/secure-data-proxy instead of birdhouse/optional-components/stac-data-proxy/config/magpie to make it effective only when both are combined? secure-data-proxy already depends on magpie

Yes I think that makes more sense

mishaschwartz:
I think that it is misleading to name this component "stac" since it really doesn't have anything to do with STAC (it just serves files that _could_ be referenced by a STAC catalog but doesn't need to be at all).

I think that we should use the `birdhouse-fileserver-proxy` name instead.

> It is better to leave the option to mount each service's data individually for maximum flexibility.

I agree with this.

However, having to make a new optional component every single time we need a new mountpoint is going to be very confusing. It would be better to make this component configurable so that it can support multiple mountpoints with different nginx location definitions if needed.

fmigneault:
> I think that it is misleading to name this component "stac" since it really doesn't have anything to do with STAC (it just serves files that could be referenced by a STAC catalog but doesn't need to be at all).

Component wise, this is true for now with minimal nginx config. However, I believe STAC has an important role being an eventual default component that could justify it. As shown by the component combination table [here](https://github.com/bird-house/birdhouse-deploy/commit/812e42ddfd0b073c1afc628c5134ff911eae1a50?short_path=d1bd7c6#diff-d1bd7c60b458436b881e8172ca76ae01f0e5a868e9815b1b9faaa0089aec9871), the multiple flavors in which STAC can be configured is already somewhat complex depending on use cases the node what to support. I think we can help new node manager configure their server more efficiently and quickly by providing explicit references to STAC rather than more extra variables to configure.

> It would be better to make this component configurable so that it can support multiple mountpoints with different nginx location definitions if needed.

Wouldn't that mean that each of those mount point and nginx configuration will require to define a new component anyway to apply relevant `default.env` and docker-compose volume mounts of (service-specific) nginx configs? So instead of having each component managing their own alias, data mount and nginx-specific optimizations, we would have a `birdhouse-fileserver-proxy` dependency with a generic location which the sole purpose is to be overridden by other components variables? I don't think that makes it easier. 

Just to make sure my point is understood properly. The idea of having a separate nginx config and component for STAC data specifically is that any relevant optimization for accessing STAC data could be gradually defined for that path. Other optimizations or redirections could be applied as we investigate further what works best for STAC and how users interact with it. This is not the same as the case for THREDDS where all configurable datasets go through THREDDS. Since `thredds` service becomes the "bottleneck" in that case, it indeed makes more sense to make "pluggable configuration" with variables. Nginx's definitions however are directly the access point of the server. It is possible to provide much more fine-grained control of each endpoint. Because of this, I believe a new user should have something better than a generic `BIRDHOUSE_FILESERVER_NGINX_EXTRA_CONF` variable. My idea is to provide a predefined set of configurations optimized specifically for STAC (admittedly not much for now, but got to start somewhere), without asking each new node admin to try to optimize it themselves. 

I think `birdhouse-fileserver-proxy` is a good idea on its own, and could be provided as well, but the purpose of this PR is not to have a generic data location. Note also that providing an extra data access point in nginx is as simple as adding: 
```nginx
location /more-custom-data/ {
    alias /data/more-custom-data/;
}
```
Therefore, a node admin could already very easily add more data access points as needed without the need of an additional component. They are also free to choose if the (eventually) predefined optimization that would be provided by `stac-data-proxy` are relevant for them or not. It is an optional-component after all.

mishaschwartz:
Ok I'm fine with that in principle. I'll have a look in detail and leave a review shortly

birdhouse/optional-components/stac-data-proxy/config/magpie/permissions.cfg.template
mishaschwartz, line 2: This line is confusing. Do you mean: "The permissions in this file only take effect if the secure-data-proxy component is enabled" ?

birdhouse/optional-components/stac-data-proxy/config/magpie/permissions.cfg.template
```diff
@@ -0,0 +1,12 @@
...
+# Since 'secure-data-proxy' < 'stac-data-proxy', it should be loaded first.
```
mishaschwartz, line 3: The use of `<` here is confusing. Do you mean that one depends on the other? One overrides variable definitions of the other one?

birdhouse/optional-components/stac-data-proxy/default.env
```diff
@@ -0,0 +1,42 @@
...
+EXTRA_VARS='
+  $STAC_DATA_PROXY_DIR_PATH
+  $STAC_DATA_PROXY_URL_PATH
+'
+
+# extend the original 'VARS' from 'birdhouse/pavics-compose.sh' to employ them for template substitution
+# adding them to 'VARS', they will also be validated in case of override of 'default.env' using 'env.local'
+VARS="$VARS $EXTRA_VARS"
+
+# Direcotry path that will be used as volume mount for storing hosted STAC assets data
```
mishaschwartz, line 20: Directory

birdhouse/optional-components/stac-public-access/config/stac-data-proxy/permissions.cfg.template
```diff
@@ -0,0 +1,10 @@
+# NOTE:
+# Assume 'secure-data-proxy' would exist if needed.
+# Since 'secure-data-proxy' < 'stac-data-proxy-public', it should be loaded first.
```
mishaschwartz, line 3: Do you mean 'stac-public-access' not 'stac-data-proxy-public'

mishaschwartz (commented):
I just have a few comments about the documentation and wording of this. 

Other than that and the change suggested in this other comment (https://github.com/bird-house/birdhouse-deploy/pull/403/files#r1407965025) I think this looks good

tlvu:
> For the moment, permissions are somewhat generic, but I foresee they could become much more STAC-specific in follow-up PRs.

@fmigneault I wasn't aware the following up PRs that could be STAC-specific.  In this case, I agree to make this new component STAC-specific.

Also I agree it is fairly easy to add new Nginx config to handle new URL so the value to have a generic component for static file is less "valuable".

dchandan (approved):

birdhouse/optional-components/stac-public-access/config/stac-data-proxy/permissions.cfg.template
fmigneault: It is `stac-data-proxy-public` since this is the mapped named in `birdhouse/optional-components/stac-public-access/config/stac-data-proxy/docker-compose-extra.yml` that Magpie will see when listing files.

birdhouse/optional-components/stac-data-proxy/config/magpie/permissions.cfg.template
fmigneault: Sorting order by file name. Literally `__lt__` used by `sorted()`.

fmigneault: I've updated the description to be more explicit. Let me know if it is sufficient.

mishaschwartz (approved):
Looks good!

You mentioned that you could add some STAC specific settings later on in a different PR. No need to do that now but could you make an issue that outlines some of your ideas so that we don't lose track of them.

fmigneault:
run tests

Issue #47, opened 2020-05-05
Documentation: explain how to run tests and get result outputs


huard:
I suggest we recommend using owslib objects to parse the response xml and get the outputs. Access to objects on the local file system by Outputs.retrieveData will work in the upcoming version of owslib.

tlvu:
See PR https://github.com/geopython/OWSLib/pull/681

PR #404, opened 2023-11-15
Cowbird jupyter e2e test

ChaamC:
## Changes
- New optional-component `test-cowbird-jupyter-access` that executes a script to set up a test user along with different
  test files. This component is used for the related e2e test from the 
  [PAVICS-e2e-workflow-tests](https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests) repo.

## Fixes
- Updates incorrect WPS outputs resource name in the cowbird config.

## Additional Information

- Related to PR on `PAVICS-e2e-workflow-tests` : [#131](https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/131)
- Working test on Jenkins ([Build #26](http://daccs-jenkins.crim.ca/job/PAVICS-e2e-workflow-tests/job/cowbird-jupyter-e2e-test/26/)), but I have to manually clone the cowbird branch with the upcoming changes on the iac instance for now for the test to work(see PR [#47](https://github.com/Ouranosinc/cowbird/pull/47))


birdhouse/optional-components/test-cowbird-jupyter-access/test_cowbird_jupyter.py
```diff
@@ -0,0 +1,188 @@
...
+    for target_path, perms in [(os.path.join(f"{user_workspace_dir}/shapefile_datastore", f"{stem}_ro{ext}"), 0o664),
+                               (os.path.join(f"{user_workspace_dir}/shapefile_datastore", f"{stem}_rw{ext}"), 0o666)]:
+        shutil.copy2(os.path.join(GEOSERVER_TEST_DATA_DIR, filename), target_path)
+        os.chmod(target_path, perms)
+
+# ------------------------------------------------------------------
+# Add wps-outputs
+print("Creating test WPS outputs data...")
+# Add public file
+public_wpsoutputs_filepath = f"{WPS_OUTPUTS_DIR}/weaver/public/test_public_file.txt"
```
fmigneault, line 142: Now I see the source of the path from the other PR https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/131#discussion_r1394735756. 

Please adjust it just to demonstrate what would be really expected, so it represents a real use case.
Example: 
```
/data/wps_outputs
└── weaver
    ├── 0fb1371d-48bb-4c59-9939-4ec6f02954b5.log
    ├── 0fb1371d-48bb-4c59-9939-4ec6f02954b5.xml
    ├── 63de2ff8-2ee5-40d0-a254-fdb493a35d77.log
    ├── 63de2ff8-2ee5-40d0-a254-fdb493a35d77.xml
    ├── 950776da-0d99-4607-a68a-c743493b56e5.log
    ├── 950776da-0d99-4607-a68a-c743493b56e5.xml
    ├── public
    │   ├── 63de2ff8-2ee5-40d0-a254-fdb493a35d77
    │   │   └── stdout.log
    │   ├── 3dc704a8-e1f6-4fe8-8bf0-153cf1e52c7a
    │   │   └── output
    │   │       └── nc_dump_jjEIOJ.txt
    │   ├── 3dc704a8-e1f6-4fe8-8bf0-153cf1e52c7a.log
    │   └── 3dc704a8-e1f6-4fe8-8bf0-153cf1e52c7a.xml
    └── users
        ├── 1
        │   ├── 0fb1371d-48bb-4c59-9939-4ec6f02954b5
        │   │   └── nc_dump_lko95z.txt
        │   └── 950776da-0d99-4607-a68a-c743493b56e5
        │       └── stdout.log
        └── 18
            ├── 075ed2a9-7b79-4444-99a2-840092e0a26b
            │   └── output
            │       └── nc_dump_s8zoWZ.txt
            ├── 075ed2a9-7b79-4444-99a2-840092e0a26b.log
            └── 075ed2a9-7b79-4444-99a2-840092e0a26b.xml
```

fmigneault (commented):
Missing a new section in https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/optional-components/README.rst

This section should warn very explicitly that this component should not be used in non-test environments, as it opens public access for certain endpoints, defines admin-tokens for a JupyterHub user for which credentials are clearly visible in the script, and enforces use of root access for the test container. The component is for validation only. If used in a prod stack, it creates a security vulnerability.

birdhouse/optional-components/test-cowbird-jupyter-access/test_cowbird_jupyter.py
```diff
@@ -0,0 +1,188 @@
...
+    value = os.getenv(var_name)
+    if not value:
+        raise ValueError("Missing test admin credentials `{}` to run tests.".format(var_name))
+    return value
+
+TEST_MAGPIE_ADMIN_USERNAME = get_credentials("MAGPIE_ADMIN_USERNAME")
+TEST_MAGPIE_ADMIN_PASSWORD = get_credentials("MAGPIE_ADMIN_PASSWORD")
+
+TEST_USER = "testcowbirdjupyter"
+TEST_PASSWORD = "qwertyqwerty"
```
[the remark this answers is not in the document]
ChaamC: I can transfer them to the `default.env` of the optional-component.

> The CI test daccs-config could use a testcowbirdjupyter-{uuid} variant with a randomly generated password.

Although the idea is good, I am not sure how easy it is to do considering the notebook test would have to find back those values. They can be passed through JupyterHub to the 2nd notebook that is executed on the JupyterLab container started by JuptyerHub, but it's the main notebook that requires those values.

Was there an easy way to passed variables from birdhouse to the test container started by jenkins?

birdhouse/components/cowbird/config/cowbird/config.yml.template
```diff
@@ -34,7 +34,7 @@ handlers:
     jupyterhub_user_data_dir: ${JUPYTERHUB_USER_DATA_DIR}
     wps_outputs_dir: ${WPS_OUTPUTS_DIR}
     secure_data_proxy_name: ${SECURE_DATA_PROXY_NAME}
-    # wps_outputs_res_name: ${WPS_OUTPUTS_RES_NAME}
+    wps_outputs_res_name: wpsoutputs
```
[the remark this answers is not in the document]
tlvu: Just curious, did you forget to push the change to not hardcode `wps_outputs_res_name: wpsoutputs` before resolving this comment?  I see it is still hardcoded.

birdhouse/optional-components/test-cowbird-jupyter-access/default.env
```diff
@@ -0,0 +1,62 @@
...
+"
+
+# add any component that this component requires to run
+COMPONENT_DEPENDENCIES="
+  ./config/cowbird
+  ./config/geoserver
+  ./config/jupyterhub
+  ./config/magpie
+  ./optional-components/all-public-access
+  ./optional-components/secure-data-proxy
```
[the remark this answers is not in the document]
ChaamC: I hesitated to include it, because the only thing it actually uses is the test data, but it doesn't really use the actual component for the secured access.
I can still add it, since it relies on it partially.

birdhouse/optional-components/test-cowbird-jupyter-access/default.env
```diff
@@ -0,0 +1,62 @@
...
+'
+# extend the original 'VARS' from 'birdhouse/pavics-compose.sh' to employ them for template substitution
+# adding them to 'VARS', they will also be validated in case of override of 'default.env' using 'env.local'
+VARS="$VARS $EXTRA_VARS"
+
+THIS_FILE="`realpath "$0"`"
+export COWBIRD_JUPYTER_ACCESS_DIR="`dirname "$THIS_FILE"`/optional-components/test-cowbird-jupyter-access"
+
+export JUPYTERHUB_CONFIG_OVERRIDE="
+c.JupyterHub.services = [
```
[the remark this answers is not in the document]
tlvu: Same here, you probably should resolve the comment only after having pushed the corresponding fix.

birdhouse/optional-components/test-cowbird-jupyter-access/default.env
```diff
@@ -0,0 +1,62 @@
...
+    }
+]
+
+# Specify a fixed image name/version for the test
+c.DockerSpawner.image_whitelist = {'pavics-workflow-test:210216': 'pavics/workflow-tests:210216'}
+
+c.DockerSpawner.environment['TEST_USER'] = 'testcowbirdjupyter'
+c.DockerSpawner.environment['JUPYTERHUB_USER_DATA_DIR'] = os.environ['JUPYTERHUB_USER_DATA_DIR']
+
+notebook_dir = '/notebook_dir'
```
tlvu, line 49: `notebook_dir = '/notebook_dir'` this is already in the `jupyterhub_config.py.template`, why define it again?

birdhouse/optional-components/test-cowbird-jupyter-access/test_cowbird_jupyter.py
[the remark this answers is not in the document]
fmigneault: > I can transfer them to the `default.env` of the optional-component.

Yes, that is good.

>  I am not sure how easy it is to do considering the notebook test would have to find back those values.

You're right, that could be tricky. 
To close this faster, let's keep using the default. 
The CI or notebook could be adjusted manually with variable overrides for a quick test as needed.

birdhouse/optional-components/test-cowbird-jupyter-access/default.env
```diff
@@ -0,0 +1,62 @@
...
+c.DockerSpawner.environment['NOTEBOOK_DIR'] = notebook_dir
+c.DockerSpawner.environment['WORKSPACE_DIR'] = notebook_dir + '/writable-workspace'
```
tlvu, line 51: Same for this `/writable-workspace` path.  It is probably already defined in some existing variable.  If not, I'd rather you define a new variable in the `jupyterhub_config.py.template` file, then use it here.

We should not redefine the path here because if a change is made in the  `jupyterhub_config.py.template` file, you will not know about it here.

birdhouse/optional-components/test-cowbird-jupyter-access/docker-compose-extra.yml
```diff
@@ -0,0 +1,22 @@
+version: "3.4"
+services:
+  test-cowbird-jupyter-access:
+    image: pavics/workflow-tests:210216
```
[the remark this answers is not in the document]
ChaamC: I had just copied the image version that was found in the `env.local.example`, which was 210216, but it would make more sense to get the latest. It shouldn't be a problem, I will update this.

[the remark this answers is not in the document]
fmigneault: That is my intention (using the one tagged by birdhouse-deploy).
I think this was applied because our DACCS-config for the Jenkins CI running the PRs used to hardcode `pavics/workflow-tests:230601` with extra EO and NLP images from https://github.com/crim-ca/pavics-jupyter-images.
I pushed a fix yesterday that applies the upstream version from birdhouse-deploy to always apply what is currently active.

birdhouse/optional-components/test-cowbird-jupyter-access/test_cowbird_jupyter.py
```diff
@@ -0,0 +1,188 @@
+"""
+Script used to setup a test user along with different test files, for the test notebooks
+"test_cowbird_jupyter.ipynb" and "user_test_cowbird_jupyter.ipynb" found in the `PAVICS-e2e-workflow-test` repo.
+"""
+
+import json
+import os
+import requests
```
tlvu, line 8: This script basically only need to make http requests, I think any `pavics/workflow-tests` is fine, no need for the specific version `210216`.

And is this script a run once script or it has to continuously run and listen to connections?

If not, can we make it called only once by https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/scripts/bootstrap-instance-for-testsuite instead?

fmigneault: Run once and forget. 
However, the bootstrap script is not called by our Jenkins CI tests.
Only the PAVICS-E2E `testall` for running the notebooks are called on the instance swanned with this component added.

birdhouse/components/cowbird/config/cowbird/config.yml.template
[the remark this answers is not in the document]
ChaamC: Sorry, I did resolved the comment before pushing. Changes should now be pushed.

birdhouse/optional-components/test-cowbird-jupyter-access/default.env
```diff
@@ -0,0 +1,62 @@
...
+        ],
+        'services': [
+            # assign the service the above permissions
+            'service-admin',
+        ],
+    }
+]
+
+# Specify a fixed image name/version for the test
+c.DockerSpawner.image_whitelist = {'pavics-workflow-test:210216': 'pavics/workflow-tests:210216'}
```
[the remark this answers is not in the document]
ChaamC: I ended up just deleting this setting override. 
I had specified it initially to be sure it was in `pavics-workflow-test` , but I don't think it is needed after all, since the test doesn't rely on any special packages outside of python, and the features that are tested should not be different between the various images.

birdhouse/optional-components/test-cowbird-jupyter-access/docker-compose-extra.yml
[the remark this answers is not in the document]
ChaamC: Changed to use the latest image from the config.

birdhouse/optional-components/test-cowbird-jupyter-access/default.env
ChaamC: You're right. I removed both `notebook_dir` and `writable-workspace` redefinition here to use the default ones from the template.

ChaamC:
@tlvu @fmigneault 
Ready for a new review. See [comment](https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/131#issuecomment-1821738976) on the other PR.

birdhouse/optional-components/test-cowbird-jupyter-access/default.env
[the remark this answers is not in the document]
fmigneault: I doubt these components would change any time soon, but just in case that things are moved around or modified, that could help figure out the reference by the new cowbird test. 
If adding it does not break the test itself, I think it is better to include it while we are at it.

fmigneault (approved):
Looks good to me. Just check the answer from the previous review.

birdhouse/optional-components/test-cowbird-jupyter-access/test_cowbird_jupyter.py
tlvu: > However, the bootstrap script is not called by our Jenkins CI tests.

I am very surprised the boostrap script is called call by your Jenkins CI tests.  That boostrap entrypoint was specifically created for CI pipeline to call, so that new boostrap steps can be added transparently, without having to change the CI steps.

tlvu: The bootstrap entrypoint is called.

http://daccs-jenkins.crim.ca/job/DACCS-iac-birdhouse/2255/console
[RESULT] bootstrap-instance-for-testsuite script finished. Exiting.

tlvu (changes requested):
Since the boostrap entrypoint is called, see

http://daccs-jenkins.crim.ca/job/DACCS-iac-birdhouse/2255/console
`[RESULT] bootstrap-instance-for-testsuite script finished. Exiting.`

Please make it a new step in that entrypoint (a new script called by that entrypoint) instead of a new service just for a run-once and forget step.

fmigneault:
@tlvu 
> the boostrap entrypoint is called

My bad. Seems I was wrong.

However, since the optional component doesn't only create permissions and test files, but also defines specific DockerSpawner configuration overrides needed to test the feature, I think a component would be required no matter what to integrate with other `default.env` logic before the server is started. It would be more confusing to have part of the definitions under a bootstrap script and the rest under the component. IMO it is better to keep them together and togglable by a single `EXTRA_CONF_DIRS` entry than split their logic between the bootstrap script and the component.

birdhouse/optional-components/test-cowbird-jupyter-access/default.env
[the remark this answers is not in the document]
ChaamC: Already added. The test was successful, so we know it doesn't break the test now. :)

ChaamC:
@tlvu If I can time my other [PR](https://github.com/bird-house/birdhouse-deploy/pull/406) for the readme with this one, would it be alright to merge one of them on master first, and only apply the bump version on the 2nd one before merging, to have them under the same version number? Or do you prefer 2 distinct versions?

tlvu:
Same version is fine.

birdhouse/optional-components/test-cowbird-jupyter-access/docker-compose-extra.yml
```diff
@@ -0,0 +1,27 @@
+version: "3.4"
+services:
+  jupyterhub:
+    environment:
+      TEST_COWBIRD_JUPYTERHUB_USERNAME: ${TEST_COWBIRD_JUPYTERHUB_USERNAME}
```
tlvu, line 5: So this is the only change to other service (`jupyterhub`) than add the new service (`test-cowbird-jupyter-access`) just to be able to run the script once.

The `jupyterhub` service modification is not even necessary for Jenkins since Jenkins do not run on the PAVICS  platform.

So I do not see why this bootstrap script can not be called by the bootstrap entrypoint instead of being in a new component that is not really a service.

Now I am a bit lost as to why Jenkins works properly since none of all the extra changes to `jupyterhub` component is ever taken by Jenkins.  Aka, all the extra configs in `JUPYTERHUB_CONFIG_OVERRIDE` is not seen at all by Jenkins.

birdhouse/optional-components/test-cowbird-jupyter-access/docker-compose-extra.yml
```diff
@@ -0,0 +1,27 @@
...
+      MAGPIE_ADMIN_PASSWORD: ${MAGPIE_ADMIN_PASSWORD}
+      PAVICS_HOST_URL: https://${PAVICS_FQDN_PUBLIC}
+      WPS_OUTPUTS_DIR: ${WPS_OUTPUTS_DIR}
+      WORKSPACE_DIR: ${DATA_PERSIST_SHARED_ROOT}/${USER_WORKSPACES}
+      TEST_COWBIRD_JUPYTERHUB_USERNAME: ${TEST_COWBIRD_JUPYTERHUB_USERNAME}
+      TEST_COWBIRD_JUPYTERHUB_PASSWORD: ${TEST_COWBIRD_JUPYTERHUB_PASSWORD}
+    volumes:
+      - ${COWBIRD_JUPYTER_ACCESS_DIR}/test_cowbird_jupyter.py:/test_cowbird_jupyter.py
+      - ./optional-components/test-geoserver-secured-access/test-data:/geoserver-test-data:ro
+      - ${DATA_PERSIST_ROOT}:/data
```
tlvu, line 20: Shouldn't this be `${DATA_PERSIST_SHARED_ROOT}` because giving write access to the entire `${DATA_PERSIST_ROOT}` was way too much and that's the reason for  `${DATA_PERSIST_SHARED_ROOT}` to exist.

tlvu (changes requested):
I am actually still not convinced you need a new component for this, especially when the script is of bootstrap style (fire one and forget, not continuously listening to connection as a real service should).

birdhouse/optional-components/test-cowbird-jupyter-access/docker-compose-extra.yml
tlvu: > Jenkins do not run on the PAVICS platform.

Just to clarify on this point, Jenkins do test against the PAVICS platform, but the Jupyter env runtime environment that the test notebooks runs on Jenkins is not spawned by the JupyterHub component, so any changes to the JupyterHub component is not seen/tested by Jenkins.

So if the test notebook works under Jenkins, it means the changes to the JupyterHub component might not be required.  Maybe the new bootstrap script is all that is required, no need for JupyterHub config extra config changes.

Please either review the notebook or the necessity of JupyterHub extra configs.

tlvu:
Adding @mishaschwartz to this review as he might be more aware of Cowbird than me.  He might suggest other ways to make this work.

birdhouse/optional-components/test-cowbird-jupyter-access/docker-compose-extra.yml
ChaamC: The way my test notebooks work is this :

- The [main notebook](https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/blob/cowbird-jupyter-e2e-test/notebooks-auth/test_cowbird_jupyter.ipynb) that is executed first is indeed not spawned by the JupyterHub component, but it does use the JupyterHub component. I highlighted the cells that use JuptyerHub [here](https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/131/files#diff-8f663cf7d0e2fa8d8be62174e0b10b59f24e8486189d0997e9748c4dfc5e2cacR142-R196). It checks first that the test user exists on JupyterHub, gets its token, and then uses the `jhub-client` package to spawn a new JupyterLab with the usual DockerSpawner. So, the JupyterHub is indeed tested here.

- The [second notebook](https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/blob/cowbird-jupyter-e2e-test/notebooks-auth/resources/user_test_cowbird_jupyter.ipynb) is the notebook that is tested against the PAVICS platform, on the JupyterHub environment. 

So, the changes to the JupyterHub component are uses to add role permissions to be able to do the operations on the main notebook (creating a user, getting a user token), and the environment variables added to the DockerSpawner environment are used by the second notebook's test.

birdhouse/optional-components/test-cowbird-jupyter-access/docker-compose-extra.yml
ChaamC: Good catch. I think it makes sense, since I use this volume for wps outputs data and the user workspaces which should both be found in the `${DATA_PERSIST_SHARED_ROOT}`. Updated

birdhouse/optional-components/test-cowbird-jupyter-access/docker-compose-extra.yml
fmigneault: @tlvu 
Sadly, @ChaamC as already spent ~2x the amount of time originally planned for this month to complete this test validation, given how complex it can get with the double nesting of notebook executions, Jupyter API/kernel interactions, combined with Magpie/Cowbird permissions. Given the month is nearing the end, he will not have further time to refactor this.

Since there are little to no validation tests of the Jupyter/Cowbird interactions at the moment, I would like to suggest moving forward with this first implementation, and consider refactors if really judged necessary at a later point. IMO, given that part of the configuration must be done *before* the stack is started (Jupyter config/token), and other *after* (test user/permission setup), I believe that splitting them up into bootstrap/optional-component will only make this validation even more complicated than it already is. It will be much easier for someone to work on this feature if all configurations and scripts are combined under the same optional-component and executed roughly at the same time on startup.

I would like also to point out that restarting the stack should not restart that service each time if it is set to `restart: "no"`. It will also not report that it is down, but that it did a successful "Exit 0". The only thing that @ChaamC could do as final check is that conflict conditions that could make the script fail are handled gracefully in case the service is forcefully restarted.

tlvu: @fmigneault @ChaamC I wasn't against leaving it as-is. I realized it's a bit late to change the design.  I was just inquiring if any side-effects have been anticipated (ex: would re-run cause errors on force restart, anything else ...)

ChaamC: So, if the `test-cowbird-jupyter-access` container from the optional-component is restarted, its script will get reexecuted, but the first thing the script does, apart from some variable assignations and some checks, is to try to create the test user on Magpie. If the user already exists, it raises an error and stops. Considering this, I don't think there should be any side effect if the stack is restarted, as it won't do anything if the test user already exists.

The effects brought by the script are :
- the creation of a test user Magpie
- creating some files in the test user's own workspace
- creating public and user-specific WPS outputs files in `/data/wps_outputs`

tlvu (approved):

PR #406, opened 2023-11-21
Add cowbird readme

ChaamC:
## Changes
- Add new `README` file to be used on `jupyterhub` when `cowbird` is activated. The file describes to the user the 
  different directories and permissions found in its workspace.


birdhouse/components/cowbird/default.env
```diff
@@ -50,6 +50,12 @@ export COWBIRD_LOG_LEVEL=INFO
 # Subdirectory of DATA_PERSIST_SHARED_ROOT containing the user workspaces used by Cowbird
 export USER_WORKSPACES="user_workspaces"
 
+THIS_FILE="`realpath "$0"`"
```
tlvu, line 53: Do not add this `THIS_FILE` var name, which is too generic and will clash with other scripts.

The info you want is already capture by the var `COMPOSE_DIR`, set here

https://github.com/bird-house/birdhouse-deploy/blob/6584dff1574df40e897d15dc4499b63e54e82f2a/birdhouse/pavics-compose.sh#L44-L50

birdhouse/components/cowbird/default.env
```diff
@@ -50,6 +50,12 @@ export COWBIRD_LOG_LEVEL=INFO
...
+# Override JupyterHub Readme file to use Cowbird's own readme version,
```
tlvu, line 54: Override, so now we have 2 similar copies that will have to be kept in sync?

Can't we instead add an extra readme file when cowbird is enabled to further explain its capabilities?

tlvu (changes requested):

birdhouse/components/cowbird/default.env
ChaamC: I'm open to making it an extra separate readme. 
This readme should stay specific to Cowbird.

I don't know if @fmigneault on this, since I discussed this with him?

birdhouse/components/cowbird/default.env
ChaamC: This will be much cleaner, I hadn't successfully found that variable. I updated it.

birdhouse/components/cowbird/user_workspace_readme.ipynb
fmigneault, line 1: Typo/Syntax fixes:
- "subsdirectories" -> subdirectories
- "step by step" -> step-by-step
- "up to date" -> up-to-date
- genderless he/his -> they/their

Add a newline between headers and paragraph for readability. 

I think it would be worth adding an ASCII representation of the directories before their description.
Using only the header text font bold/size makes it hard to distinguish their hierarchy.
Something like: 
```shell
${HOME}
├── public
├── tutorial-notebooks
├── writable-workspace
│   ├── ...
│   └── ...
└── ...
```

birdhouse/components/cowbird/default.env
tlvu: Opps, sorry I forgot our Ouranos README is outside of the birdhouse-deploy repo, here https://github.com/Ouranosinc/PAVICS-landing/blob/master/content/jupyter-readme/README.ipynb.  We even have the matching French flavor and so we deploy both of them.

The assumption was that each organization will deploy their own README so there is not a one size-fit-all readme.

This is really annoying, we can not even try a "nb join" commands since we are missing the French translation.

Our structure, we have the additional `public/`, `mypublic/`, and `pavics-homepage/` dir.

Not sure what to do here.  I'll take @fmigneault proposal for now:
```shell
export JUPYTERHUB_COWBIRD_README="${COMPOSE_DIR}/components/cowbird/user_workspace_readme.ipynb"
export JUPYTERHUB_README="${JUPYTERHUB_README:-${JUPYTERHUB_COWBIRD_README}}"
```

ChaamC: I updated the code with @fmigneault's proposal for now. At least, it will prevent overriding a previously existing definition of the `JUPYTERHUB_README` variable.

birdhouse/components/cowbird/user_workspace_readme.ipynb
```diff
@@ -0,0 +1,130 @@
...
+   "metadata": {},
+   "source": [
+    "# README\n",
+    "\n",
+    "This is the user workspace for the JupyterLab environment.\n",
+    "\n",
+    "## Home Directory Layout\n",
+    "\n",
+    "```\n",
+    "notebook_dir\n",
```
tlvu, line 15: Can you surround `notebook_dir` with slashes, `/notebook_dir/` to imply it is at the root and it is a dir.

birdhouse/components/cowbird/user_workspace_readme.ipynb
```diff
@@ -0,0 +1,130 @@
...
+    "├── README.ipynb\n",
+    "├── public\n",
```
tlvu, line 17: Can you end each dir here with a slash `/` to immediately differentiate when it's a dir and when it's a file.

birdhouse/components/cowbird/user_workspace_readme.ipynb
```diff
@@ -0,0 +1,130 @@
...
+    "├── tutorial-notebooks\n",
+    "└── writable-workspace\n",
+    "    ├── notebooks\n",
+    "    ├── shapefile_datastore\n",
+    "    └── wps_outputs \n",
```
tlvu, line 22: I am a little bit surprised the `wps_outputs/` is under `writetable-workspace/`.  Is this only the user's output or other output shared to the user?  If also other output shared to the user, doesn't it imply it is writable while it is not?

And does the user really have write access to their own wps output since the actual author is the WPS service?

birdhouse/components/cowbird/user_workspace_readme.ipynb
```diff
@@ -0,0 +1,130 @@
...
+    "## tutorial-notebooks\n",
+    "\n",
+    "This directory contains different notebooks used to showcase different usages of the birdhouse services and shares\n",
+    "step-by-step instructions of different use cases.\n",
+    "These notebooks can be executed but cannot be saved as they are read-only.\n",
+    "\n",
+    "## writable-workspace\n",
+    "\n",
+    "This directory contains different subdirectories that are customized and specialized to the user.\n",
+    "Note that this space is generally writable for the user, so they can create their own files in this directory as desired.\n",
```
tlvu, line 42: Add that this `writable-workspace/` dir is the only location that is persisted on disk, which will survive a server shutdown.  Saving files to all other locations will be lost if the Jupyter server is shutdown and restarted.

birdhouse/components/cowbird/user_workspace_readme.ipynb
```diff
@@ -0,0 +1,130 @@
...
+    "This subdirectory is made to contain the different notebooks created by the user.\n",
+    "\n",
+    "#### shapefile_datastore\n",
+    "\n",
+    "This subdirectory is made to contain different files of the shapefile format for Geoserver. Adding new shapefiles to\n",
+    "this directory will also automatically publish the shapefile to the user's Geoserver workspace.\n",
+    "\n",
+    "#### wps_outputs\n",
+    "\n",
+    "This read-only directory contains the different WPS outputs user files. Note that this directory might not exist if the\n",
```
tlvu, line 55: Oh, now we see `wps_outputs/` dir is read-only.  Why nest it under `writable-workspace/` then?

tlvu (approved):
LGTM.  Please add ending slashes.

Other are more questions or if fixes is complex (`wps_outputs/` weirdly under `writable-worskapce/`), can be done in another PR.

birdhouse/components/cowbird/user_workspace_readme.ipynb
ChaamC: The `wps_outputs` directory under the `writable-workspace` only contains the own user's WPS outputs files. 
Any public `wps_outputs` files will be found under `/notebook_dir/public/...`.

The `wps_outputs` directory here is indeed read-only. I guess this can be quite confusing to have it under `writable-workspace`, but it seemed the easiest choice considering the different design restrictions with Cowbird and birdhouse.
Cowbird has to create hardlinks for user WPS outputs to a directory that is specific to the related user, which would logically be the user's workspace, that is mounted as `writable-workspace` here.

I guess the user's WPS outputs hardlinks would have to be stored in a new separated distinct directory, outside of the user_workspaces, so that JupyterHub could mount it outside of the `writable-workspace`? This would require some changes in Cowbird and birdhouse though.

Another option (when Cowbird is activated) would be not to use the `writable-workspace` name, but something else, such as `user-workspace`, but I'm not sure if we want to touch that dir name.

fmigneault: Given that the real nature of `writable-workspace` is to be the `user-workspace`, I would lean toward the rename solution.
The important notion to highlight with this directory is that it is the location the user should use to persist data across sessions, as mentioned in another comment. However, that does not mean this data necessarily has to be writable or managed directly by the user.

Rename does not have to be done in this PR either. Probably safer to have it separately to announce it as a change by itself, since it could impact some users' scripts.

fmigneault (approved):

PR #407, opened 2023-11-30
Allow user to access their Magpie cookie programmatically

mishaschwartz:
## Overview

When the user logs in to jupyterhub, their Magpie cookie is stored in the jupyterhub database. This allows the user to access this variable to programmatically access resources protected by magpie without having to copy/paste these cookies from their browser session or add a username and password in plaintext to the file. 

For example, to access a dataset behind a secured URL with `xarray.open_dataset` using a username and password:

```python
import requests
from request_magpie import MagpieAuth
import xarray

with requests.session() as session:
     session.auth = MagpieAuth("https://mynode/magpie", "myusername", "myverysecretpassword")
     store = xarray.backends.PydapDataStore.open("https://mynode/thredds/some/secure/dataset.nc", session=session)
     dataset = xarray.open_dataset(store)
```

And to do the same thing using the current magpie cookie already used to log in the current user (no need to include username and password)

```python
import os
import requests
import xarray

with requests.session() as session:
    r = requests.get(f"{os.getenv('JUPYTERHUB_API_URL')}/users/{os.getenv('JUPYTERHUB_USER')}", 
                     headers={"Authorization": f"token {os.getenv('JUPYTERHUB_API_TOKEN')}"})
    for name, value in r.json().get("auth_state", {}).get("magpie_cookies", {}).items():
        session.cookies.set(name, value)
    store = xarray.backends.PydapDataStore.open("https://mynode/thredds/some/secure/dataset.nc", session=session)
    dataset = xarray.open_dataset(store)        
```

Note that users who are already logged in to jupyterhub will need to log out and log in for these changes to take effect.

## Changes

**Non-breaking changes**
- New component version jupyterhub:4.0.2-20231130

**Breaking changes**
- None

## Additional Information

birdhouse_skip_ci: false


tlvu (approved):
LGTM.  Thanks, this will be useful.

fmigneault (changes requested):
I think this is a bad idea to instruct users to use env variables for cookies (note: plain text password are not better either). It can give the false impression that this is more secure. 

If the cookie is printed in the output or stored anywhere in the notebook, it could leak access.

Another issue is that cookies are not persistent. Once a refresh occurs, the value set in `MAGPIE_COOKIES` will not be valid anymore. This will confuse the users even more.

I'm not sure what would be the best solution yet, but my first impression is that `MAGPIE_COOKIES` trades a problem for another.

birdhouse/config/jupyterhub/jupyterhub_config.py.template
```diff
@@ -24,6 +24,26 @@ if os.getenv("JUPYTERHUB_CRYPT_KEY"):
     c.MagpieAuthenticator.refresh_pre_spawn = True
     c.MagpieAuthenticator.auth_refresh_age = int("${JUPYTERHUB_AUTHENTICATOR_REFRESH_AGE}")
 
+    # Allow users to access their own auth_state in order to get their own magpie cookie
+    # See https://github.com/jupyterhub/jupyterhub/issues/3588 for details
+    c.JupyterHub.load_roles = [
```
tlvu, line 29: I assume `JUPYTER_API_TOKEN` now exists because of this new config?

tlvu (approved):
LGTM

tlvu:
> * the user logs back into jupyterhub and the database value _is updated_ because it is set whenever the user logs in

Curious, when the user logs back into Jupyterhub, isn't the `MAGPIE_COOKIES` updated as well?

birdhouse/config/jupyterhub/jupyterhub_config.py.template
mishaschwartz: No I think it always existed in the environment

fmigneault (commented):
I think the new JupyterHub API alternative is a good solution. 

Please make it slightly more explicit in the CHANGES which code (first) is to avoid and the new  "STRONGLY RECOMMENDED" approach to use instead. 

That should also be described somewhere else in docs because it will get lost over time with future releases.

mishaschwartz:
> That should also be described somewhere else in docs because it will get lost over time with future releases.

I'll make sure we document it in the jupyterhub tutorials (in progress: https://github.com/DACCS-Climate/marble-tutorials/issues/18)

fmigneault (approved):

PR #410, opened 2023-12-11
Document extra reason to allow not existing dir in `EXTRA_CONF_DIRS`, fix unable to change Twitcher logging level

tlvu:
## Changes

**Non-breaking changes**
- Code documentation: provide an additional reason to not exit early if a directory listed in the `EXTRA_CONF_DIRS` variable does not exist.
- Twitcher: unable to change log level because of typo in qualname config

birdhouse_skip_ci: false


CHANGES.md
```diff
@@ -15,7 +15,12 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Changes
+- Code documentation: extra reason to allow not existing dir in `EXTRA_CONF_DIRS`
```
mishaschwartz, line 19: Can we change this to:

- Code documentation: provide an additional reason to not exit early if a directory listed in the `EXTRA_CONF_DIRS` variable does not exist.

(or similar)

birdhouse/read-configs.include.sh
```diff
@@ -126,6 +126,14 @@ source_conf_files() {
           # fix immediately.
           # The new adir with typo will not be active but at least all the existing
           # will still work.
+          #
+          # Do not exit on not existing conf dir also allow for smooth
```
mishaschwartz, line 130: Suggested:

```
# Not exiting also allows for a smooth transition when component paths change between updates.
```

mishaschwartz (changes requested):
Can we change up the wording a bit?

Maybe I'm just tired but it took me a while to understand what you meant initially.

CHANGES.md
tlvu: Thanks, updated.  English is not my native language!

birdhouse/read-configs.include.sh
tlvu: This one I decided to keep the specific part about `not existing conf dir` because it is code comment so should be more precise.  New wording: "Allowing not existing conf dir also helps for smooth transition of component path when they are new/renamed/deleted."

mishaschwartz (approved):

fmigneault (approved):

PR #412, opened 2023-12-14
Cowbird README file disappears now that cowbird is a default component

mishaschwartz:
## Overview

The settings to enable the Cowbird README file (added in version 1.41.0) assumed that the cowbird component would be loaded after the juptyerhub component. Now that the cowbird component is part of the `DEFAULT_CONF_DIRS` and therefore will always be loaded first, this updates the settings so that the README file will be enabled given the new component load order.

## Changes

**Non-breaking changes**
- Fixes a bug

**Breaking changes**

## Additional Information

birdhouse_skip_ci: false


fmigneault (commented):
I believe Ouranos overrides `JUPYTER_README`. 
It is possible that we might need to do something like 
```shell
export JUPYTERHUB_COWBIRD_README="${JUPYTER_README:-$COMPOSE_DIR/components/cowbird/user_workspace_readme.ipynb}"
```

@tlvu Need to validate this on your end.

mishaschwartz:
> I believe Ouranos overrides JUPYTER_README.

That's fine, as long as they're overriding it in the env.local file, the updated value will be used

tlvu:
> > I believe Ouranos overrides JUPYTER_README.
> 
> That's fine, as long as they're overriding it in the env.local file, the updated value will be used

Exact, we use `env.local` to override so we have the last word on this.

tlvu (approved):

ChaamC (approved):

fmigneault (approved):

PR #413, opened 2023-12-14
Update Thredds to supported version

### LINKED ISSUE #168: Migrate to THREDDS 5.4
> Unidata ended support for all TDS versions prior to 5.x on August 31st, 2022. 

> ... it is recommended that users upgrade to v5.4

https://docs.unidata.ucar.edu/tds/current/userguide/upgrade.html

mishaschwartz:
## Overview

Unidata has dropped support for TDS versions < 5.x. This updates Thredds to version 5.5.

Requires matching update:
* https://github.com/Ouranosinc/pavics-sdi/pull/341
* https://github.com/Ouranosinc/pavics-sdi/pull/342
* https://github.com/Ouranosinc/PAVICS-landing/pull/101

## Changes

**Non-breaking changes**
- Adds...
- New component version pavics/thredds-docker:5.5-unidata-2024-11-19-with-tds-plugin-jar-with-dependencies.jar

**Breaking changes**
- None

## Related Issue / Discussion

- Resolves #168 

## Additional Information

birdhouse_skip_ci: false


tlvu:
FYI, for this change, the automated pipeline test is not enough.  `.ncml` tests are not done against the new Thredds in this PR but against the production server with the current Thredds to avoid having to deploy all the big data required by `.ncml` on test servers.

We have already attempted to upgrade Thredds on Ouranos side and we have `.ncml` issues and, if my memory is right, there are also 2 notebooks not using `.ncml` failing as well which should be caught by the pipeline.

mishaschwartz:
@tlvu 

Thanks for the info. I mostly created this PR to test the pipeline with the updated version to see what happens. I can close this if you're working on it on the ouranos side.

tlvu:
Oh keep this PR, we have not yet opened a formal PR on our side, we simply override in `env.local` for testing.  We can piggyback on your PR once we sort out the various issues on our side.

tlvu:
A few problems found related to Thredds 5.4
https://github.com/Unidata/tds/issues/449
https://github.com/Unidata/tds/issues/406

Notebook changes required:
https://github.com/Ouranosinc/pavics-sdi/pull/317
https://github.com/Ouranosinc/PAVICS-landing/pull/77

tlvu:
FYI @mishaschwartz we still have not done fixing the issues on our side.  Besides there is a performce problem with 5.4 (https://github.com/Unidata/tds/issues/406) and we hope to be fixed in 5.5 so this PR is not ready (5.5 is not released).

fmigneault (approved):

tlvu:
FYI, we found more problems with version 5.  NCML, UDDC, ISO and NetcdfSubset link do not work with v5.  Have not had time to investigate.  Opendap and other links work fine.

Ex: click on the  NCML, UDDC, ISO and NetcdfSubset link here https://pavics.ouranos.ca/testthredds/catalog/testdatasets/testdata/catalog.html?dataset=testdatasets/testdata/ta_Amon_MRI-CGCM3_decadal1980_r1i1p1_199101-200012.nc

fmigneault (changes requested):
Retracting "accept PR" given https://github.com/bird-house/birdhouse-deploy/pull/413#issuecomment-2484044191

tlvu:
> FYI, we found more problems with version 5. NCML, UDDC, ISO and NetcdfSubset link do not work with v5. Have not had time to investigate. Opendap and other links work fine.
> 
> Ex: click on the NCML, UDDC, ISO and NetcdfSubset link here https://pavics.ouranos.ca/testthredds/catalog/testdatasets/testdata/catalog.html?dataset=testdatasets/testdata/ta_Amon_MRI-CGCM3_decadal1980_r1i1p1_199101-200012.nc

Related to https://github.com/Unidata/thredds-docker/issues/310

tlvu:
NCML, UDDC, ISO links fixed by adding the missing jar, see https://github.com/Unidata/thredds-docker/issues/310#issuecomment-2486367398

No other clues found for broken NetcdfSubset link, opened an issue https://github.com/Unidata/tds/issues/544

birdhouse/components/thredds/config/magpie/providers.cfg.template
```diff
@@ -33,6 +33,7 @@ providers:
           dodsC,
           wcs,
           wms,
-          ncss,
```
fmigneault, line 36: Leave the old location.
`THREDDS_VERSION` can be overridden, so a server could still employ the older variant. 

Given that, that will most probably cause a conflict in the `data_type` resolution order if all 3 are defined, so maybe alternate paths or a dynamic variable resolution will be needed here.

birdhouse/components/thredds/config/magpie/providers.cfg.template
```diff
@@ -33,6 +33,7 @@ providers:
...
+          ncss/grid,
+          ncss/point,
```
fmigneault, line 37: Did you test if this works?

I do not know if Magpie/Twitcher handles this properly.
The data types were not planned to include a `/`, so I'm not sure if a `url.split("/")` might happen somewhere in the code leading to misinterpretation of the targeted `data_type`.

fmigneault (changes requested):

birdhouse/components/thredds/config/magpie/providers.cfg.template
mishaschwartz: ugh... you're right. Ok I'll figure something out.

birdhouse/components/thredds/config/magpie/providers.cfg.template
mishaschwartz: Ok maybe we can't change the path... the URL construction docs for this service look like point or graph needs to be a subpath under ncss:

https://docs.unidata.ucar.edu/tds/current/userguide/netcdf_subset_service_ref.html#url-construction

mishaschwartz: > The data types were not planned to include a /, so I'm not sure if a url.split("/") might happen somewhere in the code leading to misinterpretation of the targeted data_type.

Yup magpie splits it up (see [here](https://github.com/Ouranosinc/Magpie/blob/dc0a9a9871f13f1c55b219d644245e9cc3e1c96e/magpie/services.py#L1372)) so our only option at this point is to keep the magpie config as it is and treat `grid/` and `point/` as directory paths if we want to differentiate their magpie permissions.

That's going to cause other confusion though... 

Right now we have URL paths like: 

- `${FQDN}/twitcher/ows/proxy/thredds/ncss/datasets/...`
- `${FQDN}/twitcher/ows/proxy/thredds/dodsC/datasets/...`

So if we want to set specific permissions on the `datasets/` directory we can do that by setting one directory permission rule on the `datasets/` subdirectory and it will work for all services.

But if we instead have:

- `${FQDN}/twitcher/ows/proxy/thredds/ncss/point/datasets/...`
- `${FQDN}/twitcher/ows/proxy/thredds/dodsC/datasets/...`

Then we would need to set the same permission rule on `datasets/` as well as `point/datasets/`. If we don't set the same rule on `point/datasets/` then the rule won't apply to resources accessed by the `ncss/point` service.

Please let me know if I'm interpreting this correctly @fmigneault 

If I'm right then I think we need to update Magpie to handle this use case.

mishaschwartz: I'm going to revert my changes for now until we figure something out.

tlvu: > I'm going to revert my changes for now until we figure something out.

Yes because in the current state (just `ncss` in `magpie/providers.cfg.template`) it already works in my test server.

I have a feeling Magpie will just let through everything after `/thredds/ncss/...` in the URL so `/thredds/ncss/grid/` and `/thredds/ncss/point/` just work.  Also this future-proof us from `/thredds/ncss/<another_split>/`.

However if we ever need to have different Allow/Deny permissions for `nccs/grid` or `ncss/point`, then we need to distinguish between them.

fmigneault: Thanks for validating. 
Yes, you're interpreting it correctly. 

You would need to duplicate the entire permission hierarchy 3 times for anything under "datasets", because the dirs for THREDDS would be defined as follows.  

````
birdhouse
  datasets
    ...  # current ncss and others
  points
    datasets
  grid
    datasets
```


This also causes a weird situation where an actual dir named "points" or "grid" under "birdhouse" (eg: `${FQDN}/twitcher/ows/proxy/thredds/dodsC/points`) would have a different set of permissions that might overlap/conflict or be irrelevant to what points/grid "datasets" define.

mishaschwartz: Alright that makes sense. We should probably make an issue for this over on the Magpie repo. Do you want to do that or shall I?

tlvu: > You would need to duplicate the entire permission hierarchy 3 times for anything under "datasets", because the dirs for THREDDS would be defined as follows.
> 
> ```
> birdhouse
>   datasets
>     ...  # current ncss and others
>   points
>     datasets
>   grid
>     datasets
> ```
> 
> This also causes a weird situation where an actual dir named "points" or "grid" under "birdhouse" (eg: `${FQDN}/twitcher/ows/proxy/thredds/dodsC/points`) would have a different set of permissions that might overlap/conflict or be irrelevant to what points/grid "datasets" define.

I am completely lost here.  Does this means we have a problem with the current Magpie config or not?

fmigneault: @mishaschwartz Yes, you can create the issue.
@tlvu Magpie cannot handle this case currently because it wasn't designed for this. It assumes THREDDS "services" don't contain a `/`, so it could rely on `{Twitcher-prefix}/{THREDDS_name}/{THREDDS_service}/{rest-as-dir/file}`.

tlvu: > @tlvu Magpie cannot handle this case currently because it wasn't designed for this. It assumes THREDDS "services" don't contain a `/`, so it could rely on `{Twitcher-prefix}/{THREDDS_name}/{THREDDS_service}/{rest-as-dir/file}`.

@fmigneault I am still lost.  What case are you talking about?

Just to be clear, in the current state, with Misha's proposed Magpie config change rollback, do we have any problems?

Been running some Jenkins on the new Thredds and I am having some weird errors.  Not sure if it's my test system or the code.  So if you foresee some problem, please let me know.

fmigneault: The `ncssGrid` and `ncssPoint` services won't be protected like other services do (ie: a common browse/read/write set for all services) since the paths are not handled.
For the moment, if you want them protected, you need to duplicate the permission hierarchy.

i.e.: 

A file accessed by `ncssGrid` via `${FQDN}/twitcher/ows/proxy/thredds/ncss/point/datasets/sub/file.nc` will actually go through `ncss` service, and `point` will be considered by Magpie as if it was any other directory (like `datasets` or `sub`).

Therefore, the same `datasets/sub/file.nc` would need 3 sets of permissions:

- datasets/sub/file.nc  => (browse,read,write) for anything currently handled
- grid/datasets/sub/file.nc  => (read) only if accessing via ncssGrid
- point/datasets/sub/file.nc  => (read) only if accessing via ncssPoint

and those (read) need to be duplicated for every user/group/dir/file combination applicable

tlvu: > The `ncssGrid` and `ncssPoint` services won't be protected like other services do (ie: a common browse/read/write set for all services) since the paths are not handled.

@fmigneault I am surprised.  If we "Deny or Allow" everything under `ncss` path, isn't both of those variant `grid/datasets/sub/file.nc` and `point/datasets/sub/file.nc` will be properly denied or allowed, because they are all under `ncss/` path?

Under `ncss` is read only no write anyways, since there is no write for that kind of path.

fmigneault: @tlvu 
Yes. 
A deny top-level will restrict everything lower. But usually, a deny is "undone" by an allow on a lower specific resource we want to grant access. The same issue also happens even when working the other way around -- allow recursive followed by restricted deny lower (see below example).

With the current implementation of THREDDS that checks the service [accses-node] (dodC, ncss, etc.) between Twitcher-proxy-path and the rest of the dir/file path, those resources would be at different "level" for the same file reference from the point of view of Magpie. So you would need to duplicate your 'allows'/'denies' across [accses-node].

```
RESOURCE        PERMISSION APPLIED / RESULT
===========================================
THREDDS         allow-recursive

ncss            [access-mode] (not a resource, cannot have permissions directly)
  datasets      deny-recursive
     sub        allow-recursve (all but sub's contents are blocked)

  public        allow-recursive
     secure     deny-recursive
       secret   allow-match only for user-1

  point         <-- THREDDS sees this as another [access-mode], but Magpie thinks its a RESOURCE!
     datasets   
       sub      <==== OH! OH! undesired full open-access (it's not under the denied "/datasets")

     public     allow-recursive (but because on THREDDS, not the one on "/dataset")
       secure   if you forget to replicate above 'secure', this is still full open acces

  grid          (repeat again, another set of duplicat permissions, for each group/user combination)
    ...
```

tlvu: > A deny top-level will restrict everything lower. But usually, a deny is "undone" by an allow on a lower specific resource we want to grant access. The same issue also happens even when working the other way around -- allow recursive followed by restricted deny lower (see below example).

Oh OK !!! Now I understand, thanks !  Top-level Allow or Deny will work.  But adding an exception under the top-level will not.

tlvu (approved):

tlvu:
@fmigneault Before you say I merge this while you had a change requested, your requested change is for Misha's magpie config change and he already reverted it.

mishaschwartz:
>  Before you say I merge this while you had a change requested, your requested change is for Misha's magpie config change and he already reverted it.

@tlvu Ok but the issue here (https://github.com/Ouranosinc/Magpie/issues/633) now becomes an actual bug instead of a potential future bug. If any deployment uses specific directory permissions for thredds, this introduces a security risk for that deployment. We really needed to resolve https://github.com/Ouranosinc/Magpie/issues/633 before merging this PR. 

I'm going to recommend that we add a warning to release 2.6.3, immediately try to fix https://github.com/Ouranosinc/Magpie/issues/633, and then release 2.7.0 with the updated Magpie changes as soon as possible.

I have 1000+ other things to do but I can put resolving the https://github.com/Ouranosinc/Magpie/issues/633 near the top of my priority list if @fmigneault can review it. If @fmigneault does not have the ability to review it shortly then we should really really revert this change and wait until https://github.com/Ouranosinc/Magpie/issues/633 can be addressed.

fmigneault:
Yeah. Same points as @mishaschwartz 
It wasn't completely reverted. The THREDDS services are added and are not functional.

I don't mind that much since I'm not using those services. Are they needed by anyone?
If this is needed soon, I can take a lot at it.

tlvu:
@mishaschwartz, @fmigneault 

Woops, I guess I should have documented my reasoning.

This Thredds v5 actually have 2 minors issues
* Magpie Allow or Deny exception under top-level exception do not work on NCSS only https://github.com/Ouranosinc/Magpie/issues/633
* Performance problem with WMS only https://github.com/Unidata/tds/issues/406

I considered them minor and not blocking because
* Magpie top-level Allow or Deny still work across the board, exception under top-level also works across the board, except for NCSS only and NCSS is not widely used
* WMS is not widely used, similar to NCSS

Other features I wanted to get from this newer Thredds
* security fixes (newer Tomcat) and the old v4 series is not even available on DockerHub anymore
  * CRIM apparently also pulled this newer Thredds v5 for another of their deployment due to security concerns
* new experimental Zarr support

On the topic of documenting, I just realized the CHANGES.md file is not up-to-date with all the discussions here in this PR.  I will make a PR amending this !

I think, for long running PR, we should double check the CHANGES.md before merge.

I understand everyone is busy so https://github.com/Ouranosinc/Magpie/issues/633 can wait unless you actually need it.

PR #418, opened 2024-01-16
Magpie: ensure that the `MAGPIE_ADMIN_USERNAME` variable is respected

### LINKED ISSUE #417: Security-related variables using hardcoded defaults
> The fact that they're hardcoded at all is a problem though. Because even without the changes you've made here, we could still override `MAGPIE_ADMIN_USERNAME` and then the `JUPYTERHUB_ADMIN_USERS` would no longer match.
> 
> Why don't we make an issue for this and we'll fix it in a later PR

_Originally posted by @mishaschwartz in https://github.com/bird-house/birdhouse-deploy/pull/415#discussion_r1450796302_

https://github.com/bird-house/birdhouse-deploy/blob/10128e58d51ca3e4b9e4022972944bfdd1f87556/birdhouse/env.local.example#L27

The above variable should use the expected reference to `MAGPIE_ADMIN_USERNAME`.
Other locations using similar configurations must be adjusted and validated.

mishaschwartz:
## Overview

We checked that all username variables are respected in the source code. We checked that there are no instances of the default being hardcoded (the assumption being that no one would ever actually change the variable from the default).

The only issue found was that the `MAGPIE_ADMIN_USERNAME` variable was not used to set the default value for `JUPYTERHUB_ADMIN_USERS` properly.

## Changes

**Non-breaking changes**
- Updates default value for `JUPYTERHUB_ADMIN_USERS` to respect other variable settings

**Breaking changes**

## Related Issue / Discussion

- Resolves #417 

## Additional Information

birdhouse_skip_ci: false


mishaschwartz:
Note: this targets the `fix-node-details` branch. If that branch is pulled in first, update the target of this PR to `master`

birdhouse/default.env
```diff
@@ -84,19 +84,14 @@ export SERVER_LICENSE_URL='${__DEFAULT__SERVER_LICENSE_URL}'
 # Those will not be set explicitly as defaults to ensure they are overridden explicitly by the instance.
 # These values would be detected only if the instance was configured using a copy of 'env.local.example'.
 export __DEFAULT__MAGPIE_SECRET=itzaseekrit
-#export __DEFAULT__MAGPIE_ADMIN_USERNAME=admin
```
fmigneault, line 87: Should this one remain to warn about it not being modified?

birdhouse/default.env
```diff
@@ -84,19 +84,14 @@ export SERVER_LICENSE_URL='${__DEFAULT__SERVER_LICENSE_URL}'
...
 export __DEFAULT__MAGPIE_ADMIN_PASSWORD=qwertyqwerty!
-#export __DEFAULT__POSTGRES_PAVICS_USERNAME=postgres-pavics
 export __DEFAULT__POSTGRES_PAVICS_PASSWORD=postgres-qwerty
-#export __DEFAULT__POSTGRES_MAGPIE_USERNAME=postgres-magpie
```
fmigneault, line 91: These might be relevant to warn as well. They should be used everywhere in the configs. 
Since this is for database endpoint connection rather than Magpie user profile, it should not be involved in other scripts than direct interactions between Magpie/Twitcher and Postgres.

birdhouse/default.env
mishaschwartz: Yeah I was debating whether to keep them or remove them....

I guess I was thinking that it's OK if the username stays as "admin" as long as we recommend they change the password.

But in the end, we're just warning the user that we recommend that it *should* be changed from the default. If they want to keep it, they can.

I guess I can put them back in

mishaschwartz:
OK there's a problem with this.... the `JUPYTERHUB_ADMIN_USERS` cannot be made a `DELAYED_EVAL` variable and keep backwards compatibility. 

This is because the current default is

```
JUPYTERHUB_ADMIN_USERS="{'admin'}"
```

which when run through the delayed evaluation algorithm resolves to `{admin}` which is not valid python.

To fix this we either need to:

- update all current deployments that use the default 
- fix this a different way by deprecating the `JUPYTERHUB_ADMIN_USERS` and either:
   1. Creating a different variable in a different format that will be parsed correctly by the delayed evaluation algorithm
   2. Not setting the jupyterhub admin users here at all but instead flagging a user as an admin in jupyterhub based on whether they are an admin in Magpie (this requires an update to https://github.com/Ouranosinc/jupyterhub but should be pretty easy to do)

mishaschwartz:
@fmigneault @tlvu @eyvorchuk 

How do you all set the `JUPYTERHUB_ADMIN_USERS` variable currently? Are you all just using the default or is anyone customizing it in `env.local`?

Do you have any objections to getting rid of the `JUPYTERHUB_ADMIN_USERS` variable and making it so that any users who are members of the `administrators` group in Magpie also get `admin` permissions in Jupterhub?

fmigneault:
We set explicitly `export JUPYTERHUB_ADMIN_USERS="{'admin'}"`. 

I think the correct fix would be to adjust the delayed eval operation. 
It looks like it is its `echo` call that gets rid of the single quotes. It should not assume them to be shell quotes, but "special characters" in this case. 

Something like the following? Not sure if `-e` is portable for the instances that needed `sh` and backticks.  

```shell
❯ X=admin
❯ X='{\"${A}\"}'
❯ eval `echo -e $X`
zsh: command not found: "admin"
❯ X='\{\"${A}\"\}'
❯ eval "echo -e $X"
{"admin"}
```

mishaschwartz:
@fmigneault 

The issue is when this is used:

```shell
$ export JUPYTERHUB_ADMIN_USERS="{'admin'}"
$ echo ${JUPYTERHUB_ADMIN_USERS}
{'admin'}                                              # quotes are still there
$ eval "echo ${JUPYTERHUB_ADMIN_USERS}"
{admin}                                                # quotes are gone
```

Note that this works:

```shell
$ eval 'echo ${JUPYTERHUB_ADMIN_USERS}'
{'admin'}
```

But it breaks for other delayed eval variables:

```shell
$ X=123
$ Y='${X}'
$ eval 'echo $Y'
${X}                                             # we expect "123" not "${X}"                      
```

So it is not a valid solution for our use-case

fmigneault:
@mishaschwartz 
My bad, I over copy-pasted my test samples, which might have caused confusion. 

What I mean to display was only: 
```shell
❯ A=admin
❯ X='\{\"${A}\"\}'
❯ eval "echo -e $X"
{"admin"}
```

This seems to work just like currently (ie: quotes are not changed from `eval "echo ${JUPYTERHUB_ADMIN_USERS}"`). The only distinction is adding `-e` to allow `\` support.

mishaschwartz:
@fmigneault 

Thanks for the clarification. To use your example the issue is still that:

```
❯ X="{'admin'}"
❯ eval "echo -e $X"
{admin}
```

Which is an issue because we want to be able to handle the case where the settings are either:

```
export JUPYTERHUB_ADMIN_USERS="{'admin'}"
```

OR

```
export JUPYTERHUB_ADMIN_USERS='{\"${MAGPIE_ADMIN_USERNAME}\"}'
```

Your example works well for the second case but doesn't work for the first case.

I'm playing around with it too and I'll see if I can come up with a solution that doesn't require us to ask people to change their env.local file. But we may have to fall back to that if we can't make this backwards compatible with the existing settings.

mishaschwartz:
@fmigneault 

Ok I think I've figured it out. There was a confusion of quoting so I added an extra step that evaluates the value of the delayed eval variable in one step and then exports it in a second step. That allows for some different quoting options and allows it to work with the old settings as well. 

See here: https://github.com/bird-house/birdhouse-deploy/pull/418/files#diff-86bb1b7f48133ba418ade714466e5d695b8ee9fb0924e2fc12752680efdb9879R261-R262

fmigneault:
@mishaschwartz Great news! I think this merits a small unittest to avoid regressions.

fmigneault (approved):

mishaschwartz:
@fmigneault want me to merge this to the `fix-node-details` branch or would you rather I wait until that is pulled in to `master`?

fmigneault:
@mishaschwartz Yes you can merge it in the branch.

PR #419, opened 2024-01-16
Autodeploy broken due to instanciated left-over files in ./config/ dir

tlvu:
The `.gitignore` syntax was wrong.  Regression from v2.0.0 (https://github.com/bird-house/birdhouse-deploy/pull/399).

**Non-breaking changes**
- Fix `.gitignore` syntax

birdhouse_skip_ci: false


birdhouse/.gitignore
```diff
@@ -1 +1 @@
-./config
+config/
```
mishaschwartz, line 1: This will match all `config` directories recursively which is *not* what we want. 

From the [gitignore documentation](https://git-scm.com/docs/gitignore#_pattern_format):

> If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.

I think we need to change it to:

```
/config/
```

instead

mishaschwartz (changes requested):

birdhouse/.gitignore
tlvu: You are absolutely right !   Thanks for catching this enormous error !!!  Fixed.

mishaschwartz (approved):
Looks good! Sorry for introducing this bug in the first place. 

Also, just mentioning that this is another good reason for #362

tlvu:
> Looks good! Sorry for introducing this bug in the first place.

We are all human, I just proved it in this own PR!  That's what peer reviewer is for.

> Also, just mentioning that this is another good reason for #362

I was already sold by your idea.  In comment https://github.com/bird-house/birdhouse-deploy/pull/362#pullrequestreview-1578879934 I was only concerned about `docker-compose`, not about the idea to cleanly separate the instantiated files to its own subtree.

PR #420, opened 2024-01-17
jupyter: new incremental build for compatibility with upcoming Thredds v5

tlvu:
See https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/134 for more info.

## Changes

**Non-breaking changes**
- New jupyter env image

birdhouse_skip_ci: false


mishaschwartz (approved):
I approve the spirit. 

I don't know enough about the workflow-tests image to have any useful insights

tlvu:
> I don't know enough about the workflow-tests image to have any useful insights

@mishaschwartz FYI

That workflow-tests image is actually our Jupyter runtime environment in production: https://pavics.ouranos.ca/jupyter !  Yeah I know bad name.

It was initially created as a Jenkins runtime test environment, hence the name, then I realized can run it in production as well so I did that to ensure we are deploying exactly what we tested.  This way I am 100% sure all our tutorial notebooks work in production.

By the way, for UofT, which Jupyter env are you deploying?  Or you are not enabling that component yet?

mishaschwartz:
> By the way, for UofT, which Jupyter env are you deploying? Or you are not enabling that component yet?

We have the defaults from `env.local.example` available but we're working on a jupyter image for general use that we're hoping can be included in other Marble nodes as an option as well.

PR #423, opened 2024-02-09
Weaver: add option to automatically unregister old providers

mishaschwartz:
## Overview

Introduces the `WEAVER_UNREGISTER_DROPPED_PROVIDERS` variable. If set to "True", Weaver providers that are no longer working (not responding when deployed) and are not named in `WEAVER_WPS_PROVIDERS` will be unregistered. This is useful when deploying Weaver with fewer providers than a previous deployment.

For example, if the stack is deployed with the Weaver, Finch, and Raven components. Then later deployed with just Weaver and Raven, the Finch provider will be unregistered from weaver.

Previously, the Finch provider would have remained as a Weaver provider despite the fact that it has been removed from the stack.

## Changes

**Non-breaking changes**
- Adds an additional config option to unregister old weaver providers 

**Breaking changes**
None

## Additional Information

birdhouse_skip_ci: false


birdhouse/components/weaver/post-docker-compose-up
```diff
@@ -333,6 +333,34 @@ for prov in ${WEAVER_WPS_PROVIDERS}; do
...
+    # Get all registered working providers
+    working_providers_resp=$( \
+        curl_cmd --insecure --silent --location \
+            -m ${REQUEST_TIMEOUT} \
+            -b "${cookie}" \
+            "${WEAVER_URL}/providers?check=true&detail=false" \
+    )
+    working_providers=$(echo "$working_providers_resp" | sed -e 's/.*"providers":\[//' -e 's/\].*//' -e 's/[",]/ /g')
+    for prov in $(echo "$all_providers_resp" | sed -e 's/.*"providers":\[//' -e 's/\].*//' -e 's/[",]/ /g'); do
+        if echo "${WEAVER_WPS_PROVIDERS}" | grep -wqv "$prov" && echo "${working_providers}" | grep -wqv "$prov"; then
```
fmigneault, line 353: Might have an issue with `-` that is not considered in the "word" characters. 

Example: `finch` replaced by some kind of `finch-new finch-old` definition. If `finch` is disabled, and therefore not in the list of enabled & working providers, `echo "finch-1 finch-2" | grep -wv "finch"` (error/if not triggered) believes that `finch` is still there and won't remove the provider.

fmigneault (commented):
Good feature. Just a minor thing to check.

birdhouse/components/weaver/post-docker-compose-up
mishaschwartz: Good catch. Thanks!

tlvu (approved):
Nice feature !   I'd say this also cover the case where we "rename" a WPS service.  This will remove the old name.

birdhouse/components/weaver/post-docker-compose-up
```diff
@@ -333,6 +333,37 @@ for prov in ${WEAVER_WPS_PROVIDERS}; do
...
+        curl_cmd --insecure --silent --location \
+            -m ${REQUEST_TIMEOUT} \
+            -b "${cookie}" \
+            "${WEAVER_URL}/providers?check=true&detail=false" \
+    )
+    working_providers=$(echo "$working_providers_resp" | tr '\n' ' ' | \
+                        sed -e 's/.*"providers":\[//' -e 's/\].*//' -e 's/[",]/ /g')
+    for prov in $(echo "$all_providers_resp" | sed -e 's/.*"providers":\[//' -e 's/\].*//' -e 's/[",]/ /g'); do
+        # Note: both WEAVER_WPS_PROVIDERS and working_providers are whitespace delimited with no newlines
+        if echo " ${WEAVER_WPS_PROVIDERS} " | grep -qv "[[:space:]]${prov}[[:space:]]" && \
```
tlvu, line 355: This will be run each time autodeploy runs, which runs inside its own container.  Please test this with autodeploy.

The reason `curl` was a `docker run` is because there are no `curl` in the autodeploy container.

fmigneault (approved):

birdhouse/components/weaver/post-docker-compose-up
mishaschwartz: can confirm this check works in the autodeploy container: `grep`, `sed`, etc. all work in that environment.

PR #426, opened 2024-02-23
decrease optional var logging level and allow to pass extra opts to `pavics-compose.sh up -d`

tlvu:
- logging: decrease logging level for empty optional vars from WARN to DEBUG

  To avoid drowning real WARN messages.  Many optional vars can be valid if empty.

- config: add sample config to configure docker-compose to remove orphans

  To remove orphans containers when components are disabled.  Also link to full
  documentations if other env var can be used.

- compose script: allow to pass extra options to `up` operation

  The previous docker-compose built-in env var was not working so had to add
  this homegrown solution.

  When disabling components, their existing containers will not be removed
  unless option `--remove-orphans` is given together with  `./pavics-compose.sh up -d`.

  This change allow any additional options, not just `--remove-orphans`.

- compose script: exit early when any errors occurred during invocation

  Before, all the `post-docker-compose-up` would still execute after
  `docker-compose` has an error.


fmigneault (approved):

tlvu:
FYI I added another very small change.  PR title and description updated.

CHANGES.md
```diff
@@ -15,7 +15,19 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Changes
+
+- logging: decrease logging level for empty optional vars from WARN to DEBUG
+
+  To avoid drowing real WARN messages.  Many optional vars can be valid if empty.
```
mishaschwartz, line 22: "drowning" ?

birdhouse/env.local.example
```diff
@@ -189,6 +189,10 @@ export GEOSERVER_ADMIN_PASSWORD="${__DEFAULT__GEOSERVER_ADMIN_PASSWORD}"
 # within the autodeploy docker container as the root user.
 #export AUTODEPLOY_CODE_OWNERSHIP="1000:1000"
 
+# Extra options for 'pavics-compose.sh up -d'.
```
mishaschwartz, line 192: This is a nitpick, but it should be `'pavics-compose.sh up'` since `-d` itself is an extra optional flag

mishaschwartz (approved):
This looks good. Just two small things in the documentation

CHANGES.md
tlvu: yes, thanks !

birdhouse/env.local.example
tlvu: Agreed !  Too used to always use `up` with `-d` that I just type it without thinking about it.  In fact, I think for my `env.local`, I will bake `-d` in that new var to ensure I never forget to use `-d` !

If you forget to use `-d`, it goes to the foreground, and the only way out is Ctrl-C which will stop the entire stack, so annoying.

birdhouse/pavics-compose.sh
```diff
@@ -116,7 +119,7 @@ if [ x"$1" = x"up" ]; then
 fi
 
 # the PROXY_SECURE_PORT is a little trick to make the compose file invalid without the usage of this wrapper script
-PROXY_SECURE_PORT=443 HOSTNAME=${PAVICS_FQDN} docker-compose ${COMPOSE_CONF_LIST} $*
```
fmigneault, line 119: Should there be some kind of duplicate filtering, or we simply let the command fail with any (possibility ambiguous) error message? 
For example, `COMPOSE_EXTRA_OPTS="-d"` and the user does `pavics-compose up -d` by mistake.

fmigneault (approved):

mishaschwartz (changes requested):
Given this point (https://github.com/bird-house/birdhouse-deploy/pull/426#discussion_r1503220670) I think that the `COMPOSE_EXTRA_OPTS` has the potential to cause a lot of confusion. 

I think that a better strategy is to encourage users to export any of the environment variables from

- https://docs.docker.com/compose/environment-variables/envvars/
- https://docs.docker.com/engine/reference/commandline/cli/#environment-variables

in their env.local file instead.

For example, instead of setting:

```
export COMPOSE_UP_EXTRA_OPTS="--remove-orphans"
```

they would set:

```
export COMPOSE_REMOVE_ORPHANS=true
```

Not all CLI options/flags are available as environment variables but I think that enough are that we can cover at least the basic use-cases and this is a well documented feature of docker compose.

Also, the precedence is clearly documented here: https://docs.docker.com/compose/environment-variables/envvars-precedence/ so we can avoid some confusion.

tlvu:
> Given this point ([#426 (comment)](https://github.com/bird-house/birdhouse-deploy/pull/426#discussion_r1503220670)) I think that the `COMPOSE_EXTRA_OPTS` has the potential to cause a lot of confusion.
> 
> I think that a better strategy is to encourage users to export any of the environment variables from
> 
>     * https://docs.docker.com/compose/environment-variables/envvars/
> 
>     * https://docs.docker.com/engine/reference/commandline/cli/#environment-variables
> 
> 
> in their env.local file instead.
> 
> For example, instead of setting:
> 
> ```
> export COMPOSE_UP_EXTRA_OPTS="--remove-orphans"
> ```
> 
> they would set:
> 
> ```
> export COMPOSE_REMOVE_ORPHANS=true
> ```
> 
> Not all CLI options/flags are available as environment variables but I think that enough are that we can cover at least the basic use-cases and this is a well documented feature of docker compose.
> 
> Also, the precedence is clearly documented here: https://docs.docker.com/compose/environment-variables/envvars-precedence/ so we can avoid some confusion.

True !  If docker-compose already have some built-in env var to solve my problem, why re-invent the wheel.

It does not cover all the options yet, but it's probably enough for now.

I'll just document this in `env.local.example` then.

mishaschwartz (approved):

tlvu:
@mishaschwartz @fmigneault 

Saddly, I tested and `COMPOSE_REMOVE_ORPHANS` do not work.  A bit of search found this https://github.com/docker/compose/issues/11374.

So how about having `COMPOSE_UP_EXTRA_OPTS` as a "hidden" option not documented in `env.local.example`?  We have a few hidden options already so this is not a precedence.

birdhouse/pavics-compose.sh
tlvu: Good catch, yes double `-d` do not work !   Weird !  Which reason would motivate such a strict cli args parsing !

mishaschwartz:
> So how about having COMPOSE_UP_EXTRA_OPTS as a "hidden" option

I don't think it needs to be hidden, but we need a better way of telling the user what the cause of an error is. Maybe something like:

```
log INFO "Executing docker-compose with options: ${COMPOSE_UP_EXTRA_OPTS} $*"
PROXY_SECURE_PORT=443 HOSTNAME=${PAVICS_FQDN} docker-compose ${COMPOSE_CONF_LIST} $*
ERR=$?
[ ${ERR} -gt 0 ] && exit ${ERR}
```

birdhouse/pavics-compose.sh
mishaschwartz: Often CLI libraries support duplicating flags to increase some level. For example, `-v` means verbose level 1 and `-v -v` means verbose level 2. 
Differentiating between options that support this duplication and those that don't are important to these argument parsing libraries.

fmigneault:
@mishaschwartz's proposal seems fine to me. As long as there's a way to figure out from the logs why the command could have failed.

birdhouse/pavics-compose.sh
tlvu: Right !  But activating this "strict count" for all options seems excessive !

tlvu:
> @mishaschwartz's proposal seems fine to me. As long as there's a way to figure out from the logs why the command could have failed.

Great, so I will un-backout my previous commit :)

Luckily I did not rush to merge this "simple" change on Friday.

birdhouse/pavics-compose.sh
```diff
@@ -115,9 +118,14 @@ if [ x"$1" = x"up" ]; then
   done
 fi
 
+log INFO "Executing docker-compose with extra options: ${COMPOSE_EXTRA_OPTS} $*"
 # the PROXY_SECURE_PORT is a little trick to make the compose file invalid without the usage of this wrapper script
-PROXY_SECURE_PORT=443 HOSTNAME=${PAVICS_FQDN} docker-compose ${COMPOSE_CONF_LIST} $*
+PROXY_SECURE_PORT=443 HOSTNAME=${PAVICS_FQDN} docker-compose ${COMPOSE_CONF_LIST} ${COMPOSE_EXTRA_OPTS} $*
```
mishaschwartz, line 123: won't this result in (for example):

```
docker-compose --remove-orphans up -d
```

I think this needs to be changed to reverse the order of `$*` and `${COMPOSE_EXTRA_OPTS}`

mishaschwartz (changes requested):

mishaschwartz (changes requested):

birdhouse/pavics-compose.sh
```diff
@@ -120,7 +120,7 @@ fi
 
 log INFO "Executing docker-compose with extra options: ${COMPOSE_EXTRA_OPTS} $*"
```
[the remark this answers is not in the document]
tlvu: Oh geez !  Thanks !!!

mishaschwartz (approved):

fmigneault (approved):

PR #427, opened 2024-02-27
Adding the marble image to notebook images list

dchandan:
## Overview

Adding the newly created [marble Jupyter image](https://hub.docker.com/repository/docker/marbleclimate/marble-jupyter-images/general) to the notebook image selections list.  

## Changes

**Non-breaking changes**
- Changes are made to `env.local.example`

birdhouse_skip_ci: false

### DIFF
--- a/birdhouse/env.local.example
+++ b/birdhouse/env.local.example
@@ -263,7 +263,8 @@ export GEOSERVER_ADMIN_PASSWORD="${__DEFAULT__GEOSERVER_ADMIN_PASSWORD}"
 #export ALLOW_UNSECURE_HTTP=""
 
 # Jupyter single-user server images
-#export DOCKER_NOTEBOOK_IMAGES="pavics/workflow-tests:210216 \
+#export DOCKER_NOTEBOOK_IMAGES="marbleclimate/marble-jupyter-images \
+#                               pavics/workflow-tests:210216 \
 #                               pavics/crim-jupyter-eo:0.3.0 \
 #                               pavics/crim-jupyter-nlp:0.4.0 \
 #                               birdhouse/pavics-jupyter-base:mlflow-proxy"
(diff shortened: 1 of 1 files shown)


birdhouse/env.local.example
```diff
@@ -263,7 +263,8 @@ export GEOSERVER_ADMIN_PASSWORD="${__DEFAULT__GEOSERVER_ADMIN_PASSWORD}"
 #export ALLOW_UNSECURE_HTTP=""
 
 # Jupyter single-user server images
-#export DOCKER_NOTEBOOK_IMAGES="pavics/workflow-tests:210216 \
+#export DOCKER_NOTEBOOK_IMAGES="marbleclimate/marble-jupyter-images \
```
fmigneault, line 266: Where is this hosted? I cannot find them on DockerHub (default for `docker pull`). Is it on a private/other registry?
If a user enables this as is without appropriate `docker login` to the appropriate registry, that will make the deployment fail due to unresolved reference, unless the Jupyter config is overridden here to allow missing dockers : 
https://github.com/bird-house/birdhouse-deploy/blob/eaa09c2819e5e5dc5bd36dae6827562f3beef1c5/birdhouse/components/jupyterhub/jupyterhub_config.py.template#L181

Also, just to make sure, is the trailing `s` of the image a typo? Is it a single docker image using this name explicitly? Images should be listed individually here.

mishaschwartz (changes requested):
Actually this is a discussion that I've been wanting to have for a while and this is a good place to start it!

I think that we should move this into a marble specific component (under optional-components) that includes all of the marble specific updates and overrides that we want to recommend/enforce for Marble nodes. That way we have a space to update those sort of changes without over-customizing this repo for Marble (like it previously was for PAVICS).

What do you think?

tlvu:
> Actually this is a discussion that I've been wanting to have for a while and this is a good place to start it!
> 
> I think that we should move this into a marble specific component (under optional-components) that includes all of the marble specific updates and overrides that we want to recommend/enforce for Marble nodes. That way we have a space to update those sort of changes without over-customizing this repo for Marble (like it previously was for PAVICS).
> 
> What do you think?

Right.  If the goal is not just providing an example but to enforce it then yes optional-component is the way to go.  Having an optional-component will also allow you to control future updates.  When updating a sample code in `env.local.example`, that change will not propagate to users have enabled the previous sample code.

Or another repo entirely to keep it clean and tight.  Ouranos very specific overrides/additions are here https://github.com/bird-house/birdhouse-deploy-ouranos.

fmigneault:
If the goal is only to apply overrides, then a separate repo like the one mentioned by @tlvu seems more appropriate. CRIM has similar private repos for `env.local` and a few extra configs. 

However, it is important to use such repo only for overrides. If new capabilities are added, they should be brought back into birdhouse-deploy as components/optional-components as appropriate.

mishaschwartz:
>  there are still several hardcoded references to PAVICS and PAVICS configuration

For sure! That is being worked on in #428

mishaschwartz:
This has been resolved in https://github.com/DACCS-Climate/marble-config/commit/fc3f016c6421060bd47de9f6971f41da9864e762 which adds this component to a marble specific configuration that is managed in a different repository.

Issue #441, opened 2024-04-03
:bulb: [Feature] Provide component service version


fmigneault:
## Description

All components that provide a service definition to be reported in `/services` of the platform should provide a `version` field when available. 
This way, the `/services` endpoint would provide a centralized view of available/deployed service versions.

The current schema definition (https://github.com/DACCS-Climate/Marble-node-registry/blob/7fc410a0fa545c11540805db0adf2d7ad576fbf6/node_registry.schema.json#L83) does not disallow adding `version`, but it could add it explicitly using the same definition (https://github.com/DACCS-Climate/Marble-node-registry/blob/7fc410a0fa545c11540805db0adf2d7ad576fbf6/node_registry.schema.json#L145) as used by the nodes.

mishaschwartz:
If we do this we need to have some concept of what a version means for a specific component or service first.

Some things to consider:

What is the value of the version here? The version of the docker image? What if we make changes to the default configuration for a service in this repo that doesn't require a change to the docker container?

fmigneault:
I would leave it up to the specific component to define what version applies. In many cases, the docker container's tag/version would be directly the version of the service, but it does not strictly need to be the case. 

The idea came from recent discussions with participants of the OGC OSPD effort. There needs to be some kind of "service and versioning listing" details where information about the specific versions of services and platform employed can be obtained. I naturally thought about our `/version` and `/service` endpoints to fulfill that requirement, but noticed respective versions of services were not available. A user needs to somehow know where those versions are available, or hope that the service provides a `links` entry with the `version` relation type. Hence, this issue.

mishaschwartz:
Another thing we should think about. We don't actually have anything that indicates a unique name for a give service. We have the "name" field but that could be any value.

For example, let's say that a node wants to deploy two versions of finch, an older one and a newer one. They can call one old-finch and one new-finch. The service definition doesn't have anything that makes it clear that these two services are the same, just with different versions.

We could introduce an "id" field which could be used with the "version" field to uniquely identify which service we're referring to and which version, no matter what the name is. So for the example above we'd have something like:

```
"services": [
  {
    "name": "old-finch",
    "id": "finch",
    "version": "1.0",
    ....
  },
  {
    "name": "new-finch",
    "id": "finch",
    "version": "2.0",
    ....
  }
]
```

fmigneault:
[...]

Issue #371, opened 2023-08-18
:bulb: Move STAC backend to stac-fastapi-pgstac


dchandan:
## Description

The recently merged STAC implementation (#297) uses a (i) lightweight implementation of the STAC API [stac-app](https://github.com/crim-ca/stac-app) that is based on [stac-fastapi](https://github.com/stac-utils/stac-fastapi) and (ii) a [PgSTAC backed](https://github.com/stac-utils/pgstac). It is not clear what advantage this setup provides since there is a dedicated, community-supported stac-fastapi implementation using the PgSTAC backend called [`stac-fastapi-pgstac`](https://github.com/stac-utils/stac-fastapi-pgstac) that could be used within Birdhouse.

Is there a need to continue using CRIM's `stac-app`? This could have been a legacy decision, which is why I am opening this issue to solicit input on whether we should change the implementation to use `stac-fastapi-pgstac`.

## Concerned Organizations

Francis and I had useful discussion on moving away from the CRIM implementation a little while back. @fmigneault what are your current thoughts on this?

fmigneault:
The last update I found regarding `stac-app` is detailed here: https://github.com/bird-house/birdhouse-deploy/pull/297#discussion_r1230273410
I would prefer using the official image if all required extensions are implemented in it to take advantage of the community bugfixes.
From the information I got, the `stac-app` was only used "temporarily" to patch missing elements from the official STAC FastAPI. 

I can see a big difference between https://github.com/crim-ca/stac-app/blob/main/filters.py and https://github.com/stac-utils/stac-fastapi-pgstac/blob/main/stac_fastapi/pgstac/extensions/filter.py, so I'm not sure how many items (if any) would still need to be added.

There are also a few differences in the Dockerfile definitions between https://github.com/crim-ca/stac-app/blob/main/Dockerfile and https://github.com/stac-utils/stac-fastapi-pgstac/blob/main/Dockerfile. We would need to see if the code changes since then work with the latest official image.

dchandan:
@fmigneault I recall you were mentioning about proposing the stac-fastapi community about using a pluggable architecture that would allow end users to add new features, such as that in https://github.com/crim-ca/stac-app/blob/main/filters.py without needing to create a separate app. Are you still leaning towards that?

fmigneault:
It depends on how reactive and willing is the stac-fastapi community to integrate proposed changes / new extensions.
[...]

Issue #346, opened 2023-06-14
Use official `radiantearth/stac-browser` image


matprov:
## Description

For the stac-browser component, use official `radiantearth/stac-browser` instead of `crim-ca/stac-browser` image once https://github.com/radiantearth/stac-browser/pull/325 is merged. It is currently being used only for container image creation and the fork doesn't add any new logic.

See https://github.com/bird-house/birdhouse-deploy/pull/297#discussion_r1166107325

fmigneault:
Relates to #371

fmigneault:
Relates to https://github.com/bird-house/birdhouse-deploy/pull/373#discussion_r1323427866

PR #445, opened 2024-04-05
add node services URI and fields

fmigneault:
## Overview

Add node services URI and fields.

## Changes

**Non-breaking changes**
- Node Services: Add definitions and variables for every service represented by
  the [DACCS-Climate/Marble-node-registry](https://github.com/DACCS-Climate/Marble-node-registry).

  - Add `version` field using the corresponding `<SERVICE>_VERSION` variables.
  - Add `types` field restricted by specific values instead of previous `keywords` expected to be extendable.
  - Add `<SERVICE>_IMAGE_URI` variables to provide `rel: service-meta` link for every service.

**Breaking changes**
- n/a

## Related Issue / Discussion

- Resolves https://github.com/bird-house/birdhouse-deploy/issues/441
- Depends on https://github.com/DACCS-Climate/Marble-node-registry/pull/40 to enforce validation

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/finch/service-config.json.template
```diff
@@ -1,6 +1,10 @@
 {
   "$schema": "https://raw.githubusercontent.com/DACCS-Climate/Marble-node-registry/main/node_registry.schema.json#service",
   "name": "finch",
+  "version": "${FINCH_VERSION}",
+  "types": [
+    "wps"
+  ],
   "keywords": [
```
mishaschwartz, line 8: should we leave these keywords as is if we're moving this information to "types". What if we borrowed some of the tags and categories from canarie-api instead. So for finch we could do something like:

```
"keywords": ["Climatology", "Cloud", "Data Slicer", "Climate Indices", ...]
```

birdhouse/components/finch/service-config.json.template
```diff
@@ -1,6 +1,10 @@
 {
   "$schema": "https://raw.githubusercontent.com/DACCS-Climate/Marble-node-registry/main/node_registry.schema.json#service",
```
mishaschwartz, line 2: Let's update this to specify version 1.2.0 as well

mishaschwartz: And then update the test in `tests/test_deployment.py` so that it gets the schema from the "$schema" keyword in the json file instead of using the `services_config_schema` pytest fixture.

That way, we can ensure that the schema is validated by the correct version of the schema each time.

birdhouse/components/raven/docker-compose-extra.yml
```diff
@@ -9,7 +9,7 @@ x-logging:
 
 services:
   raven:
-    image: pavics/raven:${RAVEN_VERSION}
```
mishaschwartz, line 12: I like that we're now being consistent with how we specify images instead of before where we had some hardcoded image names and some in variables from default.env. I like that this PR will make this more consistent

birdhouse/components/finch/service-config.json.template
```diff
@@ -20,6 +24,11 @@
       "rel": "service-desc",
       "type": "text/xml",
       "href": "https://${PAVICS_FQDN_PUBLIC}${TWITCHER_PROTECTED_PATH}/finch?service=WPS&request=GetCapabilities"
+    },
+    {
+      "rel": "service-meta",
+      "type": "application/vnd.oci.image.index.v1+json",
+      "href": "${FINCH_IMAGE_URI}"
```
tlvu, line 31: If a new var is used for template expansion it should be added to `VARS` or `OPTIONAL_VARS` in the `default.env` of the corresponding component.  Otherwise, template expansion will not work.

I guess we do not validate the value in "href", that's why we did not caught the wrong template expansion?

tlvu (changes requested):
Ensure template expansion actually work.

Also I do not think this PR introduce a breaking change (in reference to this PR description).  The new var has a default value so upgrading to this PR is 100% seamless, no manual intervention required, so nothing "breaking" as per deployment as I can see.

I defer to Misha for review of the content of the .json .

Agree with Misha this PR unified the way to specify all the images so it's a very nice side-effect.

Also, it seems all we need are additional copy/paste to make deprecated-components work so probably cheap enough to make it work as well.

mishaschwartz:
@tlvu 

> Also, it seems all we need are additional copy/paste to make deprecated-components work so probably cheap enough to make it work as well.

Cheap enough for sure, but can we also plan to stop updating these at some point? I mean, at what point do we actually treat them as deprecated instead of just in another folder that happens to be called `deprecated-components`?

> I guess we do not validate the value in "href", that's why we did not caught the wrong template expansion?

We are, it's just that the [lnks schema](https://json-schema.org/draft/2020-12/links) uses the [uri-template](https://json-schema.org/understanding-json-schema/reference/string#uri-template) string type to check the `href` value and `${some_string_here}` is a valid uri-template according to [RFC6570](https://datatracker.ietf.org/doc/html/rfc6570). There's not much we can do about that.

tlvu:
> at what point do we actually treat them as deprecated instead of just in another folder that happens to be called `deprecated-components`?

I guess at the point where it's expensive.  I agree "expensive" is very vague so up to you I guess.

birdhouse/components/finch/service-config.json.template
fmigneault: Yes. I would like to do this, but until there is a clear-cut version I can refer to, adding these to the keywords will fail schema validation.

fmigneault:
I did omit `flyingpigeon` on purpose due to deprecated, but as highlighted, it is fairly easy to update, so I can add it as well.
Will wait for a release that integrates the relevant updates to the schema to update the references.

fmigneault:
> We are, it's just that the [lnks schema](https://json-schema.org/draft/2020-12/links) uses the [uri-template](https://json-schema.org/understanding-json-schema/reference/string#uri-template) string type to check the href value and ${some_string_here} is a valid uri-template according to [RFC6570](https://datatracker.ietf.org/doc/html/rfc6570). There's not much we can do about that.

I don't think this is the cause. 
According to JSON-schema `format` is just a hint, not an enforced validation, unless otherwise specified.
```python
jsonschema.validate("^^^!!!! <<SO NOT AN URI>> ???", {"type": "string", "format": "uri-template"})  # OK
```
Looking up the format check in the validator confirms this (`uri` and `uri-template` not checked):
```python
jsonschema.draft202012_format_checker
<FormatChecker checkers=['date', 'email', 'idn-email', 'idn-hostname', 'ipv4', 'ipv6', 'regex', 'uuid']>
```
So, the real fix would be to add a `"pattern": "^https?://"` explicitly.

birdhouse/components/finch/service-config.json.template
fmigneault: Optional vars added. See other comments for the validation.

mishaschwartz:
>  According to JSON-schema format is just a hint, not an enforced validation, unless otherwise specified.

Ok that's good to know....

birdhouse/components/finch/service-config.json.template
mishaschwartz: Released! 

(https://github.com/DACCS-Climate/Marble-node-registry/releases/tag/1.2.0)

birdhouse/components/finch/service-config.json.template
mishaschwartz: Got it. Try now with version [1.2.0](https://github.com/DACCS-Climate/Marble-node-registry/releases/tag/1.2.0)?

birdhouse/components/finch/default.env
```diff
@@ -6,6 +6,7 @@
 
 export FINCH_VERSION=0.9.2
 export FINCH_IMAGE="birdhouse/finch:version-${FINCH_VERSION}"
```
tlvu, line 8: I just realized that all xxx_IMAGE and xxx_URI vars should both be delayed eval because they use xxx_VERSION var and that can be overridden by `env.local`.

birdhouse/components/finch/default.env
```diff
@@ -6,6 +6,7 @@
...
+export FINCH_IMAGE_URI="registry.hub.docker.com/${FINCH_IMAGE}"
```
tlvu, line 9: In addition to be in `DELAYED_EVAL` list, all xxx_IMAGE and xxx_IMAGE_URL have to be in single quote for delayed eval to work properly.

birdhouse/components/finch/service-config.json.template
tlvu, line 31: Just my 2 cents, since the URI, do not start with http, should we name the new field "uri" instead of "href"?  "href" would suggest starting with http usually.  Ignore if it has to be "href" for compat or other reasons.

mishaschwartz: `"href"` is required by the schema so we can't change the name

birdhouse/components/finch/default.env
fmigneault: Yes, I noticed after. In the process of rewriting with adjustment to unittests evaluating them as well.

birdhouse/components/finch/service-config.json.template
fmigneault: A non-http scheme value is a valid URI. It uniquely identifies the resource. 
I would prefer to have a `docker://` scheme or similar to be more explicit if permitted, but docker does not seem to like it. However, the values provided in the defaults can be used directly (e.g. `docker pull 'registry.hub.docker.com/pavics/weaver:5.0.0'` works, but `http://registry.hub.docker.com/pavics/weaver:5.0.0` doesn't), so I find them more useful although maybe counter-intuitive.

Maybe one way to avoid confusion would be to use `index.docker.io/pavics/weaver:5.0.0` instead, since it does not redirect to `https://hub.docker.com/` like `registry.hub.docker.com` does? This alternate location still woks with `docker pull 'index.docker.io/pavics/weaver:5.0.0'`, and it looks slightly more like the `type` definition. The idea of this link is really to have the exact value that can be plugged in the docker-compose service `image` field.

birdhouse/scripts/get-services-json.include.sh
```diff
@@ -12,7 +12,16 @@ fi
 
 for adir in ${ALL_CONF_DIRS}; do
   [ -f "${adir}/service-config.json" ] || continue
-  SERVICES="${SERVICES}$([ -n "${SERVICES}" ] && echo ',') $(cat "${adir}/service-config.json")"
+  # read and strip leading/trailing whitespaces
+  SERVICE_CONF="$(cat "${adir}/service-config.json" | sed -z 's/^\s*//;s/\s*$//')"
+  # merge whether config is an array or a single service
+  if [ "$(echo "${SERVICE_CONF}" | head -n 1)" = "[" ]; then
```
mishaschwartz, line 18: If you just left the stac/service-config.json.template file as is, all of this wouldn't be necessary.

I don't understand why this is needed

birdhouse/components/thredds/default.env
```diff
@@ -4,6 +4,7 @@
 export THREDDS_VERSION=4.6.18-unidata-2022-01
 export THREDDS_DOCKER=pavics/thredds-docker
 export THREDDS_IMAGE="${THREDDS_DOCKER}:${THREDDS_VERSION}"
```
mishaschwartz, line 6: single quote

birdhouse/components/thredds/default.env
```diff
@@ -4,6 +4,7 @@
...
+export THREDDS_IMAGE_URI="registry.hub.docker.com/${THREDDS_IMAGE}"
```
mishaschwartz, line 7: single quote

birdhouse/components/stac/service-config.json.template
```diff
@@ -1,54 +1,77 @@
...
-      "href": "https://github.com/radiantearth/stac-browser"
-    },
-    {
-      "rel": "alternate",
-      "type": "application/json",
-      "href": "https://${PAVICS_FQDN_PUBLIC}/stac/"
-    }
-  ]
-}
+[
```
mishaschwartz, line 1: I don't understand why this change is necessary, it makes the format inconsistent with the rest of these files and makes parsing them more complicated (see other comment).

If the answer is that you want this file to be valid json ... I don't think that is a good enough reason to change this, this file is for internal use only and isn't meant to be read by any external client without being processed first.

birdhouse/components/finch/service-config.json.template
mishaschwartz: Are we going to do this in this PR or in a different one?

birdhouse/components/finch/service-config.json.template
```diff
@@ -1,6 +1,10 @@
 {
-  "$schema": "https://raw.githubusercontent.com/DACCS-Climate/Marble-node-registry/main/node_registry.schema.json#service",
+  "$schema": "https://raw.githubusercontent.com/DACCS-Climate/Marble-node-registry/1.2.0/node_registry.schema.json#service",
   "name": "finch",
+  "version": "${FINCH_VERSION}",
+  "types": [
```
mishaschwartz, line 5: This hasn't been implemented yet in the schema. Adding this won't cause schema validation to fail but do we plan on updating the schema before or after this PR is merged?

birdhouse/components/hummingbird/default.env
```diff
@@ -19,7 +26,17 @@ COMPONENT_DEPENDENCIES="
...
+  $DELAYED_EVAL
+  HUMMINGBIRD_IMAGE
+  HUMMINGBIRD_IMAGE_URI
+"
+
 OPTIONAL_VARS="
   $OPTIONAL_VARS
   \$HUMMINGBIRD_VERSION
+  \$HUMMINGBIRD_DOCKER
+  \$HUMMINGBIRD_TAGGED
```
mishaschwartz, line 39: I don't think these are necessary to put here. It's not used in any templates

birdhouse/components/stac/default.env
```diff
@@ -3,6 +3,20 @@ export STAC_POSTGRES_PASSWORD=${POSTGRES_PAVICS_PASSWORD}
 export STAC_PGUSER=${POSTGRES_PAVICS_USERNAME}
 export STAC_PGPASSWORD=${POSTGRES_PAVICS_PASSWORD}
 
+# 'main' branch points at https://github.com/stac-utils/stac-fastapi/commit/d53e792
+# (see: https://github.com/crim-ca/stac-app/blob/40cad1aa7a094d58fca2d3184182761e248f781d/Dockerfile#L15-L20)
+# which corresponds to 2.4.3 release
+export STAC_VERSION=2.4.3-crim-main
+export STAC_IMAGE='ghcr.io/crim-ca/stac-app:main'
```
tlvu, line 10: Just curious, why not push the image with the actual tag `2.4.3-crim` instead of the generic tag `main`?

mishaschwartz (changes requested):

birdhouse/components/stac/default.env
```diff
@@ -3,6 +3,20 @@ export STAC_POSTGRES_PASSWORD=${POSTGRES_PAVICS_PASSWORD}
...
+export STAC_IMAGE_URI='${STAC_IMAGE}'
+
+# 'docker_image_push' branch points at https://github.com/crim-ca/stac-browser/tree/docker_image_push
+# which is some commits ahead of 'v3.0.0-beta.5' (and multiple behind latest official releases)
+# version name is slightly tweaked to fulfill schema while leaving an obvious trace
+export STAC_BROWSER_VERSION=3.0.0-beta.5-crim-docker-image-push
+export STAC_BROWSER_IMAGE='ghcr.io/crim-ca/stac-browser:docker_image_push'
```
tlvu, line 17: Same here, `3.0.0-beta.5-crim` instead of `docker_image_push`.  It is allowed to tag while on a branch.

tlvu (approved):
LGTM.  I have not done a deployment test to ensure all template expansions actually work.  I also defer to @mishaschwartz to validate the content of the various `.json` files and the test.

birdhouse/components/thredds/default.env
tlvu: Oh wow, Misha has much better eye than me this morning :D

birdhouse/components/stac/default.env
mishaschwartz: See #371 and #346

fmigneault: Yes, I could temporarily push a tag to make the reference more "semver"-like, but like @mishaschwartz highlighted, I would much rather have some work done on using the official image. The one tagged is already 9 months old. I don't want this patch tag to be considered a maintained revision.

birdhouse/scripts/get-services-json.include.sh
fmigneault: Either this, or we patch everywhere that tries to load the JSON. 
I add errors when trying to extract the `$schema` in the tests.

birdhouse/components/finch/service-config.json.template
fmigneault: If the schema gets updated to include it beforehand, then we can bump its version right away. Otherwise, it can be done later. As mentioned, it doesn't cause an issue regardless at the moment since it is not validated.

birdhouse/components/finch/service-config.json.template
fmigneault: If possible, I'd prefer to work on this only once.

birdhouse/components/stac/service-config.json.template
fmigneault: The format is also inconsistent currently. It is not a vaild JSON, so it cannot even be parsed unless hacking it before load.

fmigneault: I prefer the all list approach since it works for every use case.

birdhouse/scripts/get-services-json.include.sh
fmigneault: I purposely went without the `jq` approach to avoid introducing this extra dependency since the merge is relatively simple in this case.

One issue with the pseudo-JSON is that it is not obvious (for others) that the incorrectly formed content is intentional to work only once merged. Especially when working with IDEs, or investigating the generated JSON on the server, we get flagged about invalid JSON, which doesn't help maintain them. Since there is a way to make them valid JSON definitions and still work for the aggregation, I would favor using a valid list.

fmigneault:
run tests

mishaschwartz (approved):
Looks good! Just double checking if there's still more keywords to add or not (see comment here: https://github.com/bird-house/birdhouse-deploy/pull/445#discussion_r1559968843)

fmigneault:
For the moment, 1.2.0 still refers to `keywords` as only a specific set of values. Do you want to push a 1.3.0 right away with updated definitions allowing any value, and using `types` instead for the specific service type allowed?

mishaschwartz:
Right, we're still discussing it here: https://github.com/DACCS-Climate/Marble-node-registry/issues/39

Ok nevermind then

Issue #444, opened 2024-04-05
:bulb: [Feature] Log download stats from THREDDS server


huard:
## Description

It would be useful for reporting purposes to monitor data downloads from THREDDS:
- total volume of downloads (Gb/day);
- per-file download volume (Gb/day);
- per-datasets opendap streaming volumes (Gb/day);  

## References

This information can be parsed from NGINX logs, but those logs need to be exposed to Prometheus to be aggregated and archived within the current architecture. 

Possible solution: 
 - https://github.com/nginxinc/nginx-prometheus-exporter

fmigneault:
Consider downloads from WPS outputs and STAC data proxy endpoints as well for the same reasons.

huard:
ESGF uses Beats and Logstash to collect logs and compute their stats. See https://drive.google.com/drive/folders/1LbvoYeQ_6L_bzTsO-EEhwqjIx1jZ-G1k

fmigneault:
If the "node collector" can be located on the same instance, logstash seems like an interesting candidate. If there is no distinction between beats or logstash as "log producers", I would favor the 2nd architecture to limit the number of configurations/technologies involved.

huard:
Parser for nginx logs and prometheus counter
https://gist.github.com/huard/25ca5be3479f72546f748da54f7097e7

huard:
Thanks for the overview. I think one challenge we're having by plugging together different servers is the expertise required to configure each one. I don't think we have within our group someone fluent in Grafana for example. I'm concerned that as we add component, ẁe're going to make the problem worse. 

In that sense, I'm leaning toward your first approach, which is simple and can be easily extended without delving into yet another configuration format.

fmigneault:
I agree with @huard for the same reasons.

mishaschwartz:
Thanks @huard and @fmigneault for your input. I'm happy with that decision as well. I'll un-draft #473 and close #474.

Once we're happy with #473 we can start adding some of the other metrics discussed [here](https://github.com/bird-house/birdhouse-deploy/issues/444#issue-2228671627)

tlvu:
> I agree with @huard for the same reasons.

Same here.

PR #452, opened 2024-05-07
Bump canarie api 1.0.0

mishaschwartz:
## Overview

This version of canarie-api permits running the proxy (nginx) container independently of the canarie-api application. This makes it easier to monitor the logs of canarie-api and proxy containers simultaneously and allows for the configuration files for canarie-api to be mapped to the canarie-api containers where appropriate.

## Changes

**Non-breaking changes**
- New component version canarie-api:1.0.0

**Breaking changes**

## Related Issue / Discussion

- Resolves [issue id](url)

## Additional Information

- This might make parsing the nginx logs slightly easier as well which could help with #12 and #444 

## CI Operations

birdhouse_skip_ci: false


fmigneault:
run tests

fmigneault:
@mishaschwartz 
Not sure if I found the issue, but I think the build trigger only considered organizations `bird-house`, `Ouranosinc` and `crim-ca`. I added you manually for the time being. Maybe if you push another commit, it will run (?).

fmigneault:
run tests

birdhouse/components/twitcher/config/canarie-api/docker-compose-extra.yml
```diff
@@ -1,5 +1,6 @@
 version: "3.4"
 services:
-  proxy:
+  canarie-api: &canarie-volumes
     volumes:
     - ./components/twitcher/config/canarie-api/canarie_api_monitoring.py:${CANARIE_MONITORING_EXTRA_CONF_DIR}/twitcher_canarie_api_monitoring.py:ro
+  canarie-api-cron: *canarie-volumes
```
fmigneault, line 6: Does this anchor work when all components are combined into one big compose file? 
Doesn't it get overridden by whichever definition was loaded first/last/whatever?

birdhouse/components/canarie-api/entrypoint
```diff
@@ -1,16 +1,5 @@
...
 pip install requests-cache
 
-# chain existing cmd
-exec /bin/sh -c "\
-  env >> /etc/environment && \
-  cron && \
-  nginx && \
-  gunicorn -b 0.0.0.0:2000 --workers 1 --log-level=DEBUG --timeout 30 -k gevent canarieapi.wsgi \
-"
+exec /bin/sh -c "gunicorn -b 0.0.0.0:2000 --workers 1 --log-level=DEBUG --timeout 30 -k gevent canarieapi.wsgi"
```
fmigneault, line 5: Can this be left multi-line? It's easier to read diffs.

birdhouse/components/canarie-api/default.env
```diff
@@ -1,7 +1,7 @@
 # Folder inside "proxy" container to drop extra monitoring config
 export CANARIE_MONITORING_EXTRA_CONF_DIR="/conf.d"
 
-export PROXY_IMAGE="pavics/canarieapi:0.7.1"
+export CANARIE_IMAGE="pavics/canarieapi:1.0.0"
```
fmigneault, line 4: Use a `CANARIE_VERSION` and other `CANARIE_...` similar variables as for other components?

CHANGES.md
```diff
@@ -15,7 +15,13 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Changes
+
+- bump canarie-api version to 1.0.0
```
fmigneault, line 20: use reference URL to the tagged versions in the source repo

birdhouse/components/canarie-api/docker-compose-extra.yml
```diff
@@ -0,0 +1,45 @@
...
+    image: ${CANARIE_IMAGE}
+    container_name: canarie-api-cron
+    depends_on:
+      - canarie-api
+    # The volumes and environment should be the same as the canarie-api container except for the proxy-logs volume
+    volumes:
+      - canarie-data:/data/
+      - ./components/canarie-api/docker_configuration.py:/config/docker_configuration.py:ro
+      - ./components/canarie-api/entrypoint-cron:/entrypoint:ro
+      - proxy-logs:/logs/:ro
```
fmigneault, line 35: Might be worth adding a note that this path aligns with https://github.com/bird-house/birdhouse-deploy/blob/bd7e9b835aef0790da201dddccd581ecad430da6/birdhouse/components/canarie-api/docker_configuration.py.template#L17

fmigneault (commented):
Didn't test it, but looks ok conceptually. Just a few comments to confirm. 

Might need to push the component into the CI's `EXTRA_CONF_DIRS` to have a valid test run. It monitors the `/canarie/` endpoint to check when the instance is ready to perform test.

birdhouse/components/twitcher/config/canarie-api/docker-compose-extra.yml
mishaschwartz: Yes it works, docker compose parses yaml file individually first and then combines them into one big definition. Yaml anchors are applied at the file level.

fmigneault (approved):

birdhouse/components/canarie-api/docker-compose-extra.yml
```diff
@@ -0,0 +1,45 @@
...
+    environment:
+      CANARIE_API_CONFIG_FN: /config/docker_configuration.py
+      CANARIE_MONITORING_EXTRA_CONF_DIR: ${CANARIE_MONITORING_EXTRA_CONF_DIR}
+    restart: always
+    entrypoint: /entrypoint
+    logging: *default-logging
+
+volumes:
+  canarie-data:
+  proxy-logs:
```
tlvu, line 45: Very much like this PR.  It's much cleaner this way that the canarie-api monitoring is separate from the proxy container.

I just wonder why are you using data-volume and not direct persistence on disk.  And if you really need to use data-volume, the current way we will lose the data on `./pavics-compose.sh down -v`.

To avoid losing data on `./pavics-compose.sh down -v`, use external data-volume like the monitoring component.

tlvu (approved):
Minor feedback about data persistance.   Not a blocker because currently it's the same behavior already, if the container is delete, we lose proxy logs and canarie-api stats.

birdhouse/components/canarie-api/docker-compose-extra.yml
mishaschwartz: A few reasons...

No other component is storing logs in a persistent manner. I'm not opposed to this but this could be a larger discussion about log retention that isn't related to this PR

> the current way we will lose the data on `./pavics-compose.sh down -v`

Sure, but are you bringing the stack down with the `-v` flag often? Personally, I never ever use the `-v` flag in production for that reason, and I'll delete a volume if needed manually.

I actually find it annoying to have to delete the externally created jupyterhub, prometheus, grafana, etc. volumes when I'm testing and want to start with a blank slate. I personally would actually like to see the externally created volumes become non-external (or a bind-mount from disk) for this reason.

Like I mentioned above, we don't currently have a defined log retention mechanism/policy. I'd love to add one, but maybe we can defer the discussion of how to manage logs when we discuss that.

fmigneault: I think the way it is defined currently is the right way. 
If a platform maintainer wants to persist the logs at a specific location/drive/bind-mount, they simply need to add the relevant overrides to define the volume.
The flexibility is left up to the maintainer.

tlvu: > Sure, but are you bringing the stack down with the `-v` flag often?

I actually do use `-v` from time to time.  Otherwise, all the anonymous volume are not delete and next `up` new anonymous volumes will be created and the old anonymous volumes are only deleted with `docker volume prune --all`.

Here is an example of anonymous volume: https://github.com/bird-house/birdhouse-deploy/blob/80388564101f69d9ba93b532f41273611aa07c7a/birdhouse/components/finch/docker-compose-extra.yml#L21

tlvu:
> @tlvu @fmigneault I'm thinking this needs a minor version bump to 2.3.0. Is that ok with everyone?

Yes you are technically introducing a new component (canarie-api) so yes, a minor bump.

tlvu:
> you are technically introducing a new component (canarie-api)

Oh sorry, new container, the component already exist.

PR #453, opened 2024-05-08
Jupyter env: new full build with latest of almost everything

tlvu:
See https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/121 for more info.

birdhouse_skip_ci: false


CHANGES.md
```diff
@@ -15,7 +15,11 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Changes
+- Jupyter env: new full build with latest of almost everything
+
+  See https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/121 for more info.
```
fmigneault, line 21: Use this format to render correctly on ReadTheDocs: 
```rst
`Ouranosinc/PAVICS-e2e-workflow-tests#121 <https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/121>`_
```

mishaschwartz (changes requested):
Looks good. Can we please also update the reference in the comment here:

https://github.com/bird-house/birdhouse-deploy/blob/a095db4688b5a415aac7969d6f4521981dcaec19/birdhouse/env.local.example#L266

fmigneault (approved):

tlvu:
> Looks good. Can we please also update the reference in the comment here:
> 
> https://github.com/bird-house/birdhouse-deploy/blob/a095db4688b5a415aac7969d6f4521981dcaec19/birdhouse/env.local.example#L266

That one is just an example.  Once user copy `env.local.example` to `env.local`, they probably will never look back at `env.local.example` so I don't think there is value in keeping `env.local.example` up-to-date.  

As time goes by, the example value in their `env.local` will become outdated anyways and there is no way I can control their copy of `env.local`.

CHANGES.md
tlvu: https://github.com/bird-house/birdhouse-deploy/pull/453/commits/0528bfda66841379eea934da93ee78f7d745ea51

mishaschwartz:
> I don't think there is value in keeping env.local.example up-to-date

New deployments will definitely use `env.local.example` as a reference to define which settings they will enable. If we're going to let it get out of date then it's not a good example at all. I think it is critical for new deployments that the `env.local.example` be kept up to date.

tlvu:
> I think it is critical for new deployments that the `env.local.example` be kept up to date.

Sure, I am happy to comply, this is a simple change.

I also added "See components/jupyterhub/default.env for the most updated version of pavics/workflow-tests." so it is clear in their own copy of `env.local` that their value will eventually be not up-to-date anymore.

mishaschwartz (approved):

PR #454, opened 2024-05-10
Scripts that read configuration settings and that exit early on error fail unexpectedly

mishaschwartz:
## Overview

  - Scripts that call `set -e` before reading configuration settings were failing early because some lines are
    intentionally returning a non-zero value when setting variable defaults. This change modifies lines that may return 
    a non-zero status but should not cause the script to exit early.
  - Scripts that were exiting early prior to this change include:
    - birdhouse/deployment/fix-geoserver-data-dir-perm
    - birdhouse/deployment/fix-write-perm
    - birdhouse/deployment/trigger-deploy-notebook

## Changes

**Non-breaking changes**
- bug fix

**Breaking changes**

## Additional Information

## CI Operations

birdhouse_skip_ci: false


tests/test_read_configs_include.py
```diff
@@ -83,17 +83,21 @@ class TestReadConfigs:
...
     def run_func(
-            self, include_file: str, local_env: Union[str, dict], command_suffix: str = ""
+            self, include_file: str, local_env: Union[str, dict], command_suffix: str = "", exit_on_error: bool = True
     ) -> subprocess.CompletedProcess:
         try:
             with tempfile.NamedTemporaryFile(delete=False, mode="w") as f:
                 set_local_env(f, local_env)
 
-            env = {"BIRDHOUSE_LOCAL_ENV": f.name}
+            env = {**os.environ, "BIRDHOUSE_LOCAL_ENV": f.name}
```
fmigneault, line 92: Looks like this might lead to different test results depending on whatever is defined in anyone's personal env. Can't there be explicit variables to replicate desired test cases? Not sure what requires this addition to make the test "work" now.

tests/test_read_configs_include.py
```diff
@@ -83,17 +83,21 @@ class TestReadConfigs:
...
 
+            command_sequence = [f". {include_file}", "read_configs"]
             if command_suffix:
-                command_suffix = f"&& echo '{ENV_SPLIT_STR}' && {command_suffix}"
-            command = f". {include_file} && read_configs {command_suffix}"
+                command_sequence.extend([f"echo '{ENV_SPLIT_STR}'", f"{command_suffix}"])
+            if exit_on_error:
+                command_sequence.insert(1, "set -e")
+                command_sequence.insert(3, "set +e")
+            command = " ; ".join(command_sequence)
```
fmigneault, line 100: All tests cases omit `exit_on_error`.
So basically, the `set` commands are always added since the default is `True`.

Does that mean that `set -e` should always be added to make sure that however we edit the scripts, this special case is always considered? 

Why not have a test that validates that `set -e` is always applied in the files instead? This way, we make sure the results from actually running the scripts and evaluating them in the tests do the same thing.

birdhouse/read-configs.include.sh
```diff
@@ -208,7 +208,7 @@ check_default_vars() {
       n="${i#\$}"
       v=`eval echo "${i}" 2>/dev/null`
       default="\${__DEFAULT__${n}}"
-      d=`eval echo "${default}" 2>/dev/null`
+      d=`eval echo "${default}" 2>/dev/null` || true
```
fmigneault, line 211: Can you provide a specific example where the default evaluation caused the issue (which variable and default value)? 

I'm not quite sure to completely see the impact of ignoring the result (`|| true`) otherwise.

fmigneault (changes requested):

tests/test_read_configs_include.py
mishaschwartz: Yes, but we were always going to get different environments based on people's environments/architectures, etc.

This is specifically to set terminal related variables like `$TERM` which will raise an error if not set (and if `set -e` is set).

I could try to see if just setting `$TERM` or a smaller subset of the environment works as well.

tests/test_read_configs_include.py
mishaschwartz: > So basically, the set commands are always added since the default is True.

Yes, that way we can use existing tests to ensure that there isn't some line we're missing.

> Does that mean that set -e should always be added to make sure that however we edit the scripts

No I don't think that is necessary. `set -e` is sort of like running on strict mode, so if you run a script without it you should always get fewer errors, not more or different errors.

I like giving the scripts the flexibility of getting to choose whether to run with `set -e` or not. All this does is ensure that if they *do* run with `set -e` there won't be any unexpected errors.

fmigneault: > `set -e` is sort of like running on strict mode, so if you run a script without it you should always get fewer errors

Exactly. Under which condition we wouldn't want those reported as early as possible?

The way I see it, if this test is merged, all scripts evaluated in the tests will be strictly validated against `set -e` being enabled. Therefore, the scripts might as well include it directly so that running them on the deployments has the same behavior as evaluated here. If not applied, we could get situations where an error that *would* have been caught, will fail silently, and we'll be confused as to why the tests cannot replicate the issue.

tests/test_read_configs_include.py
fmigneault: The point of the tests is to have a reliable and replicable way to test a predefined set of configuration. 

If we need to evaluate different combinations of environments or known problematic configurations, `@pytest.mark.parameterize` could be used to test those combinations explicitly.

birdhouse/read-configs.include.sh
fmigneault: I see. Got it now.

tests/test_read_configs_include.py
mishaschwartz: Ok, should we just throw a `set -e` into the read-configs.include.sh file itself then?

birdhouse/read-configs.include.sh
tlvu: Oh wow, nice catch.  Can you add a comment saying `|| true` for when no corresponding `__DEFAULT__` var exists !

Will be hard for other people or even us to remember this details months later !

tests/test_read_configs_include.py
tlvu: Well, I am not against `set -e` everywhere but that means we should test **all** the scripts whether we need to add additional `|| true` in there !  Quite an effort.  If we really want this, I'd suggest in a separate PR.

The `set -e` here in the test only test the `read_configs` fonction works correctly with `set -e`.  It is not testing weather all the other scripts will function properly, under all unexpected circumstances, after consuming the `read_configs`.

Thinking further, enforcing `set -e` in `read-configs.include.sh` also force all external scripts in external repos that source `read-configs.include.sh` to also test for all unexpected conditions (ex: missing matching `__DEFAULT__` vars) and add  `|| true`  where needed.  That's could potentially be quite de-stabilizing.  Not sure it is worth the effort.

tests/test_read_configs_include.py
```diff
@@ -185,7 +189,7 @@ def test_delayed_eval_custom_value(self, read_config_include_file) -> None:
 
     def test_delayed_eval_quoting(self, read_config_include_file) -> None:
         """Test that the delayed evaluation functions resolve quotation marks and braces properly"""
-        extra = {"EXTRA_TEST_VAR": "\"{'123'}\"", "DELAYED_EVAL": "$DELAYED_EVAL EXTRA_TEST_VAR"}
+        extra = {"EXTRA_TEST_VAR": "\"{'123'}\"", "DELAYED_EVAL": "\"$DELAYED_EVAL EXTRA_TEST_VAR\""}
```
tlvu, line 192: Just curious, what is the unexpected conditions that needs the extra `\"` around `$DELAYED_EVAL EXTRA_TEST_VAR`?  Empty `$DELAYED_EVAL EXTRA_TEST_VAR`?

tlvu (approved):
+1 for this PR.  If extra changes, I propose to do in a separate PR to keep this one small and easy to understand.

tests/test_read_configs_include.py
fmigneault: I think external scripts sourcing `read-configs.include.sh` would somewhat expect that our own definitions are valid and don't cause any unhandled case. However, I'm fine with doing it as a separate PR to evaluate the feature on its own.

tests/test_read_configs_include.py
mishaschwartz: With the quotes, the temporary env.local looks like:

```sh
EXTRA_TEST_VAR="{'123'}"
DELAYED_EVAL="$DELAYED_EVAL EXTRA_TEST_VAR"
```

without the quotes:

```sh
EXTRA_TEST_VAR="{'123'}"
DELAYED_EVAL=$DELAYED_EVAL EXTRA_TEST_VAR
```

So when this is sourced, the second line tries to run the `EXTRA_TEST_VAR` command (which doesn't exist).

Without `set -e`, this error does not stop the script from running and the test passes (this test checks that the value of `EXTRA_TEST_VAR` does *not* change).

With `set -e`, the test fails at that line since the command cannot be found.

mishaschwartz:
@tlvu @fmigneault 

I've added some explanatory comments and I've parameterized the tests so that tests can run with and without `set -e`.

I agree that we should move the decision about whether to always run `read-configs.includ.sh` with `set -e` to another PR/Issue.

tlvu (approved):

fmigneault (approved):

PR #457, opened 2024-05-23
Make the monitoring component docker images configurable

mishaschwartz:
## Overview

  - The following variables have been added to the `components/monitoring/default.env` file and can be overridden in `env.local` to change the docker image for one of the services enabled by the monitoring component:
    - `GRAFANA_VERSION`
    - `GRAFANA_DOCKER`
    - `CADVISOR_VERSION`
    - `CADVISOR_DOCKER`
    - `PROMETHEUS_VERSION`
    - `PROMETHEUS_DOCKER`
    - `NODE_EXPORTER_VERSION`
    - `NODE_EXPORTER_DOCKER`
    - `ALERTMANAGER_VERSION`
    - `ALERTMANAGER_DOCKER`
  - Note that the defaults are the same as the previous hardcoded versions so this change is fully backwards compatible.

## Changes

**Non-breaking changes**
- Make versions configurable

**Breaking changes**
- None

## CI Operations

birdhouse_daccs_configs_branch: master
birdhouse_skip_ci: false

PR #458, opened 2024-05-23
Bump cadvisor version to the latest version: v0.49.1

mishaschwartz:
## Overview

 - See the cadvisor repo for all changes: https://github.com/google/cadvisor/compare/v0.36.0...v0.49.1
 - This updated was prompted by the fact that the previously installed version of cadvisor (v0.36.0) did not support
    newer versions of docker. When deploying this repo with recent docker version, cadvisor was unable to discover or 
    monitor running containers.

## Changes

**Non-breaking changes**
- New component version cadvisor:v0.49.1

**Breaking changes**

## Additional Information

- This PR builds on top of #457

## CI Operations

birdhouse_skip_ci: false


tlvu (approved):
LGTM.  The new version is backward compat with the settings we currently have in the docker-compose-extra.yml?  And it plays nice with the older existing Prometheus?

mishaschwartz:
> The new version is backward compat with the settings we currently have in the docker-compose-extra.yml? And it plays nice with the older existing Prometheus?

Yes it seems to based on my tests on my local test machine

tlvu:
Just to be sure, the new cadvisor version will still work with older docker-engine as well?

mishaschwartz:
@tlvu 

> Just to be sure, the new cadvisor version will still work with older docker-engine as well?

Theoretically yes. I can try to test it out but I don't have any machines that are running the old version, so it might take me some time to configure a VM to test it. If you have VMs that are running an older version please feel free to check it out too

mishaschwartz:
> Just to be sure, the new cadvisor version will still work with older docker-engine as well?

Personally, I would rather the default support the modern version fo docker engine since most new deployments will use a relatively new version anyway. 

If you need to support a legacy version of docker engine then you can always change the version of cadvisor in your deployment since the version is easily configurable since #457

tlvu:
> change the version of cadvisor in your deployment since the version is easily configurable since #457

Ah true, forgot about that one.  Merge away then.  We can override if needed.

PR #460, opened 2024-06-05
fix weaver data-source resolution for workflow steps

fmigneault:
## Overview
Weaver: Adjust invalid `data_sources.yml` definitions.

## Changes

**Non-breaking changes**
- Weaver: Adjust invalid `data_sources.yml` definitions.

  - Add the missing `data_sources.yml` volume mount for  `weaver-worker`.
  - When `weaver-worker` runs a `Workflow`, the nested `step` process locations need to be resolved according to the
    current `"localhost"` instance. However, the Web API running in `weaver` service is not visible from the worker.
    Since the configuration is shared between `weaver` and `weaver-worker`, use the public endpoint of `weaver` to
    make process URL resolution consistent, and also provide more useful references in job logs when resolution fails.

**Breaking changes**
- n/a

## Related Issue / Discussion

- identified while working on https://github.com/crim-ca/weaver/pull/661

## Additional Information

- https://pavics-weaver.readthedocs.io/en/latest/processes.html#data-source

## CI

birdhouse_skip_ci: false


birdhouse/components/weaver/config/weaver/data_sources.yml.template
```diff
@@ -2,7 +2,7 @@
 # Employed by default for looking at files available directly (e.g.: pre-fetched)
 localhost:
     netloc: "localhost"
-    ades: "http://localhost:4001"
+    ades: "https://${BIRDHOUSE_FQDN_PUBLIC}/${WEAVER_MANAGER_NAME}"
```
mishaschwartz, line 5: My worry with this change is that if the weaver-worker wants to perform an operation that is not publicly available (ie. not allowed by users in the anonymous group in Magpie) then the request will fail with a permission error. If this is what we want then let's leave it as it is. Otherwise...

If we want the `weaver-worker` container to access the `weaver` container we should be able to just use:

```
http://weaver:4001
```

since both of these containers are on the `weaver-mongodb` network.

tlvu: Agree with Misha.

Also, to be consistent with this change, should `netloc: "localhost"` right above be changed to something else?

Another question is we do have Weaver notebook test for a while but it did not catch this error until now?

tlvu (approved):
LGTM but consider Misha suggestion.

birdhouse/components/weaver/config/weaver/data_sources.yml.template
fmigneault: > operation that is not publicly available (ie. not allowed by users in the anonymous group in Magpie) then the request will fail with a permission error.

Actually, the worker should be using the public interface exactly for that purpose! 
Cookies and auth tokens are propagated to the job. Authentication should be performed for every step. Not doing so allows a user to bypass a given process access by wrapping it within another process workflow.

fmigneault: > Another question is we do have Weaver notebook test for a while but it did not catch this error until now?

There is no notebook running a workflow. The only process running in that notebook is a standalone application calling hummingbird's `ncdump`. It does not require a sub-process execution like a workflow does. 

I would like to have a notebook to test this kind of workflow use-case, but I do not have time to implement it anytime soon. If someone wants to tackle it, I can provide reference workflows and guidance (could help for knowledge transfer as well).

mishaschwartz: > If someone wants to tackle it, I can provide reference workflows and guidance (could help for knowledge transfer as well).

I would be interested but I don't think I'll be able to tackle this until after my leave. I can check back in about this in September.

mishaschwartz (approved):

birdhouse/components/weaver/config/weaver/data_sources.yml.template
fmigneault: In case you come back to this later, here is where you can find many sample applications and workflows (they mostly pass around some basic data between steps using various CWL types/features) : 
https://github.com/crim-ca/weaver/tree/master/tests/functional/application-packages

I made a generic "workflow runner" definition here: 
https://github.com/crim-ca/weaver/blob/master/tests/functional/test_workflow.py
It loads the specified Workflow by ID, and does all the setup, pre-deploy, execute, monitor, collect result operations from workflows. 

Most of it could be reused for the specific server being tested. Definitions for cookies and such should be functional as well. Most probably the https://github.com/Ouranosinc/requests-magpie/ could be reused as well to do any required auth on a live server. However, pre-configuration of the permissions is not done in those tests. Most probably https://github.com/Ouranosinc/Magpie/issues/611 would help for that aspect.

Issue #447, opened 2024-04-10
:bulb: [Feature] Add recording rules to Prometheus configuration to store hourly/daily metrics


huard:
... trying to split the metrics collection issue into small digestible bits. 

## Description

1. Select among all existing metrics recorded by Prometheus those of interest to external stakeholders:
  - Number of active users
  - CPU usage
  - Bandwidth usage
  - ...

2. Create [recording rules](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/) in our current Prometheus server to store hourly/daily resolution metrics:
  - Total number of active users per day
  - Mean hourly CPU usage
  - Mean hourly bandwidth usage
  - ...

Make sure the frequency is clearly indicated in the metrics names. 

3. Display hourly/daily data in Graphana dashboard

This doesn't solve the data retention issue, but it's a necessary step. A second Prometheus instance can [federate](https://prometheus.io/docs/prometheus/latest/federation/) the first instance and scrape the metrics that match a hourly/daily regexp pattern. 

## References

#277

mishaschwartz:
@huard you mentioned that there is an existing metric that records total number of active users per day. Do you know which metric that is (its name, how it is collected)?

huard:
The only thing I know is that it's part of the vanilla config that came with the docker image. @tlvu would know more about this.

tlvu:
> The only thing I know is that it's part of the vanilla config that came with the docker image. @tlvu would know more about this.

No we do not have that metric (total number of active users per day).

All the metrics that comes with the "vanilla" config are listed here https://github.com/bird-house/birdhouse-deploy/blob/9d9f46c497e2b00a6ad5be9f1e3ec322f85868a3/birdhouse/components/README.rst#grafana-dashboard

huard:
I think this can be used to count the number of jupyterlab containers:
```
count(rate(container_last_seen{name=~"jupyter-.+"}[1h]))
```
Because jupyter servers are named `jupyter-dhuard`

fmigneault:
> I think this can be used to count the number of jupyterlab containers

Is that a reliable measure for "number of active users"?
It works assuming they work on jupyterlab, but what about other services employed directly?
Would it make more sense to measure something like the number of logins on Magpie or another access-based check?

huard:
Not so reliable, because containers can be left open without activity and they would be counted anyway. It's an approximate measure for sure. 

I think having magpie metrics would indeed be useful. 
- Counter for new user added
[...]

PR #461, opened 2024-06-07
Add the `prometheus-longterm-metrics` and `thanos` optional components

### LINKED ISSUE #277: :bulb: [Feature] monitoring: Add a second Prometheus to scrap the first Prometheus to keep stats for long term at a lower granularity
## Description

Currently component  `./components/monitoring` scrap every 5 minutes and keep the stats for 90 days.

If we want longer stats for high level trend keeping (ex: 5 years), we could use a second Prometheus to scrap the first Prometheus daily (averaging the values) so we can keep longer stats without consuming too much disk space.

This second Prometheus should be on a different machine than PAVICS so when PAVICS is down, we can still access those longterm stats.

To explore the federation feature of Prometheus whether it is simpler to implement/deploy or gives better longterm stats results.

mishaschwartz:
## Overview

The `prometheus-longterm-metrics` component collects longterm monitoring metrics from the original prometheus instance (the one created by the ``components/monitoring`` component).

Longterm metrics are any prometheus rule that have the label ``group: longterm-metrics`` or in other words are selectable using prometheus's ``'{group="longterm-metrics"}'`` query filter. To see which longterm metric rules are added by default see the ``optional-components/prometheus-longterm-metrics/config/monitoring/prometheus.rules.template`` file.

To configure this component:

* update the ``PROMETHEUS_LONGTERM_RETENTION_TIME`` variable to set how long the data will be kept by prometheus
* update the ``PROMETHEUS_LONGTERM_STORE_INTERVAL`` variable to set how often the longterm metrics rules will be calculated. For example, setting it to ``10h`` will calculate these metrics every 10 hours.

Enabling the `prometheus-longterm-metrics` component creates the additional endpoint ``/prometheus-longterm-metrics``.

The `thanos` component enables better storage of longterm metrics collected by the ``optional-components/prometheus-longterm-metrics`` component. Data will be collected from the ``prometheus-longterm-metrics`` and stored in an S3 object store indefinitely.

When enabling this component, please change the default values for the ``MINIO_ROOT_USER`` and ``MINIO_ROOT_PASSWORD`` by updating the ``env.local`` file. These set the login credentials for the root user that runs the [minio](https://min.io/) object store.

Enabling the `thanos` component creates the additional endpoints:

* ``/thanos-query``: a prometheus-like query interface to inspect the data stored by thanos
* ``/thanos-minio``: a minio web console to inspect the data stored by minio.

This also includes an update to the prometheus version from `v2.19.0` to the current latest `v2.52.0`. This is to
required to support the interaction between prometheus and thanos.

## Changes

**Non-breaking changes**
- New component version: prometheus:v2.52.0

## Related Issue / Discussion

- Resolves #277
- Add some initial metrics as described in #447 but we should really add more (either to this PR or in a future PR) by adding more rules to the `birdhouse/optional-components/prometheus-longterm-metrics/config/monitoring/prometheus.rules.template` file.

## Additional Information

- I tested upgrading the prometheus version and there were no issues (no loss of data, no changed APIs etc.)
- Note that the thanos set up is pretty minimal but probably good enough for our purposes. We can always add more of the thanos features/components in the future if needed.

## CI Operations

birdhouse_skip_ci: false


huard (approved):
I can't comment on the specifics, but this is really great news !

Question: this assumes a MinIO server exists. Long has set one up on PAVICS, but I'm wondering if a MinIO config already officially exists on birdhouse-deploy?

mishaschwartz:
> this assumes a MinIO server exists.

If you enable the `thanos` optional component, the a minio server will be created in a docker container. It is not dependent on an externally deployed/created minio server. But if you have one already you could customize this to use that one instead.

> I'm wondering if a MinIO config already officially exists on birdhouse-deploy?

Not at the moment, you likely have a custom component that sets one up for PAVICS

huard:
Huh, that's convenient ! Ok thanks !

tlvu:
> > I'm wondering if a MinIO config already officially exists on birdhouse-deploy?
> 
> Not at the moment, you likely have a custom component that sets one up for PAVICS. Do you know if this is true @tlvu?

@mishaschwartz Yes I deployed one, just for prototyping so did not hook it up "officially" to the PAVICS stack because it's not behind the proxy and Magpie.

@huard  like Misha, I'd rather have the MinIO for the monitoring and MinIO for other usage be separate.  It much simplifies the management (custom config, version requirements, upgrade, ...).  I think Weaver or Magpie was hooking into the default Posgres at some point, then it also rolls it's own Postgres container because it requires a different version of Postgres.

@mishaschwartz Can this new component be enabled standalone?  Meaning on another physical host with only itself and the proxy, as describe in this comment https://github.com/bird-house/birdhouse-deploy/issues/277#issuecomment-2043273300?

If not yet standalone, it's okay we can do it in another PR.

huard:
Good point, makes sense.

birdhouse/optional-components/thanos/default.env
```diff
@@ -0,0 +1,50 @@
...
+
+export DELAYED_EVAL="
+  $DELAYED_EVAL
+  THANOS_IMAGE
+  MINIO_IMAGE
+  MINIO_DATA_STORE
+"
+
+COMPONENT_DEPENDENCIES="
+  ./components/monitoring
```
tlvu, line 48: Why does this thanos component depend on the original monitoring component?  It does not add additional configs to the monitoring component.  It uses variables from the monitoring component?

birdhouse/optional-components/thanos/docker-compose-extra.yml
```diff
@@ -0,0 +1,85 @@
...
+      - '--wait-interval=${THANOS_COMPACTOR_WAIT_INTERVAL}'
+      - *objstore-config
+    depends_on:
+      - minio
+    restart: always
+    logging: *default-logging
+
+  minio:
+    image: ${MINIO_IMAGE}
+    container_name: minio
```
tlvu, line 69: minio_monitoring

tlvu (commented):
Quick review just by reading the code, not tried to deploy yet.

CHANGES.md
```diff
@@ -15,7 +15,41 @@
...
+  * update the ``PROMETHEUS_LONGTERM_STORE_INTERVAL`` variable to set how often the longterm metrics rules will be
+    calculated. For example, setting it to ``10h`` will calculate these metrics every 10 hours.
+
+  Enabling the `prometheus-longterm-metrics` component creates the additional endpoint ``/prometheus-longterm-metrics``.
+
+  The `thanos` component enables better storage of longterm metrics collected by the 
+  ``optional-components/prometheus-longterm-metrics`` component. Data will be collected from the
+  ``prometheus-longterm-metrics`` and stored in an S3 object store indefinitely.
+  
+  When enabling this component, please change the default values for the ``MINIO_ROOT_USER`` and ``MINIO_ROOT_PASSWORD``
```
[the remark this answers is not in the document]
mishaschwartz: @tlvu Can you clarify your question please? 

Are you suggesting I change these variables to use the prefix `MINIO_MONITORING_` instead of just `MINIO_` ?

> So this Minio is for the monitoring only?

It is used by the thanos component, which is used for monitoring

birdhouse/optional-components/prometheus-longterm-metrics/default.env
```diff
@@ -0,0 +1,15 @@
...
+export PROMETHEUS_LONGTERM_TSDB_MIN_BLOCK_DURATION=2h
+export PROMETHEUS_LONGTERM_TSDB_MAX_BLOCK_DURATION=1d12h
+
+OPTIONAL_VARS="
+  $OPTIONAL_VARS
+  \$PROMETHEUS_LONGTERM_STORE_INTERVAL
+"
+
+COMPONENT_DEPENDENCIES="
+  ./components/monitoring
```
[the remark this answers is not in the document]
mishaschwartz: No, it collects metrics from the services that are enabled in the monitoring component.

birdhouse/optional-components/prometheus-longterm-metrics/prometheus.yml
```diff
@@ -0,0 +1,18 @@
+global:
+  external_labels:
+    instance_name: prometheus-longterm-metrics
+
+scrape_configs:
+  - job_name: 'federate'
+    scrape_interval: 15s
```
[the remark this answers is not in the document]
mishaschwartz: It collects and stores metrics that are already collected by the other prometheus service. Depending on how often the other prometheus instance is calculating those metrics, you may not need such a small `scrape_interval`. What interval would you suggest?

birdhouse/optional-components/thanos/default.env
```diff
@@ -0,0 +1,50 @@
...
+export THANOS_VERSION=v0.35.1
+export THANOS_DOCKER="thanosio/thanos"
+export THANOS_IMAGE='${THANOS_DOCKER}:${THANOS_VERSION}'
+
+export MINIO_VERSION=RELEASE.2024-05-27T19-17-46Z
+export MINIO_DOCKER=minio/minio
+export MINIO_IMAGE='${MINIO_DOCKER}:${MINIO_VERSION}'
+
+# Minio uses object storage on disk at this location
+export MINIO_DATA_STORE='${BIRDHOUSE_DATA_PERSIST_ROOT}/minio_data/'
```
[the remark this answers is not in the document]
mishaschwartz: Sure. Just to double check, this is because you have another custom component that uses minio? Or is there another reason I should be aware of?

birdhouse/optional-components/thanos/default.env
mishaschwartz: It depends on the `prometheus-longterm-metrics` component which depends on the `monitoring` component. But you're right, we don't need to explicitly state this dependency here since it is implied by it's dependence on `prometheus-longterm-metrics`.

mishaschwartz:
@tlvu 
> Can this new component be enabled standalone? Meaning on another physical host with only itself and the proxy, as describe in this comment https://github.com/bird-house/birdhouse-deploy/issues/277#issuecomment-2043273300?

Not with the current way that we deploy with without some serious additional coordination to support the networking and permissions. If we used something like kubernetes or docker swarm to deploy this would be easier to deploy on multiple hosts but we're not there yet.

birdhouse/optional-components/prometheus-longterm-metrics/prometheus.yml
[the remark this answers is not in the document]
mishaschwartz: I made it configurable

birdhouse/optional-components/thanos/default.env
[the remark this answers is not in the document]
mishaschwartz: To be consistent with other components, I'll change the prefix to `THANOS_MINIO_` and the store at `thanos_minio_data`.

birdhouse/optional-components/thanos/docker-compose-extra.yml
mishaschwartz: changed to `thanos-minio`

CHANGES.md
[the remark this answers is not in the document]
tlvu: I saw you changed to `THANOS_MINIO` prefix, all good, thanks.

birdhouse/optional-components/README.rst
```diff
@@ -443,3 +443,38 @@ How to enable X-Robots-Tag Header in ``env.local`` (a copy from `env.local.examp
...
+      calculated. For example, setting it to ``10h`` will calculate these metrics every 10 hours.
+
+Enabling this component creates the additional endpoint ``/prometheus-longterm-metrics``.
+
+Thanos
+------
+
+This enables better storage of longterm metrics collected by the ``optional-components/prometheus-longterm-metrics``
+component. Data will be collected from the ``prometheus-longterm-metrics`` and stored in an S3 object store
+indefinitely.
```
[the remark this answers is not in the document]
tlvu: Oh I see, is there a switch to disable Thanos?

Same question, in the case we want to use Thanos, how to visualize the data stored on Thanos?  I assume if Thanos is enabled, the retention duration on the Prometheus side will be very short to avoid doubling the storage so without data being stored in Prometheus, how to visualize that data stored on Thanos.

Just a question.  If another component is required,  we can do it in a follow up Pr.

birdhouse/optional-components/prometheus-longterm-metrics/default.env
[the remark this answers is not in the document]
tlvu: Say I already have a bunch of PAVICS servers running with monitoring enabled and I just want to point this new Prometheus to aggregate all the data?

Can be in a follow up PR.

birdhouse/optional-components/prometheus-longterm-metrics/prometheus.yml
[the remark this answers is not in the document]
tlvu: I do not fully understand federation yet but I had the impression the 2nd Prometheus would query the max/min/average value from the 1st Prometheus at a much longer interval.  Ex: the 1st Prometheus scrap every minute.  The 2nd Prometheus would scrap every hour, taking the max/min/average over the hour of the 1st Prometheus?

birdhouse/optional-components/thanos/default.env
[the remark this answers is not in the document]
tlvu: > another custom component that uses minio? Or is there another reason I should be aware of?

I just want each component to be as much independent as possible.  So this Thanos requires a Minio, it rolls its own.  So it controls the version, the upgrade cycle and any extra custom config of its own Minio.   To simply the management of the stack.

So renaming to THANOS_MINIO is perfect.

tlvu:
> additional coordination to support the networking and permissions.

Right.  Let's say we expose the 1st Prometheus port to the PAVICS docker host and configure the firewall on the PAVICS docker host to allow connection from the 2nd Prometheus long term only, would that would that work?

birdhouse/optional-components/README.rst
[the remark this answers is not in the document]
tlvu: > Oh I see, is there a switch to disable Thanos?

To answer my own question, Thanos is actually a separate component so it does not have to be enabled together with the Prometheus-long-term component?  The Prometheus-long-term component can function standalone of Thanos?

birdhouse/optional-components/prometheus-longterm-metrics/config/monitoring/prometheus.rules
```diff
@@ -0,0 +1,15 @@
+groups:
+    - name: longterm-metrics-hourly
+      interval: 1h
+      rules:
+        # percentage of the time, over the last hour, that all CPUs were working
+        # 1 means all CPUs were working all the time, 0 means they were all idle all the time
+        - record: instance:cpu_load:avg_rate1h
```
tlvu, line 7: So the 2nd Prometheus-longterm will scrap only these 2 new metrics or it will also scrap all the existing ones?

birdhouse/optional-components/prometheus-longterm-metrics/prometheus.yml.template
```diff
@@ -0,0 +1,18 @@
...
+    honor_labels: true
+    metrics_path: '/prometheus/federate'
+
+    params:
+      'match[]':
+        - '{group="longterm-metrics"}'
+
+    static_configs:
+      - targets:
+        - 'prometheus:9090'
```
tlvu, line 18: If you make this target list a template expansion variable, then we can easily add more hosts and override this default local Prometheus and points to other hosts, ex:
```
- existing_pavics_1:9090
- existing_pavics_2:9090
- ...
```

This assume I open up the Prometheus port on `existing_pavics_1` and `existing_pavics_2` and I handle manually the inclusion of `prometheus.rules` on on `existing_pavics_1` and `existing_pavics_2`.

Then you can remove the dependency on `./components/monitoring`.  

Just add a note in the README to enable `./components/monitoring` together with this `./optional-components/prometheus-longterm-metrics` component.

birdhouse/optional-components/README.rst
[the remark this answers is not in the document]
mishaschwartz: > Thanos is actually a separate component so it does not have to be enabled together with the Prometheus-long-term component

Yes that's right. `prometheus-longterm-metrics` collects and stores specific metrics that we want to keep for longer from `prometheus`. If you want to also enable `thanos`, then `thanos` will store those same metrics in a much more compact/efficient way so that you can store more data over a longer time period.

birdhouse/optional-components/prometheus-longterm-metrics/default.env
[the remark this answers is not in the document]
mishaschwartz: The goal of this PR is to create a method for saving existing prometheus data over the long term for a single birdhouse/PAVICS deployment. If you want something that will collect prometheus data for multiple servers I would recommend creating a new repository to host that code.

birdhouse/optional-components/prometheus-longterm-metrics/prometheus.yml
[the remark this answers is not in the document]
mishaschwartz: You don't need multiple prometheus instances for that. You just need to create recording rules that calculate the metrics that you want to store for a longer period and then use the second prometheus instance to grab those metrics for long-term storage. 

The second prometheus instance is necessary only because the data retention time is configured for each instance, so if you want to keep data longer you need a separate prometheus instance.

If you're storing data for a long time with `thanos` instead, you still need a second prometheus instance so that you can select which data `thanos` will store. `thanos` stores **all** data from a prometheus instance, you cannot pick and choose, so we get around this problem by only collecting the data we are interested in the "longterm" prometheus instance and only using `thanos` to store data from that specific instance.

The best way to think about it is:

- `prometheus`: collects all metrics we are interested in and stores them for a short period of time
- `prometheus-longterm-metrics`: selects some metrics collected by `prometheus` that we are interested in storing for a long period of time.
- `thanos`: selects **all** metrics from `prometheus-longterm-metrics` and stores them efficiently over an even longer period of time

birdhouse/optional-components/prometheus-longterm-metrics/config/monitoring/prometheus.rules
mishaschwartz: By default, only these two. I suggest we add more rules to this default list in the future though.

If you have other metrics that you want it to scrape for long term storage, you just have to give that rule the label `group: longterm-metrics`.

birdhouse/optional-components/prometheus-longterm-metrics/prometheus.yml.template
mishaschwartz: I think this is for a separate discussion and out of the scope of this PR.

We can discuss the best way to handle multiple targets later on but at the very least we need to consider the security implications of recommending users open up additional ports.

I really think that any code that handles monitoring multiple prometheus endpoints should go in a separate repo.

mishaschwartz:
> Right. Let's say we expose the 1st Prometheus port to the PAVICS docker host and configure the firewall on the PAVICS docker host to allow connection from the 2nd Prometheus long term only, would that would that work?

Yup, that would work. I think that code can go in a different repo though.

birdhouse/optional-components/README.rst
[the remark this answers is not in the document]
tlvu: Let's see how much space that will take in practice and we will adjust.  And eventually we need a way to visualize those older metrics.  Otherwise, what's the point to keep them forever if we can not visualize?

birdhouse/optional-components/prometheus-longterm-metrics/default.env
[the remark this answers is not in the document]
tlvu: Since the architecture here is pluggable, we do not need a separate repo and duplicate the work.  To deploy the 2nd Prometheus only, on a separate machine, I see we only enable the proxy and the prometheus-longterm-metrics and optionally thanos components on the new machine and that's it.

But agree we can make this "standalone" support in a separate PR.

birdhouse/optional-components/prometheus-longterm-metrics/prometheus.yml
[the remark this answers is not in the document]
tlvu: I see clearer now, thanks !

Just to confirm, the current setup `prometheus-longterm-metrics` is collecting everything from `prometheus` so we are not yet "selecting some metrics" yet, right?

It's alright we can perform the selection in a subsequent PR.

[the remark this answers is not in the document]
tlvu: > Just to confirm, the current setup `prometheus-longterm-metrics` is collecting everything from `prometheus` so we are not yet "selecting some metrics" yet, right?

Opps sorry, just read your other reply now and so to answer my own remark:  `prometheus-longterm-metrics` is collecting only metrics from `group: longterm-metrics` from `prometheus`.

tlvu (approved):
This looks pretty good for now.  Any extra improvements can be done in subsequent PRs.

mishaschwartz:
@tlvu I've made a few small changes to make it easier to deploy this on a separate server. Check out https://github.com/bird-house/birdhouse-deploy/pull/461/commits/2eab8b7aa95e94e631813b2e85d419c91a075613 for the relevant changes.

tlvu (approved):
Thanks !  This should indeed make the longterm prometheus standalone if need be.

birdhouse/optional-components/prometheus-longterm-metrics/default.env
```diff
@@ -0,0 +1,30 @@
+export PROMETHEUS_LONGTERM_VERSION='${PROMETHEUS_VERSION:-"v2.52.0"}'
+export PROMETHEUS_LONGTERM_DOCKER='${PROMETHEUS_DOCKER:-prom/prometheus}'
```
fmigneault, line 2: Should probably consider using `PROMETHEUS_VERSION` and `PROMETHEUS_DOCKER` as defaults instead of the hardcoded values. Otherwise, we must remember to update both combinations of variables each time.

birdhouse/optional-components/prometheus-longterm-metrics/default.env
```diff
@@ -0,0 +1,30 @@
...
+export DELAYED_EVAL="
+  $DELAYED_EVAL
+  PROMETHEUS_LONGTERM_VERSION
+  PROMETHEUS_LONGTERM_DOCKER
+  PROMETHEUS_LONGTERM_IMAGE
+  PROMETHEUS_LONGTERM_RULES_FILE
+"
+
+# Note that this component does not depend explicitly on the `components/monitoring` component so that this can
+# theoretically be deployed on a different machine than the `prometheus` service. This is currently untested.
```
fmigneault, line 30: Curious how that would work out?
It seems to depend on `prometheus:9090` and extends the volumes of `prometheus` service. 
I think it is more confusing to omit the dependency.

fmigneault (commented):
Hi. Sorry for the delay of the review. 
Still going through my plethora of emails following my long leave. 

Nice feature. Will rely on the configurations to have been tested on a local instance since I cannot deep dive into it yet. 

I think the README documentation with the existing monitoring component would benefit from a cross-reference of these new optional components details. A user reading about the monitoring component might not have the reflex to look for corresponding optional components definitions, and might get confused by the mixed variables/config/rule references. Also, they might not be able to figure out that `thanos` is relevant for prometheus and vice-versa. 

In the same line of thought, it is not clear yet to me what are all the impacts of the various components and how they interact between each other. There are implicit dependencies between `thanos`, `prometheus`, `prometheus-longterm-metrics`, as well as `grafana`, `alertmanager`, `cadvisor`, others(?) through `components/monitoring` and all their derived dockers/volumes, but it is really not obvious unless one has personally integrated them. Can we add some guidance in the docs to help disambiguate all of them?

CHANGES.md
```diff
@@ -15,7 +15,42 @@
...
+  The `prometheus-longterm-metrics` component collects longterm monitoring metrics from the original prometheus instance
+  (the one created by the ``components/monitoring`` component).
+
+  Longterm metrics are any prometheus rule that have the label ``group: longterm-metrics`` or in other words are
+  selectable using prometheus's ``'{group="longterm-metrics"}'`` query filter. To see which longterm metric rules are
+  added by default see the 
+  ``optional-components/prometheus-longterm-metrics/config/monitoring/prometheus.rules.template`` file.
+
+  If you do not want the default longterm-metric rules included, set the ``PROMETHEUS_LONGTERM_RULES_FILE`` to anything 
+  other than ``True`` in your ``env.local`` file.
```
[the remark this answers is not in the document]
mishaschwartz: Thanks for catching that. This variable is no longer used as of https://github.com/bird-house/birdhouse-deploy/commit/2eab8b7aa95e94e631813b2e85d419c91a075613 and I missed removing these references. I'll clean it up.

birdhouse/optional-components/prometheus-longterm-metrics/default.env
mishaschwartz: This will use those variables as defaults and only fall back to the hardcoded versions if those variables are not set.
This is necessary to allow for the use case where the longterm metrics component is deployed on a different server/stack from the prometheus component (as requested by @tlvu here: https://github.com/bird-house/birdhouse-deploy/pull/461#issuecomment-2176793022).

birdhouse/optional-components/prometheus-longterm-metrics/default.env
mishaschwartz: It's actually a completely separate service with no shared volumes. It does depend on `prometheus:9090` by default but that default can be updated by setting the `PROMETHEUS_LONGTERM_TARGETS` variable.

`PROMETHEUS_LONGTERM_TARGETS` can point to any prometheus endpoints at all. 

See the discussion about this here: https://github.com/bird-house/birdhouse-deploy/pull/461#discussion_r1644964487

birdhouse/optional-components/prometheus-longterm-metrics/default.env
fmigneault: Ok 👍

birdhouse/optional-components/prometheus-longterm-metrics/default.env
fmigneault: What I meant is that the rules are mounted under `prometheus` service here: 
https://github.com/bird-house/birdhouse-deploy/blob/longterm-monitoring/birdhouse/optional-components/prometheus-longterm-rules/config/monitoring/docker-compose-extra.yml

So, whether `PROMETHEUS_LONGTERM_TARGETS` refers to the same instance or a custom remote one, the local instance needs to have the long-term rules defined. Therefore, doesn't that mean that `prometheus` service (i.e.: using `components/monitoring`) becomes mandatory regardless, since it is needed to mount the rules?

mishaschwartz: Yes, prometheus must be running on the same host as the rest of the birdhouse stack. 

That first prometheus instance must have at least one rule that has the `longterm-metrics` group label in order for a second prometheus instance (either running on the same machine or elsewhere) to know which metrics it should monitor and store from the first prometheus instance.

fmigneault: I've taken a second look at the directory structure, and starting to see why `components/monitoring` is not a dependency. Is it only because `components/monitoring` combines cardvisor, grafana, alertmanager, prometheus, etc. all together?

Therefore, if someone wants "only" longterm-metrics, they need to define their own `prometheus` service with a definition similar to this?
https://github.com/bird-house/birdhouse-deploy/blob/3f75d496b932ab4d90857206003d0225cd20c435/birdhouse/components/monitoring/docker-compose-extra.yml#L32-L49

I think part of what makes all of this confusing is that there are 2 different, though related, component definitions (and I still do not understand why): 
- [optional-components/prometheus-longterm-metrics](https://github.com/bird-house/birdhouse-deploy/tree/longterm-monitoring/birdhouse/optional-components/prometheus-longterm-metrics)
- [optional-components/prometheus-longterm-rules](https://github.com/bird-house/birdhouse-deploy/tree/longterm-monitoring/birdhouse/optional-components/prometheus-longterm-rules)

From a user's perspective, this all seems really convoluted.
Isn't there are way to simplify this hierarchy? For example, what if there was a distinct `optional-components/prometheus` definition, and `components/monitoring` depends on it. Then, [optional-components/prometheus-longterm-metrics](https://github.com/bird-house/birdhouse-deploy/tree/longterm-monitoring/birdhouse/optional-components/prometheus-longterm-metrics) could also depend only on `optional-components/prometheus`?

mishaschwartz: @fmigneault 

> Therefore, if someone wants "only" longterm-metrics, they need to define their own prometheus service with a definition similar to this?

If someone **only** wants longterm metrics then they can have a single prometheus service that sets `--storage.tsdb.retention.time` to something longer than 90 days.

However, if you want some metrics to be stored for longer than others, you need a second prometheus instance. This is because the `--storage.tsdb.retention.time` value is set for an entire prometheus instance (see discussion here: https://github.com/bird-house/birdhouse-deploy/issues/277#issuecomment-2043273300 and here: https://github.com/bird-house/birdhouse-deploy/pull/461#discussion_r1644986997).

> I think part of what makes all of this confusing is that there are 2 different, though related, component definitions (and I still do not understand why)

These are separate in order to accommodate @tlvu's request to be able to deploy the `prometheus-longterm-metrics` component on a different server. `[optional-components/prometheus-longterm-rules](https://github.com/bird-house/birdhouse-deploy/tree/longterm-monitoring/birdhouse/optional-components/prometheus-longterm-rules) are just suggested default rules to be added to the `prometheus` server. As described in the README file:

```
This adds some default longterm metrics rules to the `prometheus` component 
```

These rules are added to the `prometheus` component on the same machine that is running the rest of the stack because that is where all the relevant metrics are collected by the rest of the monitoring components (cadvisor, nodeexporter). These rules are not necessary, they are recommended defaults but you can customize them however you like. It is easier to customize and add your own if the rules are separate. Users can choose to include the defaults, or not.

tlvu: Agree with Misha's explanations.  I just want to add that each organization needs are different so we should not force a certain "configuration".

Specifically to this PR, we should not force the 2 Prometheus to be on the same machine.  So `prometheus-longterm-metrics` should not depend on `components/monitoring`.  We should also not be forced to use `optional-components/prometheus-longterm-rules` because each org will probably have different metrics they deemed useful to keep for longterm.

During Misha's leave, a sysadmin from PCIC actually had a question about how to pull existing metrics from our PAVICS Prometheus into his own centralized Prometheus for a centralized view of all his servers in one place.  I point him to this PR for inspiration.

fmigneault: My bad. I was not clear in my explanation. What I intended to say was:
_"if someone wants only `prometheus` monitoring **with both long-term and short term metrics**, they need to define their own `prometheus` service_" [...], because they cannot activate `prometheus` on its own (i.e.: they must use the full package `components/monitoring` that includes all other services).

Having short/long-term separate, and having them configurable on different servers is perfectly fine by me. I'm all for that flexibility, and never mentioned otherwise.

It just seems from the configuration files that, in order for long-term metrics to be sent to the remote server, the *rules* must be mounted into the short-term local `prometheus` (as shown below), and since that `prometheus` must exist, then it should be a dependency. I understand `long-term-prometheus` NOT being a dependency to toggle it independently of the short-term one. This is good. Here I am referring to https://github.com/bird-house/birdhouse-deploy/tree/3f75d496b932ab4d90857206003d0225cd20c435/birdhouse/optional-components/prometheus-longterm-rules having `component/monitoring` dependency, NOT https://github.com/bird-house/birdhouse-deploy/tree/3f75d496b932ab4d90857206003d0225cd20c435/birdhouse/optional-components/prometheus-longterm-metrics. In other words, if one adds `optional-components/prometheus-longterm-rules` **ONLY** without also thinking to add `components/monitoring`, docker-compose will complain that `service: prometheus` is not defined.

https://github.com/bird-house/birdhouse-deploy/blob/3f75d496b932ab4d90857206003d0225cd20c435/birdhouse/optional-components/prometheus-longterm-rules/config/monitoring/docker-compose-extra.yml#L3-L6

birdhouse/optional-components/README.rst
```diff
@@ -443,3 +443,54 @@ How to enable X-Robots-Tag Header in ``env.local`` (a copy from `env.local.examp
...
+.. _prometheus-longterm-rules
+
+Prometheus Long-term Rules
+--------------------------
+
+This adds some default longterm metrics rules to the `prometheus` component for use by the `prometheus-longterm-metrics`
+component. These rules all have the label ``group: longterm-metrics``.
+
+To see which rules are added, check out the
+`optional-components/prometheus-longterm-rules/config/monitoring/prometheus.rules` file.
```
fmigneault, line 477: Re-reading this following other comments in the PR, I am left only more confused...

Putting ourselves in the shoes of someone having *absolutely no idea about the implementation details* of each component and how they interact with each other, what must the user do to achieve some results (aka: the "To enable this optional-component:" of each other section)?

ie: 
- If user wants SHORT only, what must be enabled (guessing only `components/monitoring`?)
- If user wants SHORT + LONG all locally, what must be enabled (`components/monitoring` + `optional-components/prometheus-longterm-metrics` ?) (+ `optional-components/prometheus-longterm-rules` depending on other dependency comment)
- If user wants SHORT local and LONG remote (`components/monitoring` + `optional-components/prometheus-longterm-rules` + IMPORTANT `PROMETHEUS_LONGTERM_TARGETS` override ?) 

--- 

Specifically regarding the "Prometheus Long-term Rules" section. 
It is mentioned that they are applied to `prometheus`. (this is fine, I understand why to apply the group tags)

However, *for a user without the details of "why"*, it is very confusing! I can see a user think they "should logically" be applied to `prometheus-longterm-rules` service given their names. A much better explanation must be given. 

Basically, something must indicate that service `prometheus` is the source that feeds the others instances using `optional-components/prometheus-longterm-rules`. Then, using  `optional-components/prometheus-longterm-metrics`, they can decide that this receiving instance is another local one (ie: `prometheus-longterm-metrics` service), or dispatch it remotely by overriding `PROMETHEUS_LONGTERM_TARGETS` accoridngly.

birdhouse/optional-components/prometheus-longterm-metrics/default.env
mishaschwartz: Ok I understand your concern now.

> In other words, if one adds optional-components/prometheus-longterm-rules ONLY without also thinking to add components/monitoring, docker-compose will complain that service: prometheus is not defined.

Docker compose won't actually complain in this case. Additional settings defined under the `config/` folder in components are only applied if the component that it references is also enabled. So in this case, if you add `prometheus-longterm-rules` as a component but not `monitoring`, then nothing will happen (since the additional docker-compose settings are in `prometheus-longterm-rules/config/monitoring`.

If we do make the `prometheus-longterm-rules` component dependent on the `monitoring` components then if someone adds the `prometheus-longterm-rules` but **not** the `monitoring` rule to the `BIRDHOUSE_EXTRA_CONF_DIRS` configuration variable, then `monitoring` **will** be added in automatically. This is also a surprising behaviour that users may not expect.

My opinion is that we should leave it as is, so that users must specify both components if they want both components. I'm open to other opinions on this subject though.

birdhouse/optional-components/README.rst
mishaschwartz: Ok... just to clarify what you're asking for here ...

You would like specific instructions about how to set up each of these three cases:

> - If user wants SHORT only ...
> - If user wants SHORT + LONG all locally ...
> - If user wants SHORT local and LONG remote ...

And you would like an additional description of how the various monitoring components interact?
Is there anything else?

birdhouse/optional-components/prometheus-longterm-metrics/default.env
fmigneault: Thanks for validating how this resolves.
I'm fine with either approach if the README indicates clearly what is supposed to happen and what to do if the one or the other situation is desired.

birdhouse/optional-components/README.rst
fmigneault: That's it. Just mitigate user expectations and ensure there is a common understanding of the components/services.

mishaschwartz:
@fmigneault 

Will you be able to look at this again or are you happy with it as it is? 
This has been sitting dormant for a while and I'd like to merge it.

fmigneault (approved):
Sorry. It completely fell off my radar to follow up.

PR #462, opened 2024-06-07
Improve deploy-data flexibility to deploy files from other checkout in same config file

tlvu:
## Overview

This is a non breaking change.

  Given the config file can specify multiple checkouts, this flexibility to
  have `SRC_DIR` be an absolute path will allow one checkout to take files from
  other checkouts, using absolute path to the other checkouts.  `SRC_DIR` can
  still be a relative path of the current checkout, as before, to preserve
  backward-compatibility.

  Possible use-case: re-organize the layout of various files from the various
  checkouts in an intermediate location before rsyncing this intermediate
  location to the final destination.

## CI Operations

birdhouse_skip_ci: false


CHANGES.md
```diff
@@ -15,7 +15,9 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Changes
+- deploy-data: allow more flexibility to deploy files from other checkout in same config file
```
fmigneault, line 19: Please improve the description to be more explicit, such as mentioning `SRC_DIR` value and the obtained behavior. Looking at this later on will be hard to understand.

fmigneault (approved):
Change looks ok. I suggest improving the changelog description.

mishaschwartz (approved):
I agree with @fmigneault. The change looks good but the description needs some clarification

CHANGES.md
tlvu: Indeed, I could re-phrase the code comment in this changelog as well.

tlvu:
Improved the changelog, let me know if it's clearer.

PR #467, opened 2024-08-22
Adapt to the new xclim-testdata structure

Zeitsperre:
## Overview

This PR updates the cloning of the xclim-testdata repo to reflect structural changes.

## Changes

**Non-breaking changes**
- Adjusts the location of the xclim-testdata data folder

## Related Issue / Discussion

- https://github.com/Ouranosinc/xclim-testdata/pull/29
- https://github.com/Ouranosinc/xclim/pull/1889

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,17 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Changes
+
+- xclim-testdata: adapt repository cloning script to the new data structure
+
+  The `xclim-testdata` repo has been restructured to include the data in a `data` subdirectory.
+  This change updates the cloning script to account for this new structure and to ensure that the
+  user experience is consistent with the previous version.
+
+  See:
+  * [xclim-testdata PR/29](https://github.com/Ouranosinc/xclim-testdata/pull/29)
+  * [xclim PR/1889](https://github.com/Ouranosinc/xclim/pull/1889)
 
 [2.5.2](https://github.com/bird-house/birdhouse-deploy/tree/2.5.2) (2024-07-19)
 ------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Zeitsperre:
@tlvu
We are currently grabbing `main` here, but the testdata is tagged according to version (which is based on the date). If we want, we could base this on the tags for less breakage potential.

The last stable tag was [v2023.12.14](https://github.com/Ouranosinc/xclim-testdata/releases/tag/v2023.12.14).

tlvu:
> @tlvu We are currently grabbing `main` here, but the testdata is tagged according to version (which is based on the date). If we want, we could base this on the tags for less breakage potential.
> 
> The last stable tag was [v2023.12.14](https://github.com/Ouranosinc/xclim-testdata/releases/tag/v2023.12.14).

@Zeitsperre 

FYI, if you want to specify the tag `v2023.12.14` you can do it here (replace `origin/main` with `v2023.12.14`) https://github.com/bird-house/birdhouse-deploy/blob/86ffba01f30187bbde85f3f12d656be7740e7785/birdhouse/deployment/deploy-data-xclim-testdata-to-thredds.yml#L4

So are we planning to only hardcode the tag just as a temporary transition measure or are we planning to go this route forward for production deployment?

Zeitsperre:
Once my PR at `xclim-testdata` is approved, I was going to tag a new commit. I don't see why we wouldn't want to use the tags going forward, though. `xclim-testdata` doesn't move very often, but this could prevent breakage for users.

tlvu:
> Once my PR at `xclim-testdata` is approved, I was going to tag a new commit. 

Oh I didn't realized that you will want to synchronize the merge of this PR with your PR so the tagging is not really required.

> I don't see why we wouldn't want to use the tags going forward, though. `xclim-testdata` doesn't move very often, but this could prevent breakage for users.

Hardcoded tags means:
* need to update here each time for new tag
* any quick change (fix existing data, add new data) will also need a tag to trigger the deploy
* changing tag is slower to autodeploy (nightly) instead of the content (every 30 mins)

Since breaking change is not often, I think this is the first time, is it worth it?  It's up to you.

tlvu (approved):

tlvu:
> Since breaking change is not often, I think this is the first time, is it worth it? It's up to you.

Sorry to clarify, we can go the temporary route just for this time, so you do not need to worry about synchronizing merge.

My concerns were about always hardcoding going forward.

Zeitsperre:
@tlvu I follow your logic. It makes sense here. Given that the structure shouldn't change after this, as well as the fact that we uniquely glob NetCDF files, I don't think we need to use tags here.

Will finalize changes on the `xclim` PR, merge `xclim-testdata`, then finally merge the remaining PRs. I don't anticipate needing to make any other changes here after this.

PR #468, opened 2024-09-09
Update PR template to avoid default regex self-match when copy-pasted to merge commit message

fmigneault:
## Overview

Update PR template to avoid default regex self-match when copy-pasted to merge commit message.

Use roundabout comment description to avoid matching the regex.

## Changes

**Non-breaking changes**
- CI maintenance, PR template

**Breaking changes**
- n/a

## Related Issue / Discussion

- Fix PR not triggering CI.

## CI Operations

birdhouse_skip_ci: false


tlvu (approved):
So instead of `[skip ci]` we should use `[<cmd> = skip ci]` in the commit message?

fmigneault:
@tlvu 
No. It remains `[skip ci]`, but if we leave `[skip ci]` as is in the PR template, this is what caused the regex to match and therefore skip the PR trigger.

PR #471, opened 2024-09-23
add migration guide to docs

mishaschwartz:
## Overview

This is a documentation change only. 

The migration guide is hidden in the CHANGES.md file which is fine if you know where to look for it but is easy to miss otherwise. This PR creates a new migration guide file that will be included in the docs. This file contains a copy of the version 2.4 migration guide (translated from markdown to reStructuredText syntax) and all future migration guides should be added here so that they are more visible to users.

## Changes

**Non-breaking changes**
None

**Breaking changes**
None

## CI Operations

birdhouse_skip_ci: true

### DIFF
--- a/docs/source/migration_guide.rst
+++ b/docs/source/migration_guide.rst
@@ -18,9 +18,11 @@ To upgrade to version 2.4 from an earlier version, please follow these steps:
 
   - Update ``env.local`` file to replace all variables that contain ``PAVICS`` with ``BIRDHOUSE``.
     Variable names have also been updated to ensure that they start with the prefix ``BIRDHOUSE_``.
-    * see `birdhouse/env.local.example <birdhouse/env.local.example>`_ (:download:`download </birdhouse/env.local.example`) to see new variable names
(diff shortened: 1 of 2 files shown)


docs/source/migration_guide.rst
```diff
@@ -0,0 +1,123 @@
+.. contents::
+
+Migration Guide
+===============
+
+Version 2.4
+-----------
+
+Version 2.4 renames files and variables that included the string PAVICS.
```
tlvu, line 9: Nice docs.  Better here than buried in the PR.  If I remember correctly the old name starting with PAVICS are still supported for backward-compatibility, it's not mentionned here?

fmigneault (approved):

docs/source/migration_guide.rst
mishaschwartz: Yes they are still supported (as this is a minor version update, not a major one). But we chose to not overly advertise this fact since we want to strongly encourage people to follow the migration guide (see example discussion here: https://github.com/bird-house/birdhouse-deploy/pull/428#discussion_r1513737235).

Backwards compatibility is described elsewhere: https://github.com/bird-house/birdhouse-deploy/blob/fd1a4b9cc29d89b2cf6205665ab387737955536c/CHANGES.md?plain=1#L153-L156 but I would like to keep the migration guide simply as a migration guide.

tlvu (approved):

PR #472, opened 2024-10-04
THREDDS: add more options to configure catalog.xml

mishaschwartz:
## Overview

- Currently the default THREDDS configuration creates two default datasets, the Service Data dataset and the 
  Main dataset. The Service Data dataset is used internally and hosts WPS outputs. The Main dataset is the
  place where users can access data served by THREDDS. Both of these are configured to serve files with the following 
  extensions: .nc .ncml .txt .md .rst .csv

- In order to allow the THREDDS server to serve files with additional extensions, this introduces two new
  variables: 
  - `THREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS`: this allows users to specify additional [filter 
      elements](https://docs.unidata.ucar.edu/tds/current/userguide/tds_dataset_scan_ref.html#including-only-desired-files) to the Service Data dataset. This is especially useful if a WPS 
      outputs files with an extension other than the default (eg: .h5) to the `wps_outputs/` directory.
  - `THREDDS_DATASET_DATASETSCAN_BODY`: this allows users to specify the whole body of the main dataset's 
     [`<datasetScan>`](https://docs.unidata.ucar.edu/tds/current/userguide/tds_dataset_scan_ref.html) element.
     This allows users to fully customize how this dataset serves files.

- We limit the configuration options for the Service Data dataset more than the main dataset because the Service
  Data dataset requires a basic configuration in order to properly serve WPS outputs. Making significant changes
  to this configuration could have unexpected negative impacts on WPS usage.

- The defaults for these new variables are fully backwards compatible. Without changing these variables, the THREDDS
  server should behave exactly the same as before. 

## Changes

**Non-breaking changes**
- Adds more configuration options to THREDDS catalog configuration (backwards compatible)

**Breaking changes**
- None

## CI Operations

birdhouse_skip_ci: false


birdhouse/env.local.example
```diff
@@ -476,6 +476,40 @@ export THREDDS_ADDITIONAL_CATALOG=""
...
 #"
 
+# Additional file filters to add for the Service Data THREDDS dataset. By default, the Service Data dataset will only
+# serve files with the following extensions: .nc .ncml .txt .md .rst .csv
+# If you need this dataset to serve other files you should update the THREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS to add
+# additional file filters.
+# This may be useful to set if a WPS outputs files to the wps_outputs/ directory (hosted under the Service Data dataset) 
+# in a file format other than one of the defaults.
+# See the example below which would also enable serving .png and .h5 files.
+#export THREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS="
```
tlvu, line 486: The example here should be using single quote so the double quote inside do not have to be escaped, same as in the `default.env` file.

tlvu (approved):
LGTM, juste ensure the example is working out of the box.

tlvu:
Oh wait, can you do the same changes for optional-components/testthredds as well?

mishaschwartz:
@tlvu 

> Oh wait, can you do the same changes for optional-components/testthredds as well?

Isn't this set up so that we can run tests against a different THREDDS server? If the tests don't require a different configuration why do we need to change this as well?

tlvu:
> Isn't this set up so that we can run tests against a different THREDDS server? 

To test different version yes.

> If the tests don't require a different configuration why do we need to change this as well?

I meant to allow the same customizations for the test.  Currently we are testing Thredds v5 using this testthredds on our production host.  By the same token we could at the same time test additional configs.  So the same customizations would be useful.

So I meant to add `TESTTHREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS` and `TESTTHREDDS_DATASET_DATASETSCAN_BODY`.

CHANGES.md
```diff
@@ -15,7 +15,29 @@
...
+
+- THREDDS: add more options to configure catalog.xml
+  - Currently the default THREDDS configuration creates two default datasets, the Service Data dataset and the 
+    Main dataset. The Service Data dataset is used internally and hosts WPS outputs. The Main dataset is the
+    place where users can access data served by THREDDS. Both of these are configured to serve files with the following 
+    extensions: .nc .ncml .txt .md .rst .csv
+
+  - In order to allow the THREDDS server to serve files with additional extensions, this introduces two new
+    variables: 
+    - `THREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS`: this allows users to specify additional [filter
```
fmigneault, line 28: This will most probably need to be 2 variables to align with Magpie definitions.

CHANGES.md
```diff
@@ -15,7 +15,29 @@
...
+  - Currently the default THREDDS configuration creates two default datasets, the Service Data dataset and the 
+    Main dataset. The Service Data dataset is used internally and hosts WPS outputs. The Main dataset is the
+    place where users can access data served by THREDDS. Both of these are configured to serve files with the following 
+    extensions: .nc .ncml .txt .md .rst .csv
+
+  - In order to allow the THREDDS server to serve files with additional extensions, this introduces two new
+    variables: 
+    - `THREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS`: this allows users to specify additional [filter 
+        elements](https://docs.unidata.ucar.edu/tds/current/userguide/tds_dataset_scan_ref.html#including-only-desired-files) to the Service Data dataset. This is especially useful if a WPS 
+        outputs files with an extension other than the default (eg: .h5) to the `wps_outputs/` directory.
```
fmigneault, line 30: I'm curious about the use case - not against it, just making sure the intended use is appropriate.

Is there any advantage of exposing those HDF5 files via THREDDS rather than accessing them directly by the WPS-outputs dir? If anything, I would expect Nginx to provide much better/faster responses, as well potentially additional support of `Content-Range` requests if enabled.

birdhouse/components/thredds/default.env
```diff
@@ -17,7 +17,22 @@ export THREDDS_SERVICE_DATA_LOCATION_NAME='Birdhouse'
...
+      </metadata>
+
+      <filter>
+        <include wildcard="*.nc" />
+        <include wildcard="*.ncml" />
+        <include wildcard="*.txt" />
+        <include wildcard="*.md" />
+        <include wildcard="*.rst" />
+        <include wildcard="*.csv" />
+      </filter>
```
fmigneault, line 34: `THREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS` should probably be embedded by default within `THREDDS_DATASET_DATASETSCAN_BODY`? 

I think the original intention was to always update the 2 datasets with similar combinations of extensions. If not desired, the user can still override everything, but it does need for them to override everything if they only want to add 1 other extension via `THREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS`.

fmigneault (changes requested):
The definitions need to align with Magpie definitions to protect the contents accordingly. 

See https://github.com/bird-house/birdhouse-deploy/blob/3d7c8d64230c0798e4362239ff1064ef005ec681/birdhouse/components/thredds/config/magpie/providers.cfg.template#L11-L35
and 
https://pavics-magpie.readthedocs.io/en/latest/services.html#servicethredds
for details.

mishaschwartz:
@fmigneault 

> The THREDDS definitions need to align with Magpie definitions to protect the contents accordingly.

What was the intention for how the Magpie permissions were supposed to interact with additional catalogs introduced by setting the `THREDDS_ADDITIONAL_CATALOG` variable? 

Is there a solution in place for that already or do I have to come up with a solution that accounts for arbitrary catalog definitions as well?

fmigneault:
@mishaschwartz 

For Magpie permissions, it does not really care about how the catalogs (the `datasetScan` blocks) are defined. It works only with the resolved URL paths. The "`catalog`" is the default "browsing" `service` that THREDDS uses when navigating its hierarchy, and the URL are all formed as `/thredds/{service}/{nested-dirs...}`. So, on Magpie side, the resources are defined as `{thredds-service-type}/{nested-dirs-resources...}/{file-resource}` (see the `{service}` is omitted). Adding more catalogs would only mean to reflect these new directories how they are resolved by URL under the Magpie THREDDS service.

The `service` combinations are defined directly on the Magpie THREDDS service configuration. Depending on the "`prefixes`" (i.e.: the `service`), it handles `GET` requests either as `browse` or `read` permissions. All `service` and file extensions that are classified as providing metadata are typically `browse`, and the actual data access are `read`. In some edge cases, some services can be both (eg: WMS and WCS that can describe or get the data based on `request` parameter), and are therefore placed in the `read` category.

mishaschwartz:
@fmigneault 

Thanks for the detailed explanation. I understand the Magpie configuration and how its "prefix" definitions relate to the URLs in THREDDS.

My concern is more about whether we need to be able to customize the "file_patterns" definitions in the Magpie configuration files to handle duplicate file extensions other than .nc and .ncml

The other concern is that users can define custom service definitions if they'd like other than the ones listed here:

https://github.com/bird-house/birdhouse-deploy/blob/3d7c8d64230c0798e4362239ff1064ef005ec681/birdhouse/components/thredds/catalog.xml.template#L6-L15

Or they could potentially modify the `base` attribute so that the URL path no longer matches the prefix defined in Magpie.

I propose we either do:

- come up with some solution so that all catalogs must only employ the predefined services without modification (through sort of XML validation)
- write a bunch of warnings in the documentation that tell users not to modify any of the service definitions (easy to do but doesn't enforce anything)

I'm working on a solution but if you have any insight into this issue let me know

fmigneault:
If custom service types are added, they must be provided in the browse/read section accordingly for Magpie to grant/deny access to them as expected. Similarly, additional file patterns (or extensions) must also be provided.

If another location than `/twitcher/ows/proxy/thredds/{service}/` is employed, it will not be managed by Magpie/Twitcher. However, if the prefix uses the same `/twitcher/ows/proxy/thredds/` prefix (it should since it is under this THREDDS docker service), it will default to DENY access, unless "full-access" was granted on THREDDS service to an anonymous group. Therefore, I don't think modifying `base` is the right approach. (BTW, they should probably also inherit from `TWITCHER_PROTECTED_PATH` rather than hard-coded)

Rather than having `THREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS` directly with the XML, maybe we should have a `THREDDS_SERVICE_DATA_EXTRA_FILE_BROWSE_EXTENSIONS` and `THREDDS_SERVICE_DATA_EXTRA_FILE_READ_EXTENSIONS`, and generate `THREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS` from them? There are actually already some discrepancies (eg: missing `.md`, `.rst`, `.csv` in Magpie, but listed in THREDDS). A few warning/descriptions would help explain how to customize them.

birdhouse/components/thredds/default.env
```diff
@@ -74,5 +74,4 @@ export DELAYED_EVAL="
   THREDDS_SERVICE_DATA_LOCATION_ON_HOST
   THREDDS_IMAGE
   THREDDS_IMAGE_URI
-  THREDDS_DATASET_DATASETSCAN_BODY
```
mishaschwartz: Note to reviewers: This cannot be a delayed eval variable because of the way that the delayed evaluation process removes quotes. If we change the delayed eval process to allow this, it will break other functionality (especially in jupyterhub).

mishaschwartz:
After several iterations, I don't think that there is an easy way to get the flexibility we want by defining these variables and also enforce the Magpie settings as well. So the compromise I went with is to add some defaults and make the Magpie settings configurable as well so that they can be updated as needed. I also added some instructions/warnings about how to configure Magpie to match changes to THREDDS.

birdhouse/env.local.example
```diff
@@ -456,26 +456,85 @@ export GEOSERVER_ADMIN_PASSWORD="${__DEFAULT__GEOSERVER_ADMIN_PASSWORD}"
...
+# This may be useful if you are creating a catalog that only provides a subset of the services defined in the
+# compound service named "all" (see birdhouse/components/thredds/catalog.xml.template).
+# DO NOT define any non-compound services in THREDDS_ADDITIONAL_CATALOG that is not an exact copy of one of the
+# variables defined in "all"! Especially, do not change the "base" attribute of any existing service. 
+# Doing so may break the way that access permissions are enforced when accessing data through this service. 
+
+# Additional file filters to add for the Service Data THREDDS dataset. By default, the Service Data dataset will only
+# serve files with the following extensions: .nc .ncml .txt .md .rst .csv
+# If you need this dataset to serve other files you should update the THREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS to add
+# additional file filters.
```
fmigneault, line 488: Mention the corresponding `THREDDS_MAGPIE_...` variables as well?

fmigneault (commented):
I think the configuration is overcomplicated, and that is reflected in the example comments trying to explain what everything does in this process that is already complicated in itself. 

For example, if I wanted to configure THREDDS with only `fileServer` for `.nc` and `.txt`, and no other extension, there are way too many variables to override.

```env
export THREDDS_MAGPIE_EXTRA_METADATA_PREFIXES='".+\\.txt"'
export THREDDS_MAGPIE_EXTRA_DATA_PREFIXES='".+\\.nc"'
export THREDDS_DEFAULT_FILE_FILTERS=''
export THREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS='
    <include wildcard="*.txt" />
    <include wildcard="*.nc" />
'
```

And still, "somehow" override the Magpie providers file, since all "default" `metadata_type` and `data_type` I do not want would still be enabled otherwise. 

Is there really any advantage of having duplicate sets of `THREDDS_DEFAULT_...` and `THREDDS_..._EXTRA_...`  variables?
Can't everything be simplified by having a single one, which uses the `THREDDS_DEFAULT_...` values by default?

mishaschwartz:
> I think the configuration is overcomplicated

I think I agree that this has gotten out of hand.

The main issue is that I don't want to make it possible to break the service catalog which would break other things internally for the rest of the components in the stack. But I still don't fully understand how that is used...

> For example, if I wanted to configure THREDDS with only fileServer for .nc and .txt

I think that your set up here is overly complicated actually (which highlights your point). I don't think you'd need to set `THREDDS_MAGPIE_EXTRA_DATA_PREFIXES`  and I don't think we ever want people to modify `THREDDS_DEFAULT_FILE_FILTERS` (it's not provided in env.local.example as a settable option).

> Is there really any advantage of having duplicate sets...

I think I agree with this. But the defaults vs. the extras were requested in the discussion here https://github.com/bird-house/birdhouse-deploy/pull/472#discussion_r1790748137 Do you no longer think that's a concern?

fmigneault:
I think it might be worthwhile to remove the defaults vs extras duplication to make things easier for users in general. If one wants to preserve the defaults, it is easy to copy-paste its value and add the "extra" that is desired within a single variable.

birdhouse/components/thredds/default.env
```diff
@@ -17,7 +17,25 @@ export THREDDS_SERVICE_DATA_LOCATION_NAME='Birdhouse'
 export THREDDS_DATASET_URL_PATH='datasets'
 export THREDDS_SERVICE_DATA_URL_PATH='birdhouse'
 
+export THREDDS_MAGPIE_EXTRA_METADATA_PREFIXES='".+\\.txt", ".+\\.md", ".+\\.rst"'
+export THREDDS_MAGPIE_EXTRA_DATA_PREFIXES=''
 
+export THREDDS_SERVICE_DATA_EXTRA_FILE_FILTERS=''
```
fmigneault, line 23: How about using explicitly: 
```env
export THREDDS_MAGPIE_METADATA_PREFIXES='
[
  null,              # note: special YAML value evaluated as `no-prefix`, use quotes if literal value is needed
  "\\w+\\.gif",      # threddsIcon, folder icon, etc.
  "\\w+\\.ico",      # favicon
  "\\w+\\.css",      # tds.css
  "catalog\\.\\w+",  # note: special case for `THREDDS` top-level directory (root) accessed for `BROWSE`
  catalog,
  ncml,
  uddc,
  iso,
]
'
export THREDDS_MAGPIE_DATA_PREFIXES='
[
  fileServer,
  dodsC,
  wcs,
  wms,
  ncss,
]
'
```

With those variables defined by default next to the below `THREDDS_DATASET_DATASETSCAN_BODY`, it is easy for a user that desires to extend the capabilities to override the variables by a simple copy-paste. The doc can also more easily describe that overriding `THREDDS_MAGPIE_DATA_PREFIXES` for example would require the corresponding filter to be added to `THREDDS_DATASET_DATASETSCAN_BODY`. 

The issue I have with the current definitions is that modifying the `THREDDS_DATASET_DATASETSCAN_BODY` variable to remove the defaults will not reflect in the Magpie config unless it is completely overridden.

mishaschwartz: But we don't want users to remove any of the existing metadata prefixes or it will break THREDDS.

Metadata prefixes:
- `null`: without this they can't access `example.com/thredds/`
- `.gif`, `.ico`, `.css`: without this, the catalog pages won't render properly
- `catalog\\.\\w+`, `catalog`: without this we can't view any of the catalog.html or catalog.xml pages (see catalog services https://docs.unidata.ucar.edu/tds/current/userguide/services_ref.html#catalog-services)
- the rest: these are metadata services offered by thredds https://docs.unidata.ucar.edu/tds/current/userguide/services_ref.html#nciso-metadata-services

Data prefixes:
- these are data services offered by thredds https://docs.unidata.ucar.edu/tds/current/userguide/services_ref.html#data-access-services

If the user changes any of those they either can't access crucial parts of the THREDDS catalog or they access metadata as data or vice versa. I don't think we want to let users do that.

> The doc can also more easily describe that overriding THREDDS_MAGPIE_DATA_PREFIXES for example would require the corresponding filter to be added to THREDDS_DATASET_DATASETSCAN_BODY

But if they modify filters in `THREDDS_DATASET_DATASETSCAN_BODY` they don't have to update `THREDDS_MAGPIE_DATA_PREFIXES` since new file types can still be accessed by the same THREDDS services that are described by the existing prefixes. They only need to modify `THREDDS_MAGPIE_DATA_PREFIXES` if:

- they create a brand new THREDDS service with a different prefix (which we discourage)
- they want a specific file extension to always be treated as metadata or as data regardless of the THREDDS service they are using to access the data.

In both of these cases, they only will ever need to add more fields, not remove or modify existing ones.

> The issue I have with the current definitions is that modifying the THREDDS_DATASET_DATASETSCAN_BODY variable to remove the defaults will not reflect in the Magpie config unless it is completely overridden.

But it does not need to be overridden. 

For example, if we remove the ncml service from `THREDDS_DATASET_DATASETSCAN_BODY` then a user trying to access a file using this service will no longer see the associated ncml URL as an option in the catalog. This does not affect the magpie permissions since the URL that would trigger the relevant Magpie permissions is no longer being offered by THREDDS.

Another example, if we change the filters in `THREDDS_DATASET_DATASETSCAN_BODY` to include files with a `.json` extension then we also don't need to modify anything in Magpie because the current magpie configuration already covers the case where we want to view the file in the catalog (the metadata `catalog` prefix) and where we want to access the file as data (the data `fileServer` prefix).

As usual, please let me know if I've misunderstood something about how Magpie works.

fmigneault: Good point about the `catalog` and above items for rendering THREDDS browsing. Those can be left. 

For the others, they are relevant as hardcoded services because they happen to be enabled in https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/components/thredds/threddsConfig.xml.template

So, I was wondering, if anyone is trying to modify which services/files are available or not by the various variables, isn't there a chance they are also overriding that config to modify which file extensions are accessible by certain services or not, to avoid weird side-effects?

For example, if enabling `.h5`, is it really relevant for this to be accessed via `wms`, `wcs`, etc.? Anything else than `fileServer` will most probably cause random errors.

_**Note**_: I do not have a strong feeling about this proposal. Modifying enabled services is a strong configuration change that I'm not planing to do. I was just shooting out the idea that if someone does try to modify them, they probably need to override almost everything anyway, and it's probably easier to do so with full-overrides rather than many concatenations of disparate variables. That is maybe out of scope for this PR.

mishaschwartz: > For example, if enabling .h5, is it really relevant for this to be accessed via wms, wcs, etc.? Anything else than fileServer will most probably cause random errors.

Good question. In this example they would likely create a configuration that looks something like this:

```xml
<metadata inherited="true">
  <serviceName>fileServer</serviceName>
</metadata>

<filter>
  <include wildcard="*.h5" />
</filter>
```

This tells the `datasetScan` block to only serve files with an .h5 extension and only serve those with the `fileServer` service. 

This interacts with Magpie in the following ways:

- if a user tries to view the catalog for these files it will follow the Magpie rules for the `catalog\\.\\w+` prefix (browse permissions)
- if a user tries to access an .h5 file it will follow the Magpie rules for viewing the `fileServer` prefix (read permissions)
- if the user tries to access an .h5 file with another service, let's say `wms`, Magpie will follow the rules for viewing the `wms` prefix but then THREDDS will return a 404 error since there is no valid `wms` URL for the file.

The only time we need to modify the Magpie configuration is if we change the services themselves (which we discourage). However, with this change, that is technically possible by updating the  `THREDDS_MAGPIE_EXTRA_METADATA_PREFIXES` and `THREDDS_MAGPIE_EXTRA_DATA_PREFIXES` variables.

To summarize ... I don't think that users will really have to modify the Magpie configuration very much even if they're modifying the THREDDS configuration.

fmigneault: Ok. Sounds good. Let's keep it the way it is currently.

fmigneault (approved):

PR #475, opened 2024-11-13
Jupyter: update to release py311-241111

tlvu:
## Overview

Update Jupyter env to latest version, see https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/137 for more info.

## Changes

**Non-breaking changes**
- Update Jupyter env to version `py311-241111`.

**Breaking changes**
- This new Jupyter env requires the Jenkins host to be updated.
  Jenkins host upgrade problem found and fix on Ouranos side can be found here https://github.com/Ouranosinc/jenkins-config/pull/16.

## Related Issue / Discussion

- Resolves [issue id](url)

## Additional Information

- [ ] Things to do...

## CI Operations

birdhouse_skip_ci: false


tlvu:
> Looks good. I think there's just one more notebook that needs to be updated so that the tests pass

@mishaschwartz 
Sorry this error is misleading.  Bumping the Jupyter env here do not bump the Jupyter env used by Jenkins to actually running the tests.

It's the merge in the PR https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/137 that will make effective this new Jupyter env as the runtime env for Jenkins.

This PR is simply to bump the Jupyter env deployed on PAVICS.

Also a head-ups, Jenkins pipeline on CRIM side will be broken with the go-live of this new Jupyter env.  This new Jupyter env requires an updated kernel which requires an OS upgrade !  CRIM has already been notified for a while.

fmigneault:
Build: http://daccs-jenkins.crim.ca/job/DACCS-iac-birdhouse/2887/
Using:
- https://github.com/bird-house/birdhouse-deploy/tree/update-jupyter-env
- https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/tree/new-docker-build

tlvu:
https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/137 has been merged and the new Jupyter env is live now.

@fmigneault 
This Jenkins IAC pipeline will be broken until CRIM update the Jenkins host.

Jenkins host upgrade problem found and fix on Ouranos side can be found here https://github.com/Ouranosinc/jenkins-config/pull/16.

fmigneault:
@tlvu 
Patch version bump for a knowingly breaking jupyter-env version seems like a poor choice.
Should be a minor.

tlvu:
> @tlvu Patch version bump for a knowingly breaking jupyter-env version seems like a poor choice. Should be a minor.

Well the actual Jupyter env itself is not breaking for the end-user per say.  It's just a Jenkins host that need updated.  This changelog is for the end-user.

fmigneault:
Strongly disagree.

The PR changes the default `JUPYTERHUB_DOCKER_NOTEBOOK_IMAGES` which does a major bump of the underlying jupyterhub dependencies, which could break depending on the docker engine and OS where the platform is running. Even worst, nothing about that is explicitly mentioned in the CHANGES, so anyone doing an update as if it was a simple security/bug fix could potentially break their platform.

tlvu:
> The PR changes the default `JUPYTERHUB_DOCKER_NOTEBOOK_IMAGES` which does a major bump of the underlying jupyterhub dependencies

But that's how all the previous Jupyter env release was done.  I am surprised only now you bring this up.

Just to be clear, this release breaks on Jenkins but not on PAVICS.

For me that change log was for PAVICS only.

mishaschwartz:
I think @fmigneault's point is that if the default `JUPYTERHUB_DOCKER_NOTEBOOK_IMAGES` is enabled, a user's jupyterlab instance is now being run with your updated image without warning. Their notebook may now run differently because the underlying image/kernel is different.

If the image should only be changed for jenkins, then we should have a different mechanism for updating the image that jenkins uses to test the stack. By changing it here, we are making a user-facing change.

fmigneault:
Only now? 
I've been saying that since the moment that the jupyter env update was proposed, since it breaks the CI. How can I give a better example of this being a breaking change than the CI itself breaking for that exact reason? 

Any instance that was relying on `JUPYTERHUB_DOCKER_NOTEBOOK_IMAGES` with the workflow tests as the default image is breaking.

If PAVICS instance uses another default image, yeah it doesn't break that instance, but that's just Ouranos' instance. Still breaks all the others that don't explicitly override `JUPYTERHUB_DOCKER_NOTEBOOK_IMAGES`.

fmigneault:
@tlvu @mishaschwartz 
I made a temporary and ugly patch that (at least) lets the notebooks run on Jenkins in the meantime.
When opening PRs, the description should define: `birdhouse_daccs_configs_branch: pin-jupyter-env`
See outputs ⬆️ : https://github.com/bird-house/birdhouse-deploy/pull/475#issuecomment-2480337580

⚠️ Expect output errors in the results. 
I'm pinning this override: https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/blob/714f75205c125cfcb44b8fd8f795484303842e1f/Jenkinsfile#L13 
on branch https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/tree/pin-jupyter-env
Therefore, the actual outputs from the notebooks on `master` are not exactly as expect from back when 
that docker was employed.

Example of minor diffs that cause the failures: 

```
21:28:55    - <xarray.Dataset>
21:28:55    + <xarray.Dataset> Size: 418GB
21:28:55      Dimensions:       (rlat: 800, rlon: 706, time: 14245)
21:28:55      Coordinates:
21:28:55    -   * rlat          (rlat) float32 -46.17 -46.08 -45.99 ... 25.56 25.65 25.74
21:28:55    +   * rlat          (rlat) float32 3kB -46.17 -46.08 -45.99 ... 25.56 25.65 25.74
```

And, just to reiterate, I do still believe this does deserve more than a PATCH version release and more details in changelogs warning about it. It definitely wasn't just a simple patch to apply. 

Reference (for myself to find details): https://crim-ca.atlassian.net/browse/DAC-586

tlvu:
@mishaschwartz I think we have probably have a different definition of what is the end-user (intended audience) of the semver bump and by extension the changelog.

For me, the intended audience is the sysadmin responsible for maintaining the deployment of the stack.

I think you also include the scientist that actually will use the stack, the Jupyter server, as your intended audience.

For the sysadmin, he wants to know "do I need to do anything manual other than `git pull` and `./pavics-compose.sh up -d`" to pull the change.

Hence we document backward incompatible changes like manual scanning and updating Magpie DB when the unique email constraint was enforced or editing the `env.local` file when the component path changes for the sysadmin and we bump minor or major of the semver, but not patch, to signal to the sysadmin he has manual steps to perform.

For the scientist, those breaking DB constraints or component path change never impacted him.  In fact, all the previous minor and even major version bumps never impacted him.

So that's why I always thought our intended audience for those version bump and changelog file has always been the sysadmin.

@fmigneault when I said "But that's how all the previous Jupyter env release was done. I am surprised only now you bring this up." I meant since the past 4 years, every single Jupyter env release has been trying to update to the latest packages and accompany by patch version bump (because no manual upgrade step for the sysadmin).  This is not something new.

If your Jupyter users were broken by python version change or xarray or any other packages behavior change, it must have happened already several times in the past and you would already know about this, not just now.

Since you never bought this up in the past, but only this time, that's why I was surprised.

CRIM has different releases of your NLP and EO Jupyter image as well.  You never mentioned any compatibility warnings in the changelog when they change.

Given the stack should be generic and this Jupyter image is pretty Ouranos specific, how about we do not activate this by default in the Jupyterhub component and only list it in the `env.local.example`, like the extra NLP and EO images from CRIM, see https://github.com/bird-house/birdhouse-deploy/blob/1528ba14db5c476fca11033a97f7df99b74b0fc4/birdhouse/env.local.example#L264-L268

The sysadmin activating the Jupyterhub will have to choose the Jupyter image(s) and this won't change unless he decided to.

Previously the component Jupyterhub was enabled by default so we needed to have a default Jupyter image enabled so it works "out of the box".  But it is not the case anymore, so there is no real reason to enable an Ouranos specific image by default.

Thanks for bringing this up.  This is a little detail we missed when making the platform fully modular and minimalist.

mishaschwartz:
@tlvu 

Both the sysadmin and the scientist are users of this software.

If there are breaking changes for either of them, we need to make sure they are aware of those changes and how to update their configurations or code accordingly.

fmigneault:
@tlvu
I have no problem with the usual strategy of updating latest packages in the jupyter env in general. It's just this time, the upgrade implies a necessary OS update to run that default docker. 

In a generic CRIM instance (eg: hirondelle.crim.ca), I really don't care about the jupyter instance (I could even disable them entirely), because it isn't really the main use case for that server. However, the `pavics/workflow-tests` happens to be the instance that is running for the CI, which runs on our side.

As already obvious by my response time to update jenkins, it is really low in my list of to-dos, and sadly, because it affects the CI, this change affects all of us indirectly. Changing the default `JUPYTERHUB_DOCKER_NOTEBOOK_IMAGES` doesn't matter. To patch the CI, I had to override the checkout of the repository during the pipeline execution such that all the tests launch using the old docker (i.e.: https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/compare/master...pin-jupyter-env), regardless of the actual default docker-notebook configured on the server.

tlvu:
> Why don't we continue discussion about this in #476

Yes agreed.

PR #482, opened 2024-11-27
Deprecate `stac-populator` optional component

mishaschwartz:
## Overview

This component has not been functional since it was deprecated in https://github.com/crim-ca/stac-populator/pull/12

## Changes

**Non-breaking changes**
- deprecates non-functional component

**Breaking changes**
None

## Additional Information

Links to other issues or sources.

## CI Operations

birdhouse_daccs_configs_branch: master
birdhouse_skip_ci: false

PR #483, opened 2024-11-27
Correct docker image for `stac-populator` optional component

mishaschwartz:
## Overview

This sets the docker image for the `stac-populator` component to a version that actually contains the code that is executed when `stac-populator` is called. The previous image no longer contained the relevant code.

## Changes

**Non-breaking changes**
- stac-populator image changed to version 0.0.0

**Breaking changes**
None

## Related Issue / Discussion

- Closes #482

## CI Operations

birdhouse_skip_ci: false


tlvu (approved):
Thanks !!!  I never liked using "evolving" tags like `master` or `latest`.  It's just a timing bomb that causes surprise in the future.

fmigneault (commented):
Just planing ahead, should we add `${STAC_POPULATOR_VERSION}`, `${STAC_POPULATOR_DOCKER}` and `${STAC_POPULATOR_IMAGE}` variables as in the case of the other services?

fmigneault (approved):

PR #484, opened 2024-12-04
Enable local deployment to improve development and testing

mishaschwartz:
## Overview

  The Birdhouse stack can now be deployed locally and accessed on a browser on the host machine without the need 
  for an SSL certificate. This is useful for local development and for running tests against the full stack while developing
  and in CI environments.

  To enable this, add the new `optional-components/local-dev-test` component to `BIRDHOUSE_EXTRA_CONF_DIRS` and
  set the following environment variables in the local environment file:

  * `export BIRDHOUSE_FQDN=host.docker.internal`
  * `export BIRDHOUSE_HTTP_ONLY=True`

  You should also add ``host.docker.internal`` to your ``/etc/hosts`` file pointing to the loopback address so 
  that URLs generated by Birdhouse that refer to ``host.docker.internal`` will resolve properly in a browser:

  ```
  echo '127.0.0.1    host.docker.internal' | sudo tee -a /etc/hosts
  ```

  After deploying the stack, you can now interact with the Birdhouse software at ``http://host.docker.internal`` 
  from the machine that is the docker host.

  In order to implement the changes above, the following non-breaking changes have been made to the deployment code:

  - added a configuration variable `BIRDHOUSE_HTTP_ONLY` which is not set by default. If set to `True` the `proxy` component will only serve content over `http` (not `https`).
  - added the following configuration variables. These should not be set directly unless you really know what you're doing:
    - `BIRDHOUSE_PROXY_SCHEME`: default is `http` if `BIRDHOUSE_HTTP_ONLY` is `True`, otherwise `https`
    - `PROXY_INCLUDE_HTTPS`: default is unset if `BIRDHOUSE_HTTP_ONLY` is `True`, otherwise `include /etc/nginx/conf.d/https.include;`
  - changed the default values for the following configuration variables:
    - `BIRDHOUSE_ALLOW_UNSECURE_HTTP`: default is now `True` if `BIRDHOUSE_HTTP_ONLY` is `True`
  - logs are written to stderr by default. Previously they were written to stdout.
    - this allows us to call scripts and programmatically use their outputs. Previously log entries would need to be 
      manually filtered out before program outputs could be used.
    - added the `--log-stdout` and `--log-file` flags to the `bin/birdhouse` interface to allow redirecting logs to 
      stdout or to a specific file instead.
    - log redirection can also now be set using environment variables:
      - `BIRDHOUSE_LOG_FD` can be used to redirect logs to a file descriptor (ex: `BIRDHOUSE_LOG_FD=3`)
      - `BIRDHOUSE_LOG_FILE` can be used to redirect logs to file (ex: `BIRDHOUSE_LOG_FILE=/some/file/on/disk.log`)
      - Note that the variables here should not be set in the local environment file since that file is sourced **after**
        some logs are written. Instead, set these by exporting them in the parent process that calls `bin/birdhouse`.
    - for backwards compatibility, if scripts are not called through the `bin/birdhouse` interface, logs will still be 
      written to stdout.

## Changes

**Non-breaking changes**
- allows users to run the stack locally without the need for an ssl certificate
- logs are now written to stderr when called through the `bin/birdhouse` interface
- new flags for `bin/birdhouse` and environment variables to set logging options 

**Breaking changes**
- If you have an external script or component that you want to run with a locally deployed stack that contains a URL with a hardcoded scheme set to `https`, this may need to be changed to use the `BIRDHOUSE_PROXY_SCHEME` environment variables instead (replace `https` with `${BIRDHOUSE_PROXY_SCHEME}`)

## Related Issue / Discussion

- see a previous (failed) attempt at this in #361
- Further discussion:
  - Where else should we document this? Maybe in `CONTRIBUTING.rst`?
  - Should we also be adding some sort of `start-local` option to the Makefile?

## Additional Information

This change is necessary if we are serious about improving the testing infrastructure for this stack.

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/stac/config/proxy/conf.extra-service.d/stac.conf.template
```diff
@@ -1,10 +1,10 @@
 
-    location /stac {
+    location /stac/ {
```
fmigneault, line 2: I'm worried about this edit. 

I do not remember why it was omitting the `/`, but I remember having a lot of problems dealing with its redirects in STAC, because its own internal prefix redirect does weird stuff we cannot override explicitly, hence why there is dual `/stac` stuff happening. 

I would rather keep this a separate PR and test it more thoroughly.

birdhouse/components/twitcher/twitcher.ini.template
fmigneault, line 1: Let's use the `BIRDHOUSE_FQDN_PUBLIC` since that's what is intended.

birdhouse/components/weaver/config/canarie-api/canarie_api_monitoring.py.template
```diff
@@ -19,7 +19,7 @@ SERVICES['Weaver'] = {
         'releasenotes': 'https://github.com/crim-ca/weaver/blob/master/CHANGES.rst',
         'support': 'https://github.com/crim-ca/weaver/issues',
         'source': 'https://github.com/crim-ca/weaver',
-        'tryme': 'https://${BIRDHOUSE_FQDN_PUBLIC}/weaver/',
+        'tryme': '${BIRDHOUSE_PROXY_SCHEME}://${BIRDHOUSE_FQDN_PUBLIC}/weaver/',
```
fmigneault, line 22: Just saw that other configs use `/${WEAVER_MANAGER_NAME}` rather than `/weaver` explicitly, so we might want to adjust while at it.

bin/birdhouse
fmigneault, line 1: These log changes are fine on their own.
Depending on how the rest of SSL-related discussions go, could be a separate PR.

birdhouse/components/proxy/conf.d/frontend.conf.template
```diff
@@ -27,22 +27,4 @@ server {
...
-
-    include /etc/nginx/conf.d/all-services.include;
-
-    # redirect server error pages to the static page /50x.html
-    error_page   500 502 503 504  /50x.html;
-    location = /50x.html {
-        root   /usr/share/nginx/html;
-    }
-}
+${PROXY_INCLUDE_HTTPS}
```
fmigneault, line 30: See other comment in proxy `default.env` for more context.

I think it is better to leave it there regardless and let the HTTPS redirection or not deal with which "server" config gets hit by the requests.

birdhouse/components/proxy/docker-compose-extra.yml
```diff
@@ -12,12 +12,10 @@ services:
     image: ${PROXY_IMAGE}
     container_name: proxy
     ports:
-      - "80:80"
-      - "443:${PROXY_SECURE_PORT}"
```
fmigneault, line 16: I think the 443 could be left and is simply ignored.
Makes sense to keep this edit depending on other discussions regarding how the redirects are handled and how variables get defined.

birdhouse/scripts/logging.include.sh
fmigneault, line 1: I'm not sure the changes are entirely backward compatible. 

Before, the `1>&2` redirects were done for specific log levels (WARN and above). 
This allowed logging outputs to stdout or some file, and only potentially problematic messages to another file. It seems this distinction is not possible anymore, since all goes either to stdout, stderr or the logfile. 

Also, if we want both stdout/logfile, we need to ignore both options the script provides, and do `birdhouse [...] | tee logfile` ourselves. It seems like specifying both options should do the same as `tee` using both outputs simultaneously.

fmigneault (changes requested):
Not sure why https://github.com/bird-house/birdhouse-deploy/pull/361 was considered a failed attempt? My current dev setup is running the stack locally with a self-signed certificate, and everything works just fine. 

Personally, I have found that bypassing SSL via HTTP lead to code that worked locally but didn't once deployed on actual servers enforcing HTTPS, because requests are not doing exactly the same checks and code paths in requests libraries, and that introduces slight variations in what actually happens when resolving requests. 

See also other comments about the strategy employed for the HTTP access. 
I have tried using the current `master` branch with options already available, and I'm able to access the services in HTTP-only mode. Just need to have the dummy SSL cert file define to patch the volume mount (that on its own would be OK to adjust). I'm interested to see if the operations can be simplified. 

Also interested in @tlvu's approach for local dev.

mishaschwartz:
> Not sure why https://github.com/bird-house/birdhouse-deploy/pull/361 was considered a failed attempt?

See closing comment here https://github.com/bird-house/birdhouse-deploy/pull/361#issuecomment-1745270480 and the comment it references here https://github.com/bird-house/birdhouse-deploy/pull/361#discussion_r1334753368

> My current dev setup is running the stack locally with a self-signed certificate, and everything works just fine.

That's great! Can you share your setup with us so that we can replicate it? What is your `BIRDHOUSE_FQDN` variable set to? If it's localhost, how do you differentiate between the container's localhost and the host machine's localhost?

> and that introduces slight variations in what actually happens when resolving requests.

You're right and that's why testing in a production-like environment is always going to be useful. This helps developers easily test 99% of the functionality locally and avoids a lot of potential regressions early in development.

> Just need to have the dummy SSL cert file define to patch the volume mount

Which volume mount are you referring to here? The SSL cert file itself? What needs to be patched?

> Also interested in @tlvu's approach for local dev.

me too

birdhouse/components/monitoring/docker-compose-extra.yml
```diff
@@ -21,7 +21,7 @@ services:
     image: ${NODE_EXPORTER_IMAGE}
     container_name: node-exporter
     volumes:
-      - /:/host:ro,rslave
+      - /:/host:ro
```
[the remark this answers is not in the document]
mishaschwartz: This is not strictly necessary in production and it causes problems when deploying locally with Docker Desktop as it is affected by the way that Docker Desktop mounts containers to the VM that runs docker. 

I can make this configurable if needed.

birdhouse/components/jupyterhub/default.env
```diff
@@ -104,6 +104,7 @@ OPTIONAL_VARS="
   \$JUPYTER_IDLE_SERVER_CULL_TIMEOUT
   \$JUPYTER_IDLE_KERNEL_CULL_TIMEOUT
   \$JUPYTER_IDLE_KERNEL_CULL_INTERVAL
+  \$JUPYTERHUB_DOCKER_EXTRA_HOSTS
```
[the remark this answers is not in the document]
mishaschwartz: Sure, this is a variable that is only used internally so users really shouldn't set it. But I can add a comment here to help future developers out.

birdhouse/components/jupyterhub/jupyterhub_config.py.template
```diff
@@ -203,6 +203,7 @@ c.Spawner.args = [
 c.DockerSpawner.extra_host_config = {
     # start init pid 1 process to reap defunct processes
     'init': True,
+    'extra_hosts': dict(host_mapping.split(":") for host_mapping in "${JUPYTERHUB_DOCKER_EXTRA_HOSTS}".split())
```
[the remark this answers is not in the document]
mishaschwartz: yes

birdhouse/components/proxy/default.env
```diff
@@ -6,23 +6,31 @@ export PROXY_IMAGE="nginx:1.23.4"
...
 # Content of "location /" in file config/proxy/conf.d/all-services.include.template
 # Useful to have a custom homepage.
 # Note that the default homepage will become the jupyterhub login page if the jupyterhub component is enabled.
 # If the jupyterhub component is not enabled, it is highly recommended to create a custom homepage since the magpie
 # landing page is not the most user-friendly option.
-export BIRDHOUSE_PROXY_ROOT_LOCATION="return 302 https://\$host/jupyter/hub/login;"
+export BIRDHOUSE_PROXY_ROOT_LOCATION='return 302 ${BIRDHOUSE_PROXY_SCHEME}://\$host/jupyter/hub/login;'
 
-export INCLUDE_FOR_PORT_80='$([ x"$BIRDHOUSE_ALLOW_UNSECURE_HTTP" = x"True" ] && echo "include /etc/nginx/conf.d/all-services.include;" || echo "include /etc/nginx/conf.d/redirect-to-https.include;")'
+export PROXY_INCLUDE_FOR_PORT_80='$([ x"$BIRDHOUSE_ALLOW_UNSECURE_HTTP" = x"True" ] && echo "include /etc/nginx/conf.d/all-services.include;" || echo "include /etc/nginx/conf.d/redirect-to-https.include;")'
```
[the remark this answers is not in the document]
mishaschwartz: This change here is only changing the variable name to follow the naming convention introduced in 2.4.0. This doesn't need to be in the PR but as long as we're fixing things I brought the variable name up to date.

> Granted, the BIRDHOUSE_PROXY_SCHEME updates would be needed to report the http rather than https where it was harcoded

Yes, this is the important change.

> but it seems the https.include rewrite is not needed (the HTTPS server definition is just ignored).

It's needed if you don't have a self signed SSL certificate. Otherwise the proxy container won't start since nginx fails to parse a valid SSL certificate in the https block.

birdhouse/components/proxy/pre-docker-compose-up.include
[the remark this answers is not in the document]
mishaschwartz: Docker compose doesn't let you do that. See example:

```yaml
# docker-compose.yml
services:
  test:
    image: bash:5.1.4
    command: ls
    volumes:
      - "./test.txt:/test.txt"
      - ${VOLUME_MAPPING}

```

```
touch test.txt
touch test2.txt
VOLUME_MAPPING='./test2.txt:/test2.txt' docker compose up # this works
docker compose up # this doesn't: ERROR: invalid empty volume spec
```

If we instead change the syntax in the docker-compose.yml file to 

```yaml
...
    volumes:
      - "./test.txt:/test.txt"
      ${VOLUME_MAPPING}
```

Then we get the following syntax error in every case since the yaml itself is invalid: `could not find expected ':'`

If you have another strategy in mind. Please share.

birdhouse/optional-components/local-dev-test/pre-docker-compose-up.include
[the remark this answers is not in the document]
mishaschwartz: see https://github.com/bird-house/birdhouse-deploy/pull/484#discussion_r1871484079

birdhouse/components/stac/config/proxy/conf.extra-service.d/stac.conf.template
mishaschwartz: This has always bothered me and I'm 99% sure that this is an oversight and it always should have been `/stac/`.

But you're right, this isn't strictly related to this PR so I can remove it for now.

birdhouse/components/proxy/conf.d/frontend.conf.template
mishaschwartz: See comment here: https://github.com/bird-house/birdhouse-deploy/pull/484#discussion_r1871460056

Specifically: 

>> but it seems the https.include rewrite is not needed (the HTTPS server definition is just ignored).

> It's needed if you don't have a self signed SSL certificate. Otherwise the proxy container won't start since nginx fails to parse a valid SSL certificate in the https block.

birdhouse/components/proxy/docker-compose-extra.yml
mishaschwartz: > I think the 443 could be left and is simply ignored.

True, but better not to open up ports if it's not necessary

> Makes sense to keep this edit depending on other discussions

Sure that makes sense

birdhouse/scripts/logging.include.sh
mishaschwartz: > Before, the 1>&2 redirects were done for specific log levels (WARN and above).

This is not exactly true. **Some** logs at WARN level and above were redirected to stderr but other's weren't. The previous approach was inconsistent. 

> This allowed logging outputs to stdout or some file, and only potentially problematic messages to another file. 

This can be implemented but it should be done in `logging.include.sh`, not on an ad hoc basis in the files that call the logger. I'm happy to implement this functionality if it's useful.

> It seems like specifying both options should do the same as tee using both outputs simultaneously.

Sure, we can make that work

fmigneault:
> See closing comment here https://github.com/bird-house/birdhouse-deploy/pull/361#issuecomment-1745270480 and the comment it references here https://github.com/bird-house/birdhouse-deploy/pull/361#discussion_r1334753368

I'm not sure why specifically that comment would be the cause of the PR to be invalid (specifically regarding the self-signed SSL certificate approach). 

In that case, the proposed `magpie.url = http://magpie:2001` didn't make sense because that is the exposed service URL, so `magpie:2001` is not a valid hostname/port from the host. However, using `https://${BIRDHOUSE_FQDN}/magpie` (current config) with a self-signed SSL certificate works (that's what I do on my dev instance). 

> Can you share your setup with us so that we can replicate it?  What is your BIRDHOUSE_FQDN variable set to? If it's localhost, how do you differentiate between the container's localhost and the host machine's localhost?

The self-signed SSL certificate is obtained using the following steps.
(note: some metadata values to adjust based on your use case / what you want the cert to indicate)

```shell
host="$${1:-localhost}"
sudo mkdir -p /etc/nginx/ssl
sudo openssl req -x509 -nodes -days 10000 -newkey rsa:2048 \
    -keyout /etc/nginx/ssl/${host}.key \
    -out /etc/nginx/ssl/${host}.crt \
    -subj "/CN=$${host}/C=CA/ST=Quebec/L=Montreal/O=CENTRE DE RECHERCHE INFORMATIQUE DE MONTREAL INC." \
    2>/dev/null
sudo cat /etc/nginx/ssl/${host}.crt /etc/nginx/ssl/${host}.key | sudo tee /etc/nginx/ssl/cert.pem 1>/dev/null
sudo chmod 0600 /etc/nginx/ssl/*
```

And then, adjust the env file with 
```shell
export BIRDHOUSE_SSL_CERTIFICATE="/etc/nginx/ssl/${host}.pem"
```

In my specific case, I want to keep the DACCS "localhost" distinct from other local dev apps I'm working on.
Therefore, I use `BIRDHOUSE_FQDN=<my.vm.ip.addr>`.

Using `BIRDHOUSE_FQDN_PUBLIC=localhost` or `127.0.0.1` seems to work in some cases, but using those could cause ambiguity if a service inside a docker calls `https://localhost/...` because it thinks it is itself (inside docker) rather than the host server, so better to avoid those values altogether.

I'm guessing you could also use another custom host by editing the `/etc/hosts` as proposed in the PR, but I didn't need to do that in my case (the local IP is already accessible).

All other configs are as per usual.

> You're right and that's why testing in a production-like environment is always going to be useful.

Yes. Regardless of the approach selected, everything needs to be tested in an actual deployment state. However, I find using a self-signed SSL certificate is closer to the real behavior of the deployed server than bypassing them entirely. If you get weird SSL-related issues, you get them right away while developing rather than waiting all the way until deployment breaks. That is a much longer development feedback loop.

birdhouse/scripts/logging.include.sh
fmigneault: If some >WARN logs were not redirected, those were unintended bugs.

I agree it should be implemented in `logging.include.sh`. 

Which levels to log under with std/file output could be controlled by different CLI options, so it doesn't use any hardcoded redirects.

birdhouse/components/proxy/default.env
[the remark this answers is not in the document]
fmigneault: Sorry. My comment wasn't clear.
The variable rename is fine for alignment with 2.x, and so is the `BIRDHOUSE_PROXY_SCHEME` addition.

What I was referring to is the contents of the variable. I placed the comment here because I found it better illustrated the `BIRDHOUSE_ALLOW_UNSECURE_HTTP=True` I was referring to.

Agreed, the SSL certificate file is needed for `nginx` config to resolve.
I think that if you use the `BIRDHOUSE_ALLOW_UNSECURE_HTTP=True` approach, you don't even need the contents to be an actual SSL cert file. It just needs to be "something" for the volume mount to work (but I'm not sure, to validate).

Might be worth checking out if the following works:
```bash
touch ssl-dont-care.pem

BIRDHOUSE_ALLOW_UNSECURE_HTTP=True \
BIRDHOUSE_SSL_CERTIFICATE="$(realpath ssl-dont-care.pem)" \
	birdhouse compose up -d 
```

Otherwise, the self-signed certificate described in https://github.com/bird-house/birdhouse-deploy/pull/484#issuecomment-2521037780 can be used to have a more "legit" file.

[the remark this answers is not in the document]
mishaschwartz: Oh I see...
Yes, it has to be a well-formatted certificate file or else nginx complains and won't start.

tlvu (approved):
LGTM.

Let me summarize the high-level changes here, correct me if I am wrong.

With self-signed SSL cert, the difficulty is inter-container communication.  Between the Proxy and the browser, the browser can accept the self-signed cert, no problem there.  But between Twitcher and the Proxy (to get to Magpie) for example, self-signed cert breaks down because Twitcher sees the self-signed cert from Proxy as invalid.

You solve the inter-container problem by making all communications `http` instead of `https`.

But since inter-container communications is going through the Proxy, it uses the `BIRDHOUSE_FQDN` and not the container name so you had to set `BIRDHOUSE_FQDN=host.docker.internal` and add that extra hostname to all the containers so they can reach the Proxy.

Just curious, have you explored to make inter-container communication not going through the Proxy but using directly the internal container name so internally it will default to `http` anyways and no more problems with self-signed cert for inter-container?  Or somehow make all the containers trust the self-signed cert and keep using `https` for inter-container?

I previously also wanted to use self-signed cert for development, that's why the flag `VERIFY_SSL` was introduced.  Then I quickly faced the inter-container communication problems with self-signed cert.

Then I also had the need to let other access my test VM for demo.  So my test VM is accessible on the internet.  Since it is on the internet, I can use real LetsEncrypt cert and don't have to worry about self-signed anymore.  The production also uses LetsEncrypt cert so I prefer my test VM to also use it so I catch problems with LetsEncrypt earlier, if any.

Downsides with my method are 
  1) I need to ask IT to open my VM on the internet (port 80 and 443)
  2) my VM has to be at work and not on my laptop because can not open 80 and 443 with residential internet provider (well my laptop is not powerful enough to run the full stack anyways so personally this downside did not affect me).

So I can see your method is an improvement over my 2 downsides.

As for redirecting all STDOUT to STDERR for parsing the output, I am not sure how to use it yet.

birdhouse/components/proxy/docker-compose-extra.yml
```diff
@@ -12,12 +12,10 @@ services:
...
+      - "80:${PROXY_HTTP_PORT}"
```
tlvu, line 15: I've been searching the diff in this PR, is `PROXY_HTTP_PORT` configurable or always hardcoded to 80?  I only see it hardcoded.

If always hardcoded, might as well add a comment here that his variable is not configurable, to not give false impression to other.

tlvu:
> > Also interested in @tlvu's approach for local dev.
> 
> me too

I use Vagrant VM with real LetsEncrypt SSL cert for my local dev.  The constrain is my VM has to be at work and not on my laptop so I can open port 80 and 443 on the internet to get LetsEncrypt SSL cert.

The requirements https://github.com/bird-house/birdhouse-deploy/blob/225f3c35e91952a6133dbd87ea3d91d95722acd5/birdhouse/README.rst#requirements and quick-start https://github.com/bird-house/birdhouse-deploy/blob/225f3c35e91952a6133dbd87ea3d91d95722acd5/birdhouse/README.rst#quick-start sections document this.

birdhouse/components/proxy/docker-compose-extra.yml
tlvu: Oh, going over the PR again, `PROXY_HTTP_PORT` is not new.  It is just a rename of `PROXY_SECURE_PORT` which is also hardcoded and not configurable and is a trick to force people to use the wrapper `pavics-compose.sh` and not call `docker-compose directly.

Forget my previous comment.

birdhouse/birdhouse-compose.sh
```diff
@@ -123,12 +123,17 @@ if [ x"$1" = x"up" ]; then
...
+      . "$COMPONENT_PRE_COMPOSE_UP_INCLUDE"
+    fi
   done
 fi
 
 log INFO "Executing docker-compose with extra options: $* ${COMPOSE_EXTRA_OPTS}"
-# the PROXY_SECURE_PORT is a little trick to make the compose file invalid without the usage of this wrapper script
-PROXY_SECURE_PORT=443 HOSTNAME=${BIRDHOUSE_FQDN} docker-compose ${COMPOSE_CONF_LIST} $* ${COMPOSE_EXTRA_OPTS}
+# the PROXY_HTTP_PORT is a little trick to make the compose file invalid without the usage of this wrapper script
+PROXY_HTTP_PORT=80 HOSTNAME=${BIRDHOUSE_FQDN} docker-compose ${COMPOSE_CONF_LIST} $* ${COMPOSE_EXTRA_OPTS}
```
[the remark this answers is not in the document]
tlvu: Resolved, see comment https://github.com/bird-house/birdhouse-deploy/pull/484#discussion_r1871938579

mishaschwartz:
> How did you make all the containers, or at least Twitcher and Magpie, trust the self-signed cert from the Proxy?

@tlvu 
I'm pretty sure if you set `BIRDHOUSE_VERIFY_SSL=false` then twitcher won't care about the ssl certificate.

However, this doesn't solve the problem in JupyterLab where all requests to `BIRDHOUSE_FQDN_PUBLIC` need to skip SSL verification. So for example, you need to do `requests.get(os.getenv("BIRDHOUSE_HOST_URL"), verify=False)` everywhere.

This means that your notebook code run locally is going to look drastically different from your code run with a proper SSL certificate.

Contrast this with the `http` option where your code can look like  `requests.get(os.getenv("BIRDHOUSE_HOST_URL")` in development and production and it will still work.

fmigneault:
@tlvu 
Yes. I am not observing the issue about inter-docker connectivity that was mentioned. On my local dev instance, everything runs as if it was deployed on any other server, but using self-signed cert to simulate the HTTPS verification. I don't think I do anything special beside using an explicit IP to avoid the dockers being confused by `localhost`/ `127.0.0.1`.

mishaschwartz:
> The self-signed SSL certificate is obtained using the following steps.

Thanks @fmigneault I can confirm that this works for me as well (with some minor changes to the permissions due to some specifics of my VM). 

It mostly works as intended but there are a few issues with this approach as described here https://github.com/bird-house/birdhouse-deploy/pull/484#issuecomment-2521348979

birdhouse/components/monitoring/docker-compose-extra.yml
```diff
@@ -21,7 +21,7 @@ services:
     image: ${NODE_EXPORTER_IMAGE}
     container_name: node-exporter
     volumes:
-      - /:/host:ro
+      - /:/host:ro,${NODE_EXPORTER_HOST_BIND_PROPOGATION}
```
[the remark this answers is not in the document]
mishaschwartz: No, the propagation setting is `rslave` already (on the master branch). But this breaks things if you deploy the stack using Docker Desktop so this makes this propagation setting optional

mishaschwartz:
> SSL is enabled:

Alright, them I'm with @tlvu  on this one... how is twitcher not complaining about your self-signed certificate?

fmigneault:
I'm passing it: 
```yaml
services:
  twither:
    environment:
      REQUESTS_CA_BUNDLE: "${SSL_CERTIFICATE}"
    volumes:
      - "${SSL_CERTIFICATE}:${SSL_CERTIFICATE}:ro"
```

(repeat for any relevant services)

mishaschwartz:
> I'm passing it:

Very interesting. Is this documented anywhere?

If this change is required, my point here (https://github.com/bird-house/birdhouse-deploy/pull/484#issuecomment-2521371247) still stands since we're still diverging from a production-like environment to make this work.

tlvu:
> I'm passing it:
> 
> ```yaml
> services:
>   twither:
>     environment:
>       REQUESTS_CA_BUNDLE: "${SSL_CERTIFICATE}"
>     volumes:
>       - "${SSL_CERTIFICATE}:${SSL_CERTIFICATE}:ro"
> ```
> 
> (repeat for any relevant services)

@fmigneault 
OH !!!  Very interesting.  So this way, inter-container comm think the self-signed cert is actually valid !  Nice hack.  Please document this !


> However, this doesn't solve the problem in JupyterLab where all requests to `BIRDHOUSE_FQDN_PUBLIC` need to skip SSL verification. So for example, you need to do `requests.get(os.getenv("BIRDHOUSE_HOST_URL"), verify=False)` everywhere.
> 
> This means that your notebook code run locally is going to look drastically different from your code run with a proper SSL certificate.

@mishaschwartz A while back when I was still trying to get self-signed cert to work, I added this to Jenkins https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/blob/5d816cf9582e8db84ad62d5c5e01918c1b8a8af4/runtest#L37-L39 to avoid warning about self-signed cert.  Hopefully it still works.

mishaschwartz:
> A while back when I was still trying to get self-signed cert to work, I added this to Jenkins

Thanks @tlvu. This suppresses the warning message but there's still the error that gets raised unfortunately

tlvu:
> > A while back when I was still trying to get self-signed cert to work, I added this to Jenkins
> 
> Thanks @tlvu. This suppresses the warning message but there's still the error that gets raised unfortunately

Just curious, what kind of error?   Can you share?   It's been a very long time since I attempted to make self-signed cert work.

fmigneault:
@mishaschwartz @tlvu 
It's not documented anywhere in birdhouse. I had even forgotten this was the trick before actually digging in the docker-compose config output. Feel free to document it for future reference (maybe even self-document it via code as the way to implement the local-dev feature if that seems easier).

mishaschwartz:
> Just curious, what kind of error? Can you share? It's been a very long time since I attempted to make self-signed cert work.

@tlvu It's a `SSLError` if you're using the `requests` library and a `URLError` if using `urllib`

```python
>>> import requests
>>> import os
>>> requests.get(os.getenv("BIRDHOUSE_HOST_URL") + "/stac")
...
SSLError: HTTPSConnectionPool(host='<my-ip-addr>', port=443): Max retries exceeded with url: /stac (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate (_ssl.c:1091)')))
>>> import urllib.request
>>> urllib.request.urlopen(os.getenv("BIRDHOUSE_HOST_URL") + "/stac")
URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate (_ssl.c:1091)>
```

mishaschwartz:
Ok so I just want to summarize a few things here:

#### Strategies for supporting local development:

||self-signed certificate|certificate from LetsEncrypt|http using host.docker.internal|
|-|-|-|-|
|used by:|@fmigneault|@tlvu|@mishaschwartz|
|set up:|- create a self-signed certificate<br>- set `REQUESTS_CA_BUNDLE` env variable in all containers as described here (https://github.com/bird-house/birdhouse-deploy/pull/484#issuecomment-2521402666)|- ensure host machine has a static public IP<br>- get an SSL certificate from Let's Encrypt|- add `host.docker.internal` as an extra host on all containers (optional if using Docker Desktop)<br>- send requests over `http`|
|security risks:|- exposes SSL certificates to jupyterlab containers (users can see private key in plain text)|- no additional|- creates an environment where sending requests over http is possible|
|possible security mitigation:|- we could add a bunch of warnings if self-signed certs are used<br>- we could create a development environment which must be enabled for the code to permit self-signed certificates in this way|- N/A|- we could add a bunch of warnings if http is used<br>- we could create a development environment which must be enabled for the code to permit http in this way|
|documentation:|- None (should be added)|- [README](https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/README.rst)|- documented in this PR|
|additional comments|- your local IP changes if your network changes. This means you have to change `BIRDHOUSE_FQDN` if you're developing at home vs. at the office, etc.<br>- requires some additional code changes to set `REQUESTS_CA_BUNDLE` variables|- does not work from most residential internet provider plans (can't set static public IP)<br>- requires no additional code changes|- requires changes described in this PR to work<br>- does not require any additional set up to create SSL certificates<br>- works from any network location without the need for any extra steps|

My personal view on this is as follows:

- "certificate from LetsEncrypt" is too difficult to set up and doesn't actually allow development from home (a lot of developers including myself work from home)
- "self signed certificate" and "http using host.docker.internal" are both good options, both introduce some security risks that can be mitigated and both diverge from a production environment.
- I would prefer to use "http using host.docker.internal" for the following reasons:
    - it is easier to set up, this lowers the barrier to entry for future developers
    - it allows me to work both from home and at work easily
- I would be open to also documenting the self-signed SSL certificate option so that developers can choose to use that one instead _if they wish_.

fmigneault:
I don't mind either `REQUESTS_CA_BUNDLE` or the `host.docker.internal` approach, but I am not a fan of the (already complicated to read) nginx config being split up only for the purpose of local dev. The 443/HTTPS server portion can just be left where it was and would be ignored since requests use `http` anyway.

IMO, if the code needs to be adjusted only to support local dev, that is not a good justification to introduce changes.

I think the security aspects are not really problematic in either case since those are intended for local development and are purposely bypassing SSL checks with an override variable. You have to know what you're doing to be making setting those configs. We can add warnings (like the variables that printed when default security values are left unchanged) to be clear.

tlvu:
> I don't mind either `REQUESTS_CA_BUNDLE` or the `host.docker.internal` approach, but I am not a fan of the (already complicated to read) nginx config being split up only for the purpose of local dev. The 443/HTTPS server portion can just be left where it was and would be ignored since requests use `http` anyway.
> 

I think Francis has a point here.

I also see Misha point to not open unnecessary port since this will be directly on his laptop and he could be working in public spaces (airport, hotel, coffee shop, ...)

I let you guys decide but if you decide to restore the 443 into the main `docker-compose-extra.yml`, leave the new `pre-docker-compose-up.include` there, it could be useful someday in the future.

> IMO, if the code needs to be adjusted only to support local dev, that is not a good justification to introduce changes.

Yes this makes sense.

However, as someone who have previously also spent quite a long time trying to get a fully working dev environment, I also understand Misha's desire to have an easier time setting up a local dev env.

How about adding a warning the for the `optional-components/local-test-dev` that standard notebooks in the Jenkins suite will not work due to hardcoded `https` URL but at least Misha will be able to have his own test notebooks that will be able to hit and validate his stack.  At least Misha will have some stack running instead of nothing.

Let's consider this local-test-dev as the poor-man sharing, an intermediate working solution with some drawbacks.  It's not ideal but it's better than having nothing at all.

Misha, in order to use LetsEncrypt, all you need is a VM or machine stationary at work.  If no space for additional VM, you can buy a used server for really cheap.  My machine at work was actually a used machine.  Yes, you heard me right.  When I started at Ouranos, I actually requested a used machine to leave budget for more RAM and HDD because I knew I would be runnings VM.

Let's take this one for example https://deltaserverstore.com/product/dell-precision-5820/, Upgraded cpu to 6 cores, ram to 64G, disk to 1T SSD, it's $455 at the time of this writing.  This machine can easily allow you to run 3 concurrent VMs at the same time.  You'll need to find a network port that is in the DMZ, reserve some IP at work, open port 80 and 443 on the internet and you're set.  I think your needs justify to make this expense request at work.

This would be my suggestion once you've pushed the limit of local-test-dev.

mishaschwartz:
@fmigneault 
> The 443/HTTPS server portion can just be left where it was and would be ignored since requests use http anyway.

I'm not sure that's true since you are then forced to create a valid SSL certificate that you aren't going to use. I can see if there are other ways to work around this that don't require these files to be split up.

> IMO, if the code needs to be adjusted only to support local dev, that is not a good justification to introduce changes.

Are the changes you're objecting to here about splitting up the nginx config files here or introducing the BIRDHOUSE_PROXY_SCHEME variable?

>  We can add warnings ....

Agreed and I'm happy to add those

mishaschwartz:
@tlvu 

> This would be my suggestion once you've pushed the limit of local-test-dev.

Yes, I have a used computer at work that is set up exactly as you describe.

If we're serious about this being a community developed project then this is too high of a bar for entry to allow people to develop this software with us.
If we require that people configure a dedicated machine to properly develop this stack, no one will do it.

tlvu:
> Yes, I have a used computer at work that is set up exactly as you describe.

Great !  I wasn't aware of this.

> If we're serious about this being a community developed project then this is too high of a bar for entry to allow people to develop this software with us. If we require that people configure a dedicated machine to properly develop this stack, no one will do it.

Agreed !  That's why I already approved this PR.  We should give choices and not force someone into a particular workflow.

As for the Jenkins testsuite, in the current state, all the notebooks and the matching Jupyter image are pretty Ouranos specific.  So I would not count this as a blocker for not adopting a deployment/development workflow because the Jenkins testsuite is not working. 

Jenkins is pretty modular enough to run other notebooks from any repos as well so it's not a complete dead-end if someone wants to integrate Jenkins into their workflow.

mishaschwartz:
>> The 443/HTTPS server portion can just be left where it was and would be ignored since requests use http anyway.

> I'm not sure that's true since you are then forced to create a valid SSL certificate that you aren't going to use. I can see if there are other ways to work around this that don't require these files to be split up.

I can confirm that a valid SSL certificate is required in this case. I can think of more complicated ways to do this but the current proposed solution is the least complicated method I can think of. 

I really don't think that this solution is that complex. It's exactly what we do for all-services.include, cors.include, etc. and all the files are in one place. It's not like it's complicated for a developer to look at one more file when they already have to look at 4+ to understand the configuration.

birdhouse/scripts/logging.include.sh
mishaschwartz: Changes are now fully backwards compatible as of e655404d79bfbae4bf03d8865a76acacd9cb13ff

birdhouse/components/proxy/pre-docker-compose-up.include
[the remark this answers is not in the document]
fmigneault: OK. Thanks for investigating.

fmigneault (commented):
I think we only need to document the alternatives from https://github.com/bird-house/birdhouse-deploy/pull/484#issuecomment-2523413110. The rest seems good to go.

bin/birdhouse
```diff
@@ -18,14 +29,19 @@ Commands:
...
+  -b, --backwards-compatible                              Run in backwards compatible mode
+  -e, --env-file string                                   Override the local environment file, default is ${COMPOSE_DIR}/env.local
+  -s, --log-stdout                                        Write logs to stdout, default is to write to stderr
+  -s, --log-stdout {DEBUG|INFO|WARN|ERROR|CRITICAL}       Write logs to stdout for the given log level only, default is determined by --log-stdout (this option can be repeated)
+  -l, --log-file path                                     Write logs to this file path, default is to write to stderr
+  -l, --log-file {DEBUG|INFO|WARN|ERROR|CRITICAL} path    Write logs to this file path for the given log level only default is determined by --log-file (this option can be repeated)
+  -q, --quiet                                             Do not write logs to stdout or stderr. Logs will still be written to a file if --log-file is set
+  -q, --quiet {DEBUG|INFO|WARN|ERROR|CRITICAL}            Do not write logs to stdout or stderr for the given log level only, default is determined by --quiet (this option can be repeated)
```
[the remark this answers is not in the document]
mishaschwartz: Thanks, I've updated the description to be clearer.

> If I do the following, what happens exactly? 

Almost... all log levels will be written to stdout, errors will be written to errors.log and all other log levels will be written to all.log

Is that a use-case that you want? To be able to write to multiple log files simultaneously?

mishaschwartz:
> I think we only need to document the alternatives from https://github.com/bird-house/birdhouse-deploy/pull/484#issuecomment-2523413110. The rest seems good to go.

Great! I'll work on that today and re-request a review shortly.

bin/birdhouse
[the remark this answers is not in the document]
fmigneault: That is fine if that's how it is implemented. I just wanted to be sure. I wasn't certain if additional `--log-file` worked as extra fd redirects or override fd per matching level.

birdhouse/README.rst
```diff
@@ -294,17 +294,32 @@ Starting and managing the lifecycle of the VM:
...
+
+- ``export BIRDHOUSE_FQDN=host.docker.internal``
+- ``export BIRDHOUSE_HTTP_ONLY=True``
+
+This will allow you to access the Birdhouse software in a browser on your local machine using 
+the URL ``http://host.docker.internal`` without the need for an SSL certificate or to expose ports 80 and 443 
+publicly.
+
+Use a Self-Signed SSL certificate
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
fmigneault, line 319: Need to validate the output. I think RST will complain if the `^` don't span as far as the text.

mishaschwartz: It works as it is:

https://birdhouse-deploy--484.org.readthedocs.build/en/484/deployment.html#use-a-self-signed-ssl-certificate

read-the-docs always builds for each PR and you can see it by clicking the "Details" link here: 

image

fmigneault: Ok. Might be only the linters that complain about it in that case.

PR #487, opened 2024-12-13
fix incorrect geoserver URL in cowbird configuration

mishaschwartz:
## Overview

The Geoserver URL in the Cowbird configuration was incorrect as it used a non-standard port.

This change removes the port so that the URL will resolve properly. This allows Cowbird to execute webhooks that target Geoserver.

## Changes

**Non-breaking changes**
- Bug fix

**Breaking changes**
- None

## Additional Information

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/cowbird/config/cowbird/config.yml.template
```diff
@@ -14,8 +14,7 @@ handlers:
     admin_password: ${MAGPIE_ADMIN_PASSWORD}
   Geoserver:
     active: true
-    # FIXME: replace by the protected/proxy endpoint once component is made available
-    url: http://${BIRDHOUSE_FQDN_PUBLIC}:8765/geoserver
+    url: http://${BIRDHOUSE_FQDN_PUBLIC}/geoserver
```
fmigneault, line 17: Shouldn't that be either the internal access (`http://geoserver:????`) or using HTTPS?
A mix of `http` and `${BIRDHOUSE_FQDN_PUBLIC}` seems incorrect.

Ideally, it would use the internal connection to avoid the unnecessary loop-back.

mishaschwartz: So I thought so too but the above comment is `"replace by the protected/proxy endpoint once component is made available"` which says to me that someone intentionally wanted this to be the public URL. 

I'm happy to change it to the `http://geoserver:...` version since I don't see why we have to use the public one really.

fmigneault: Yes, that's an old comment from myself related to the removal of protected ports. I think at that time GeoServer was still accessed by HTTP:PORT so it couldn't be updated. Most probably not applicable anymore. 

Since Cowbird acts "behind the scene" with admin-only access specific to GeoServer, I think it's safe to skip the extra Magpie/Twitcher layer.

tlvu (approved):
LGTM as I do not really know how Cowbird integrate with Geoserver.

CHANGES.md
```diff
@@ -15,7 +15,13 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Fixes
+
+- Fix incorrect Geoserver URL in Cowbird configuration
+
+  The Geoserver URL in the Cowbird configuration was incorrect as it used a non-standard port.
+  This change removes the port so that the URL will resolve properly. This allows Cowbird to execute
```
fmigneault, line 23: The message needs adjustment based on latest change.
Good to merge after.

fmigneault (approved):

PR #491, opened 2025-01-07
Add integration test framework

mishaschwartz:
## Overview

This update adds a framework for testing the deployed stack using pytest. This will allow developers to check
that their changes are consistent with the existing stack and to add additionally tests when new functionality
is introduced. 

Changes to implement this include:

- existing unit tests are moved to the `tests/unit/` directory
- new integration tests are written in the `tests/integration/` directory. More tests will be added in the
  future!
- `conftest.py` scripts updated to bring the stack up/down in a consistent way for the integration tests.
- unit tests updated to accomodate new testing infrastructure as needed.
- unit tests updated to test logging outputs better
- `birdhouse` interface script updated to support testing infrastructure (this should not change anything for
  other end-users).
- additional documentation added to `birdhouse` interface to improve user experience.
- docker healthchecks added to more components so that the readiness of the stack can be determined with or
  without the use of the `canarie-api` component.

Next steps:

- add more integration tests as needed
- add a framework for testing migrating the stack from one version to another 

## Changes

**Non-breaking changes**
- Adds integration tests and healthchecks

**Breaking changes**
- None

## Related Issue / Discussion

- Introduces a mechanism for testing the stack locally so that we don't need to rely on integration tests run by notebooks through Jenkins. Since the Jenkins CI tests have been broken lately.
- This also gives developers a mechanism to tests their work locally and establishes a simple way to write integration tests (without the need to wrap them in jupyter notebooks).
- This PR only introduces the framework for creating these tests. Once this is approved, the actual work of writing the tests can start.
- Further discussion/planning is needed if we want to eventually run these tests in the CI pipeline

## CI Operations

birdhouse_skip_ci: false


mishaschwartz:
closed in favour of #491 which is an exact copy but using a source branch from this repository so that the CI actions will run properly.

PR #494, opened 2025-01-20
fix invalid weaver healthchecks

fmigneault:
## Overview

Fix Weaver healthchecks.

## Changes

**Non-breaking changes**
- Weaver: adjust missing `WEAVER_MANAGER_NAME` in healthchecks expecting matching `SCRIPT_NAME` configuration

  * Weaver's own `healthcheck` definition did not resolve to the expected endpoint.
  * Canarie-API `monitoring` endpoint for Weaver did not resolve to the expected endpoint.

**Breaking changes**
- n/a

## Related Issue / Discussion

- relates to https://github.com/bird-house/birdhouse-deploy/pull/492

## CI Operations

birdhouse_skip_ci: false


mishaschwartz (approved):
Just to double check... did the previous healthcheck pass when it shouldn't have? Because it doesn't seem to fail.

fmigneault:
> Just to double check... did the previous healthcheck pass when it shouldn't have? Because it doesn't seem to fail.

Both Weaver Docker and the Canarie API report 500 on my end. 

{E2CD4082-CA9A-4D99-820A-73972F4CBF92}

On `/canarie/node/service/stats`

{39F92DF8-B6A0-4ABC-8A17-D37537CCFCE4}

I'm also investigating because we are working on the new DACCS-jenkins instance to bring back the CI with latest Jupyter kernel, but the initial check of the server's status on CanarieAPI's endpoint fails since around the time https://github.com/bird-house/birdhouse-deploy/pull/492 was merged.

birdhouse/components/weaver/config/canarie-api/canarie_api_monitoring.py.template
```diff
@@ -26,7 +26,7 @@ SERVICES['Weaver'] = {
     "monitoring": {
         "Weaver": {
             'request': {
-                'url': 'http://weaver:4001/'
+                'url': 'http://weaver:4001/${WEAVER_MANAGER_NAME}/'
```
tlvu, line 29: Pour votre info, Ouranos production https://pavics.ouranos.ca/canarie/node/service/stats have the old                 `'url': 'http://weaver:4001/` and it reports success.

tlvu (approved):
Same as Misha, I find it weird that the old monitoring url without the weaver context root also worked !

But if you think this update is better, I have no objections.

birdhouse/components/weaver/config/canarie-api/canarie_api_monitoring.py.template
fmigneault: Ouranos server is on 1.42.2, so it doesn't include the changes from https://github.com/bird-house/birdhouse-deploy/releases/tag/2.8.0 and https://github.com/bird-house/birdhouse-deploy/releases/tag/2.7.3 that, when combined cause the problem, but not detected when tested separately on their own.

https://github.com/bird-house/birdhouse-deploy/blob/18929112c0ef14a210d47cb5a31f3242d2deee8c/birdhouse/components/weaver/docker-compose-extra.yml#L21-L25

fmigneault: @mishaschwartz ⬆️ 

This why you didn't notice it as wrong. 
The healthcheck was created before that change (so it worked), while I was testing Weaver before it was integrated, so never saw the error.

mishaschwartz: Ah right! That makes sense

PR #500, opened 2025-03-06
deprecate docker-compose default command

mishaschwartz:
## Overview

Sets the default compose command to `docker compose` using the new `DOCKER_COMPOSE` environment variable. This uses compose v2 by default (instead of v1 which is very old and has been deprecated for a while). For backwards compatibility, you can set `DOCKER_COMPOSE=docker-compose` in the local environment file to use the previous default.

## Changes

**Non-breaking changes**

**Breaking changes**
- If you do not have a version of docker that has a compose subcommand, you will have to set `DOCKER_COMPOSE=docker-compose` in env.local

## Additional Information

This means that new installations will not need to download a separate `docker-compose` binary.

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,14 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Changes
+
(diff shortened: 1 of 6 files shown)


fmigneault (approved):

mishaschwartz:
I didn't create a new version for this because I think we should do a minor version update that includes this PR, #504, and #506 since they all involve deprecating old code.

PR #503, opened 2025-03-07
Make scheduler jobs configurable

mishaschwartz:
## Overview

The scheduler component automatically enables three jobs (autodeploy, logrotate, notebookdeploy). If someone wants to use the scheduler component but does not want these jobs, there is no obvious way to disable any one of these jobs.

This change makes it possible to enable/disable jobs as required by the user and adds documentation to explain how to do this.

**Breaking Change**: 
- the three jobs that were automatically enabled previously are now no longer enabled by default.
- to re-enable these three jobs, source the relevant `*.env` files at the end of your local environment file
      (see documentation for further details)

What about... ?
  - just schedule these jobs for a non-existant day like February 31st?
      - Answer: This would technically work but is not obvious to the user. It is better to make this explicit.
  - just set the schedule to the `'#'` string?
      - Answer: This is a hack that would work based on the specific way that the docker-crontab image sets schedules.
                However, this is not obvious to the user and is unreliable since it is not documented.

## Changes

**Non-breaking changes**

**Breaking changes**
- See above

## Additional Information

I would really like to use the scheduler for other things but I do not want to have to enable the autodeploy mechanism.
It seems like the scheduler component was designed with the autodeploy mechanism in mind since that is added by default but it could be much more useful if it was more configurable.

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/scheduler/pre-docker-compose-up.include
```diff
@@ -0,0 +1,3 @@
+if [ -z "${BIRDHOUSE_SCHEDULER_NOTEBOOKDEPLOY_JOB}${BIRDHOUSE_SCHEDULER_AUTODEPLOY_JOB}${BIRDHOUSE_SCHEDULER_LOGROTATE_JOB}${BIRDHOUSE_AUTODEPLOY_EXTRA_SCHEDULER_JOBS}" ] ; then
+    log WARN 'The scheduler job has been enabled but no jobs are configured to be run. Please enable some jobs or disable the scheduler'
```
fmigneault, line 2: Maybe use `components/scheduler` in the message to be explicit about what it referred here, just in case there are other "scheduler" logs happening somewhere else.

birdhouse/components/README.rst
```diff
@@ -8,6 +8,76 @@ Birdhouse Components
...
 =========
 
+The scheduler component runs specific jobs on a schedule. This is similar to using the `cron`
+service but this runs in docker containers and is specifically designed to interact with the
+Birdhouse stack.
+
+Scheduler jobs can be enabled by sourcing specific ``*.env`` files in the ``components/scheduler``
+directory. The jobs that come included with the Birdhouse code are as follows:
+
+* Enable automatic deployment:
```
fmigneault, line 18: Spacing for all of these does not seem to render correctly.
Probably needs a newline between the nested lists/code blocks.

I could see an anchor to specific subsections of the doc being useful here, so maybe consider using `^^^` header/separators here as well.

birdhouse/components/README.rst
```diff
@@ -8,6 +8,76 @@ Birdhouse Components
...
+    .. code-block:: shell
+      . $COMPOSE_DIR/components/scheduler/autodeploy_job.env
+
+* Automatically rotate logs:
+  * to enable this component add the following to your local environment file:
+    .. code-block:: shell
+      . $COMPOSE_DIR/components/scheduler/logrotate_job.env
+
+* Automatically update tutorial Jupyter notebooks:
+  * note that this only can be enabled if the ``jupyterhub`` component is also enabled
```
fmigneault, line 30: Should the dependency be applied dynamically when sourcing its env file?

This could be relevant if the sourced env makes use of a `jupyterhub`-specific config that must be sourced first.

Same concept of the below `thredds` cases.

birdhouse/components/scheduler/config.yml.template
```diff
@@ -1,54 +1,6 @@
...
-    --env JUPYTERHUB_USER_DATA_DIR=${JUPYTERHUB_USER_DATA_DIR}
-    --env CODE_OWNERSHIP=${CODE_OWNERSHIP}
-    --env BIRDHOUSE_LOCAL_ENV=${BIRDHOUSE_LOCAL_ENV}
-    --env BIRDHOUSE_COMPOSE=${BIRDHOUSE_COMPOSE}
-    --env BIRDHOUSE_LOG_DIR=${BIRDHOUSE_LOG_DIR}
-    --env AUTODEPLOY_SILENT=true${AUTODEPLOY_PLATFORM_EXTRA_DOCKER_ARGS}
-  image: 'pavics/docker-compose-git:docker-18.09.7-compose-1.25.1'
+${BIRDHOUSE_SCHEDULER_NOTEBOOKDEPLOY_JOB}
+${BIRDHOUSE_SCHEDULER_AUTODEPLOY_JOB}
+${BIRDHOUSE_SCHEDULER_LOGROTATE_JOB}
```
fmigneault, line 4: I don't mind this approach particularly, but it seems somewhat a random combination whether specific scheduler config-vars are used vs the `BIRDHOUSE_AUTODEPLOY_EXTRA_SCHEDULER_JOBS` one. 

Should the provided scheduler configs all use explicit variables, and `BIRDHOUSE_AUTODEPLOY_EXTRA_SCHEDULER_JOBS` would be left empty for users to easily extend with their private configs as they deem fit, without dealing with prepend/append other ones added dynamically?

mishaschwartz: Yeah this wasn't my first choice either but I opted for not changing the original env scripts too much. I can revisit this and make it more explicit though... let me revisit this option.

mishaschwartz: Ok done

birdhouse/components/scheduler/deploy_data_job.env
```diff
@@ -96,17 +108,8 @@ $BIRDHOUSE_AUTODEPLOY_EXTRA_SCHEDULER_JOBS
 
 fi
 
+)
```
fmigneault, line 111: I suggest adding a comment here (similar to the opening `(`) because the subshell and nested YAML within multiline `echo ""` is hard to track without indents. Maybe add a comment before the `echo` as well.

tlvu (changes requested):
Very nice PR.  I spotted some suspicious change below.

Have you tried enabling all the 6 new `.env` file and see the cron daemon fire the jobs according to schedule (ie, no syntax error with the generation of the jobs)?

birdhouse/components/scheduler/deploy_data_job.env
```diff
@@ -96,17 +108,8 @@ $BIRDHOUSE_AUTODEPLOY_EXTRA_SCHEDULER_JOBS
...
-DEPLOY_DATA_JOB_SCHEDULE=""
-DEPLOY_DATA_JOB_JOB_NAME=""
-DEPLOY_DATA_JOB_CONFIG=""
-DEPLOY_DATA_JOB_CHECKOUT_CACHE=""
-DEPLOY_DATA_JOB_LOGFILE=""
-DEPLOY_DATA_JOB_JOB_DESCRIPTION=""
-DEPLOY_DATA_JOB_DOCKER_IMAGE=""
-DEPLOY_DATA_JOB_GIT_SSH_IDENTITY_FILE=""
-DEPLOY_DATA_JOB_DOCKER_RUN_EXTRA_OPTS=""
-
```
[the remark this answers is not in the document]
mishaschwartz: I see, that should be handled in the code that calls it though. I'll make that change

birdhouse/components/README.rst
```diff
@@ -8,6 +8,96 @@ Birdhouse Components
 Scheduler
 =========
 
+The scheduler component runs specific jobs on a schedule. This is similar to using the `cron`
```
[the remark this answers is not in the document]
mishaschwartz: OK I made a few small changes and it should be working now

birdhouse/components/scheduler/logrotate.conf
```diff
@@ -0,0 +1,9 @@
+/var/log/birdhouse/*.log {
+    rotate ${BIRDHOUSE_SCHEDULER_LOGROTATE_ROTATE_DAILY_COUNT}
```
tlvu, line 2: Accidental commit of this file?

birdhouse/components/scheduler/deploy_xclim_testdata_to_thredds.env
```diff
@@ -1,10 +1,14 @@
...
 
-DEPLOY_DATA_JOB_SCHEDULE="7,37 * * * *"  # UTC
-DEPLOY_DATA_JOB_JOB_NAME="deploy_xclim_testdata_to_thredds"
-DEPLOY_DATA_JOB_CONFIG="${COMPOSE_DIR}/deployment/deploy-data-xclim-testdata-to-thredds.yml"
-DEPLOY_DATA_JOB_JOB_DESCRIPTION="Auto-deploy Xclim testdata to Thredds for Finch and Xclim tutorial notebooks."
+print_xclim_cron_config() {
+    DEPLOY_DATA_JOB_SCHEDULE="7,37 * * * *"  # UTC
+    DEPLOY_DATA_JOB_JOB_NAME="deploy_xclim_testdata_to_thredds"
+    DEPLOY_DATA_JOB_CONFIG="${COMPOSE_DIR}/deployment/deploy-data-xclim-testdata-to-thredds.yml"
+    DEPLOY_DATA_JOB_JOB_DESCRIPTION="Auto-deploy Xclim testdata to Thredds for Finch and Xclim tutorial notebooks."
```
tlvu, line 8: Even if you do this, I think it's still much safer to reset all the vars at the end of `deploy_data_job.env` because if any external custom cron job do not follow this pattern, there will be cross contamination between successive invocations.

birdhouse/components/scheduler/notebookdeploy_job.env
```diff
@@ -9,7 +9,7 @@ export BIRDHOUSE_SCHEDULER_NOTEBOOKDEPLOY_JOB="
     --rm --name notebookdeploy
     --volume /var/run/docker.sock:/var/run/docker.sock:ro
     --volume ${BIRDHOUSE_LOG_DIR}:${BIRDHOUSE_LOG_DIR}:rw
-    --volume ${COMPOSE_DIR}:${COMPOSE_DIR}:ro
+    --volume \${COMPOSE_DIR}:\${COMPOSE_DIR}:ro
```
tlvu, line 12: I was referring to all the `$` some are escaped and some are not, not just `$COMPOSE_DIR`.  I don't remember to what the difference between escaping or not.  What happen when you test those?

birdhouse/components/README.rst
```diff
@@ -12,6 +12,9 @@ The scheduler component runs specific jobs on a schedule. This is similar to usi
 service but this runs in docker containers and is specifically designed to interact with the
 Birdhouse stack.
 
+Available jobs
+-------------
```
tlvu, line 16: The file is rendering properly on Github now.  So it's the addition of this 2nd level that fixes the rendering?

birdhouse/components/README.rst
```diff
@@ -76,14 +85,18 @@ as well as the individual ``*.env`` files listed above.
...
+
 To add custom jobs to the scheduler component, update the ``BIRDHOUSE_AUTODEPLOY_EXTRA_SCHEDULER_JOBS``
 environment variable in the local environment file to contain a YAML string that describes the job to run.
 
 For example a simple additional job might look like:
 
 .. code-block:: shell
 
-  export BIRDHOUSE_AUTODEPLOY_EXTRA_SCHEDULER_JOBS="
+  if [ -z "$(echo "$BIRDHOUSE_AUTODEPLOY_EXTRA_SCHEDULER_JOBS" | grep 'example job')" ]; then
```
tlvu, line 98: Nice, putting the pattern directly in the example.

tlvu (changes requested):

birdhouse/components/scheduler/config.yml.template
```diff
@@ -1,54 +1,9 @@
...
-  schedule: '${BIRDHOUSE_AUTODEPLOY_PLATFORM_FREQUENCY}'
-  command: '${COMPOSE_DIR}/deployment/triggerdeploy.sh ${COMPOSE_DIR}'
-  dockerargs: >-
-    --rm --name autodeploy${BIRDHOUSE_AUTODEPLOY_EXTRA_REPOS_AS_DOCKER_VOLUMES}
-    --volume /var/run/docker.sock:/var/run/docker.sock:ro
-    --volume ${BIRDHOUSE_LOG_DIR}:${BIRDHOUSE_LOG_DIR}:rw
-    --volume ${COMPOSE_DIR}/..:${COMPOSE_DIR}/..:rw
-    --volume ${BIRDHOUSE_AUTODEPLOY_DEPLOY_KEY_ROOT_DIR}:${BIRDHOUSE_AUTODEPLOY_DEPLOY_KEY_ROOT_DIR}:ro
-    --volume ${JUPYTERHUB_USER_DATA_DIR}:${JUPYTERHUB_USER_DATA_DIR}:rw
-    --volume ${BIRDHOUSE_LOCAL_ENV}:${BIRDHOUSE_LOCAL_ENV}
```
tlvu, line 43: This is the previous code, there are no `$` escape.  Is there any reasons to start escaping now?

birdhouse/components/scheduler/pre-docker-compose-up.include
```diff
@@ -0,0 +1,3 @@
+if [ -z "${BIRDHOUSE_SCHEDULER_NOTEBOOKDEPLOY_JOB}${BIRDHOUSE_SCHEDULER_AUTODEPLOY_JOB}${BIRDHOUSE_SCHEDULER_LOGROTATE_JOB}${BIRDHOUSE_AUTODEPLOY_EXTRA_SCHEDULER_JOBS}" ] ; then
+    log WARN 'The components/scheduler job has been enabled but no jobs are configured to be run. Please enable some jobs or disable this component.'
```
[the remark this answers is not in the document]
tlvu: I just though about this, now that you added 3 more built-in explicit vars ... those 3 new have to be added here as well !

birdhouse/components/scheduler/notebookdeploy_job.env
mishaschwartz: Yes, this is intended.
Some variables are set when this file is sourced and others are set by environment variables passed to the container. This is the previous behaviour but made explicit here.

Is this the best way to do this? Probably not... but I'm trying to not break backwards compatibility too much. If you're OK with me breaking backwards compatibility even more I'm sure I can think of a better way to construct these YAML fields.

birdhouse/components/scheduler/config.yml.template
mishaschwartz: Yes.. see https://github.com/bird-house/birdhouse-deploy/pull/503#discussion_r1991448930

birdhouse/components/scheduler/notebookdeploy_job.env
mishaschwartz: You know what... I'm going to make this better/more inline with how the stack actually works now... hold on a bit I'm going to add some more changes

tlvu: FYI it's been a while I touched this code but maybe the all the env var passed to the container are probably for extra invocations by the entrypoint script.  It's probably not to keep the volume-mount directives as env var.

However, this is prior to the recent changes that all scripts will source `read-configs.include.sh` so maybe we need less of those env var.  Only a real test will confirm this.

birdhouse/components/scheduler/deploy_xclim_testdata_to_thredds.env
tlvu: > Resetting the values to an empty string will override any values that the user explicitly wants to set. 

Yes, that is why they are all prefixed with `DEPLOY_DATA_JOB_` so someone using this knows what they are doing.

> You're assuming one very specific way to call these scripts which I don't think is the most obvious to most users.

I was assuming something like in `env.local.example`:

```
. $COMPOSE_DIR/components/scheduler/deploy_raven_testdata_to_thredds.env
. $COMPOSE_DIR/components/scheduler/deploy_data_job.env

(need a reset in the middle)

. $COMPOSE_DIR/components/scheduler/deploy_xclim_testdata_to_thredds.env
. $COMPOSE_DIR/components/scheduler/deploy_data_job.env

(...)
```

You have a new way with a function and you also assume a certain way to call this. Is this more obvious to most users?

To keep back-ward compatibility with the previous ways I would like to keep the reset because these vars are prefixed so someone will not just randomly use them for no reason.

External repos with existing external jobs are using the old way.  I'd rather not force the changes to all the existing external jobs.

> If I explicitly set a variable in a previous script I do not want a file that I source later on to magically unset that variable.

I agree.  But in this specific case, someone setting those vars are intended only to be used by `deploy_data_job.env` and all those vars are prefixed properly to avoid name clash or mis-understanding.

birdhouse/components/README.rst
fmigneault: Yeah. RST doesn't use a specific character to infer heading levels, but rather, relies on the order by which they are specified in the file.

fmigneault (approved):

mishaschwartz:
@fmigneault 

I know you already approved this but I completely changed the approach so you probably want to take another look

birdhouse/components/scheduler/config.yml.template
mishaschwartz: This is no longer relevant since [b923730](https://github.com/bird-house/birdhouse-deploy/pull/503/commits/b923730b6b8c1befd35677bab7946444c794e63a)

birdhouse/components/scheduler/notebookdeploy_job.env
mishaschwartz: This is no longer relevant since [b923730](https://github.com/bird-house/birdhouse-deploy/pull/503/commits/b923730b6b8c1befd35677bab7946444c794e63a)

birdhouse/components/scheduler/deploy_xclim_testdata_to_thredds.env
mishaschwartz: This is no longer relevant since [b923730](https://github.com/bird-house/birdhouse-deploy/pull/503/commits/b923730b6b8c1befd35677bab7946444c794e63a)

birdhouse/optional-components/scheduler-job-deploy_raven_testdata/config.yml.template
```diff
@@ -0,0 +1,14 @@
+- name: deploy_raven_testdata_to_thredds
```
tlvu, line 1: So for all the "deploy-data" job style, you are repeating the boilerplate config.yml ?  This is the reason that `deploy_data_job.env` exists: to generate the boilerplate with different schedule, name, command, and so on.

This is so that if the `deploy-data` script change, we have one single location to update the boilerplate template.

How about have one single boilerplate template in the `components/scheduler`, a new file next to `deploy_data_job.env` and use that boilerplate template to generate all `.yml` file for jobs that use `deploy-data` script?

birdhouse/optional-components/scheduler-job-renew_letsencrypt_ssl_cert/config.yml.template
```diff
@@ -0,0 +1,15 @@
...
+    --rm --name renew_letsencrypt_ssl
+    --volume /var/run/docker.sock:/var/run/docker.sock:ro
+    --volume ${COMPOSE_DIR}:${COMPOSE_DIR}:ro
+    --volume ${BIRDHOUSE_LOG_DIR}:/var/log/birdhouse:rw
+    --volume ${SCHEDULER_JOB_RENEW_SSL_CERT_SSL_CERT_DIR}:${SCHEDULER_JOB_RENEW_SSL_CERT_SSL_CERT_DIR}:rw
+    --env COMPOSE_DIR=${COMPOSE_DIR}
+    --env FORCE_CERTBOT_E2E=1
+    --env CERTBOT_RENEW=1
+    --env CERTBOTWRAPPER_LOGFILE=/var/log/birdhouse/renew_letsencrypt_ssl.log ${SCHEDULER_JOB_RENEW_SSL_CERT_LOCAL_ENV_DIR_VOLUME}
+  image: 'pavics/docker-compose-git:docker-18.09.7-compose-1.25.1'
```
tlvu, line 15: Please make this image configurable as well, same reason as the autodeploy, for compatibility with the other deprecate docker-compose v1 PR.

Also do not hard code `/var/log/birdhouse`.  I think we have a variable for that.

For consistency, please do a mass find and replace for that `/var/log/birdhouse`.

tlvu (changes requested):
Much like this component approach better and thanks a lot for keeping backward compat with jobs in external repos still using the older `BIRDHOUSE_AUTODEPLOY_EXTRA_SCHEDULER_JOBS`

I'll have to try this on out to be sure our numerous jobs still using the old style still work.

birdhouse/components/scheduler/deploy_data_job.env
```diff
@@ -2,8 +2,26 @@
...
+# $BIRDHOUSE_AUTODEPLOY_EXTRA_SCHEDULER_JOBS
+# $(print_cron_config)
+# "
 ##############################################################################
 
+# WARNING!! Using this file to update BIRDHOUSE_AUTODEPLOY_EXTRA_SCHEDULER_JOBS is deprecated.
+#           Please convert any custom jobs that use this script to components. See the following
+#           components for examples on how to do this:
+#           - optional-components/scheduler-job-deploy_raven_testdata
+#           - optional-components/scheduler-job-deploy_xclim_testdata
```
tlvu, line 21: Since we have a new way using "components", can you revert this file to the previous state to avoid all the external existing jobs?  Here are some of our external jobs using `deploy-data`, we have more that are private: https://github.com/bird-house/birdhouse-deploy-ouranos/tree/master/scheduler-jobs

You can keep this warning.

mishaschwartz:
> Is the change fairly transparent if the optional-components are applied?

Yes it should be completely transparent.

> Could this be considered a major change? 

I definitely think this deserves a minor version bump but not a major one.

birdhouse/components/scheduler/entrypoint
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+cat /scheduler-job-configs/*.yml > /opt/crontab/config.yml
```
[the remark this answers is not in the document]
mishaschwartz: As far as I'm aware, the way that docker bind-mounts work is as follows:

- docker registers the inode of a file or directory that it wants to mount to a container
- if the inode refers to a directory, the inode itself will contain a list of inodes of the directory's children. If one of the children's content changes or an inode is added/removed from the list, the change is reflected in the container right away.
- if the inode refers to a file and the files content changes, the inode itself will stay the same so the change will be reflected in the container
- if the inode refers to a file and the inode changes (the file path on the host is deleted and recreated) then the inode that docker is aware of no longer exists and the container needs to be restarted for the container to be aware of the new inode at that path.

When we change templated files, we're only updating their content, not the inode so this means that we don't actually have to restart the containers for the file change to be visible in the container.

However, in the case of the scheduler (and other components), the files that are mounted are **only read when the container starts** so to modify the behaviour of the container, you have to modify the file and then restart the container so the new configuration options are applied. 

I hope that makes sense

birdhouse/optional-components/scheduler-job-renew_letsencrypt_ssl_cert/config.yml.template
mishaschwartz: /var/log/birdhouse is the location in the container. The location on the host is determined by the `BIRDHOUSE_LOG_DIR` which is used in the volume mount above.

mishaschwartz: > Please make this image configurable as well, same reason as the autodeploy, for compatibility with the other deprecate docker-compose v1 PR.

done!

birdhouse/optional-components/scheduler-job-deploy_raven_testdata/config.yml.template
tlvu: Let me think of something, I'll submit a PR on top of this PR.

mishaschwartz:
@tlvu please don't force push to an open PR, especially if multiple people are working on it. I have to reset my local branch now instead of being able to just pull your changes.

birdhouse/README.rst
```diff
@@ -7,16 +7,23 @@ Docker instructions
 Requirements
 ^^^^^^^^^^^^
 
-* Centos 7, RockyLinux 8, Ubuntu 18.04, 20.04, 22.04, known to work.
+* Centos 7, RockyLinux 8, Ubuntu 20.04, 22.04 known to work.
```
fmigneault, line 10: ubuntu 24 as well :)

fmigneault (approved):

mishaschwartz:
@tlvu 

Can you please let me know if you're ok with this PR so that we can merge it since we have a few others that are dependent on this one.

If you're not going to have time to review it more then can we please pull it in and we can make additional changes later if needed. I don't think this will affect PAVICS since you're still running version 1.42.1

tlvu (approved):
@mishaschwartz 

Very sorry for the delay.  I took a few days off the past 2 weeks.

The only thing I didn't like was the job repetition for all "deploy-data" jobs (deploy xclim and raven testdata).  These are just example usage of "deploy-data" style of jobs.

In-house we have a ton of other "deploy-data" jobs and it's much more convenient to update the jobs to match any deploy-data code changes when all the jobs are generated from the same template instead of having to separately update 10+ separate jobs in 6-7 different external repos.  Having one common template will make job update completely transparent for all external jobs in all external repos.

I agree this is not show stopper.  I will send a separate PR for this.  I was hoping I could send this PR sooner but all the back-compat config vars problems prevented me from really testing autodeploy so my attention was to fix autodeploy first.

I quickly skimmed over the PR again and there some env var removed I am not sure, I probably missed during my first review.  The rest is fine.

tlvu:
> @tlvu please don't force push to an open PR, especially if multiple people are working on it. I have to reset my local branch now instead of being able to just pull your changes.

@mishaschwartz 
Total mistake.  I thought I was on my perso branch (I branched off your branch for a common job template).  I usually use force push only when a branch is not shared with other.

birdhouse/components/scheduler/docker-compose-extra.yml
```diff
@@ -1,15 +1,14 @@
...
-    image: pavics/crontab:0.6-rc2
+    image: ${SCHEDULER_IMAGE}
     container_name: scheduler
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock:ro
-      - ./components/scheduler/config.yml:/opt/crontab/config.yml:ro
-    environment:
-      COMPOSE_DIR: ${PWD}
-      BIRDHOUSE_AUTODEPLOY_DEPLOY_KEY_ROOT_DIR: ${BIRDHOUSE_AUTODEPLOY_DEPLOY_KEY_ROOT_DIR}
-      CODE_OWNERSHIP: ${BIRDHOUSE_AUTODEPLOY_CODE_OWNERSHIP:-}
```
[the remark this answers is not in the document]
mishaschwartz: They're added in templated files so they're not needed as environment variables in the scheduler container directly.

tlvu:
> Did you bring the stack down before you migrated to the new version?

Did not bring the entire stack down, but I did delete the `scheduler` container before `./pavics-compose.sh up -d`.   That was not enough?

mishaschwartz:
Ok, there's nothing special about those files, they're just template files like we have elsewhere in the stack.

I'm not really sure what would be different about those ones but in general the code I mentioned here (https://github.com/bird-house/birdhouse-deploy/pull/503#issuecomment-2956886848) in birdhouse-compose.sh should handle that case where directories are created.

If you figure out why that isn't working as intended please let me know so we can add a fix.

In the meantime, I'd recommend always bringing the stack down before you update the source code just so you know that you're starting fresh and docker doesn't get put in an unexpected state.

mishaschwartz:
What is in the non-empty directories?

> Am I the only person with all of these enabled at the same time? 

I have everything enabled too

tlvu:
> So the issue is that it's owned by root?

Oh true ... but that's not supposed to happen since I have `AUTODEPLOY_CODE_OWNERSHIP="1000:1000"` in `env.local`.

That's probably because it was created by the volume-mount because the instantiated template file is not there.  And the instantiated was not there because the parent is also owned by root.  Weird permission problem.

I can now start the stack finally.

PR #504, opened 2025-03-12
Remove compose v1 syntax

mishaschwartz:
## Overview

The version 1 compose specification that docker uses is out of date and no longer maintained by docker.
We currently have a mix of version 1 syntax and version 2 (specifically 2.20.2+) syntax in our docker compose files.

This means that the stack will not properly run with a docker compose version <2.20.2. For any version >2.20.2 the stack runs properly but displays lots of deprecation warnings about deprecated version strings and external volume and network definitions.

This PR updates all version 1 syntax so that these deprecation warnings are not displayed. Documentation has been updated to make this dependency on a modern version of docker explicit.

**Breaking Change**: as of birdhouse-deploy version 2.7.3, the stack could not be deployed with a docker compose version <2.20.2. However, that was not specified in the release notes so we're stating it here.

## Changes

**Non-breaking changes**
- None

**Breaking changes**

- None but this documents a previous breaking change that occurred in version 2.7.3 for deployments using a docker compose version <2.20.2 

## Additional Information

@tlvu I think you'll be especially interested in this

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,23 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+# Changes
+
+- Update Docker Compose syntax to version 2 in all docker compose files
+
+  The version 1 compose specification that docker uses is out of date and no longer maintained by docker.
+  We currently have a mix of version 1 syntax and version 2 (specifically 2.20.2+) syntax in our docker compose
+  files.
+
+  This means that the stack will not properly run with a docker compose version <2.20.2. For any version >2.20.2 
+  the stack runs properly but displays lots of deprecation warnings about deprecated version strings and external
+  volume and network definitions.
+
+  This PR updates all version 1 syntax so that these deprecation warnings are not displayed. Documentation has 
+  been updated to make this dependency on a modern version of docker explicit.
+
+  **Breaking Change**: as of birdhouse-deploy version 2.7.3, the stack could not be deployed with a docker compose
+  version <2.20.2. However, that was not specified in the release notes so we're stating it here.
 
 [2.10.1](https://github.com/bird-house/birdhouse-deploy/tree/2.10.1) (2025-03-10)
 ------------------------------------------------------------------------------------------------------------------
--- a/birdhouse/README.rst
+++ b/birdhouse/README.rst
@@ -15,8 +15,15 @@ Requirements
 * User running ``BIRDHOUSE_COMPOSE`` below must not be ``root`` but a regular user
   belonging to the ``docker`` group.
 
-* Install latest docker-ce and docker-compose for the chosen distro (not the
-  version from the distro).
+* `Install latest version of docker <https://docs.docker.com/engine/install/>`_ for the chosen distro 
+   (not the version from the distro).
+
+  * Please ensure the latest versions of the following packages and any of their dependencies
+    are installed for your distro:
+
+    * docker engine
+    * docker CLI
+    * docker compose plugin (v2.20.2+ is required)
   
 * Have a real SSL Certificate, self-signed SSL Certificate do not work properly.
   Let's Encrypt offers free SSL Certificate.
--- a/birdhouse/components/canarie-api/config/proxy/docker-compose-extra.yml
+++ b/birdhouse/components/canarie-api/config/proxy/docker-compose-extra.yml
@@ -1,4 +1,3 @@
-version: "3.4"
 services:
   proxy:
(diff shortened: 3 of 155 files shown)


CHANGES.md
fmigneault, line 1: For the various `<2.20.2`, `>2.20.2`, `2.20.2+`, please use code quotes.

For 2.7.3 reference, the tag reference should be used (like the headers).

I think it would also help to guide as to what to expect. Notably, the unsupported `start_interval`, etc. parameters being the cause of the incompatibility.

birdhouse/README.rst
```diff
@@ -15,8 +15,15 @@ Requirements
...
-  version from the distro).
+* `Install latest version of docker <https://docs.docker.com/engine/install/>`_ for the chosen distro 
+   (not the version from the distro).
+
+  * Please ensure the latest versions of the following packages and any of their dependencies
+    are installed for your distro:
+
+    * docker engine
+    * docker CLI
+    * docker compose plugin (v2.20.2+ is required)
```
fmigneault, line 26: Good idea to have this explicitly. 
Are there any specific docker engine/CLI versions required to work with compose `2.20.2+`?

CHANGES.md
```diff
@@ -15,7 +15,23 @@
...
+# Changes
+
+- Update Docker Compose syntax to version 2 in all docker compose files
+
+  The version 1 compose specification that docker uses is out of date and no longer maintained by docker.
+  We currently have a mix of version 1 syntax and version 2 (specifically 2.20.2+) syntax in our docker compose
+  files.
+
+  This means that the stack will not properly run with a docker compose version <2.20.2. For any version >2.20.2 
+  the stack runs properly but displays lots of deprecation warnings about deprecated version strings and external
```
fmigneault, line 27: Here, the `version` should be highlighted since this message out of context can be ambiguous. 
i.e.: link to https://docs.docker.com/reference/compose-file/version-and-name/

CHANGES.md
```diff
@@ -15,7 +15,23 @@
...
+  volume and network definitions.
+
+  This PR updates all version 1 syntax so that these deprecation warnings are not displayed. Documentation has 
+  been updated to make this dependency on a modern version of docker explicit.
+
+  **Breaking Change**: as of birdhouse-deploy version 2.7.3, the stack could not be deployed with a docker compose
+  version <2.20.2. However, that was not specified in the release notes so we're stating it here.
```
fmigneault, line 34: Should be at the top of the change, so it is immediately visible.

fmigneault (commented):
Changes look ok.

I will actually try them on our server that needs the update, and report back if I find anything that breaks.

birdhouse/README.rst
fmigneault: Found something... 

```
docker-compose-plugin is already the newest version (2.18.1-1~ubuntu.18.04~bionic).
```

This change enforces Ubuntu 20.04+

mishaschwartz:
@fmigneault 

Thanks for the documentation suggestions. I'll add those in shortly.

> I will actually try them on our server that needs the update, and report back if I find anything that breaks.

Thanks, I'm making some further changes to turn these into actual components so that we can make use of all of the various build tools that we already have in the stack (`*.template` files, `DELAYED_EVAL`, `VARS` and `OPTIONAL_VARS`, etc.) so you might want to wait until that change has been added.

birdhouse/README.rst
mishaschwartz: I think I'm OK with that. 18.04 was EOL in 2023. I'll update the README.

mishaschwartz: >  Are there any specific docker engine/CLI versions required to work with compose 2.20.2+?

I tried to find any documentation about that but I wasn't able to find anything specific. Certain versions of Docker Desktop come with specific versions of docker compose but as far as the engine/CLI I can't find anything about specific version compatibilities.

I'm just assuming that if you have a relatively modern version of docker engine you'll be compatible with a relatively modern docker compose. :shrug:

tlvu (approved):
LGTM to please doc in the README or the Migration doc that the docker image in the scheduler component, for the autodeploy job, will need to be overridden to use a newer docker-compose as well.

And also all the external components in external repos will also need to remove the `version` string so please document all this in clear backward incompatible section that needs manual intervention for those pulling this PR in.

fmigneault (approved):
If it's blocking on your end, you can go ahead. 
I need to keep working on CRIM's instance to migrate data and upgrade the OS to validate the updates (planed early next week).

fmigneault:
CRIM's server is now updated https://hirondelle.crim.ca/versions/

It combines this PR and #509. All seems to work correctly. 

I'm proposing we merge all "patch PRs" (including #511), and then bump all of this together into `2.11`.

PR #506, opened 2025-03-12
remove deprecated catalog crawler code

mishaschwartz:
## Overview

Code related to the solr catalog in the `birdhouse-compose.sh` file has been deprecated for a while and is not currently usable with the current stack anyway so it is removed

Also there is commented out old code that relates to the presence of an SSL certificate which is not use and no longer relevant as of version [2.7.0](https://github.com/bird-house/birdhouse-deploy/releases/tag/2.7.0).

## Changes

**Non-breaking changes**
- cleanup unused code

**Breaking changes**
- None

## Additional Information

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,14 @@
   For backwards compatibility, you can set `DOCKER_COMPOSE=docker-compose` in the local environment file to 
   use the previous default.
 
+- Remove deprecated catalog crawler code
+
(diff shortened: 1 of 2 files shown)


fmigneault (approved):

fmigneault (commented):
Please add a changelog entry, but ok with the change.

PR #510, opened 2025-03-18
Introduce a scheduler job to delete old files that may accumulate over time

mishaschwartz:
## Overview

Creates the `optional-component-clean_old_files` job that deletes old THREDDS log files and WPS output files. Allows individual cleanup jobs to be enabled for each of `raven`, `finch`, `hummingbird`, and `thredds` components. Allows the user to configure how old a file must be before it is deleted (age in days) and how to calculate the age of the file (time since last modified, time since last accessed, time since created).

(see `env.local.example` or the `scheduler` documentation for details).

## Changes

**Non-breaking changes**
- New component

**Breaking changes**

## Additional Information

## CI Operations

birdhouse_skip_ci: false


fmigneault (commented):
I'm missing a bit of context to do the review. 

Are the targeted log files the ones generated by https://github.com/bird-house/birdhouse-deploy/blob/293645aaad726fb71524bef5427c9a2a7ffc29f0/birdhouse/components/thredds/docker-compose-extra.yml#L3-L7 and equivalent in other components? Can't the configuration be overridden on the relevant component's service definition and let docker handle more powerful log rotation capability?

Or are they other logs produced by some other process?
If so, don't these service have specific log-rotate configurations?

Specifically for WPS-outputs, I'm curious about which ones are involved. The logs of older jobs might be relevant for traceability.

mishaschwartz:
@fmigneault 

Good questions... 

> Are the targeted log files the ones generated by 

No, those are handled by docker and currently only last for the lifetime of the docker container (which is a whole other issue that needs to be addressed in a completely different way).

> Or are they other logs produced by some other process?

Yes: 

- THREDDS generates separate logs that are accumulated in the `thredds_persistence` volume. These logs are rotated but old files are never cleaned up. The rotation policy is determined by a log4j configuration file that is hardcoded in the THREDDS docker container so it can't be easily modified directly. The THREDDS documentation simply recommends archiving or deleting old log files generated here.
- WPS outputs are not log files but are files that can accumulate over time in the `wps_outputs` volume if they're not cleaned up.
- If there are other locations where you know that files accumulate and are not currently cleaned up let me know and I'll add them to this PR

>  If so, don't these service have specific log-rotate configurations?

THREDDS does but not in a way that is easily overridable (see above) and WPS outputs do not. 

> Specifically for WPS-outputs, I'm curious about which ones are involved. 

Whichever ones you have enabled I guess, right now on my test machine I have raven and finch enabled so there are output directories for those. 

I guess weaver puts its outputs in the data persist directory as a bind mount ... I should probably add an option to clean those ones up too right?

> The logs of older jobs might be relevant for traceability.

I'm not sure what you mean... the user gets to set the policy for how long to keep old files. If you're concerned about losing old files then you should just not enable this feature.

fmigneault:
I think it is reasonable for the THREDDS case.

For the WPSs, it should be up to them to do the relevant cleanup, since those files are the output of potentially long processing, and the service could be "smarter" about its persistence. 

I don't think PyWPS provides something explicit about cleanup, but Weaver that also uses the WPS outputs has explicit 

https://pavics-weaver.readthedocs.io/en/latest/cli.html#dismiss
https://pavics-weaver.readthedocs.io/en/latest/cli.html#logs
https://pavics-weaver.readthedocs.io/en/latest/cli.html#provenance%20(prov)

mishaschwartz:
You say:

> For the WPSs, it should be up to them to do the relevant cleanup

but then

> I don't think PyWPS provides something explicit about cleanup

So do you mean that the solution needs to be a change to PyWPS itself or that the change I propose here is good enough for now until a solution to PyWPS is available?

I'm happy to let weaver handle its own thing. Can you explain how someone would configure weaver to delete outputs older than 1 year (for example)?

fmigneault:
It's good enough if it is configurable per service, not globally over the volume/directory.
I'm not a fan of the default blindly working over all THREDDS datasets or all WPS outputs. 

Maybe consider also using the `atime` (or make it configurable) over `mtime`. I would rather have old unused files be wiped first than actively accessed ones suddenly disappearing after a while.

mishaschwartz:
> I'm not a fan of the default blindly working over all THREDDS datasets

This removes THREDDS logs, not datasets

> Maybe consider also using the atime (or make it configurable) over mtime

It is configurable... I can make the choice between atime and mtime configurable as well though.

mishaschwartz:
Let's remove this from the 2.11 milestone. I have a bunch more scheduler jobs that I'm working on and I don't think the individual jobs need to be part of the next minor release

birdhouse/components/thredds/default.env
```diff
@@ -38,6 +38,12 @@ export THREDDS_DATASET_DATASETSCAN_BODY='
       </filter>
 '
 
+export THREDDS_DELETE_FILES_OLDER_THAN_DAYS=
+export SCHEDULER_JOB_CLEAN_OLD_FILES_OPTIONS="
+  ${SCHEDULER_JOB_CLEAN_OLD_FILES_OPTIONS}
+  \$([ -n \"\${THREDDS_DELETE_FILES_OLDER_THAN_DAYS}\" ] && echo \"thredds_persistence:/thredds|/thredds/logs/threddsServlet.*.log|\${THREDDS_DELETE_FILES_OLDER_THAN_DAYS}\")
```
tlvu, line 44: Should we stay away from appending to variable to avoid escaping and potentially duplicate problem.

I much like the dropping a config file approach.  It's more visible to debug too, than having to track the content of a variable.

How about the script `clean-old-files.sh` read `/some-dir/*.conf` and iterate over them?  

Sample .conf format:
```
LOCATION=/path/to/dir
AGE=num

... and we can easily add more config var later and we avoid parsing problem.
```

Each component just drop a .conf file at the location where `clean-old-files.sh` expects.

birdhouse/components/README.rst
```diff
@@ -59,6 +59,21 @@ component directory to the ``BIRDHOUSE_EXTRA_CONF_DIRS`` variable in your local
...
+  * In order to remove THREDDS log files the ``thredds`` component needs to be enabled.
+
+    * Set the ``THREDDS_DELETE_FILES_OLDER_THAN_DAYS`` variable in the local environment file to an integer specifying
+      how old a THREDDS log file needs to be before it is deleted.
+
+  * In order to remove WPS output files, at least one of the WPS components needs to be enabled.
+
+    * Set the ``WPS_OUTPUTS_DELETE_FILES_OLDER_THAN_DAYS`` variable in the local environment file to an integer specifying
+      how old a WPS output file needs to be before it is deleted.
+
```
tlvu, line 75: Forgot "component location"?

birdhouse/env.local.example
```diff
@@ -204,6 +204,22 @@ export GEOSERVER_ADMIN_PASSWORD="${__DEFAULT__GEOSERVER_ADMIN_PASSWORD}"
...
+#
+# Delete THREDDS log files older than X days (e.g. X=20):
+#export THREDDS_DELETE_FILES_OLDER_THAN_DAYS=20
+#
+# Delete WPS output files older than X days (e.g. X=90):
+#export WPS_OUTPUTS_DELETE_FILES_OLDER_THAN_DAYS=90
+# 
+# Set cron schedule for the clean old files job (how often the job runs).
+# By default it runs weekly on Sunday at 2:05 am:
+#export SCHEDULER_JOB_CLEAN_OLD_FILES_FREQUENCY="5 2 * * 0"
```
tlvu, line 217: I would avoid anything close to 2 and 3 AM to avoid problems on Daylight saving time switch twice a year.  Make it 1:05 am or 4:05 am to be sure.

birdhouse/optional-components/scheduler-job-clean_old_files/default.env
```diff
@@ -0,0 +1,24 @@
+export SCHEDULER_JOB_CLEAN_OLD_FILES_IMAGE='${BASH_IMAGE}'
+
+export SCHEDULER_JOB_CLEAN_OLD_FILES_FREQUENCY="5 2 * * 0" # weekly on Sunday at 2:05
+
+export SCHEDULER_JOB_CLEAN_OLD_FILES_VOLUMES='$(for opt in ${SCHEDULER_JOB_CLEAN_OLD_FILES_OPTIONS}; do printf " --volume %s:rw " "$(echo $opt | cut -d\| -f 1)"; done)'
```
tlvu, line 5: Oh ! Additional volume mount as well !  I have the same problem with adapting `deploy-data` job.  I am testing my change, once working, I'll send a PR.

tlvu (commented):
Just quick comments.

birdhouse/components/thredds/default.env
mishaschwartz: The problem is that we're not mounting config files to a container that is run by birdhouse-deploy when it starts up. The scheduler component runs the container so all the additional configurations need to be specified in the `scheduler-job-clean_old_files/config.yml` file.

No matter what we do, we need to generate that file dynamically using our template mechanism. In order to do that, we need to add additional information in an environment variable.

We could create configuration files and mount them to the container, but we'd still be appending to a variable in order to tell the scheduler which configuration files to mount. Going through a configuration file just adds extra code/overhead and doesn't buy us much.

If you want to debug this easily I'd recommend:

```
bin/birdhouse configs -c 'echo $SCHEDULER_JOB_CLEAN_OLD_FILES_OPTIONS'
```

mishaschwartz: Just thinking about this again ... 

I guess we could have a separate scheduler job for each cleanup option ... that would get a little strange because of the way that component dependencies work but maybe I can introduce something here that would make resolving those dependencies a bit easier. Let me think about this a bit

mishaschwartz: Ok I've managed an implementation where individual cleanup jobs are their own scheduler job. @tlvu please check it out when you have a minute.

mishaschwartz:
@fmigneault 

I've made it so that it is individually configurable for each service and I have set it so that the default mode is `atime` for WPS outputs and `mtime` for logs. Please take another look when you have a minute.

birdhouse/components/README.rst
```diff
@@ -59,6 +59,15 @@ component directory to the ``BIRDHOUSE_EXTRA_CONF_DIRS`` variable in your local
...
   * component location: ``optional-components/scheduler-job-deploy_raven_testdata``
 
+* Automatically remove old files
+
+  * Removes files generated by other components that may accumulate over time and are not manage automatically by those components.
+
+  * Currently supports removing WPS output files from the ``finch``, ``raven``, and ``hummingbird`` components as well as log files
+    from the ``thredds`` component.
+
+  * component location: ``optional-components/scheduler-job-clean_old_files``
```
fmigneault, line 69: Should be in the optional-components' README. 

Should also include mentions of the specific per-service configuration variables, just to provide insights that there are options to enable/disable them individually, and further options to control their cron/cleanup/fs-meta strategies.

fmigneault (commented):
Just one item remaining regarding documentation, but looks fine for the feature itself.

birdhouse/components/README.rst
mishaschwartz: All of the optional scheduler jobs are documented here and in env.local.example (for configuration options).

If we want to improve the documentation for all these jobs that's fine but we should do so for all, not just one, and maybe in a different PR.

birdhouse/optional-components/scheduler-job-clean_old_files/config/finch/config.yml.template
```diff
@@ -0,0 +1,9 @@
+- name: clean_old_files_finch
+  comment: clean old WPS output files generated by Finch
+  schedule: '${SCHEDULER_JOB_CLEAN_OLD_FILES_FINCH_FREQUENCY}'
+  command: 'sh /clean-old-files.sh "${SCHEDULER_JOB_CLEAN_OLD_FILES_FINCH_OLDER_THAN_DAYS}" "${SCHEDULER_JOB_CLEAN_OLD_FILES_FINCH_TIME_MODE}" /wps_outputs/finch'
+  dockerargs: >-
+    --rm --name scheduler-job-clean_old_files_finch
+    --volume ${COMPOSE_DIR}/optional-components/scheduler-job-clean_old_files/clean-old-files.sh:/clean-old-files.sh:ro
+    --volume "${COMPOSE_PROJECT_NAME}_wps_outputs:/wps_outputs:rw"
+  image: '${SCHEDULER_JOB_CLEAN_OLD_FILES_IMAGE}'
```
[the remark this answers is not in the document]
mishaschwartz: I actually have a question about that.... 

Now that we are making more use of the scheduler to run additional jobs, why are we writing additional logs to `BIRDHOUSE_LOG_DIR` for the scheduler jobs instead of the usual pattern of logging to the stdout/stderr of the scheduler container? Then docker would manage the logs for us (and the logs can easily be backed up using the backup scheduler job).

birdhouse/components/README.rst
tlvu: > reads those files and generates a default configuration file for the user on demand.

Yes this clearly solve the duplicate docs problem but might render the tracking of what config var changed (added, renamed, delete) harder for the simple user.  Example with JupyterHub, I had to `jupyterhub --generate-config` to have this one single file so I can diff against my current config.  Given I am not familiar with their code, I can not see the difference easily simply when browsing the code.

If we take the simple road of simply referencing the config var doc to the proper `default.env` file, we also fix the duplicate doc problem and we provide a one single file for new users to look up for the possible list of "public" config vars.  

This one single file can also be diffed to find out when a new config var has been added, renamed or deleted.

But with only the reference to the `default.env`, user will have to make an additional lookup to the matching `default.env` to read the docs, if the var name is unclear.

Both (one single file for easy browsing vs generated docs) have pros and cons.  I do not have a strong opinion on either of those approaches.

birdhouse/optional-components/scheduler-job-clean_old_files/config/finch/config.yml.template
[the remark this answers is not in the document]
tlvu: > logging to the stdout/stderr of the scheduler container?

That is fine for regular service container but here if all the jobs all log to the same scheduler container we can have interleaving log if jobs runtime happen to overlap.  Even without overlap, it will make searching for the desired log harder because all the jobs logs to the same place.  With each separate job logging to separate file, we avoid the search for the desired job log problem and the possible overlap problem.  The more jobs we have, the more likely the problems will happen.

So I much prefer each job to its own separate log file.

tlvu (changes requested):
Great generic update to handle empty dirs for any structure.  A few left-over points from previous review.  Looking good !

tlvu (approved):
Very nice DRYing of code :D  In the future we can add more birds and the addition of clean up will be simpler !

I haven't deployed this to test but LGTM, hopefully no typo during the refactoring.

PR #513, opened 2025-03-19
remove `$` from logged error variable

fmigneault:
## Overview

Logged errors in case of required variable include a `$`, leading to this kind of display: 

```sh
ERROR:    Required variable $ALERTMANAGER_SMTP_SERVER is not set. Check env.local file.
```

This gives the impression that something was not parsed or interpreted correctly. 

## Changes

**Non-breaking changes**
- remove the prefixed `$` from the variable in the log message, if any is applied, as in the case of `VAR` definitions having literal `$<VAR>`.

**Breaking changes**
- n/a

## Related Issue / Discussion

- Relates to https://github.com/bird-house/birdhouse-deploy/issues/514

## CI Operations

birdhouse_skip_ci: true

### DIFF
--- a/birdhouse/read-configs.include.sh
+++ b/birdhouse/read-configs.include.sh
@@ -340,7 +340,7 @@ check_required_vars() {
         v="${i}"
         if [ -z "`eval "echo ${v}"`" ]
         then
-            log ERROR "Required variable $v is not set. Check env.local file."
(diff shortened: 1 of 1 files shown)


mishaschwartz (approved):
Should this be included in the 2.11 version bump as well?

fmigneault:
merging as is since only visual log message adjustment

PR #516, opened 2025-03-20
Fix bug where compose directory can't be found in `bin/birdhouse` script

mishaschwartz:
## Overview

The `COMPOSE_DIR` variable cannot be discovered properly if:

- the `bin/birdhouse` script is called from with the `configs --print-config-command` options.
- the result of that call is `eval`ed in order to load the birdhouse configuration settings into the calling process's environment.
- this is done from a directory outside of the birdhouse-deploy source code directory.

This is fixed by explicitly giving a value for the `COMPOSE_DIR` variable when using the `--print-config-command` option. The value is already correctly set in the `bin/birdhouse` script so it is easy to pass that value on to the user. 

## Changes

**Non-breaking changes**
- bug fix

**Breaking changes**
- None

## Additional Information

## CI Operations

birdhouse_skip_ci: true

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -70,6 +70,21 @@
       those docker compose files as well so that you aren't bombarded by deprecation warnings whenever you start
       the birdhouse stack.  
 
+## Fixes
+
+  - Fix bug where compose directory can't be found in `bin/birdhouse` script
+
+    The `COMPOSE_DIR` variable cannnot be discovered properly if:
+    
+    - the `bin/birdhouse` script is called from with the `configs --print-config-command` options.
+    - the result of that call is `eval`ed in order to load the birdhouse configuration settings into 
+      the calling process's environment.
+    - this is done from a directory outside of the birdhouse-deploy source code directory.
+
+    This is fixed by explicitly giving a value for the `COMPOSE_DIR` variable when using the `--print-config-command`
+    option. The value is already correctly set in the `bin/birdhouse` script so it is easy to pass that 
+    value on to the user. 
+
 [2.10.1](https://github.com/bird-house/birdhouse-deploy/tree/2.10.1) (2025-03-10)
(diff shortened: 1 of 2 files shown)


fmigneault (approved):

CHANGES.md
```diff
@@ -70,6 +70,21 @@
       those docker compose files as well so that you aren't bombarded by deprecation warnings whenever you start
       the birdhouse stack.  
 
+## Fixes
+
+  - Fix bug where compose directory can't be found in `bin/birdhouse` script
+
+    The `COMPOSE_DIR` variable cannnot be discovered properly if:
+    
+    - the `bin/birdhouse` script is called from with the `configs --print-config-command` options.
```
tlvu, line 79: So `bin/birdhouse` recursively calls itself, is that why this bug happen?  Is that intended or an accident?

tlvu (approved):
LGTM although I do not understand the original reason to need a recursive call.  But the design of the script is outside the scope of this fix.

CHANGES.md
mishaschwartz: No the --print-config-command option just prints to stdout. There's nothing recursive here.

tlvu: Reading "the `bin/birdhouse` script is called from with the `configs --print-config-command` options" somehow I understood
* the `configs` command is from `bin/birdhouse`
* that `configs` command calls `bin/birdhouse`
* so it sounds like `bin/birdhouse` calls itself recursively

Maybe rephrase to make it clearer?

tlvu (approved):
Ah, with the same code code it's clearer, thanks !

So to complete my understanding, this happens for 2 reasons:
* the user accidentally set `COMPOSE_DIR` to the wrong location
* function `discover_compose_dir` in `read-configs.include.sh` fails to auto-detect the proper value

PR #518, opened 2025-03-21
 Add template expansion to backward compatibility mode

tlvu:
## Overview

Add template expansion to backward compatibility mode.

Otherwise old external repos using template expansion is broken.

## Changes

**Non-breaking changes**
- Add template expansion to backward compatibility mode.

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -70,6 +70,8 @@
       those docker compose files as well so that you aren't bombarded by deprecation warnings whenever you start
       the birdhouse stack.  
 
+- Improve backward-compatibility mode to also template expand old variable names
+
 [2.10.1](https://github.com/bird-house/birdhouse-deploy/tree/2.10.1) (2025-03-10)
 ------------------------------------------------------------------------------------------------------------------
 
--- a/birdhouse/birdhouse-compose.sh
+++ b/birdhouse/birdhouse-compose.sh
@@ -24,6 +24,15 @@ VARS='
   $BIRDHOUSE_LOG_DIR
 '
 
+# For backward-compatibility.
+OLD_VARS='
+  $PAVICS_FQDN
+  $DOC_URL
+  $SUPPORT_EMAIL
+  $DATA_PERSIST_ROOT
+  $DATA_PERSIST_SHARED_ROOT
+'
+
 # list of vars to be substituted in template but they do not have to be set in env.local
 #   their default values are from 'default.env', so they do not have to be defined in 'env.local' if values are adequate
 #   they usually are intended to provide additional features or extended customization of their behavior
@@ -41,6 +50,22 @@ OPTIONAL_VARS='
   $BIRDHOUSE_RELEASE_NOTES_URL
   $BIRDHOUSE_SUPPORT_URL
   $BIRDHOUSE_LICENSE_URL
+'
+
+# For backward-compatibility.
+OLD_OPTIONAL_VARS='
+  $PAVICS_FQDN_PUBLIC
+  $SSL_CERTIFICATE
+  $EXTRA_PYWPS_CONFIG
+  $SERVER_NAME
+  $SERVER_DESCRIPTION
+  $SERVER_INSTITUTION
+  $SERVER_SUBJECT
+  $SERVER_TAGS
+  $SERVER_DOCUMENTATION_URL
+  $SERVER_RELEASE_NOTES_URL
+  $SERVER_SUPPORT_URL
+  $SERVER_LICENSE_URL
 '
 
 THIS_FILE="$(readlink -f "$0" || realpath "$0")"
@@ -94,6 +119,13 @@ if [ x"$1" = x"up" ] || [ x"$1" = x"restart" ] || [ x"${BIRDHOUSE_COMPOSE_TEMPLA
       fi
     fi
     cat "${FILE}" | envsubst "$VARS" | envsubst "$OPTIONAL_VARS" > "${DEST}"
+
+    if [ x"${BIRDHOUSE_BACKWARD_COMPATIBLE_ALLOWED}" = x"True" ]; then
+      cp "${DEST}" "${DEST}.tmp"
+      cat "${DEST}.tmp" | envsubst "$OLD_VARS" | envsubst "$OLD_OPTIONAL_VARS" > "${DEST}"
+      rm "${DEST}.tmp"
+    fi
+
   done
   [ "$?" -eq 1 ] && exit 1  # re-raise the subshell error of the while loop
 else


mishaschwartz (commented):
I like the idea but I think we need to fix it in a different way. This solution has the following issues:

- it duplicates the list of backwards compatible variables that is already present in default.env. If we deprecate more variables, we now have to update 2 places.
- it adds another location in the source code where backwards compatible variables are processed.

I would recommend instead adding the code to modify `VARS` and `OPTIONAL_VARS` in the process_backwards_compatible_variables function in read-configs.include.sh

tlvu:
> I would recommend instead adding the code to modify `VARS` and `OPTIONAL_VARS` in the process_backwards_compatible_variables function in read-configs.include.sh

Indeed this is a better solution.  To test autodeploy I was enabling one by one all our numerous external repos on my VM and in the heat to get it to work quickly I did not think everything through.

tlvu:
FYI, in addition to `VARS` and `OPTIONAL_VARS`, for full back-compat, we also need all the old vars in `DELAYED_EVAL` as well.  

Probably only the initial ones in `default.env` or `pavics-compose.sh` matters because if an old var is added to the list by an external repos, they will already be added to the proper vars `VARS`, `OPTIONAL_VARS` or `DELAYED_EVAL` under the old name.

Just trying to get the full stack to start properly, I'll submit a proper PR later.

tlvu:
> I don't think so. The backwards compatible variables are converted before we process the delayed eval variables:
> 
>     * PAVICS_XXX becomes BIRDHOUSE_XXX
> 
>     * BIRDHOUSE_XXX is a DELAYED_EVAL variable so it gets processed afterwards

I just got an external repo using `PAVICS_FQDN_PUBLIC`, which normally resolve to `$PAVICS_FQDN` but now it is `$BIRDHOUSE_FQDN`.  

But since `PAVICS_FQDN_PUBLIC` is not in `DELAYED_EVAL` it stays at the string `$BIRDHOUSE_FQDN`.

My quick fix was to add  `PAVICS_FQDN_PUBLIC` to `DELAYED_EVAL` in `env.local` and it works.

tlvu:
> Of course if you have custom delayed eval variables then you might have to update them in your local environment file.

Yes.  It has always been the case for external custom vars.

Just a further clarification, it's the "default" built-in list of vars that needs back-compat.  So that's why I said probably only the list in `default.env` and `pavics-compose.sh` are needed.   Anything that is appended after the initial built-in list are already fine.

birdhouse/read-configs.include.sh
```diff
@@ -302,19 +336,42 @@ process_backwards_compatible_variables() {
...
             log WARN "Deprecated variable [${old_var}] can be removed as it has been superseded by [${new_var}]. Check env.local file."
           else
             log WARN "Deprecated variable [${old_var}] is present but ignored in favour of [${new_var}]. Check env.local file."
           fi
         fi
       fi
+
+      if [ x"${BIRDHOUSE_BACKWARD_COMPATIBLE_ALLOWED}" = x"True" ] && [ ! "$1" = "pre-components" ]; then
+        # Enable corresponding old var for template expansion if new var is in template expansion.
+        if echo "${VARS}" | grep -q "\s\$${new_var}\b"; then
```
mishaschwartz, line 355: `\s` and `\b` are not very portable. I'd recommend:

```
"^[[:space:]]*\$${new_var}[[:space:]]*$"
```

(see above for other examples)

tests/unit/test_read_configs_include.py
```diff
@@ -547,6 +567,156 @@ def test_allowed_no_override_old_variables_when_set(self, read_config_include_fi
...
+        """
+
+        env_local = '''
+# Add custom backward compatible var mapping.
+BIRDHOUSE_BACKWARDS_COMPATIBLE_VARIABLES="$BIRDHOUSE_BACKWARDS_COMPATIBLE_VARIABLES
+    MY_OLD_VAR=MY_NEW_VAR"
+
+# Add new var to template expansion
+VARS="$VARS
+  \$MY_NEW_VAR"
```
mishaschwartz, line 608: You need to escape the `\` character in the python string: 

```
    \\$MY_NEW_VAR
```

and the same below

tests/unit/test_read_configs_include.py
```diff
@@ -442,11 +463,10 @@ def test_not_allowed_substitution_all(self, read_config_include_file, exit_on_er
...
             exit_on_error=exit_on_error,
         )
-        expected = set()
-        for k in self.new_vars:
-            expected.add(f"{k}=new")
         actual = [re.sub(r"[\s\n]+", " ", val.strip()) for val in get_command_stdout(proc).split(ENV_SPLIT_STR_ALT)]
-        assert all(val != "new" for val in actual)
+        # "val" is like "NEW_VAR=" without the "new" value because it is initially unset and
+        # old var are not allowed to override so it stays unset.
+        assert all("new" not in val for val in actual)
```
mishaschwartz, line 469: can we leave the comparison as `val != "new"` since `"new" not in val` may match unnecessary things

tests/unit/test_read_configs_include.py
```diff
@@ -547,6 +567,156 @@ def test_allowed_no_override_old_variables_when_set(self, read_config_include_fi
...
+        proc = self.run_func(
+            read_config_include_file,
+            extra,
+            f'echo "${to_name}"',
+            exit_on_error=exit_on_error,
+        )
+        print(proc.stdout)
+        assert "\n".join(get_command_stdout(proc).split("\n")[-4:]) == expected
+
+    def test_template_expansion_enabled_for_old_var(self, read_config_include_file, exit_on_error):
```
mishaschwartz, line 596: This should be split into multiple test functions. You've got multiple calls to proc.run_func and multiple asserts for each. 

For organizational purposes you could even move this to its own class and make some of the repeated code into fixtures.

Otherwise, it is unclear what this function is testing because it is testing too many things at once.

tests/unit/test_read_configs_include.py
```diff
@@ -547,6 +567,156 @@ def test_allowed_no_override_old_variables_when_set(self, read_config_include_fi
...
+        """
+        Test that template expansion is enabled for corresponding old var if new var is enabled.
+        """
+
+        env_local = '''
```
mishaschwartz, line 601: I would prefer to either make these global constants so that the indentation within the function doesn't get all confusing. If you want to keep them in the function itself then please use something like:

```python
        env_local = "\n".join([
            "",  
            "# Add custom backward compatible var mapping.",
            "BIRDHOUSE_BACKWARDS_COMPATIBLE_VARIABLES=\"$BIRDHOUSE_BACKWARDS_COMPATIBLE_VARIABLES\"",
            "    MY_OLD_VAR=MY_NEW_VAR"",
            "",
            ...
        ])
```

mishaschwartz (commented):
In general I think this is a great idea!

I worry that there's some extra stuff added in here though that affects logging and I'm a little confused about some of the design choices. Please see my questions and comments below:

birdhouse/default.env
```diff
@@ -118,8 +151,28 @@ __DEFAULT__PHOENIX_PASSWORD="phoenix_pass"
...
-# Note: if adding to this later on, make sure that you add new overrides to the *end* of this list so that they will be
+#
+# Note: if adding to this later on, make sure that you add new overrides to the **END** of this list so that they will be
 #       parsed in the correct order.
+#
+# Example: SSL_CERTIFICATE renamed to SERVER_SSL_CERTIFICATE then renamed to BIRDHOUSE_SSL_CERTIFICATE
+#
+#          Then we need the following mapping in this **SPECIFIC** order:
+#
+#            SSL_CERTIFICATE=SERVER_SSL_CERTIFICATE             # chain 1st name to 2nd rename
```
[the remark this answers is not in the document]
fmigneault: It comes from an incorrect rewrite of the variables in https://github.com/bird-house/birdhouse-deploy/pull/428/files#diff-2cb30d913fb346ad995e3fbd70f3eec0c6170a9a0f5ebbe072239c393277dd33 when adding `BIRDHOUSE_` prefix.

birdhouse/read-configs.include.sh
```diff
@@ -23,6 +23,39 @@
...
+#                 SSL_CERTIFICATE="/ssl_cert/ouranos_cert.pem"
+#   TRACE_VARS:  after set_old_backwards_compatible_variables:
+#       BIRDHOUSE_SSL_CERTIFICATE="${__DEFAULT__BIRDHOUSE_SSL_CERTIFICATE}"
+#          SERVER_SSL_CERTIFICATE="${__DEFAULT__BIRDHOUSE_SSL_CERTIFICATE}"
+#                 SSL_CERTIFICATE="/ssl_cert/ouranos_cert.pem"
+#   TRACE_VARS:  after process_backwards_compatible_variables:  <-- error introduced at this step
+#       BIRDHOUSE_SSL_CERTIFICATE="/ssl_cert/ouranos_cert.pem"
+#          SERVER_SSL_CERTIFICATE="${__DEFAULT__BIRDHOUSE_SSL_CERTIFICATE}"
+#                 SSL_CERTIFICATE="/ssl_cert/ouranos_cert.pem"
+log_trace_vars_cmd() {
```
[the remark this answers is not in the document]
fmigneault: I agree, it should reuse the existing method. If needed, logging methods can be extended. Toggle of these variables should be controllable with the same logging level logic (here, clearly a DEBUG).

tests/unit/test_read_configs_include.py
fmigneault: Is it possible that an erroneous `"new"` was not caught because newlines were not preserved correctly? Now that they should be, the value might not be *exactly equal* to `"new"`?

tests/unit/test_read_configs_include.py
fmigneault: Or, use test files, and employ actual formatting without misinterpretation/mangling by strings.

birdhouse/default.env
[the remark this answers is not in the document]
tlvu: > Why can't we just have:
> 
> ```
> SSL_CERTIFICATE=SERVER_SSL_CERTIFICATE
> SERVER_SSL_CERTIFICATE=BIRDHOUSE_SSL_CERTIFICATE
> ```
> 
> Also this seems to imply that `SSL_CERTIFICATE` won't get added to the `VARS` and `OPTIONAL_VARS` list unless it's declared twice (here and the line below)... is that right? That doesn't seem right...

Yes ... that was my first attempt, if we do not want to change the data format.  To put everything on one line, we probably need `SSL_CERTIFICAT=>SERVER_SSL_CERTIFICATE=>BIRDHOUSE_SSL_CERTIFICATE`.  I was planning to make this switch after I finish writing additional tests to avoid adding regressions when switching the code to match the new data format.

birdhouse/read-configs.include.sh
tlvu: Ah nice !

tests/unit/test_read_configs_include.py
tlvu: Here is the list of the names of the var, it's not the actual content of the var itself.  Are you referencing to preserve special character and string formatting for template expansion to not break python code?

tests/unit/test_read_configs_include.py
tlvu: I changed it because of the comment right above.  `val` is like `NEW_VAR=` so even if `new` happen to be the value of `val`, it would look like `NEW_VAR=new` so the original comparison predicate `val != "new"` will always be true meaning the test will never catch any bugs.

This fix has nothing to do with all my code changes. I simply catch the bug that the test is invalid.

tests/unit/test_read_configs_include.py
tlvu: Yeah let me try to clean this up.

tests/unit/test_read_configs_include.py
tlvu: These are specific config for each test, why make them global constant?

And the syntax I use the equivalent of "here doc" in Python to make multi lines string as simple as possible.  Why surround with "\n".join() if we don't need to.  Using here-doc, the will be less "misinterpretation/mangling by strings".

I'd rather choose simplicity over strict indentation and possibility of error-prone.

birdhouse/read-configs.include.sh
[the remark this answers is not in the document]
fmigneault: When tracing to debug variables, I somewhat expect it to be extra verbose. I think the amount of debug outputs produced is a non issue in this case.

I'm not sure to understand how `echo "TRACE_VARS: ...` differs from `log DEBUG "TRACE_VARS: ...` in terms of the output produced by that specific call. The string obtained by `BIRDHOUSE_DEBUG_VARS_TRACE_CMD` should be passed to `log` the same way it is passed to `echo`. If that is not the case, then something is misbehaving in the log message handling, and this should be addressed instead of bypassing with another command.

If each section had a `log DEBUG "=== read-configs.include.sh sourced ==="`, wouldn't it be possible to quickly filter the produced log with a few `grep` of the variables of interest, and find at which step the change happened? There wouldn't be a need for a complex injection of `BIRDHOUSE_DEBUG_VARS_TRACE_CMD` just to debug.

tests/unit/test_read_configs_include.py
mishaschwartz: Oh I see. Ok thanks for the catch. 

Please update it to:

```python
assert all(val.split("=")[-1] != "new" for val in actual)
```

so that the intent of the test is kept (this is testing that the value of the variable is not set to `new`, not that `new` doesn't occur in either the value or the variable name).

birdhouse/default.env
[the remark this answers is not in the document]
mishaschwartz: > It comes from an incorrect rewrite of the variables in https://github.com/bird-house/birdhouse-deploy/pull/428/files#diff-2cb30d913fb346ad995e3fbd70f3eec0c6170a9a0f5ebbe072239c393277dd33 when adding BIRDHOUSE_ prefix.

Alright I guess my bad then.

If it never made it into any actual part of the source code can we get rid of it?

tests/unit/test_read_configs_include.py
mishaschwartz: > why make them global constant?

Make them a fixture then.

Python code that is suddenly de-indented mid function is hard to read and makes it difficult to easily determine the nesting hierarchy.

I agree with @fmigneault that using a test file would work well for this too (we already have some in the `fixtures` directory ... feel free to add more).

birdhouse/default.env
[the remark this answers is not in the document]
fmigneault: Fine with me if that removes the resolution complexity of the `SSL_CERTIFICAT=>SERVER_SSL_CERTIFICATE=>BIRDHOUSE_SSL_CERTIFICATE` chain. 

Was that the only case that was problematic in this way?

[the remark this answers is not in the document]
mishaschwartz: I think so. 

But we still need to be able to handle variables that change multiple times in case we rename things again in the future (I hope we don't but better safe then sorry).

CHANGES.md
```diff
@@ -15,7 +15,157 @@
...
+    from the new var to the old var is missing all the new lines and also
+    generate broken code.
+
+  - Wrong old var overriding new var during child invocation, even when new
+    var is set properly in `env.local`.
+
+    This bug is mostly visible with child invocation.  First level
+    invocation, to be affected, need to use the old var name, which means
+    external repos only because the current repo has all switched to use the
+    new var names.
```
fmigneault, line 54: This seems to indicate that setting new/old variable is supported bidirectionally. Is this correct?  

If that is the case, I wonder if this is fundamentally something we want to handle? Considering that the main reason old variable renames were applied is because they are _**deprecated**_, and that we want to move away from them. The backward-compatibility is there only to lower maintenance burden until they can be migrated. They are not intended to have long term support or be used simultaneously.

More specifically, I believe the case of setting the old name from the new one should _**not**_ be a concern. If a platform relies on older variable names due to external scripts limitations, it should stick **only** to the old names in its `env.local`, and let `new = old` happen automatically to patch local scripts expecting the new var. If the platform decides to use new names in `env.local`, than it should migrate its external configurations by dropping old names entirely. There shouldn't be duplicate/mixed use of names corresponding to the same variable.

In which situation are new var names used in combination with old ones in `env.local` instead of just using the old name needed by older scripts. Why not just use those old names, as the backward-compatibility mode expects?

fmigneault (changes requested):

birdhouse/read-configs.include.sh
```diff
@@ -352,20 +352,20 @@ process_backwards_compatible_variables() {
 
       if [ x"${BIRDHOUSE_BACKWARD_COMPATIBLE_ALLOWED}" = x"True" ] && [ ! "$1" = "pre-components" ]; then
         # Enable corresponding old var for template expansion if new var is in template expansion.
-        if echo "${VARS}" | grep -q "\s\$${new_var}\b"; then
+        if echo "${VARS}" | grep -q "[[:space:]]\$${new_var}[[:space:]]*"; then
```
mishaschwartz, line 355: This is better but you're going to want to be able to handle VARS that start or end at the edge of the line as well (see the suggestion here: https://github.com/bird-house/birdhouse-deploy/pull/518#discussion_r2052929257)

Right now for example, if `new_var='$SOME_VAR'`, this won't match:

```sh
VARS='$SOME_VAR'
VARS='
$SOME_VAR'
```

and it will match:

```sh
VARS=' $SOME_VAR_OTHER_THING_ENTIRELY'
VARS='
  $SOME_VAR_ANOTHER_ONE
'
```

both of which you don't want.

tlvu: > if `new_var='$SOME_VAR'`, this won't match:

`VARS` is an append only variable so setting it directly like that should never happen.  Not only the user will zap their own variable, they will zap all the variables previously accumulated there !



> ```shell
> VARS=' $SOME_VAR_OTHER_THING_ENTIRELY'
> ```

This is why I previously had `\b` in `${new_var}\b` which will match the edge of the word and would take care of this scenario, but then you said it is not portable.

Each append to `VARS` is on a new line is just a readability thing.  It's never enforced or even documented.  Anyone can do `VARS="$VARS \$NEW_1 \$NEW_2 \$NEW_3"` and it is still valid.

Unfortunately I have not found `\b` portable equivalent.  So to handle this case we can use `\b`?

mishaschwartz: Or you can do:

```sh
echo ${VARS} | grep -q '\(^\|[[:space:]]\)\$${new_var}\([[:space:]]\|$\)'
```

Or:

```sh
echo "|$(echo ${VARS}| tr ' ' \|)|" | grep "|\$${new_var}|"
```

Or:

```sh
echo " ${VARS} " | tr '\n' ' ' | grep "[[:space:]]\$${new_var}[[:space:]]"
```

tlvu: > ```shell
> echo " ${VARS} " | tr '\n' ' ' | grep "[[:space:]]\$${new_var}[[:space:]]"
> ```

Ah nice, adding extra "space" padding to normalize everything to space !  Why haven't I though about that!  I like this last one, it's the simplest of all the other variations.  

The `|` padded version might not handle the case when 
```
VARS="$VARS
\$NEW"  # no space in front to turn to '|'
```

The other 2 with or `\|` operator and meta `\(` and `\)` just seems too easy to make a typo mistake and takes more time to "read" and "understand".

mishaschwartz: All of them work. I'll leave it up to you which you want to use.

birdhouse/default.env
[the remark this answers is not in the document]
mishaschwartz: These are different though. They're not a chain of variable1 -> variable2 -> variable3, they're two variables that got merged: variable1 -> variable3 and variable2 -> variable3.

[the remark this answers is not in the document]
tlvu: > two variables that got merged: variable1 -> variable3 and variable2 -> variable3.

That is the work-around to handle the multi-rename variable1 -> variable2 -> variable3 when the current algo and current syntax only support one rename.

I plan to remove var2 therefore all this extra work-around is not needed anymore.

tests/unit/test_read_configs_include.py
tlvu: done in 6b280b88811c34171e57c161d7f03d212fc1a6f8

CHANGES.md
tlvu: > In which situation are new var names used in combination with old ones in `env.local` instead of just using the old name needed by older scripts. Why not just use those old names, as the backward-compatibility mode expects?

The case when we have migrate `env.local` but all the numerous external repos still use old names.  This ease the transition of each piece separately without having to change everything at the same time.   It also allows for "roll-back" when something at the tip broke, we rollback the repo but we still keep the latest `env.local` with all the newer var names.

I'd say it's a good safety insurance policy for production deployment that Misha already implemented initially, I do not see why we would want to remove this now.

CHANGES.md
```diff
@@ -15,7 +15,157 @@
...
+    `set_old_backwards_compatible_variables()` are set from new var values or
+    default values so setting "new_var = old_value" back in
+    `process_backwards_compatible_variables()` is basically setting the same value again.
+
+- Fix process_delayed_eval() losing new lines and leading empty space after being eval'ed
+
+  Example: if `AUTODEPLOY_PLATFORM_EXTRA_DOCKER_ARGS` is set in `env.local` and
+  added to `DELAYED_EVAL` in `env.local` because it uses other variables,
+  the new lines and the 4 front spaces must be kept for proper yaml syntax when
+  template expanding in `components/scheduler/config.yml.template`.
```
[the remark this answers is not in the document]
tlvu: Once I'll remove the intermediate rename that never made it out to any release, maybe we can simplify this changelog.

birdhouse/read-configs.include.sh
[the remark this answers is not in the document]
fmigneault: That's fine if it is for a specific use case. A comment to avoid circling back to this question later would be relevant.

`TRACE` is a good idea. It is fairly common and would allow controlling all logging behaviors from a common parameter.

CHANGES.md
mishaschwartz: > I'd say it's a good safety insurance policy for production deployment that Misha already implemented initially,

Did I? I don't think I did. Right now we can run in "backwards compatible mode" which sets new variables based on the values of the old variables in env.local.  But I don't think it ever sets the old variable names based on the new variables.

> The case when we have migrate env.local but all the numerous external repos still use old names.

The compromise we originally came up with was that env.local could contain both the old and new variables names simultaneously so that you could "roll back" without having to change the local env file. But this was only ever meant as a temporary measure while you were testing out a new deployment. It's not meant so that you never update your env.local file and rely on the backwards compatibility mode.

fmigneault: I think setting back old variables names based on new ones fundamentally goes against "backward compatibility". The idea is to phase away from old names, but provide at least considerations about them to not break everything and slow down development. If they are inversely assigned, we will never move away from them, which would make the integration of new names pointless.

tlvu: > But I don't think it ever sets the old variable names based on the new variables.

Yes, you did, see below (I did not write that function and that comment).

https://github.com/bird-house/birdhouse-deploy/blob/c209fbacfa193b9ee726873b943f0ae648f3087a/birdhouse/read-configs.include.sh#L286-L289

> It's not meant so that you never update your env.local file and rely on the backwards compatibility mode.

During the big rename PR, Francis wanted his IaC pipeline to "just work" without any modifications to his `env.local` file, otherwise he will need to manage 2 branches for his config files.

Same for me, I just want to simply catch-up with tip of master without managing a bunch of branches.  Because in addition to the `env.local`, Ouranos have a bunch of external repos so I'd rather not manage a bunch of branches as well.  Once the go-live with the catch is over, I can migrate one by one.   Right now I just want to focus on getting everything to run as before so the least moving parts, the better.

> I think setting back old variables names based on new ones fundamentally goes against "backward compatibility".

In this PR I was just fixing things that are already here.  Not adding, changing, or removing any behavior.

I am just trying to merge this PR so I can actually properly test autodeploy because Ouranos massively use it.  This is the block issue for us.

mishaschwartz: Ok thanks for pointing that out. I had forgotten.

Let's worry about whether to keep this or not going forward after this PR.

> I am just trying to merge this PR so I can actually properly test autodeploy because Ouranos massively use it. This is the block issue for us.

OK, let's do the minimal plan here (https://github.com/bird-house/birdhouse-deploy/pull/518#issuecomment-2956794071)

fmigneault: On our end (and the CI), the only remaining old variables are `PAVICS_FQDN` and the various metadata `SERVER_<>` to their `BIRDHOUSE_<>` counterparts. The `SERVER_<>` ones probably can be migrated safely since its just metadata. The only problematic/blocking one is `PAVICS_FQDN` kept only because of notebook tests that employ it in the CI (not our platform itself).

My expectation from backward compatibility is that `PAVICS_FQDN`'s value defined in `env.local` is set to the desired `BIRDHOUSE_FQDN` so that all templates using `BIRDHOUSE_FQDN` work seamlessly, just so I don't need to set both of them myself.

However! 
It is my assumption that, if at some point we take the time to update `BIRDHOUSE_FQDN`,  any script (tests) sourcing the `env.local` are also modified to use `BIRDHOUSE_FQDN`. In order words, making the switch implies (to me) that `PAVICS_FQDN` would be dropeped everywhere relevant for the CI. I am NOT expecting the inverse `PAVICS_FQDN <= BIRDHOUSE_FQDN` value propagation, given that my only motivation to switch to `BIRDHOUSE_FQDN` is to get rid of `PAVICS_FQDN` altogether. If that is never possible, I will simply keep the `PAVICS_FQDN` as is.

My impression is that, by supporting `PAVICS_FQDN <= BIRDHOUSE_FQDN` value propagation, we will never catch up with the variable renames, since there is no incentive to do so, and that we have other more urgent things to do (since the platform is not broken by leaving `PAVICS_FQDN`).

mishaschwartz:
@tlvu 

Do you have any comments about my question here? https://github.com/bird-house/birdhouse-deploy/pull/518#issuecomment-2956794071

PR #519, opened 2025-03-21
Stop build when a build step fails

### LINKED ISSUE #512: :bug: [BUG]: Silently failing birdhouse-deploy operations
## Summary

Some operations executed by `birdhouse-deploy.sh` can fail silently.

This issue is opened as a general discussion about how to handle this kind of situation as a whole (what is our philosophy about managing these cases?), but providing a specific use case that was encountered.

Do we consider these to be edge cases that we don't manage? Do we reinforce strict error handling? Or, do we have some middle ground best-practice to put forward in the docs?

## Details

For example, if any of the following commands fail, for example, because `"${THIS_DIR}/static"` already exists, but with inadequate user/group permissions causing the files to not be created, the `birdhouse-deploy.sh` execution will happily resume it operations without any apparent failure.

https://github.com/bird-house/birdhouse-deploy/blob/293645aaad726fb71524bef5427c9a2a7ffc29f0/birdhouse/components/proxy/pre-docker-compose-up#L9-L13

At the very least, permissions errors such as in this case would be relevant to be displayed on stderr. 

In this specific case, this leads to a seemingly "functional" platform that boots and where services respond without problem, but the informative `/version`, `/services` and `/components` endpoints cause 404 errors. 

### To Reproduce

1. Create a `birdhouse/components/proxy/static` directory with some limited access (eg: root or specific user read-only)
2. Run the stack with `birdhouse-deploy` to achieve a "running" state, and notice that no errors are reported.
3. Access any of the  `/version`, `/services` or `/components` endpoint.

## Environment

all

mishaschwartz:
## Overview

If a command exits with a non-zero exit code when deploying the stack (i.e. running `birdhouse-compose.sh`) the build process should stop and report that an unexpected error occurs.

This is enforced by setting the `-e` flag when `birdhouse-compose.sh` is run by default. Several commands have been updated so that they are followed by `|| true` so that if they exit with a non-zero status it will not stop the script.

Also a new environment variable `BIRDHOUSE_DEBUG_MODE` is introduced which can be set to `true` to set the `-x` flag which will write every command that is run by `birdhouse-compose.sh` to stderr. Previously, setting the `BIRDHOUSE_LOG_LEVEL` to `DEBUG` would set the `-x` flag whenever `pre-docker-compose-up` and `post-docker-compose-up` scripts are executed. Please use the `BIRDHOUSE_DEBUG_MODE` instead from now on. `BIRDHOUSE_LOG_LEVEL` should only be used to set the log level of the birdhouse logger.

## Changes

**Non-breaking changes**
- new env var: `BIRDHOUSE_DEBUG_MODE`

**Breaking changes**
- `BIRDHOUSE_LOG_LEVEL=DEBUG` no longer sets the `-x` flag on pre/post compose-up scripts

## Related Issue / Discussion

- Resolves #512

This is one possible solution to the problem discussed in #512. If you have thoughts on an alternative approach let me know.

## CI Operations

birdhouse_skip_ci: false


birdhouse/birdhouse-compose.sh
```diff
@@ -100,15 +128,10 @@ else
...
-SHELL_EXEC_FLAGS=
-if [ "${BIRDHOUSE_LOG_LEVEL}" = "DEBUG" ]; then
-  SHELL_EXEC_FLAGS=-x
-fi
-
 create_compose_conf_list # this sets COMPOSE_CONF_LIST
 log INFO "Displaying resolved compose configurations:"
 log INFO "COMPOSE_CONF_LIST="
-log INFO ${COMPOSE_CONF_LIST} | tr ' ' '\n' | grep -v '^-f'
+log INFO "$(echo ${COMPOSE_CONF_LIST} | tr ' ' '\n' | grep -v '^-f')"
```
fmigneault, line 134: Should `${COMPOSE_CONF_LIST}` be in quotes as well, to avoid possible issues like a `>` or `|` character making its way there somehow?

birdhouse/components/weaver/post-docker-compose-up
```diff
@@ -374,10 +374,10 @@ mkdir -p "${BIRDHOUSE_LOG_DIR}"
 # note: use 'tee' instead of capturing in variable to allow displaying results directly when running command
 ${BIRDHOUSE_COMPOSE} exec weaver bash "${CELERY_HEALTHCHECK}" | tee "${BIRDHOUSE_LOG_DIR}/weaver.log"
 ret_weaver=$?
-out_weaver=$(cat "${BIRDHOUSE_LOG_DIR}/weaver.log" | tail -n 1 | grep -c "ERROR")
+out_weaver=$(cat "${BIRDHOUSE_LOG_DIR}/weaver.log" | tail -n 1 | grep -c "ERROR") || true
 ${BIRDHOUSE_COMPOSE} exec weaver-worker bash "${CELERY_HEALTHCHECK}" | tee "${BIRDHOUSE_LOG_DIR}/weaver-worker.log"
 ret_worker=$?
-out_worker=$(cat "${BIRDHOUSE_LOG_DIR}/weaver-worker.log" | tail -n 1 | grep -c "ERROR")
+out_worker=$(cat "${BIRDHOUSE_LOG_DIR}/weaver-worker.log" | tail -n 1 | grep -c "ERROR") || true
```
fmigneault, line 380: Fine if the error was from grep no-match, but just making sure it's not from missing logs?

birdhouse/components/jupyterhub/pre-docker-compose-up
```diff
@@ -1,8 +1,7 @@
 #!/bin/sh
 
 # this is external in docker-compose.yml so have to create here
-# no error if already exist, just an error message
-docker network create jupyterhub_network
+docker network create jupyterhub_network || true
```
fmigneault, line 4: I think a warning log (or maybe debug since "intended behavior") should be reported here. Maybe the jupyter network already existing could be the cause of an underlying issue that would be hard to identify by silencing the "duplicate".

birdhouse/birdhouse-compose.sh
```diff
@@ -135,53 +158,40 @@ fi
...
 
+if [ x"$1" = x"up" ]; then
+  # we restart the proxy after an up to make sure nginx continue to work if any container IP address changes
+  PROXY_HTTP_PORT=80 HOSTNAME=${BIRDHOUSE_FQDN} ${DOCKER_COMPOSE} ${COMPOSE_CONF_LIST} restart proxy
 
-while [ $# -gt 0 ]
-do
-  if [ x"$1" = x"up" ]; then
-    # we restart the proxy after an up to make sure nginx continue to work if any container IP address changes
-    PROXY_HTTP_PORT=80 HOSTNAME=${BIRDHOUSE_FQDN} ${DOCKER_COMPOSE} ${COMPOSE_CONF_LIST} restart proxy
```
fmigneault, line 158: By removing the while loop, doesn't that have the side effect of removing support of options to docker compose? Something like `birdhouse-compose --compatibility --dry-run up -d` would not be possible? 

Not sure if these options were ever handled actually, since the first option assumed to be the docker compose command seems to have spread across other places as well.

birdhouse/birdhouse-compose.sh
```diff
@@ -13,6 +13,34 @@
 # list of all variables to be substituted in templates
 #   some of these variables *could* employ provided values in 'default.env',
 #   but they must ultimately be defined one way or another for the server to work
+
+SHELL_EXEC_FLAGS=-e
+if [ x"${BIRDHOUSE_DEBUG_MODE}" = x"true" ]; then
+  SHELL_EXEC_FLAGS=-ex
+fi
```
fmigneault, line 20: I'm fine with this being the default, but just raising the question up for debate.

Is there any situation where one might want to purposely `set +e; set -x` to debug all problems at once, or simply `set +e` to purposely ignore some known errors?

birdhouse/birdhouse-compose.sh
```diff
@@ -135,53 +158,40 @@ fi
...
-  exit ${ERR}
-fi
+PROXY_HTTP_PORT=80 HOSTNAME=${BIRDHOUSE_FQDN} ${DOCKER_COMPOSE} ${COMPOSE_CONF_LIST} $* ${COMPOSE_EXTRA_OPTS} || expect_exit
 
 # execute post-compose function if exists and no error occurred
-type post-compose 2>&1 | grep 'post-compose is a function' > /dev/null
-if [ $? -eq 0 ]
+if type post-compose 2>&1 | grep 'post-compose is a function' > /dev/null
 then
-  [ ${ERR} -gt 0 ] && { log ERROR "Error occurred with docker-compose, not running post-compose"; exit $?; }
```
fmigneault, line 149: Depending on the use case of `set +e; set -x`, might need to leave that in.

birdhouse/components/weaver/post-docker-compose-up
mishaschwartz: Yeah it's just from the grep no match.

Since the `pipefail` flag is not set the `e` flag will only cause the script to exit if the last command in the pipe fails.

birdhouse/birdhouse-compose.sh
mishaschwartz: I removed the while loop for a few reasons:

- the last test in the while loop will always be falsy which causes the script to exit when the `e` flag is set
- we don't do this loop to check for `up` anywhere else in this script where it could be relevant
- it is prone to false positives like: `birdhouse compose run --name up stac ls`

If we want to be able to handle the `birdhouse compose --compatibility --dry-run up -d` case then we need to figure out a better way to do this and to apply this everywhere in this script.... 

Something like this could work:

```
compose_command=$(echo "$@" | grep -o '[[:space:]][[:alpha:]]\{1,\}' | head -1 | sed 's/[[:space:]]//g')
if [ "$compose_command" = "up" ]; then
...
```

I don't know if that's what we actually want anymore though since now we can do something like:

```
DOCKER_COMPOSE='docker compose --compatibility --dry-run' birdhouse compose up -d
```

I feel like this needs to be in a different PR though. Let me think about it.

birdhouse/birdhouse-compose.sh
mishaschwartz: Maybe in development while we're debugging but my feeling is with this PR and the issues raised in #512 we really want this script to always run without errors so we always want to enforce the `-e` flag.

If you're debugging in development you can manually edit the flags I guess. If we want a shortcut we can always change this to:

```
SHELL_EXEC_FLAGS="${SHELL_EXEC_FLAGS:-"-e"}
```

what do you think?

tlvu: I like the fact that we do not force this default `-e` but then `SHELL_EXEC_FLAGS="${SHELL_EXEC_FLAGS:-"-e"}` means if the user do not want `-e` and leave `SHELL_EXEC_FLAGS` empty, he will still get `-e` !

I don't have a quick suggestion on top of my head how to honor user not wanting any default value in this rather unusual situation!

birdhouse/birdhouse-compose.sh
```diff
@@ -13,6 +13,34 @@
...
+fi
+
+set ${SHELL_EXEC_FLAGS}
+
+exit_handler() {
+  exit_code="$?"
+  [ "$exit_code" -eq 0 ] && exit
+  [ "$BIRDHOUSE_EXPECTED_EXIT" = "true" ] && exit "$exit_code"
+  err_msg="An unexpected error occurred. See additional messages on stderr for details or rerun this command with BIRDHOUSE_DEBUG_MODE=true."
+  if command -v log >/dev/null; then
```
tlvu, line 29: `command` is a built-in command?  I didn't know that !  `man command` and `info command` returns nothing !

birdhouse/birdhouse-compose.sh
```diff
@@ -13,6 +13,34 @@
...
+  fi
+}
+
+expect_exit() {
+  exit_code="${1:-$?}"
+  export BIRDHOUSE_EXPECTED_EXIT=true
+  exit "${exit_code}"
+}
+
+trap 'exit_handler' EXIT
```
tlvu, line 42: Should this `exit_handler` and `expect_exit` be relocated to `read-configs.include.sh` so other scripts can also use it?

birdhouse/birdhouse-compose.sh
tlvu: Perso I have never had the need for `+e` but that's probably because `-e` was never enforced !

So how about flipping this upside down and have an escape route to not force `-e` so then we do not really need to support `+e`?

birdhouse/birdhouse-compose.sh
tlvu: `BIRDHOUSE_COMPOSE_DEBUG_MODE`?  Because `BIRDHOUSE_DEBUG_MODE` sort of implies all scripts in the stack will honor this, which is not true because only `birdhouse-compose.sh` is honoring this flag for now.

Or are we planning to make this platform wide and move this to `default.env` with some helper functions in `read-configs.include.sh` so all script can enable it to be compliant?

birdhouse/birdhouse-compose.sh
```diff
@@ -135,53 +158,40 @@ fi
 
 log INFO "Executing docker-compose with extra options: $* ${COMPOSE_EXTRA_OPTS}"
 # the PROXY_HTTP_PORT is a little trick to make the compose file invalid without the usage of this wrapper script
-PROXY_HTTP_PORT=80 HOSTNAME=${BIRDHOUSE_FQDN} ${DOCKER_COMPOSE} ${COMPOSE_CONF_LIST} $* ${COMPOSE_EXTRA_OPTS}
-ERR=$?
-if [ ${ERR} -gt 0 ]; then
-  log ERROR "docker-compose error, exit code ${ERR}"
```
tlvu, line 141: Add an optional `$2` to `expect_exit` to keep ability to log error message?

birdhouse/birdhouse-compose.sh
tlvu: Ah wow, I never noticed there is while loop here !  Not sure the reason for the while loop anymore.  Was there before me.

birdhouse/components/jupyterhub/pre-docker-compose-up
tlvu: I think this just print the error message to console.  Whether with or without `|| true` the message will still be printed anyways so no "silencing" risk here.  Or I might have misunderstood the effect of `|| true`?

birdhouse/scripts/get-services-json.include.sh
```diff
@@ -16,7 +16,7 @@ for adir in ${ALL_CONF_DIRS}; do
   SERVICE_CONF="$(cat "${adir}/service-config.json" | sed -z 's/^\s*//;s/\s*$//')"
   # remove the leading/trailing [] to get a pseudo-json of nested objects to extend the list
   SERVICE_CONF="$(echo "${SERVICE_CONF}" | sed -z 's/^\s*\[\s*//;s/\s*\]\s*$//')"
-  SERVICES="${SERVICES}$([ -n "${SERVICES}" ] && echo ',') ${SERVICE_CONF}"
+  SERVICES="${SERVICES}$([ -n "${SERVICES}" ] && echo ',' || echo '') ${SERVICE_CONF}"
```
tlvu, line 19: Adding `|| echo ''` is for the first time when `${SERVICES}` is empty?

tlvu (changes requested):

birdhouse/components/jupyterhub/pre-docker-compose-up
mishaschwartz: @tlvu yes that is correct

birdhouse/birdhouse-compose.sh
mishaschwartz: > SHELL_EXEC_FLAGS="${SHELL_EXEC_FLAGS:-"-e"} means if the user do not want -e and leave SHELL_EXEC_FLAGS empty, he will still get -e

A user could do `SHELL_EXEC_FLAGS=' '` but I agree that this is not exactly obvious and can be very confusing.

What if we're just really explicit about everything:

```sh
BIRDHOUSE_DEBUG_MODE=false # default false
BIRDHOUSE_FAIL_FAST=true # default true

SHELL_EXEC_FLAGS=""
[ "$BIRDHOUSE_DEBUG_MODE" = "$true" ] && SHELL_EXEC_FLAGS="$SHELL_EXEC_FLAGS -x"
[ "$BIRDHOUSE_FAIL_FAST" = "$true" ] && SHELL_EXEC_FLAGS="$SHELL_EXEC_FLAGS -e"
```

birdhouse/birdhouse-compose.sh
mishaschwartz: This (https://github.com/bird-house/birdhouse-deploy/pull/519#discussion_r2018773906) could solve that problem as well I think

`+e` would be the same as setting `BIRDHOUSE_FAIL_FAST=false`

birdhouse/birdhouse-compose.sh
mishaschwartz: I only worry that then we can't actually use it until read-configs.include.sh is sourced. But I agree that it's a useful tool to provide if other scripts want to use it. Let me see what I can do.

birdhouse/birdhouse-compose.sh
mishaschwartz: I don't think it's necessary in this case since docker will write any relevant error message itself. The error message here doesn't really add any additional information that is relevant to the user.

birdhouse/birdhouse-compose.sh
mishaschwartz: The problem is that I want this to be defined early so that we get relevant error messages.

If we put it in read-configs.include.sh then we need to make sure that `COMPOSE_DIR` is defined first. In order to set `COMPOSE_DIR` there's a bunch of code that I'd like to be able to trap when it exits.

So I think I'd rather leave it here and other scripts can define their own error handling if needed. I just don't think it's worth making this less useful for birdhouse-compose.sh.

birdhouse/birdhouse-compose.sh
mishaschwartz: > BIRDHOUSE_COMPOSE_DEBUG_MODE

Yeah I agree let's change the name to this. I don't want to make this stack-wide because of https://github.com/bird-house/birdhouse-deploy/pull/519#discussion_r2018838047

birdhouse/birdhouse-compose.sh
mishaschwartz: In a different case if you wanted an additional error message you could still do:

```
some_command_that_might_fail
ERR=$?
if [ ${ERR} -gt 0 ]; then
  log ERROR "error message"
  expect_exit "$ERR"
fi
```

birdhouse/components/jupyterhub/pre-docker-compose-up
fmigneault: Ok, that is fine then.

birdhouse/birdhouse-compose.sh
fmigneault: @tlvu 
The while loop was there exactly for this reason that the docker compose command is not necessarily first, though this logic has not been maintained properly everywhere that it is checked for conditional operations.

@mishaschwartz 
I think both examples of potential false positives are somewhat of a stretch. They are almost asking for things to break if one is actually doing this. I don't think neither are representative of reasonable and typical scenarios. Therefore, I would prefer we opt for a solution that best reduces the chances of side effects from not-so-obvious shell behaviors, such as how while-loop handles errors. If using the grep approach has limited chances of failure, but saves us major refactoring tasks and special error handling each time a new problem is found, I would be willing to live with this tradeoff.

birdhouse/birdhouse-compose.sh
fmigneault: `SHELL_EXEC_FLAGS="${SHELL_EXEC_FLAGS:-"-e"}` is what I had in mind, but I like the `BIRDHOUSE_DEBUG_MODE` and `BIRDHOUSE_FAIL_FAST` alternatives as well. 

> Or are we planning to make this platform wide

That is a great question. 

I would like to avoid reaching a point were both `BIRDHOUSE_DEBUG_MODE` and `BIRDHOUSE_COMPOSE_DEBUG_MODE` exist.

Also, from an external user's point of view, even if the variable is named `BIRDHOUSE_COMPOSE_DEBUG_MODE`, I think anyone would assume that everything would respect the flags, since the other scripts that are called is an implementation detail that users shouldn't care about, just like we do not care how `docker compose` dispatches to `docker-compose-up` and others. You simply expect the common flags to make their way across the tool's ecosystem. There is no distinction between `BIRDHOUSE_COMPOSE_...` or `BIRDHOUSE_...` for the user calling `birhouse-compose.sh`.

birdhouse/birdhouse-compose.sh
fmigneault: I support using `BIRDHOUSE_FAIL_FAST=true` by default. It seems to be the most reasonable default. If disabled explicitly, things can break and the user is aware of it since they disabled it.

In the specific case here, it might still be relevant to have some "checkpoint logs", even if they do not fail the execution. Of course, we cannot have all errors logged since there's an infinity of combinations, but the main pre/"core"/post-compose steps seems relevant at least to narrow down to which step one should investigate where errors happen. 
If not enough of a concern, I am fine to leave it as is (removing the log).

birdhouse/birdhouse-compose.sh
fmigneault: I agree to leave it there to take advantage of error handling as early as possible.
Otherwise, we get situations similar to the logs were they cannot be called yet since the file is not sourced, and we have a few `echo <ERROR> <message>` mimicking its behavior in the meantime.

birdhouse/birdhouse-compose.sh
fmigneault: It might be relevant to use this last snippet, with a message along the lines of `"failure with docker compose command:  PROXY_HTTP_PORT=80 HOSTNAME=${BIRDHOUSE_FQDN} ${DOCKER_COMPOSE} ${COMPOSE_CONF_LIST} $* ${COMPOSE_EXTRA_OPTS}"` (or a simplified variant).  

I'm not sure what is the entire scope of error messages docker compose can throw at us in the console, but I wouldn't be surprised that some might be hard to understand without the context of the entire command that caused it.

birdhouse/birdhouse-compose.sh
tlvu: I am fine with `grep` for `up` if that simplify things.  We have been doing `if [ $1 = up ]` instead of the "proper" `while` loop for years and nothing broke catastrophically yet.

birdhouse/birdhouse-compose.sh
tlvu: > In a different case if you wanted an additional error message you could still do:
> 
> ```
> some_command_that_might_fail
> ERR=$?
> if [ ${ERR} -gt 0 ]; then
>   log ERROR "error message"
>   expect_exit "$ERR"
> fi
> ```

In this specific example, with `FAIL_FAST=true`, isn't the `some_command_that_might_fail` will exit immediately and never execute the `log ERROR` bellow? If we change to `some_command_that_might_fail || expect_exit`, it will still not reach the `log ERROR` line.

That's why I suggest adding an optional `$2` to `expect_exit` if we ever need to provide a helpful error message for some corner case.

birdhouse/birdhouse-compose.sh
fmigneault: If their scope remains under `birdhouse-compose.sh`, I agree to use `BIRDHOUSE_COMPOSE_FAIL_FAST` and `BIRDHOUSE_COMPOSE_DEBUG_MODE`. 

However, will they _**really never**_ make their way in other places? I think it is only a matter of time until we encounter a situation where we would like the debug flags to propagate or fail-fast to happen sooner under specific components/tools, just like the logs were originally limited only to the compose script, but are now everywhere via `read-configs.include.sh`.

tlvu (approved):

birdhouse/birdhouse-compose.sh
mishaschwartz: Ok what about this... I have played around with adding additional log messages to expect_exit but it really feels like we're asking the function to do too much when we already have a good way of getting this information already. 

When there is a failure that is handled by expect_exit the error message already contains the line:

> See additional messages on stderr for details or rerun this command with BIRDHOUSE_DEBUG_MODE=true

Which gives the user lots of info about how to get more details about what went wrong.

If we want to always give additional information about a command that might fail we should either:

- log the information before the command is run, which we already do in a few places (including this one):

```sh
log DEBUG "running: command_that_might_fail arg1 arg2"
command_that_might_fail arg2 arg2
```

- catch the error (@tlvu you're right that my previous example would fail with FAIL_FAST):

```sh
if ! some_command_that_might_fail arg1 arg2; then
  ERR=$?
  log ERROR "error message"
  expect_exit "$ERR"
fi
```

tlvu (approved):

PR #517, opened 2025-03-20
Remove deprecated version field from generated docker-compose files

mishaschwartz:
## Overview

The `env-local` optional component generates a docker compose file that contained a version field which are deprecated. This fixes the issue by removing the code that generates the deprecated field.

## Changes

**Non-breaking changes**
- None

**Breaking changes**
- None

## Related Issue / Discussion

- This should have been included in #504

## CI Operations

birdhouse_daccs_configs_branch: master
birdhouse_skip_ci: false

PR #520, opened 2025-03-25
Fix bug where generated docker compose file is appended to not written

mishaschwartz:
## Overview

Fixes a bug introduced when the version string was removed from the generated docker compose file. The previous line used `>` which truncated the file before writing. Now that the previous line is removed, the truncation logic needed to be applied elsewhere. 

## Changes

**Non-breaking changes**
- bug fix

**Breaking changes**

## Related Issue / Discussion

- Fixes a bug introduced in #517

## CI Operations

birdhouse_skip_ci: false


tlvu (approved):
LGTM

fmigneault (approved):

PR #522, opened 2025-03-26
Make sure that authentication routes use the correct scheme

mishaschwartz:
## Overview

Two components that were added after the `BIRDHOUSE_PROXY_SCHEME` environment variable was introduced did not use it when checking whether a user was authenticated to view a resource using ``twitcher``'s verify route. This is now fixed so that the proper scheme is used.

## Changes

**Non-breaking changes**
- bug fix for development/test mode

**Breaking changes**
- None

## Additional Information

## CI Operations

birdhouse_skip_ci: false


tlvu (approved):
LGTM

fmigneault (approved):

PR #523, opened 2025-03-27
Fix setting `BIRDHOUSE_LOCAL_ENV` in `deploy.sh` to the wrong argument

mishaschwartz:
## Overview
  
The `deploy.sh` script was setting the `BIRDHOUSE_LOCAL_ENV` variable to the first argument to the script instead of the second one. This is now fixed.

## Changes

**Non-breaking changes**
- Bug fix

**Breaking changes**
- None
 
## Additional Information

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/birdhouse/deployment/deploy.sh
+++ b/birdhouse/deployment/deploy.sh
@@ -67,7 +67,7 @@ else
     shift
 fi
 
-BIRDHOUSE_LOCAL_ENV="${1:-${BIRDHOUSE_LOCAL_ENV:-"${COMPOSE_DIR}/env.local"}}"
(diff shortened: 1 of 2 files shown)


birdhouse/deployment/deploy.sh
```diff
@@ -67,7 +67,7 @@ else
     shift
 fi
 
-BIRDHOUSE_LOCAL_ENV="${1:-${BIRDHOUSE_LOCAL_ENV:-"${COMPOSE_DIR}/env.local"}}"
+BIRDHOUSE_LOCAL_ENV="${2:-${BIRDHOUSE_LOCAL_ENV:-"${COMPOSE_DIR}/env.local"}}"
```
tlvu, line 70: I'll test but probably not it because there is a `shift` right above, meaning `$2` becomes `$1`.

tlvu: I tested, it's not the fix.

PR #524, opened 2025-03-28
deprecate portainer

### LINKED ISSUE #521: :question: [Question]: Is anyone using portainer?
## Questions

I'm not sure what the benefit is and even after #335 was resolved it remains a security issue. If no one is using it I suggest we deprecate it.

If people are using it then let's discuss how it's being used so that we can figure out what to do with it.

## References

See discussion here: https://github.com/bird-house/birdhouse-deploy/issues/335#issuecomment-1701895067

## Environment

| Server/Platform URL   | 
| Version Tag/Commit    |  
| Related issues/PR     | 
| Related components    | portainer

mishaschwartz:
## Overview

The portainer component is not currently being used and is not actually usable outside of a very specific host machine configuration. This change deprecates the component by moving it to the `deprecated-components` directory. It can still be enabled from that path if desired.

## Changes

**Non-breaking changes**
- None

**Breaking changes**
- Deprecates portainer

## Related Issue / Discussion

- Resolves #521 

## Additional Information

- I think this should require a minor version bump just in case someone out there is using portainer since this will mean that they'll need to manually change the entry in their local environment file.

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Old paths from deprecated components. Keep these so that old config files remain uncommittable after updates.
+portainer/config/proxy/conf.extra-service.d/portainer.conf
```
fmigneault, line 2: Should `components/portainer` be added as well?

mishaschwartz: That's not necessary really. All the other files in the portainer directory are tracked by git so they will be deleted by this change. Only the file that was in the .gitignore will be left behind.

tlvu (approved):
LGTM

birdhouse/components/.gitignore
fmigneault: I'm thinking in terms of any place that could have them still present on this could re-inject them by mistake. Though, not blocking, since this could be detected and reverted if it happens in another PR.

fmigneault (approved):

PR #533, opened 2025-05-01
Update `stac` service to use `crim-ca/stac-app:1.1.0` image

mishaschwartz:
## Overview

  This updates the version of `stac-fastapi` to version 5 (currently the latest) and resolves and issue
  where paging links did not work properly.

  See more details [here](https://github.com/crim-ca/stac-app/pull/27).

  Also fixes:

  The nginx configuration for `twitcher` was creating a `Forwarded` header to help `stac` construct a `base_url`
  behind the reverse proxy. However, with newer versions of `stac-fastapi` (the application running the `stac`
  service), the `Forwarded` header is being parsed incorrectly which means that the `base_url` was incorrectly
  formed.

  This change removes the problematic `Forwarded` header and instead send the information to the `stac` application
  using the `X-Forwarded-Port`, `X-Forwarded-Proto`, and `X-Forwarded-Host` headers. This technique allows `stac` 
  to generate the correct `base_url` for all versions (up to the current version 5).

## Changes

**Non-breaking changes**
- update stac to version 1.1.0
- Bug fix

**Breaking changes**
- None

## Related Issue / Discussion

- The issue this resolves was first raised here: https://github.com/crim-ca/stac-app/pull/27#pullrequestreview-2809985820
- A description of the issue can be found here: https://github.com/crim-ca/stac-app/pull/27#issuecomment-2845558493

## Additional Information

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/twitcher/config/proxy/conf.extra-service.d/twitcher.conf
fmigneault, line 1: Please leave it as `.template`, even if there is no more actual templating happening in this case. Going between different commits will cause unnecessary diffs and conflicts with the file that was ignored before, and we might just need to switch it back to a template as some later date.

fmigneault (changes requested):

fmigneault:
New tag is available: https://github.com/crim-ca/stac-app/pkgs/container/stac-app/406793397?tag=1.1.0

fmigneault (approved):

mishaschwartz:
> New tag is available: https://github.com/crim-ca/stac-app/pkgs/container/stac-app/406793397?tag=1.1.0

Do you suggest we update it in this PR?

fmigneault:
Yes. Since we just did a test and validated they work together, might as well do it to have the paging links fixes as well.

fmigneault (approved):

PR #542, opened 2025-06-04
Add prometheus rules for long-term metrics

### LINKED ISSUE #447: :bulb: [Feature] Add recording rules to Prometheus configuration to store hourly/daily metrics

huard:
## Overview

New long-term metrics.

## Changes

**Non-breaking changes**
- Adds rules for Prometheus to store hourly and daily metrics. 

**Breaking changes**
- Removed metric `instance:network_bytes_transmitted:sum_rate1h` and replaced it with the transmit and receive metrics. I thought that "compositions" could be done in Grafana, and prometheus should store "simple" metrics. 

## Related Issue / Discussion

- Resolves #447 

## Additional Information

- [ ] I think I've checked all metrics manually, but this is using the PAVICS instance which is behind master and does not include the thredds download metric. 

## CI Operations

birdhouse_skip_ci: false


mishaschwartz (approved):
Thank you for adding these.
Please also add an entry in the changelog

tlvu (approved):
LGTM.  You're the ones who knows best which metrics to keep for long term !

I'll see if I can use my current VM that I used to test tip of PAVICS to test your PR.  I guess this PR depend on your other PR that update the line parsing?

huard:
@tlogan2000 Could you check the list of long-term rules and note missing information you'd like to have for reporting ?

huard:
@tlvu PR #540 is not a hard dependency, because master has the metric, it's just meaningless.

huard:
Question: would the test suite failed if a prometheus expression was malformed ?

tlvu:
@huard I've deployed this extra longterm rules together with the 2nd longterm Prometheus.  

I found a slight typo in the rule: https://github.com/bird-house/birdhouse-deploy/pull/551

But what's puzzling me is how to see those longterm rules on the 2nd Prometheus side.  Maybe @mishaschwartz can chime in?

mishaschwartz:
@tlvu 

All the new metrics should be visible at both the prometheus/rules and prometheus-longterm-metrics/rules endpoint.

By default though the longterm metrics instance only scans once per hour so the rules won't be visible there for a while once the new rules are applied. You'll also have to wait a full day before the daily metrics start showing up.

mishaschwartz:
@tlvu 

I think that your instance is actually correct in only showing the rules in `prometheus/rules` since they're not rules defined for the `prometheus-longterm-metrics` instance (which is only consuming those rules that are exported, not calculating those rules itself). 

The fact that you only see the the 1d variables but not the 1h ones is a different story and I'm not sure why that is. I have tested it out on my end and I'm able to get both the 1h and 1d metrics on the `prometheus-longterm-metrics` instance.

One thing to think about is that if there is no data for a given variable then it won't show up. So you could experiment by:

- choose a rule that doesn't appear in the `prometheus-longterm-metrics` instance
- go to the `prometheus` instance and in the "Graph" tab manually calculate the rule and see if there is any data associated with it
- for example: if you're not seeing data for the `instance:cpu_load:avg_rate1h` rule you can manually calculate `avg by(instance) (rate(node_cpu_seconds_total{mode!="idle"}[1h]))` in the "graph" tab and see if it gives you any data.

tlvu:
> The fact that you only see the the 1d variables but not the 1h ones is a different story and I'm not sure why that is.

I `docker volume rm prometheus_longterm_persistence` and restart from scratch.  Maybe it was just fluke.  Let's see tomorrow if I will have both 1d and 1h stats.

huard:
Note that Grafana is not really designed to display hourly values... It assumes data is valid for 5 minutes, so if we scrape every hour, there are 55 minutes without any data. There might be ways out of this, but we might be going against the current here.

huard:
The long-term metrics show up in the rules, but not in the "metrics explorer"
See 
/prometheus/graph?g0.expr=instance%3Acpu_load%3Aavg_rate1h&g0.tab=0&g0.display_mode=lines&g0.show_exemplars=0&g0.range_input=1w

mishaschwartz:
It works for me:

```sh
$ docker exec -it proxy curl -G --data-urlencode 'match[]={group="longterm-metrics"}' http://prometheus:9090/prometheus/federate
```

And it shows up in the `prometheus-longterm-metrics` instance. 

Other things you can check:

- are the node-exporter and cadvisor components working?
- you can check on the prometheus instance if the targets are all being polled properly
- do the rules defined in `prometheus` show data when you query them in the Graph tab?

Issue #536, opened 2025-05-09
:bulb: [Feature] Backup *representative* data functionality


fmigneault:
## Description

Using https://github.com/bird-house/birdhouse-deploy/pull/532, options are provided for backup/restore of databases. This is perfect for recreating VMs with exact configurations. However, as detailed in https://github.com/bird-house/birdhouse-deploy/pull/534, any DB migration that might be required to switch versions might make this backup data unusable. 

Given the varying complexity of various DB migration technologies, one approach could be to simply `export` the data in a format that is appropriate for the API service that uses the database. Although not as efficient as direct backup/restore of the database, going through the API and underlying data representation can help migrate data across versions, and even potentially across servers not using `birdhouse-deploy`, while leveraging similar APIs.

Therefore, the `backup [create,restore]` CLI mode provided by https://github.com/bird-house/birdhouse-deploy/pull/532 should be further extended to provide an `export`, which would perform the relevant API/FileSystem/whatever calls to get a "relevant data representation" for that API. An `import` could be considered as well.

An example use case is STAC API. The exported data should be a nested hierarchy of collections/items. Those could then be reimported in another server / migrated DB using [`STACpopulator.implementations.DirectoryLoader`](https://github.com/crim-ca/stac-populator/tree/master/STACpopulator/implementations/DirectoryLoader). 

## References

- https://github.com/bird-house/birdhouse-deploy/pull/532
- https://github.com/bird-house/birdhouse-deploy/pull/534
- https://github.com/crim-ca/stac-populator/tree/master/STACpopulator/implementations/DirectoryLoader

## Concerned Organizations

all

mishaschwartz:
I really like this idea and it should be relatively easy to adapt the backup feature introduced in #532 to accommodate this.

It would be useful here to identify which services could be exported/imported in this way. So far I'm thinking:

- `stac` as JSON files
- all the postgres data (WPS, geoserver, magpie, ...) using `COPY ... WITH (FORMAT CSV)`
   - though this will break if a later migration changes the table definitions (column names, data types, etc.)
- all the mongodb data (weaver, cowbird) using `mongoexport` and `mongoimport`:
   - this probably has the same issue as the postgres one 

Is STAC the only safe one? Am I missing something?

I've already started working on an export solution for `stac` as part of the `stac-populator` project so that could be used as part of this. I'll link that here once I've made a PR for that.

tlvu:
Great idea !

[...]

PR #548, opened 2025-06-12
Add option to backup "representative" application data

mishaschwartz:
## Overview

Representative data is an application agnostic version of the stateful data used by components to store the current state of the running service.

This includes an option to backup and restore representative data for the `stac` component. Other components should be added in future updates.

## Changes

**Non-breaking changes**
- Adds a new backup data type

**Breaking changes**
- None

## Related Issue / Discussion

- completes the first step of #536 but does not complete it since backup restore functionality should also be added for components other than STAC (see: https://github.com/bird-house/birdhouse-deploy/issues/536#issuecomment-2873310426)

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/stac/backup
```diff
@@ -13,3 +13,41 @@ restore_application() {
...
   backup_restore_runner 'stac postgis database' _do_restore 'stac' 'stac'
 }
+
+backup_representative() {
+  _do_backup() {
+    docker run --rm \
+               --network "${COMPOSE_PROJECT_NAME}_default" \
+               -v ${BIRDHOUSE_BACKUP_VOLUME}:/backup \
+               --user root \
+               --env PYESSV_ARCHIVE_HOME=/home/stac/.esdoc/pyessv-archive \
```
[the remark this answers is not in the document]
mishaschwartz: It doesn't need to be set but there's a warning that is unrelated so it could confuse the user into thinking something is wrong.

Ideally it would be set in the image, yes

birdhouse/README.rst
```diff
@@ -504,6 +497,68 @@ For details about the backup and restore commands run any of the following:
...
+  * backing up and restoring representative data will probably take a much longer time than
+    application data.
+
+  * for example: STAC objects from the ``stac`` component stored as JSON files.
+
+* user data
+
+  * data created directly by users of birdhouse.
+
+  * for example: files written to the ``jupyterhub`` component's user workspaces
```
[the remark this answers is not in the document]
fmigneault: Yes, that's fine as app data and within another PR. User permissions are specific to Cowbird/Magpie, so they would take care of this aspect based on the backed-up/restored structure.

bin/birdhouse
```diff
@@ -115,6 +115,7 @@ Options:
   -a, --component-application-data    Backup application data for a given component (e.g. magpie). This can be specified multiple times. Use '*' to select all components.
   -u, --component-user-data           Backup user data for a given component (e.g. cowbird). This can be specified multiple times. Use '*' to select all components.
   -l, --component-logs-data           Backup logs data for a given component (e.g. thredds). This can be specified multiple times. Use '*' to select all components.
+  -r, --representative-data           Backup representative data for a given component (e.g. stac). This can be specified multiple times. Use '*' to select all components.
```
tlvu, line 118: Since this option is only for `stac` and the moment, could we add "Not all components are ready for this mode.  Please double check before enabling this mode"?  Just to clarify user expectations.

Also, what happen if the use really does `-r '*'` but only `stac` is supported?

bin/birdhouse
```diff
@@ -149,6 +150,7 @@ Options:
   -a, --component-application-data    Restore application data for a given component (e.g. magpie). This can be specified multiple times. Use '*' to select all components.
   -u, --component-user-data           Restore user data for a given component (e.g. cowbird). This can be specified multiple times. Use '*' to select all components.
   -l, --component-logs-data           Restore logs data for a given component (e.g. thredds). This can be specified multiple times. Use '*' to select all components.
+  -r, --representative-data           Restore representative data for a given component (e.g. stac). This can be specified multiple times. Use '*' to select all components.
```
tlvu, line 153: I never noticed but we have duplicate identical help text?

birdhouse/components/stac/backup
```diff
@@ -13,3 +13,41 @@ restore_application() {
...
+  _do_restore() {
+    ${BIRDHOUSE_EXE} --quiet compose rm -f stac-db
+    docker volume rm "${COMPOSE_PROJECT_NAME}_stac-db"
+    ${BIRDHOUSE_EXE} --quiet compose create stac-db
+    ${BIRDHOUSE_EXE} --quiet compose start stac-db stac
+    count=0
+    while [ "$(docker inspect --format='{{.State.Health.Status}}' stac)" != "healthy" ]; do
+      [ "$count" -gt 5 ] && log ERROR "stac service is not healthy after 10 seconds" && return 1
+      sleep 2
+      count="$(( $count + 1 ))"
```
tlvu, line 41: On slow server, could stac startup take more than 10 sec?  Maybe make this timeout configurable?  Not critical.  Can keep hardcode for now until we actually need it configurable.

birdhouse/components/stac/default.env
```diff
@@ -15,6 +15,14 @@ export STAC_BROWSER_VERSION=3.0.0-beta.5-crim-docker-image-push
 export STAC_BROWSER_IMAGE='ghcr.io/crim-ca/stac-browser:docker_image_push'
 export STAC_BROWSER_IMAGE_URI='${STAC_BROWSER_IMAGE}'
 
+export STAC_POPULATOR_BACKUP_DOCKER=ghcr.io/crim-ca/stac-populator
```
tlvu, line 18: Is this the same as `STAC_POPULATOR_{DOCKER,VERSION` in `optional-components/stac-populator/default.env`?  

Should we use the same version (default to the other variable `STAC_POPULATOR_{DOCKER,VERSION` if it is enabled) or it creates problems, so 2 separate version is better?

bin/birdhouse
tlvu: Ah ok, I didn't know that "note" section is there.  All fine then.

bin/birdhouse
tlvu: OH sorry, read too fast, it's the exact same text except "Backup" becomes "Restore" so yes they are not identical.

birdhouse/components/stac/default.env
tlvu: Ah ok, I wasn't aware of the state of the stac-populator component.  Fine then.

birdhouse/components/stac/backup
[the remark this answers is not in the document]
mishaschwartz: https://github.com/crim-ca/stac-populator/pull/98

PR #549, opened 2025-06-16
update restic docs

mishaschwartz:
## Overview

Clarify/add to the restic documentation for backups.

Also removes the read-only flag from the `BIRDHOUSE_BACKUP_SSH_KEY_DIR` bind mount used by the `do_restic` command. This is necessary to support unique SSH keypairs generated for restic.

## Changes

**Non-breaking changes**
- documentation update

**Breaking changes**
- None

## CI Operations

birdhouse_skip_ci: false


bin/birdhouse
```diff
@@ -453,7 +453,7 @@ extra_args() {
 # Run a restic command.
 do_restic() {
   source_env
-  restic_cmd="docker run --rm --env-file='${BIRDHOUSE_BACKUP_RESTIC_ENV_FILE}' --env 'RESTIC_PASSWORD_COMMAND=' -v '${BIRDHOUSE_BACKUP_SSH_KEY_DIR}:/root/.ssh:ro' -v ${BIRDHOUSE_BACKUP_VOLUME}:/backup ${BIRDHOUSE_BACKUP_RESTIC_EXTRA_DOCKER_OPTIONS} $(extra_args) ${RESTIC_IMAGE} $@"
+  restic_cmd="docker run --rm --env-file='${BIRDHOUSE_BACKUP_RESTIC_ENV_FILE}' --env 'RESTIC_PASSWORD_COMMAND=' -v '${BIRDHOUSE_BACKUP_SSH_KEY_DIR}:/root/.ssh' -v ${BIRDHOUSE_BACKUP_VOLUME}:/backup ${BIRDHOUSE_BACKUP_RESTIC_EXTRA_DOCKER_OPTIONS} $(extra_args) ${RESTIC_IMAGE} $@"
```
fmigneault, line 456: Why does this need write access now? Seems dangerous, given the defaults is `${HOME}/.ssh`.

Would be less critical if it was a specific key for restic only, but seems too broad for all user keys (which should also be accessible only to the user, no read/write group/other permissions).

mishaschwartz: > Why does this need write access now?

https://github.com/bird-house/birdhouse-deploy/blob/e68caa767a8dda253c94fd73f73b77283a5afc20/birdhouse/README.rst?plain=1#L606-L628

> given the defaults is ${HOME}/.ssh

That's not the default. The default is an empty directory.

fmigneault: By bad. Read too quickly the line https://github.com/bird-house/birdhouse-deploy/blob/02206e1cc5cb48e22103ac9e61bf667cd96263ef/birdhouse/env.local.example#L689-L690

However, the issue still stands. Does restic actually need write access to it? The described steps seems to indicate to prepare everything in advance, so restic is expected to simply read it, no?

mishaschwartz: `ssh-keygen` needs to be run inside the docker container in the examples given. If we can't write to the mounted directory then we can't save a the key files.

CHANGES.md
```diff
@@ -60,12 +62,12 @@
   Create a new scheduler job at `optional-components/scheduler-job-backup` which runs the `bin/birdhouse backup create` 
   command at regular intervals to ensure that the birdhouse stack's data is regularly backed up.
 
-  To configure this job you may set the following variables:
+  To configure this job you must set the following variables:
     - `SCHEDULER_JOB_BACKUP_FREQUENCY`:
       - Cron schedule when to run this scheduler job (default is `'1 1 * * *'`, at 1:01 am daily)
     - `SCHEDULER_JOB_BACKUP_ARGS`:
```
tlvu, line 68: If these var changed from "may" to "must", they could be added to `VARS` to enforce they are "must"?

tlvu (changes requested):
I am fine if we absolutely need to keep write-access for the ssh key, just clarify to the user that restic will have write-access so much safer to not give the entire `$HOME/.ssh` but some dedicated folder specific for this task only.

CHANGES.md
mishaschwartz: https://github.com/bird-house/birdhouse-deploy/blob/02206e1cc5cb48e22103ac9e61bf667cd96263ef/birdhouse/optional-components/scheduler-job-backup/default.env#L16-L20

tlvu (approved):
LGTM

bin/birdhouse
mishaschwartz: > autodeploy is running as root inside its container so I never had this problem!

the restic container is also running as root 

>  if you've had stated the real reason earlier, we would not have such a long conversation :)

https://github.com/bird-house/birdhouse-deploy/blob/15b691cea5da3b077765a4f38cd8dff57bc9ddc0/birdhouse/README.rst?plain=1#L565-L566

Sorry I should have provided the comment above earlier.

> if you make your ssh-key outside owned by the same UID and GID of restic container, would that make restic happy?

No, changing the UID/GID of the files and folders does not have any affect.

This only happens when remotely accessing a server with specific SSH server (openssh-server ... I'm not sure which version). It's possible that github uses something else that doesn't reject the keys.

## Follow up

Ok so after further extensive experimentation it looks like this is an issue only if using a private key generated with the rsa algorithm which was recently deprecated because it's too hackable. If you use the new default algorithm (ECDSA) or any of the other supported algorithms then even new ssh servers won't complain.

The initial bug that I couldn't figure out seemed to be due to the fact that the `ssh-keygen` command on my host machine was an older version than the one in the restic container so the behaviour of generating a key inside or outside the container was different.

With this in mind, I am now happy to keep the volume read-only but I'm going to throw in a bunch of additional warnings so that people can debug this issue themselves.

PR #563, opened 2025-07-04
Clean Prometheus rules

huard:
## Overview

Whlie trying to access the long-term metrics, I realized that the names were not matching the Prometheus conventions outlined in https://prometheus.io/docs/practices/naming/

Conventions also suggest to use base units (e.g. bytes, not kb) in metrics, so I changed the THREDDS download counter. 

## Changes

- Prometheus TDS counter records bytes instead ok kb, as per convention. 
- Renamed long-term rules to adhere to naming conventions.

## Related Issue / Discussion

- Resolves [issue id](url)

## Additional Information

- [ ] Things to do...

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,10 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Fixes
+
+- Modify Prometheus Thredds counter to record bytes instead of kb. 
+- Change longterm rule names to follow naming conventions.
 
 [2.16.4](https://github.com/bird-house/birdhouse-deploy/tree/2.16.4) (2025-07-03)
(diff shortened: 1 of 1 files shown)


tlvu:
Deployed this change on my test VM.  Feel free to hit my VM to generate data for the monitoring.

fmigneault (approved):

mishaschwartz (approved):

huard:
The only thing that I'm not sure is working is the `instance:thredds_transfer_size_bytes:increase1h`

Is the log-parser option component enabled? There are stricty no change over time to the counter values, even after I've downloaded files from your server.

tlvu (approved):
No error with these changes on my test system.  Not sure about the `instance:thredds_transfer_size_bytes:increase1h` issue.  I'll run full Jenkins against my test system to see if I get some clues.

tlvu:
> @tlvu These ones? {56ABA400-A949-4EB9-8BCE-59105B5CF78A}
> 
> There was no new commit since then, so no new run.

Oh so tests were actually ran but we never noticed they failed !  Opps !

PR #573, opened 2025-08-14
User kernels directory should always be writable

mishaschwartz:
## Overview

The destination used to symlink user kernels (`/usr/local/share/jupyter`) is not always writable depending on the jupyterlab docker image that is used to spawn the jupyterlab containers. To ensure that it is always writable this places the link under `/var/tmp` which is guaranteed to be writable.

## Changes

**Non-breaking changes**
- bugfix

**Breaking changes**
- None

## CI Operations

birdhouse_skip_ci: false


fmigneault (approved):
If JupyterLab able to detect that directory, this seems OK.

tlvu (approved):

tlvu:
> recognize kernels in the home directory but the pavics images recognize it fine

Is it because in our PAVICS Jupyter image, the username is `jenkins` and the ones from [jupyter base-notebook](https://quay.io/repository/jupyter/base-notebook) is `jovian` (I might have misspelled it)?

tlvu:
> some issue with the jupyter images that we've luckily managed to avoid for all these years.

Here's another hint: I tried to replicate the same "environment" when building the PAVICS Jupyter image so they look as similar as possible to the [jupyter base-notebook](https://quay.io/repository/jupyter/base-notebook) ones.  Obviously there are a few differences (ex: we need a different base image because we install more stuff, we need a different username so it is compatible with Jenkins testing).  By same "environment" I mean I took their Dockerfile and I tried to copy the same steps in there, to the best I could, at that time.

I have not reviewed since then.  It is possible they change their "steps" which broke this feature.

PR #579, opened 2025-08-27
Drop unsupported `pytest-lazy-fixture` python package for tests

mishaschwartz:
## Overview

This package is no longer maintained and breaks for `pytest` versions 8+. We do not need it for our tests so it was dropped. In the future if we need to support lazy fixtures we should use the `pytest-lazy-fixtures` package instead (which is actively maintained).

## Changes

**Non-breaking changes**
- minor test changes

**Breaking changes**
- None

## Related Issue / Discussion

- Fixes issue of failed tests in #578 

## Additional Information

@fmigneault I believe you added the lazy-fixture to that specific test. I've removed it as I don't see why this had to be made a lazy fixture. If there is a good reason why it should not be removed please let me know.

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,13 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Changes
+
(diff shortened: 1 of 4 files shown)


Zeitsperre (approved):
LGTM

tlvu (approved):
LGTM, pending @fmigneault if whether the lazy fixtures is really needed.

PR #577, opened 2025-08-26
Revert CI labeler action

fmigneault:
## Overview

The V5 version set in https://github.com/bird-house/birdhouse-deploy/pull/574 does not align with the provided configuration.

See for example on an active PR: 
https://github.com/bird-house/birdhouse-deploy/actions/runs/17250273776/job/48950581278?pr=576
{DA04A23E-1E16-42E1-8860-BF68C3682F34}

## CI Operations

birdhouse_daccs_configs_branch: master
birdhouse_skip_ci: true

PR #580, opened 2025-08-27
Fix labeler configuration

Zeitsperre:
## Overview

This PR fixes an oversight that wasn't caught when upgrading the `actions/labeler` action from v2.0 (released in 2019) to v5.0 (last released in 2023). The vocabulary has changed such that there is much more control over the trigger for labelling pull requests based on changed/unchanged files.

For our purposes, we simply require the following in entries:
```YAML
<label>
  - changed-files:
    - any-glob-to-any-file:
      - file.txt
      - ...
```

## Changes

**Non-breaking changes**
- Updates the `labeler.yml` configuration to use modern conventions.

## Related Issue / Discussion

- Closes #577 

## Additional Information

The reason for performing this update is not solely "new = better", but also because the underlying dependencies (primarily node.js) used in the action will eventually be removed from `ubuntu-latest`. These conventions have been stable for the last two major releases of `actions/labeler`.

## CI Operations

birdhouse_skip_ci: false


Zeitsperre:
There isn't a way, no. This configuration resembles what I've used in other repositories using the same version of `actions/labeler`, so I don't imagine we'll have any big surprises.

For updates, there hasn't been a major release in a while (though development is still occurring). I don't imagine there being huge breaking changes for at least a few years.

tlvu (approved):
LGTM

Zeitsperre:
@fmigneault Looks good! `label (pull_request)` is working as intended. Pull request target should work fine once on master.

fmigneault:
@Zeitsperre 
FYI. You are not obligated to bump the version on every PR, especially if they are CI-only and do not impact the platform. You can stack a few PR updates together and make a release if they are relevant together, or leave it up for the next release if only CI related.

Zeitsperre:
Ah, got it. I was under the impression that all pushes to `master` should be versioned. Will be a bit more conservative on those bumps moving forward.

PR #582, opened 2025-09-06
fix weaver landing page media-type links

fmigneault:
## Overview

Fix weaver landing page media-type links reported in `/services`.

Weaver only had JSON responses before (<5.7), but it offers an HTML interface since.
Because browsers push HTML Accept header by default, the assumed links/media-types were not adequate anymore. 

## Changes

**Non-breaking changes**
- Apply the format specifiers to return the corresponding media-types.

**Breaking changes**
- n/a

birdhouse_skip_ci: true


tlvu (approved):
LGTM

PR #584, opened 2025-09-11
add services item endpoint

fmigneault:
## Overview

Improve access to services metadata.

Example: https://hirondelle.crim.ca/services/weaver

Preview the updated metadata at a glance here: https://hirondelle.crim.ca/

### Open Question ❓ 

Should the `service-config.json.template` be modified to be the objects directly instead of a list? 
I have tested it quickly, and the code works already since the join operation strips the `[]`. There is only an issue for STAC that defines 2 services. That would require some kind of loop in the script to deal with multiple `service-config.{service}.json.template`, or involving a `docker ghcr.io/jqlang/jq:latest` call to extract them from current lists (which would be used also for all services without any change to the array form).

## Changes

**Breaking changes**

- STAC: Split the services API/UI into distinct `components/stac` and  `components/stac-browser`.
  - Enabling only `components/stac` will not provide `/stac-browser/` endpoint anymore. It must be explicitly enabled.
  - Enabling `components/stac-browser` will enforce enabling `components/stac` as a dependency.
  - Add missing `STAC_BROWSER_DOCKER` variable to respect the standard naming convention used by other components.
  - Each component has their respective `service-config.json.template` definition.

**Non-breaking changes**

- Cowbird: Define variables in the standard naming convention used by other components.
  - `COWBIRD_DOCKER`, `COWBIRD_IMAGE`, `COWBIRD_IMAGE_API`, `COWBIRD_IMAGE_WORKER`, `COWBIRD_IMAGE_URI`

- Services: Add more links and references.
  - Add a `/services/{serviceId}` endpoint to retrieve individual service metadata definitions.
  - Add missing `service-config.json.template` files for Magpie, Twitcher and Cowbird.
  - Extend the provided metadata links for multiple services. 
  - Change all `service-config.json.template` definition to be directly the JSON object rather than an array.
    This allows reporting these objects directly on `/services/{serviceId}`, while `/services` combines them.

**Breaking changes**
- n/a

## CI Operations

birdhouse_skip_ci: false


fmigneault:
@mishaschwartz 
Sounds good. I'll open a PR to add the new property. Then we can bump the schema version, so I can include it in these service configs.

fmigneault:
@mishaschwartz 
Linking this as it is relevant: 
- https://github.com/DACCS-Climate/Marble-node-registry/issues/39

Currently, the `types` are not even considered (or at least shouldn't, since they are not officially part of the schema). They were predefined in preparation of that issue.

Filtering should be based on `keywords`, and `other` is used in this PR to allow omitting the new "auth-typed" services.

fmigneault:
run tests 

Was failing because of the missing `stac-browser` now a separate component

PR #587, opened 2025-09-18
fix http->https redirects on other methods then GET

fmigneault:
## Overview

When doing any other request then `GET` using `http://`, the proxy was returning a `302`, leading to lost of the HTTP method. Returning `308` ensures that the HTTP method is preserved during the redirection.

Below is shown the before/after where `302` caused `HTTP POST` to become `HTTPS GET` leading to unexpected response in the context of the original request.

The fixed variant with `308` correctly leads to the `HTTPS POST` (and then the expected unauthorized response in this case).

{5A6B23CC-F275-4138-91CC-B91411EDB69B}

## Changes

**Non-breaking changes**
- Fix the `HTTP->HTTPS` redirection when the method is not `GET`

**Breaking changes**
- If any script was looking explicitly at 302, it won't work anymore, though it shouldn't be doing so in the first place to be resilient to specific 3xx code changes.

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,9 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Fixes
+
+- Proxy: Fix lost HTTP method on redirect to HTTPS.
 
 [2.17.2](https://github.com/bird-house/birdhouse-deploy/tree/2.17.2) (2025-09-12)
 ------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mishaschwartz (approved):

tlvu (approved):
Just curious, 308 is permanent redirect and 307 is temporary redirect.  What's the difference?   A redirect is a redirect, how can it be permanent or temporary?  How "sticky" is this permanent?

mishaschwartz:
@tlvu 

Permanent vs. temporary redirects get interpreted differently by the client. A client that gets a temporary redirect will keep trying the _original_ location in future calls, but if it gets a permanent redirect it will update the request on the client side and make requests to the _new_ location going forward.

For example:

### 307 
  - a client sends a request to http://example.com
  - the server returns a 307 (temporary redirect) with location=https://example.com
  - the client still sends the next request to http://example.com

### 308
   - a client sends a request to http://example.com
   - the server returns a 308 (permanent redirect) with location=https://example.com
   - the client changes the next request to http://example.com to https://example.com (on the client side) before sending the request

fmigneault:
I'm merging this along #584 to bump once.

PR #594, opened 2025-10-02
Document how to allow Jupyterlab containers to access GPUs

mishaschwartz:
## Overview

Documentation update to describe steps to enable GPU support for jupyterhub.

## Changes

**Non-breaking changes**
- none
**Breaking changes**
- none

## Additional Information

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/birdhouse/components/README.rst
+++ b/birdhouse/components/README.rst
@@ -828,25 +828,27 @@ If the host machine has GPUs and you want to make them available to the docker c
 
 1. ensure that the GPU drivers on the host machine are up to date
 2. install the `NVIDIA container toolkit`_ package on the host machine
-3. configure docker to use the NVIDIA as a container run-time (See the "configure docker" section in `NVIDIA container toolkit`_)
+3. `configure docker`_ to use the NVIDIA as a container run-time
 4. restart docker for the changes to take effect
 5. add the following to the ``JUPYTERHUB_CONFIG_OVERRIDE`` variable in your local environment file:
 
-.. code-block:: shell
+.. code-block:: python
 
-    export JUPYTERHUB_CONFIG_OVERRIDE='
     # enable GPU support
     import docker
 
     c.DockerSpawner.extra_host_config["device_requests"] = [
         docker.types.DeviceRequest(count=-1, capabilities=[["gpu"]])
     ]
-    '
+    
 
 This will allow the docker containers to access all GPUs on the host machine. To limit the number of GPUs you want to make available
 you can change the ``count`` value to a positive integer or you can specify the ``device_ids`` key instead. ``device_ids`` takes a list
 of integers representing the devices (GPUs) that you want to enable. Device IDs for each GPU can be inspected by running the ``nvidia-smi``
 command on the host machine.
+
+The `driver capabilities`_ setting indicates that this device request is for GPUs (as opposed to other devices that may be 
+available such as TPUs). 
 
 For example, if I only want to make available GPUs with ids 1 and 4 you would set:
 
@@ -855,7 +857,8 @@ For example, if I only want to make available GPUs with ids 1 and 4 you would se
    docker.types.DeviceRequest(device_ids=[1, 4], capabilities=[["gpu"]])
 
 .. _NVIDIA container toolkit: https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html 
-
+.. _configure docker: https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html#configuring-docker
+.. _driver capabilities: https://docs.docker.com/reference/compose-file/deploy/#capabilities
 
 How to Enable the Component
 ---------------------------


birdhouse/components/README.rst
```diff
@@ -821,6 +821,42 @@ Usage
...
+
+1. ensure that the GPU drivers on the host machine are up to date
+2. install the `NVIDIA container toolkit`_ package on the host machine
+3. configure docker to use the NVIDIA as a container run-time (See the "configure docker" section in `NVIDIA container toolkit`_)
+4. restart docker for the changes to take effect
+5. add the following to the ``JUPYTERHUB_CONFIG_OVERRIDE`` variable in your local environment file:
+
+.. code-block:: shell
+
+    export JUPYTERHUB_CONFIG_OVERRIDE='
```
tlvu, line 837: Should append to `JUPYTERHUB_CONFIG_OVERRIDE` instead because you never know if that variable is used somewhere before.

birdhouse/components/README.rst
```diff
@@ -821,6 +821,42 @@ Usage
...
+    # enable GPU support
+    import docker
+
+    c.DockerSpawner.extra_host_config["device_requests"] = [
+        docker.types.DeviceRequest(count=-1, capabilities=[["gpu"]])
+    ]
+    '
+
+This will allow the docker containers to access all GPUs on the host machine. To limit the number of GPUs you want to make available
```
tlvu, line 846: So all Jupyter users will have access to the GPU?  And if they happen to all use the GPU at the same time, will they step on each other foot?

tlvu (approved):
LGTM, except the part that override `JUPYTERHUB_CONFIG_OVERRIDE`, should be appending.

birdhouse/components/README.rst
```diff
@@ -821,6 +821,42 @@ Usage
...
 same user name and password as Magpie. They will then be able to launch a personal jupyterlab server.
 
+GPU support for jupyterlab containers
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If the host machine has GPUs and you want to make them available to the docker containers running Jupyterlab:
+
+1. ensure that the GPU drivers on the host machine are up to date
+2. install the `NVIDIA container toolkit`_ package on the host machine
+3. configure docker to use the NVIDIA as a container run-time (See the "configure docker" section in `NVIDIA container toolkit`_)
```
fmigneault, line 831: Better to link directly (https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html#configuring-docker), assuming this is the one you meant?

birdhouse/components/README.rst
```diff
@@ -821,6 +821,42 @@ Usage
...
+    # enable GPU support
+    import docker
+
+    c.DockerSpawner.extra_host_config["device_requests"] = [
+        docker.types.DeviceRequest(count=-1, capabilities=[["gpu"]])
+    ]
+    '
+
+This will allow the docker containers to access all GPUs on the host machine. To limit the number of GPUs you want to make available
+you can change the ``count`` value to a positive integer or you can specify the ``device_ids`` key instead. ``device_ids`` takes a list
```
fmigneault, line 847: `count=-1`: That was my impression even before reading this part. 

Requesting all available GPUs would essentially lock-out any second user trying to use a kernel.

The example should probably use `count=1` instead and make a stronger warning about this situation, to let server maintainers know to not raise the value too much unless they got some really big GPU cluster.

fmigneault (commented):
Nice addition. Looking forward to advanced applications of it.

birdhouse/components/README.rst
```diff
@@ -821,6 +821,42 @@ Usage
...
+
+.. code-block:: shell
+
+    export JUPYTERHUB_CONFIG_OVERRIDE='
+    # enable GPU support
+    import docker
+
+    c.DockerSpawner.extra_host_config["device_requests"] = [
+        docker.types.DeviceRequest(count=-1, capabilities=[["gpu"]])
+    ]
```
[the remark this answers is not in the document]
mishaschwartz: > Does that mean that every single user notebook automatically gets assigned a GPU?

No, it means that every container has access to _all_ GPUs on the host. This PR doesn't introduce any solutions for allocating different GPU resources to different users. That is a much more complex thing that I'll have to try to figure out at a later date (because I don't really understand it yet).

> Is there a way to have both GPU/CPU-only simultaneously (adding more DeviceRequest variants to the list?), and have it selected somehow by the user when starting the kernel?
> Maybe even more specific GPU definitions, like providing the ones with just 8GB VRAM vs others with 48GB separately?

I think so, it would require a pretty good understanding of the nvidia toolkit and docker settings. I'm still reading about it and I can continue to update these docs as we figure different possible configurations.

birdhouse/components/README.rst
mishaschwartz: Let me clarify... this gives each container access to all GPUs on the host and they **share** them as a resource. The same way they're sharing access to the CPUs, memory, etc.
This will not stop a user from starting up their container.

Note that if user A proceeds to max out some of the GPUs then user B can't use them until they are done but managing that goes beyond the scope of this documentation so far.

birdhouse/components/README.rst
mishaschwartz: Yup they will definitely step on each others feet. Just the same as if a user hogs any other resource (CPU, memory, etc.)

We definitely need a better way to manage resource over-use  but the problem isn't specific to GPUs.

birdhouse/components/README.rst
fmigneault: Unless you have big GPUs like A100 to actually provide virtual-GPU VRAM, it won't take that much to cause all users to crash their respective processes from OutOfMemoryError. 

I'm nowhere near an expert on the matter, but I know that our clusters leverage some vGPUs to allow some kind of splitting this way. I don't know if that would play nice with multiple dockers trying to access the same GPU. Doesn't it do some kind of lock/reservation when assigned to a particular kernel?

birdhouse/components/README.rst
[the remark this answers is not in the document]
fmigneault: I wonder if something like https://jupyterhub-dockerspawner.readthedocs.io/en/latest/api/index.html#dockerspawner.SwarmSpawner.group_overrides could be used to dynamically apply the GPU request on specific users/conditions, therefore allowing having GPU or CPU-only setup.

Documentation is very sparse, so definitely very hard to figure out 😅 
All promising features nonetheless. Thanks for looking into them.

birdhouse/components/README.rst
mishaschwartz: I don't think that you need a vGPU setup for docker. Since there is no true VM (no hypervisor) with docker the containers can just access the GPU directly.

One thing you could do is use MIG (Multi-Instance GPU) or MPS (Multi-Process Service) to split up resources but not all GPUs support these (none of ours do unfortunately).

> Doesn't it do some kind of lock/reservation when assigned to a particular kernel?

As far as I know it manages context-switching the same way a CPU would that was running multiple processes/threads. So everything gets slowed down and/or on-GPU memory gets filled up but it won't necessarily just immediately fail from a user perspective.

Issue #593, opened 2025-10-02
:bug: [BUG]: Nginx/Canarie logging parsing bottleneck


fmigneault:
## Summary

There is an issue with the logging/parsing strategy of nginx access logs.

## Details

Since the conversion of CanarieAPI with separate cron/nginx/api definitions, the `cron` command that was invoked (https://github.com/Ouranosinc/CanarieAPI/blob/39c32e7ee2b21b5a3a6ba671ab41aa6195fde5b7/Dockerfile#L39) that led to some log rotation to occur (since `logrotate` config is available by default, but not `cron` or `logrotate` utilities, see below), is not being done anymore (current `proxy` only calls `nginx -g daemon off;`).

```bash
> docker exec -ti proxy bash 
> service --status-all
 [ ? ]  hwclock.sh
 [ + ]  nginx
 [ + ]  nginx-debug
> cat /etc/logrotate.d/nginx 
/var/log/nginx/*.log {
        daily
        missingok
        rotate 52
        compress
        delaycompress
        notifempty
        create 640 nginx adm
        sharedscripts
        postrotate
                if [ -f /var/run/nginx.pid ]; then
                        kill -USR1 `cat /var/run/nginx.pid`
                fi
        endscript
}
```

This leads to, eventually, very large logs to be collected.
```shell
3.9G Oct 2 06:26 access_file.log
```

This was not "that much a big deal" before.

However, since the fix of the log parser (https://github.com/Ouranosinc/CanarieAPI/pull/53) and its integration (https://github.com/bird-house/birdhouse-deploy/pull/588), the operation ran by CanarieAPI actually runs rather than preemptively failing. With this big file, it then proceeds to parse the log, which can take ~4min to parse based on the above size:
```log
[2025-10-02 06:24:44,766] [13] [INFO] canarieapi.app_object : Loading log file : /var/logs/access_file.log
[2025-10-02 06:28:43,812] [13] [INFO] canarieapi.app_object : Compiling stats from 27483724 records
```

Provided that this command is launched every minute (https://github.com/Ouranosinc/CanarieAPI/blob/master/docker/canarieapi-cron), this leads to an ever-growing amount of duplicate tasks in parallel.

I have received memory monitoring alarms from our server, and found that dozens of cron jobs were hogging the memory and bringing the server to a halt. 

### Potential Strategies to Fix

>[!NOTE]
> I can think of a few strategies to fix the problem.
> Maybe others are possible, suggestions/concerns welcomed.

[...]

PR #597, opened 2025-10-17
Autodeploy unable to git pull the newer `env.local` file.

tlvu:
## Overview

- Autodeploy unable to git pull the newer `env.local` file.

  Previously the `env.local` file could be a symlink to a file anywhere on the
  disk.

  Now `env.local` can only be a symlink to a file under a folder listed in
  `BIRDHOUSE_AUTODEPLOY_EXTRA_REPOS`.

  This is a new restriction from the newer autodeploy docker image in the
  previous release `2.18.5` below.  We can not volume-mount both the parent dir
  and the file at the same time and have write access.  We can only
  volume-mount one of the two.  For read-only access, it is still okay to
  volume-mount both.

## Changes

**Non-breaking changes**
- Fix autodeploy problem from using newer official Docker image

**Breaking changes**
- None

## Related Issue / Discussion

- Problem was introduced in the previous PR https://github.com/bird-house/birdhouse-deploy/pull/595

## CI Operations

birdhouse_skip_ci: false


tlvu:
@mishaschwartz feel free to comment again on this PR and if anything we can open a follow up PR to address your concerns.

However I will have to merge this PR now because this is blocking Ouranos from catching up with tip of this repo in production.

tlvu:
>  * why do we need read access to the env.local file? What needs to be able to write to it?

Autodeploy is a 2 stages process.

First stage is to check whether any repos need to be updated.  This needs to read `env.local` for `BIRDHOUSE_AUTODEPLOY_EXTRA_REPOS` in order to have this list of repos to check.

Second stage is actually performing the autodeploy (stop the full stack, git pull all the repos, restart the full stack).

This second stage is skipped if no repo needs update.  If this stage is actually triggered, to stop/start it uses `birdhouse-compose.sh` which needs `env.local` as you are all familiar with and to git pull the newer version of `env.local`, autodeploy will need write access to `env.local`.

> * why not just volume mount `BIRDHOUSE_LOCAL_ENV_REAL_PATH` to the location of `BIRDHOUSE_LOCAL_ENV` in the container.

We never had the need because when volume-mounting a symlink, it actually follow the link and volume-mount the real path already.  The not writable error I got is actually for  the real path.

> * why did you remove the volume mount of `BIRDHOUSE_LOCAL_ENV`? Won't that just mean that the container can't find the local env file at all?

Because the repo that contain `env.local` is already volume-mount entirely so autodeploy can git pull it.

Recall this is the recommanded layout if `env.local` is sourced-controlled:

https://github.com/bird-house/birdhouse-deploy/blob/3e007821e9fd013334ab47e90e4567e5cdc7c041/birdhouse/README.rst?plain=1#L118-L131

And we suggest in `env.local.example` to add that repo containing `env.local` to `BIRDHOUSE_AUTODEPLOY_EXTRA_REPOS`:

https://github.com/bird-house/birdhouse-deploy/blob/3e007821e9fd013334ab47e90e4567e5cdc7c041/birdhouse/env.local.example#L142

And in the autodeploy scheduler job we volume-mount all repos in the var `BIRDHOUSE_AUTODEPLOY_EXTRA_REPOS`:

https://github.com/bird-house/birdhouse-deploy/blob/3e007821e9fd013334ab47e90e4567e5cdc7c041/birdhouse/optional-components/scheduler-job-autodeploy/default.env#L10

https://github.com/bird-house/birdhouse-deploy/blob/3e007821e9fd013334ab47e90e4567e5cdc7c041/birdhouse/optional-components/scheduler-job-autodeploy/config.yml.template#L1-L6

So basically we were double volume-mounting the folder containing `env.local` and `env.local` itself.  With the older version of docker-cli it looks like less strict and let this slide.  Now the newer version, if you double mount, you can have read-only access but not write access !  So in order to have write access, I need to remove one of the two volume-mount !

> Unless I'm missing something I've very confused about this fix.

Yes it took me a while to get this too because I could not understand why this suddenly do not work.  I rebooted my VM twice thinking something got stuck on my VM.  Then I had to reproduce on another VM to be sure it's not my first VM that suddenly gone mad.

tlvu:
> > Still using birdhouse-compose.sh for day to day. But then how do you persist this?
> 
> I really recommend you start using bin/birdhouse.

Eventually yes.  I have many external scripts and components that uses `birdhouse-compose.sh` so they all have to transition before I can switch.

That's pretty low on my list.  Right now I want to go to prod with tip of BH and there is https://github.com/bird-house/birdhouse-deploy/issues/593 and https://github.com/bird-house/birdhouse-deploy/pull/598 blocking me.  Then after that I want all the new metrics and longterm metrics storage to work.

> > But then how do you persist this?
> 
> Well there are a few usual strategies for this...
> 
>     1. you put your local env in a default location
> 
>     2. your automation script calls the command line option with the location of the local env
> 
>     3. you put a symlink to the file in a default location (your current strategy)
> 
> 
> Note that the scheduler job is essentially doing strategy 2. It notes the location of the local env file when the stack is first started up and saves that location (by writing a template file) for the scheduler job to use.

The scheduler job can save the location of the local env file when the stack is first started up because it has a job to save to !  Not all scripts have this availability (no jobs associated) or does this explicitly in their job.  So option 2 is not a 100% foolproof, option 3 is.  Once the symlink is there, it is completely transparent to any scripts or jobs authors.

> > I thought allowing env.local path to be set by env var BIRDHOUSE_LOCAL_ENV was just for testing and you want to keep this ability hidden? You said it yourself here [#304 (comment)](https://github.com/bird-house/birdhouse-deploy/pull/304#discussion_r1163)
> 
> I think a lot of your issues could be solved if you used this environment variable and put it in your `.profile` or `.bashrc` file so you wouldn't have to worry about symlinking.

Setting `BIRDHOUSE_LOCAL_ENV` in `.profile` or `.bashrc` is another option for persistence but that means one single checkout of birdhouse-deploy on a machine.  The symlink trick is more flexible, each checkout can have their symlink.  This could be an option if user want to avoid symlink, which I do not see why for the moment.

So I guess we can use `bin/birdhouse --env-file` to quickly test out multiples variants of `env.local` but the "default" `env.local` should have the symlink point to it to facilitate usage by other scripts/components, etc.  Is this a reasonable conclusion?

tlvu:
> But, the change you made in this PR breaks things for others that set things up differently from you.

Yes I will for sure try what you suggest.  If being able to support both methods (the symlink and `BIRDHOUSE_LOCAL_ENV` override) is the best scenario.

What I am trying to say is `BIRDHOUSE_LOCAL_ENV` override is not the most transparent solution unless there is somehow a reason why the symlink solution can not be used.  Currently overriding `BIRDHOUSE_LOCAL_ENV`  is used during testing scenarios and that's fine.  For production use, the symlink solution should be used as it is more transparent and compatible with a wider range of external scripts and jobs.

tlvu:
> dictated to all users

Woo, if I was dictating my choice, I would have remove that variable !  I think you are mis-understanding here.

I am simply saying between the 2 approaches, the symlink one is more transparent and compatible with wider range of external scripts and jobs.

PR #599, opened 2025-10-20
allow custom cors headers for stac

mishaschwartz:
## Overview

- Allow each service to specify values for `Access-Control-Allow-Origin`

  Previously, if a `location` block in the `nginx` configuration for a given service included the cors helper
  configuration (with `include /etc/nginx/conf.d/cors.include;`) then all origins were allowed by default.

  This was done by setting the header `Access-Control-Allow-Origin: *` which works well but is a bit too permissive
  since it allowed __all__ origins.

  This change introduces a mechanism to specify specific additional allowed origins by setting the 
  `$access_control_allow_origin` nginx variable in the `location` block before including the `cors.include` file.

  For example:

  ```
  set $access_control_allow_origin http://example.com;
  include /etc/nginx/conf.d/cors.include;
  ```

  will set the value of the `Access-Control-Allow-Origin` response header to `http://example.com`.

  By default, the header value will be `*` if `$access_control_allow_origin` is not set (to maintain backwards
  compatibility).

  To specify multiple allowed origins, use a `map` directive (see the implementation for `components/stac` for an
  example).

- Set allowed CORS origins for `stac` through an environment variable

  This change implements this flexibility for the `components/stac` component. By setting the `STAC_CORS_ORIGINS`
  variable a user can specify allowed origins for responses from the `components/stac` component.

  For example, setting the following:
  
  ```
  export STAC_CORS_ORIGINS='https://example.com ~^https?://(www\.)?other\.example\.com$' 
  ```

  then requests from https://example.com and http://other.example.com will get a response with the 
  `Access-Control-Allow-Origin header` set to their origin, but http://example.ca will not.

  Note that this breaks backwards compatibility slightly since previously all origins were allowed for `/stac` by
  default. To keep the backwards compatible behaviour you can set:

  ```
  export STAC_CORS_ORIGINS='~.*' 
  ```

  to match all origins.

## Changes

**Non-breaking changes**
- Adds mechanism to allow services to have more control over CORS headers

**Breaking changes**
- responses from `/stac` no longer set `Access-Control-Allow-Origin: *` by default

## Related Issue / Discussion

- As discussed in #450

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/stac/config/proxy/conf.extra-service.d/stac.conf.template
```diff
@@ -9,7 +9,8 @@
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Host $host:$server_port;
         proxy_buffering off;
-        include /etc/nginx/conf.d/cors.include;
+        proxy_hide_header 'Access-Control-Allow-Origin';
+        add_header Access-Control-Allow-Origin $stac_origin_allowed;
```
fmigneault: I think the `cors.include` should still be there since it adds other definitions (allowed headers, allowed methods, etc.) which are also required.

I think `proxy_hide_header 'Access-Control-Allow-Origin';` will remove it.
Does `add_header` automatically reapply? I was under the impression explicit `proxy_pass_header` was needed to undo the hide. 

On top of this header, an `add_header Vary Origin;` would be required to advertise (as required by the standard) that it is Origin-dependant.

birdhouse/components/stac/default.env
```diff
@@ -45,6 +45,8 @@ export STAC_POPULATOR_BACKUP_IMAGE='${STAC_POPULATOR_BACKUP_DOCKER}:${STAC_POPUL
 # This must match the "stac_version" value in the current STAC catalog.
 export PYSTAC_STAC_VERSION_OVERRIDE=1.0.0
 
+export _STAC_ADDITIONAL_CORS_ORIGINS='$(for origin in $STAC_ADDITIONAL_CORS_ORIGINS; do echo "$origin \$http_origin;"; done;)'
```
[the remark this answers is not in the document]
mishaschwartz: It can either be a regex (if it starts with `~`) or just a string to match. Theoretically we could use either one.

birdhouse/components/stac/default.env
```diff
@@ -71,6 +73,7 @@ export DELAYED_EVAL="
   STAC_MIGRATION_DOCKER
   STAC_MIGRATION_IMAGE
   STAC_POPULATOR_BACKUP_IMAGE
+  _STAC_ADDITIONAL_CORS_ORIGINS
```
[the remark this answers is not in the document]
mishaschwartz: Sure, I like `_` for private variables like in python but you're right that we're not doing it consistently so we can leave it out.

birdhouse/components/stac-browser/default.env
```diff
@@ -24,6 +24,8 @@ export STAC_BROWSER_IMAGE_URI='${STAC_BROWSER_IMAGE}'
 #   - https://github.com/radiantearth/stac-browser/blob/main/docs/docker.md
 export STAC_BROWSER_PATH_PREFIX='/stac-browser/'
 
+export STAC_ADDITIONAL_CORS_ORIGINS='https://geojson.io'
```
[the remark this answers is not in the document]
mishaschwartz: Yeah I'll need to update comments and documentation explaining the format of everything.

> I think it might also need the ${BIRDHOUSE_FQDN_PUBLIC} since stac-browser self-pings the stac API.

`${BIRDHOUSE_FQDN_PUBLIC}` is always allowed because it is the default value for the nginx map. So by default, requests from all domains (including ${BIRDHOUSE_FQDN_PUBLIC}) will allow it.

> Should this parameter be in stac/default.env, or this one should update the variable additively?

It doesn't need to be. The default (withouth stac-browser) is that it is empty/unset.

CHANGES.md
```diff
@@ -15,7 +15,56 @@
...
+  example).
+
+- Set allowed CORS origins for `stac` through an environment variable
+
+  This change implements this flexibility for the `components/stac` component. By setting the `STAC_CORS_ORIGINS`
+  variable a user can specify allowed origins for responses from the `components/stac` component.
+
+  For example, setting the following:
+  
+  ```
```
fmigneault: add `shell`

CHANGES.md
```diff
@@ -15,7 +15,56 @@
...
+  `Access-Control-Allow-Origin header` set to their origin, but http://example.ca will not.
+
+  Note that this breaks backwards compatibility slightly since previously all origins were allowed for `/stac` by
+  default. To keep the backwards compatible behaviour you can set:
+
+  ```
+  export STAC_CORS_ORIGINS='~.*' 
+  ```
+
+  to match all origins.
```
fmigneault: Can reword the sentence `To match all origins and keep backwards compatible behaviour you can set:` to avoid sliting it.

fmigneault (commented):
Mostly editing stuff to review. Feature wise looks good.

birdhouse/components/README.rst
```diff
@@ -698,31 +698,60 @@ information.
...
 
     pip install pystac-client
-    stac-client search $PAVIS_FQDN/stac -q "variable_id=txgt_32" "scenario=ssp585"
+    stac-client search $BIRDHOUSE_FQDN_PUBLIC/stac -q "variable_id=txgt_32" "scenario=ssp585"
 
 Calls to the STAC API pass through Twitcher in order to validate authorization. Unauthenticated users will have 
-read-only access by default to STAC API resources while members of the `stac-admin` group can create and modify 
-resources. STAC Browser is not protected by any authorization mechanism.
+read-only access to STAC API resources while members of the `stac-admin` group can create and modify 
+resources if the ``optional-components/stac-public-access`` component is enabled.
```
fmigneault: The wording makes it seem like the optional component is needed for `stac-admin` to create/modify resources, but they always have access to do it. The component mention should be placed after the "Unauthenticated users will have read-only access" part.

birdhouse/components/stac/config/proxy/conf.extra-directives.d/stac.conf.template
```diff
@@ -0,0 +1,7 @@
+    map $http_origin $stac_origin_allowed {
```
tlvu: Remind me what the Nginx `map` directive do again?

This block means if the `http_origin` variable exists, create `stac_origin_allowed` variable with value `${BIRDHOUSE_PROXY_SCHEME}://${BIRDHOUSE_FQDN_PUBLIC}; ${STAC_ADDITIONAL_CORS_ORIGINS}` ?

birdhouse/components/stac/config/proxy/conf.extra-service.d/stac.conf.template
```diff
@@ -9,7 +9,9 @@
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Host $host:$server_port;
         proxy_buffering off;
+        set $access_control_allow_origin $stac_origin_allowed;
         include /etc/nginx/conf.d/cors.include;
+        add_header Vary Origin;
```
tlvu: What does the Vary header do again?  Why we need to add it here?

tlvu (changes requested):
I am not very familiar with all this CORS headers stuff although I had to deal with it in the past, but I forgot the details.

So I trust you guys with all this.  Just please ensure the behavior is backward-compatible with other components already using `proxy/conf.d/cors.include`.

As for how this CORS header should behave for `stac`, I'll follow your lead.

I have a few very minor change requested.  The rest looks good.

birdhouse/components/README.rst
mishaschwartz: It is needed. The magpie permissions set by `components/stac` give write permissions to the administrators group. The stac-admin group is created by the `optional-components/stac-public-access` component.

birdhouse/components/README.rst
```diff
@@ -698,31 +698,60 @@ information.
...
 
 .. code-block:: shell
 
     pip install pystac-client
-    stac-client search $PAVIS_FQDN/stac -q "variable_id=txgt_32" "scenario=ssp585"
+    stac-client search $BIRDHOUSE_FQDN_PUBLIC/stac -q "variable_id=txgt_32" "scenario=ssp585"
 
 Calls to the STAC API pass through Twitcher in order to validate authorization. Unauthenticated users will have 
-read-only access by default to STAC API resources while members of the `stac-admin` group can create and modify 
-resources. STAC Browser is not protected by any authorization mechanism.
```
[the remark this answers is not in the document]
mishaschwartz: It's technically true but it's misleading. Access to the stac-browser endpoint is not checked by twitcher but if the user doesn't have permissions to read the stac data (because the stac endpoint **is** protected by twitcher) then the stac-browser page shows a 403 error anyway.

So from the user's perspective it is protected by an authorization mechanism, it's the same one that protects the stac endpoint.

birdhouse/components/stac/config/proxy/conf.extra-service.d/stac.conf.template
fmigneault: The thing though is that `Vary` should be added only if the `Origin` (or any other header) actually affects the response (assuming no other header already distinguishes the response like the `Content-Type`). Adding it systematically (eg: when `Access-Control-Allow-Origin: *`) could result in duplicate cache entries for each domain accessing the server although nothing changes between them. It is used purposely to generate a "diff" in the cached response according to its domain-specific value, so that each domain is managed separately. I think we don't have a choice to apply it on case-by-case basis.

tlvu: > Adding it systematically (eg: when `Access-Control-Allow-Origin: *`)

I was suggesting to add `Vary` only when `Access-Control-Allow-Origin` is **not** `*`.  Hopefully the if-statement in Nginx allows this.  Otherwise we should put a big reminder somewhere so other users notice it.

fmigneault (approved):
Looks good. There is one comment remaining about the `stac-admin` magpie group, but that can be done in another PR if desired.

birdhouse/components/stac/config/proxy/conf.extra-service.d/stac.conf.template
tlvu: > `add_header Vary "";` the empty value won't cause any side-effect and is a legit value allowed by HTTP?

I just re-read Francis response earlier (https://github.com/bird-house/birdhouse-deploy/pull/599#discussion_r2482567612) ? "nginx omits header if empty".  Could you add this comment to the change so it is clear for other readers too.

mishaschwartz: For sure! Added in https://github.com/bird-house/birdhouse-deploy/pull/599/commits/2aa0caed5087250ed7682c9cc5c3b9cc46573dc7

PR #604, opened 2025-11-12
Fix `bin/birdhouse configs --default` flag usage description

mishaschwartz:
## Overview

Previously the description made it seem like the local environment file would __not__ be read if this option was specified. In fact, it's the components that won't be read.

Also rename the `--default` flag to `--basic` to reflect the function it calls (`read_basic_configs_only`) but keep the old version for backwards compatibility.

## Changes

**Non-breaking changes**
- fix usage message
- deprecate --default flag in favour of --basic

**Breaking changes**

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/bin/birdhouse
+++ b/bin/birdhouse
@@ -260,7 +260,7 @@ parse_configs_args() {
       shift
       parse_configs_args --basic "$@"
       ;;
-    -b|--basic|-d|--default)
+    -b|--basic)
       READ_CONFIGS_CMD=read_basic_configs_only
       shift
       parse_configs_args "$@"


bin/birdhouse
```diff
@@ -42,19 +42,20 @@ Options:
...
 Options:
   -h, --help                   Print this message and exit
-  -d, --default                Only load/print a command for the default configuration settings, not those specified by the local environment file
+  -b, --basic                  Only load/print a command for the basic configuration settings, not those specified by additional components
   -p, --print-config-command   Print a command that can be used to load configuration settings as environment variables
   -c, --command string         Execute the given command after loading configuration settings
   --print-log-command          Print a command that can be used to load the 'log' function used by birdhouse
 Deprecated Options:
   -q, --quiet                  Suppress stdout when loading configuration settings for the '--command' option. [DEPRECATED: use the --quiet option directly under birdhouse instead]
+  -d, --default                Same as the --basic flag. [DEPRECATED: use the --basic flag instead]
```
fmigneault: Should `log WARN` be done in the CLI if detected?

PR #605, opened 2025-11-14
Fix a few recent regressions to Autodeploy 

tlvu:
## Changes

- README: remind the user to source control `env.local` securely as it may contains passwords.

## Fixes

- Autodeploy broken due to new config variable precedence order.

  Broken since `2.18.8` (https://github.com/bird-house/birdhouse-deploy/pull/600).

  Example breakage scenario:

  Autodeploy starts and calls `read_configs`, then git pull a newer
  version of `env.local` (which might modify `EXTRA_CONF_DIRS` to enable
  new components for example).  But since the new var precedence locked
  all the values to the first call of `read_configs`, the new change to
  `EXTRA_CONF_DIRS` (or any vars in `env.local`) never went into
  effect at the end of the autodeploy process when we start the full
  stack.

  This is because `read_configs` returns all variables as environment
  variables and the new variable precedence order gives environment
  variables precedence higher than the values in `env.local`.  So after
  the first call to `read_configs` (or `read_basic_configs_only`), all the
  values are "locked" for all child processes even if they try to call
  `read_configs` again to get the newest values from `env.local`.

  The fix is to avoid getting these config variables as environment
  variables which will avoid "locking" their values for child processes.

- Autodeploy broken due to attempt to be compat with changing the location of `env.local` via `BIRDHOUSE_LOCAL_ENV`.

  Broken since `2.18.9` (https://github.com/bird-house/birdhouse-deploy/pull/601).

  It was working for the first run of autodeploy but on the next run, the value of
  `BIRDHOUSE_LOCAL_ENV` in `optional-components/scheduler-job-autodeploy/config.yml`
  is wrong !

  Good value in first run:

  `--volume /real/path/to/env.local:/tmp/birdhouse-local-env`

  Bad value in subsequent run:

  `--volume /tmp/birdhouse-local-env:/tmp/birdhouse-local-env`

  This is because we override `--env BIRDHOUSE_LOCAL_ENV=/tmp/birdhouse-local-env` in
  `optional-components/scheduler-job-autodeploy/config.yml.template` as env var so on
  the subsequent runs the bad value persists because it is an env var.

  This `BIRDHOUSE_LOCAL_ENV` is meant to be set "outside" the stack as env var
  so there is nothing in the stack that will fix this value if it is bad.

  The fix is to revert that attempt.  So autodeploy only supports changing the
  location of `env.local` via symlink, or by manually adding the folder
  containing the file `env.local` to `BIRDHOUSE_AUTODEPLOY_EXTRA_REPOS` if
  `env.local` is set by `BIRDHOUSE_LOCAL_ENV` environment variable and not via
  symlink.

- One feature of autodeploy is broken due to a missing mapping in back-compat config var name change.

  Broken since `2.4.0` but only if back-compat var name mode is required.

  The `AUTODEPLOY_CODE_OWNERSHIP` old name mapping is missing in `BIRDHOUSE_BACKWARDS_COMPATIBLE_VARIABLES`.

  Autodeploy is missing the `chown` step to the good code owner because it does
  not "see" the value of the old variable name since the mapping to the new
  variable name is missing.

- Missing default value for `BIRDHOUSE_HTTP_ONLY` causing inconsistent behavior
  when the var is used in `env.local` then removed from `env.local`.  The
  behavior never revert back as if it has never been initially used in
  `env.local`.  See [discussion](https://github.com/bird-house/birdhouse-deploy/pull/600#issuecomment-3528058708).

## CI Operations

birdhouse_skip_ci: false


birdhouse/optional-components/scheduler-job-autodeploy/config.yml.template
```diff
@@ -9,11 +9,10 @@
     --volume ${COMPOSE_DIR}/..:${COMPOSE_DIR}/..:rw
     --volume ${BIRDHOUSE_AUTODEPLOY_DEPLOY_KEY_ROOT_DIR}:${BIRDHOUSE_AUTODEPLOY_DEPLOY_KEY_ROOT_DIR}:ro
     --volume ${BIRDHOUSE_SSL_CERTIFICATE}:${BIRDHOUSE_SSL_CERTIFICATE}:ro
-    --volume ${BIRDHOUSE_LOCAL_ENV_REAL_PATH}:/tmp/birdhouse-local-env
```
mishaschwartz: What about:

```
--volume ${BIRDHOUSE_LOCAL_ENV_REAL_PATH}:${BIRDHOUSE_LOCAL_ENV_REAL_PATH}
--volume ${BIRDHOUSE_LOCAL_ENV}:${BIRDHOUSE_LOCAL_ENV}
--env BIRDHOUSE_LOCAL_ENV=${BIRDHOUSE_LOCAL_ENV}
```

mishaschwartz (approved):

birdhouse/optional-components/scheduler-job-autodeploy/config.yml.template
tlvu: This will likely not work because originally it had `--volume ${BIRDHOUSE_LOCAL_ENV_REAL_PATH}:${BIRDHOUSE_LOCAL_ENV_REAL_PATH}` and I had to remove it because we can not double volume-mounting the folder containing `env.local` and `env.local` itself and keep write-access (readonly-access is fine) with the newer docker image, see https://github.com/bird-house/birdhouse-deploy/pull/597#issuecomment-3426793138 

But I could give `--volume ${BIRDHOUSE_LOCAL_ENV}:${BIRDHOUSE_LOCAL_ENV}` a try, although I expect the same result since the symlink points to the same place anyways.

PR #609, opened 2025-11-24
add information about docker-rootless mode for first-time setup

fmigneault:
## Overview

Small documentation update about docker rootless to guide new users.

No need for version update.

## Changes

**Non-breaking changes**
- add docker rootless documentation details.

**Breaking changes**
- n/a

## CI Operations

birdhouse_skip_ci: true

### DIFF
--- a/birdhouse/README.rst
+++ b/birdhouse/README.rst
@@ -354,24 +354,31 @@ Docker rootless configuration
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 If you are using `Docker Rootless Mode <https://docs.docker.com/engine/security/rootless/>`_ on your machine,
-you *might* need to execute one of the following commands to allow the `proxy` (Nginx) service to connect to the relevant
+you *might* need to execute the following command to allow the `proxy` (Nginx) service to connect to the relevant
 HTTP ports.
 
 .. code-block:: shell
-    sudo setcap cap_net_bind_service=ep $(which rootlesskit)
-
-.. code-block:: shell
     sudo sysctl -w net.ipv4.ip_unprivileged_port_start=80
-
-.. warning::
-    Using the ``sysctl`` call grants access to any user able to interact with the machine's IP which could cause security concerns,
-    whereas the ``setcap`` call only grants access to the port by the Docker rootless executable. Developers are free to choose the
-    approach that best suits their use case.
 
 Notably, this could be required when encountering errors such as the following when invoking ``bin/birdhouse compose up -d``.
 
 .. code-block:: text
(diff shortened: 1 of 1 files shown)


birdhouse/README.rst
```diff
@@ -350,6 +350,21 @@ not able to access protected URLs:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you are using `Docker Rootless Mode <https://docs.docker.com/engine/security/rootless/>`_ on your machine,
+you might need to execute the following command to allow the `proxy` (Nginx) service to connect to the relevant
+HTTP ports.
+
+.. code-block:: shell
+    sudo sysctl -w net.ipv4.ip_unprivileged_port_start=80
+
+This could be required when encountering errors such as the following when invoking ``birdhouse compose up -d``.
```
mishaschwartz: I usually use `bin/birdhouse` in the documentation so that it's clear what we're talking about (in case the user hasn't added the interface script to their PATH)

birdhouse/README.rst
```diff
@@ -350,6 +350,21 @@ not able to access protected URLs:
...
 
+Docker rootless configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you are using `Docker Rootless Mode <https://docs.docker.com/engine/security/rootless/>`_ on your machine,
+you might need to execute the following command to allow the `proxy` (Nginx) service to connect to the relevant
+HTTP ports.
+
+.. code-block:: shell
+    sudo sysctl -w net.ipv4.ip_unprivileged_port_start=80
```
mishaschwartz: There are some security implications that we should mention here (all users will be able to access ports >=80 from now on).

We could also mention another alternative which is to use setcap:

```sh
sudo setcap cap_net_bind_service=ep $(which rootlesskit)
```

which only allows the rootlesskit executable to access privileged ports instead.

We could also mention that if users aren't comfortable making any of these privileged ports available then they could set up another reverse proxy in front of birdhouse or use iptables rules:

```
    sudo iptables -t nat -A PREROUTING -p tcp --dport 80 -j REDIRECT --to-port 8080
```

This would require allowing birdhouse to use a port other than 80 and 443 (which would require some additional code changes) so I don't know if it's worth mentioning that for now.

mishaschwartz (approved):
This is very good information. Some suggestions mostly to do with warning users about security implications:

birdhouse/README.rst
fmigneault: Good point. Will adjust.

fmigneault:
Merging as is for now since it is not blocking or affecting the stack's code.
The only working approach that I could validate on my end is put forward. Others are left for exploration with an indication that they are not expected to work straight away - and should be updated once evaluated.

PR #611, opened 2025-11-27
add `BIRDHOUSE_PROXY_CORS_ALLOW_ORIGIN` setting

### LINKED ISSUE #610: Issues with STAC CORS options
https://github.com/bird-house/birdhouse-deploy/blob/ee70e5683fb4c61efa80beb52c2535f716520a7f/birdhouse/components/stac/default.env#L56-L57

Since the addition of the above (and other corresponding code/files to use them in Nginx), the `proxy` service emits a lot of warnings:

```
proxy  | 2025/11/25 00:44:03 [warn] 20#20: *1509 using uninitialized "access_control_allow_origin" variable, client: 132.217.140.249, server: localhost, request: "GET /cowbird/ HTTP/1.1", host: "hirondelle.crim.ca"
proxy  | 2025/11/25 00:44:03 [warn] 22#22: *1519 using uninitialized "access_control_allow_origin" variable, client: 132.217.140.208, server: localhost, request: "GET /cowbird/ HTTP/1.1", host: "hirondelle.crim.ca"
proxy  | 2025/11/25 00:44:05 [warn] 22#22: *1548 using uninitialized "access_control_allow_origin" variable, client: 132.217.140.249, server: localhost, request: "GET /magpie/version HTTP/1.1", host: "hirondelle.crim.ca"
proxy  | 2025/11/25 00:44:05 [warn] 22#22: *1546 using uninitialized "access_control_allow_origin" variable, client: 132.217.140.208, server: localhost, request: "GET /magpie/version HTTP/1.1", host: "hirondelle.crim.ca"
```

These are triggered often as they are part of the service's healthchecks and a lot of permission checks.

I have tried setting `set $access_control_allow_origin '*';` explicitly (instead of conditionally as below), and that avoids the warnings. 
https://github.com/bird-house/birdhouse-deploy/blob/e2040fdbc18d292e5517d93d7635e464b3b02470/birdhouse/components/proxy/conf.d/cors.include#L2-L6

Should another variable be defined, such as `BIRDHOUSE_CORS_ORIGINS=*` by default, which would be set there explicitly and have to be overridden for servers trying to limit the origins?

fmigneault:
## Overview

Add `BIRDHOUSE_PROXY_CORS_ALLOW_ORIGIN` setting.

>[!NOTE]
> The `cors.include.template` diff looks massive, but the actual concrete change is only adding: 
> ```nginx
> set $access_control_allow_origin "${BIRDHOUSE_PROXY_CORS_ALLOW_ORIGIN}";
> ```

## Changes

**Non-breaking changes**
- Proxy: Allow `Access-Control-Allow-Origin` header to be configured using `BIRDHOUSE_PROXY_CORS_ALLOW_ORIGIN` variable.
  - The `cors.include` file is converted to a `cors.include.template` file to allow variable expansion.
  - The default `BIRDHOUSE_PROXY_CORS_ALLOW_ORIGIN="*"` is used to retain the previous behaviour.
  - The `BIRDHOUSE_PROXY_CORS_ALLOW_ORIGIN` variable can reference other variables to allow dynamic configuration
    (notably, to reference `BIRDHOUSE_FQDN_PUBLIC` for same-origin allowance).
  - Align the documentation with corresponding `STAC_CORS_ORIGINS` header implications.
  - Avoids Nginx warnings flagged from using uninitialized `access_control_allow_origin` variable.

**Breaking changes**
- n/a

## Related Issue / Discussion

- fixes #610 
- relates to #599

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/proxy/conf.d/cors.include.template
```diff
@@ -0,0 +1,42 @@
+    proxy_hide_header 'Access-Control-Allow-Origin';
+
+    set $access_control_allow_origin "${BIRDHOUSE_PROXY_CORS_ALLOW_ORIGIN}";
+    if ( $access_control_allow_origin ~ "^$" ) {
+       set $access_control_allow_origin '*';
+    }
```
tlvu: So this fix is required because the if-statement did not work?  

If the if-statement had worked, it would already set `$access_control_allow_origin` to `*` if that value is empty initially.  Of the uninitialised warning is also generated by the if-statement itself?

birdhouse/components/proxy/conf.d/cors.include.template
```diff
@@ -0,0 +1,42 @@
...
+
+    set $vary_origin ""; # nginx will omit this header value if set to the empty string
+    if ( $access_control_allow_origin != '*' ) {
+       set $vary_origin Origin;
+    }
```
tlvu: If problem with the if-statement to set default value, does this `$vary_origin` also generate uninitialised warning too?

tlvu (changes requested):
A few minor changes requested but looks good otherwise.

birdhouse/components/stac/default.env
```diff
@@ -53,6 +53,8 @@ export PYSTAC_STAC_VERSION_OVERRIDE=1.0.0
 # https://example.com and http://other.example.com will get a response with the Access-Control-Allow-Origin header set
 # to their origin, but http://example.ca will not.
 # By default all origins are allowed. Set STAC_CORS_ORIGINS in the local environment file to limit the allowed origins.
+# Default is aligned with BIRDHOUSE_PROXY_CORS_ALLOW_ORIGIN="*", any adjustment there should be reflected here too.
+# However, this is left as manual step to allow flexibility in case a different CORS policy is desired for STAC.
 export STAC_CORS_ORIGINS="~.*"
```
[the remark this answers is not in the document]
fmigneault: Added [`85e13f4` (#611)](https://github.com/bird-house/birdhouse-deploy/pull/611/commits/85e13f475ea6d0040797c47499fe4f516eaa1754).

birdhouse/components/proxy/default.env
```diff
@@ -6,6 +6,12 @@ export PROXY_IMAGE="nginx:1.23.4"
 # Any WPS processes taking longer than this should use async mode.
 export PROXY_READ_TIMEOUT_VALUE="240s"
 
+# CORS settings for the proxy
+# Set to "*" to allow all origins by default
+# If you want to restrict access, set this to the space-delimited allowed domain(s)
+# Can use other variables if needed (e.g.: '${BIRDHOUSE_PROXY_SCHEME}://${BIRDHOUSE_FQDN_PUBLIC}')
+export BIRDHOUSE_PROXY_CORS_ALLOW_ORIGIN="*"
```
mishaschwartz: One very small last thing:

Can we add a note here that reminds the user that if this is unset (or set to the empty string) then it will default to "*" (because of the if block in cors.include). 

That might be something that would surprise users if they haven't looked too closely at cors.include

tlvu: On the other hand, if the user explicitly override it to empty string in `env.local`, maybe he wants to **not** allow '\*' and if we still set to '\*' in `cors.include`, is that a good thing?!

Is there any scenarios where the user really want that value to be empty string and we should not block that?

fmigneault:
> By the way, I think you could have avoided all the diff noise if you `git mv cors.include cors.include.template; git commit`, then update the content in a separate commit.

It wasn't done in one step on my end. It was multiple commits spread out over a few days.

PR #617, opened 2025-12-11
Fix proxy lost log rotation, filling up disk and hogging CPU and RAM to parse large logs on restart

tlvu:
## Fixes

- Proxy lost log rotation, filling up disk and hogging CPU and RAM to parse large logs on restart

  Broken since `2.3.0` (https://github.com/bird-house/birdhouse-deploy/pull/452).

  The root cause is the log rotation has been removed from the python code in
  https://github.com/Ouranosinc/CanarieAPI/pull/18 without any replacement.

  Fixes https://github.com/bird-house/birdhouse-deploy/issues/593.

  The fix is to restore logrotation in the proxy container using `cron` and
  `logrotate` instead of the old python code.  Retention and frequency are the
  same as the previous python code.  For this we needed a custom build of the
  official Nginx docker image + `cron` + `logrotate`.

  This quick and least disruptive fix to get the production server out of the
  water should be a temporary solution until a better solution using container
  STDOUT parsing is implemented for the CanarieAPI and prometheus-log-parser
  (https://github.com/bird-house/birdhouse-deploy/issues/618).  Then we can
  switch back to the regular official Nginx image.

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,7 +34,7 @@
   This quick and least disruptive fix to get the production server out of the
   water should be a temporary solution until a better solution using container
   STDOUT parsing is implemented for the CanarieAPI and prometheus-log-parser
-  (https://github.com/bird-house/birdhouse-deploy/pull/617).  Then we can
+  (https://github.com/bird-house/birdhouse-deploy/issues/618).  Then we can
   switch back to the regular official Nginx image.
 
 


mishaschwartz (commented):
Thank you for this quick fix @tlvu!
I think there's a much simpler way to do this though, especially since this is meant as a temporary fix:

We can just add to the `scheduler-job-logrotate` component to handle the nginx logs as well. We already have the infrastructure in place to do log rotation, why not just add to that?

tlvu:
> We can just add to the `scheduler-job-logrotate` component to handle the nginx logs as well. We already have the infrastructure in place to do log rotation, why not just add to that?

I had thought of that but it does not work for 3 reasons:

1. `proxy` is a default component, going that route means it should now depend on `scheduler-job-logrotate` component, which is bad since default component should be standalone.
2. `logrotate` has to do a `kill -USR1 <Nginx pid>` to tell Nginx to re-open the new log file after the previous one has been rotated, so the logrotate should run in the same container as nginx to send the kill signal.
3. The cron frequency is much different between `scheduler-job-logrotate` and the one we want.  I'd rather not touch `scheduler-job-logrotate` now, preferring the least disruptive route to go fastest to production with least chance of regression.

mishaschwartz:
> I had thought of that but it does not work for 3 reasons

Ok but...

> proxy is a default component, going that route means it should now depend on scheduler-job-logrotate component, which is bad since default component should be standalone.

This is a temporary workaround and it's OK if we need to include a separate component for now. Also, when we eventually implement #618 it will be much easier to deprecate a single component instead of having to revert a bunch of code that has been added to the proxy component.

> logrotate has to do a kill -USR1 <Nginx pid> to tell Nginx to re-open the new log file after the previous one has been rotated, so the logrotate should run in the same container as nginx to send the kill signal.

Good point. This can be handled by the scheduler jobs with a [trigger](https://github.com/willfarrell/docker-crontab/tree/main?tab=readme-ov-file#config-file)

> The cron frequency is much different between scheduler-job-logrotate and the one we want. I'd rather not touch scheduler-job-logrotate now, preferring the least disruptive route to go fastest to production with least chance of regression.

You're right. This makes more sense as a separate scheduler job, not as an add on to the existing scheduler-jog-logrotate job.

tlvu:
Replaced by https://github.com/bird-house/birdhouse-deploy/pull/619

PR #620, opened 2025-12-18
Document development conventions for config vars

tlvu:
## Overview

Added development conventions for config variables in `env.local`, focusing on minimizing update friction and enhancing user experience.

## Related Issue / Discussion

- https://github.com/bird-house/birdhouse-deploy/pull/616#discussion_r2631306096

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/birdhouse/README.rst
+++ b/birdhouse/README.rst
@@ -407,6 +407,13 @@ In order to set up the development environment, run the following commands:
 Development Conventions
 ^^^^^^^^^^^^^^^^^^^^^^^
 
+The aim should be to **avoid** backward **incompatible** changes to decrease manual effort between each
+update. When the effort is low, users will be more likey to stay up to date and not "fear" updates.
+
+We should also aim to be **user-friendly** by being consistent with the way we grow the platform and by
+requiring as little setup/steps (sensible default values) as possible when spin up a fresh new stack
+for onboarding new users.
+
 Adding a new config var to ``env.local``
 ========================================
 * **Default value** should be in a corresponding ``default.env`` so it is easy for the user to find them.


birdhouse/README.rst
```diff
@@ -403,6 +403,79 @@ In order to set up the development environment, run the following commands:
...
+for onboarding new users.
+
+Adding a new config var to ``env.local``
+========================================
+* **Default value** should be in a corresponding ``default.env`` so it is easy for the user to find them.
+  The default value should not be burried in the code.  The commented out value in ``env.local.example``
+  is only an example and can not count as a default value.
+
+* **Documentation** for the new var can be both in ``default.env`` and ``env.local.example`` to be
+  most user-friendly. If you do not wish to duplicate the info because it is big, you can put in one of
```
mishaschwartz: I think we need to choose which one we prefer. I would say that we should always prefer `env.local.example` for variables that are use facing and `default.env` for the internal-only variables.

birdhouse/README.rst
```diff
@@ -403,6 +403,79 @@ In order to set up the development environment, run the following commands:
...
+
+  * adding to the ``BIRDHOUSE_BACKWARDS_COMPATIBLE_VARIABLES`` mapping in ``birdhouse/default.env`` or
+  * trying to support both names in the code at the same time for a few releases
+
+  to give time for all users to update all existing ``env.local`` on all existing deployments.
+
+* **Document migration path** clearly in ``CHANGES.md``.
+
+* Bump **minor** version, **not patch** version on release to signal **backward incompatible** change
+  requiring manual update to all existing ``env.local`` on all existing deployments.
```
mishaschwartz: Also mention that they should add an entry to `docs/source/migration_guide.rst`

birdhouse/README.rst
```diff
@@ -403,6 +403,79 @@ In order to set up the development environment, run the following commands:
...
+
+Changing the default value for the expected format of an existing config var in ``env.local``
+=============================================================================================
+* Try to **avoid** this scenario as this is **backward incompatible** with all existing ``env.local`` on 
+  all existing deployments.
+
+* Changing the format is potentially more disruptive than the default value.
```
mishaschwartz: Why? I'm not sure I understand

mishaschwartz (changes requested):

birdhouse/README.rst
```diff
@@ -403,6 +403,79 @@ In order to set up the development environment, run the following commands:
...
+* **Default value** should be in a corresponding ``default.env`` so it is easy for the user to find them.
+  The default value should not be burried in the code.  The commented out value in ``env.local.example``
+  is only an example and can not count as a default value.
+
+* **Documentation** for the new var can be both in ``default.env`` and ``env.local.example`` to be
+  most user-friendly. If you do not wish to duplicate the info because it is big, you can put in one of
+  the two files and reference in the other file.
+
+  * Documenting in ``env.local.example`` is the most user-friendly for new user starting out because they
+    will have to copy ``env.local.example`` to ``env.local``.
```
fmigneault: As per https://github.com/bird-house/birdhouse-deploy/pull/620/files#r2666114501, I think the overloaded `env.local.example` actually produces the opposite effect of being too noisy from unnecessary definitions, which makes it *less* user-friendly and more daunting for new users.

birdhouse/README.rst
```diff
@@ -403,6 +403,79 @@ In order to set up the development environment, run the following commands:
...
+
+Renaming or deleting an existing config var in ``env.local``
+============================================================
+* Try to **avoid** this scenario as this is **backward incompatible** with all existing ``env.local`` on
+  all existing deployments.
+
+* If you must, **soften the blow** by 
+
+  * adding to the ``BIRDHOUSE_BACKWARDS_COMPATIBLE_VARIABLES`` mapping in ``birdhouse/default.env`` or
+  * trying to support both names in the code at the same time for a few releases
```
fmigneault: I would rather have the convention of always using `BIRDHOUSE_BACKWARDS_COMPATIBLE_VARIABLES`, and never have both names in the code. The code should only use the "final / official" name established from that change and consider the deprecated/backward-compatible name as non-existent to make code maintenance easier.

birdhouse/README.rst
```diff
@@ -403,6 +403,79 @@ In order to set up the development environment, run the following commands:
     # Activate the conda environment
     conda activate birdhouse-dev
 
+
+Development Conventions
+^^^^^^^^^^^^^^^^^^^^^^^
```
fmigneault: Seems like all of this should be in https://github.com/bird-house/birdhouse-deploy/blob/master/CONTRIBUTING.rst instead. Cross-referencing to it here should be sufficient.

birdhouse/README.rst
```diff
@@ -403,6 +403,79 @@ In order to set up the development environment, run the following commands:
...
+
+The aim should be to **avoid** backward **incompatible** changes to decrease manual effort between each
+update. When the effort is low, users will be more likey to stay up to date and not "fear" updates.
+
+We should also aim to be **user-friendly** by being consistent with the way we grow the platform and by
+requiring as little setup/steps (sensible default values) as possible when spin up a fresh new stack
+for onboarding new users.
+
+Adding a new config var to ``env.local``
```
fmigneault: Use the long-form "configuration" and "variable"

Here and in all other places in the document.

birdhouse/README.rst
```diff
@@ -403,6 +403,79 @@ In order to set up the development environment, run the following commands:
...
+
+  to give time for all users to update all existing ``env.local`` on all existing deployments.
+
+* **Document migration path** clearly in ``CHANGES.md``.  If default value is changed, explain the impact
+  as user could previously be relying on the default value and not setting the variable explicitly in their
+  ``env.local``.  If they understand the impact and wish to retain the previous behavior, they now have to
+  explicitly set the previous default value in their ``env.local``.
+
+* Bump **minor** version, **not patch** version on release to signal **backward incompatible** change
+  requiring manual update to all existing ``env.local`` on all existing deployments.
```
fmigneault: cross-ref tagging policy

PR #625, opened 2025-12-22
Refactor Jupyterhub configuration files

### LINKED ISSUE #614: :question: [Question]: Can we clean up the jupyterhub_config.py code
## Questions

The code in `jupyterhub_config.py.template` is difficult to read and edit.  Issues are:

- some customization are template variables and others are environment variables (no clear reason why except maybe historical?)
- inconsistent modification of spawner and authenticator attributes. For example, some modify attributes for `c.Spawner` or `c.DockerSpawner` or `c.CustomSpawner` (all work but its confusing why we pick one over the other)
- custom authenticator is defined in https://github.com/Ouranosinc/jupyterhub but the custom Spawner is defined in `jupyterhub_config.py.template`
- no linting, no tests

In the past this might have worked but this file has gotten complex enough that it makes more sense to treat it like a proper python project

## Proposed Solution

Move the `CustomSpawner` implementation and most of the hardcoded settings to to https://github.com/Ouranosinc/jupyterhub.

Keep `jupyterhub_config.py.template` for customizations that can differ for different deployments as well as hardcoded values for the `c.JupyterHub` instance.

This will make this file easier to maintain and also put the code for both the `CustomSpawner` and `MagpieAuthenticator` in the same place so that they can be maintained together.

## Environment

| Server/Platform URL   | 
| Version Tag/Commit    |  
| Related issues/PR     | 
| Related components    |

mishaschwartz:
## Overview

Previously the jupyterhub configuration was defined in `components/jupyterhub/jupyterhub_config.py.template` except for the custom authenticator which was included in the `pavics/jupyterhub` docker image. This was fine, except that the configuration was split across two projects and as the configuration got more complex, maintaining these was getting more difficult.

This moves all configuration code including the authenticator to a python package named `jupyterhub_custom` located at `components/jupyterhub/jupyterhub_custom/`. It moves all configurations for the authenticator to the `magpie_authenticator.py` file and all the configurations for the spawner to the `custom_spawner.py` file. All variables that are settable by environment variables are moved to the `constants.py` file. This makes it much easier to see which variables are configurable using environment variables all in one place.

This change introduces unit tests and style policies (enforced by [ruff](https://docs.astral.sh/ruff/) and [pre-commit](https://pre-commit.com/)) for this package to encourage better code practices.

This change should be fully backwards compatible with the exception of the change to how settings for `deprecated-components/catalog` is handled (see below). However, it does deprecate some environment variables in favour of better configuration solutions:

  - using the `JUPYTERHUB_ENABLE_MULTI_NOTEBOOKS` variable to set the `DockerSpawner.allowed_images` variable is deprecated.
    - why?: this variable could be used to insert any python code into the middle of the `jupyterhub_config.py.template` file. This makes it too easy to accidentally insert code that breaks the configuration settings later on. We should avoid code insertion like this whenever possible.
    - new method: by default `DockerSpawner.allowed_images` will be set based on the values of `JUPYTERHUB_IMAGE_SELECTION_NAMES` and `JUPYTERHUB_DOCKER_NOTEBOOK_IMAGES`. If more than one image is specified by those variables then the user will be able to select which one they want. If you want to specify images other than those in the default those can be set using the `JUPYTERHUB_ALLOWED_IMAGES` which is a yaml or JSON mapping of image names to jupyterlab docker image tags.
  - using the `JUPYTERHUB_ADMIN_USERS` variable to set the `DockerSpawner.admin_users` variable is deprecated.
    - why?: this also executes arbitrary code (like above). Also, jupyterhub encourages assigning admin permissions
      based on [roles](https://jupyterhub.readthedocs.io/en/latest/tutorial/getting-started/authenticators-users-basics.html#configure-admins-admin-users), not by assigning them to the `admin_users` attribute. This makes it possible to easily revoke admin permissions as well and does not require us to restart Jupyterhub to do so.
    - new method: a new Magpie group is created. Its name is determined by the `JUPYTERHUB_ADMIN_GROUP_NAME` variable (default is "jupyterhub-admin"). Add users to this group in Magpie in order to give them admin permissions on Jupyterhub.
  - using lowercase constants in `JUPYTERHUB_CONFIG_OVERRIDE` is deprecated.
    - why?: [PEP8 recommends](https://peps.python.org/pep-0008/#constants) constants be written with capitals with underscores separating them.
    - new method: all the constants that were previously named with lowercase have an uppercase equivalent in `constants.py`. For example: `notebook_dir == constants.NOTEBOOK_DIR`. The uppercase version is preferred.

  Note: if your deployment is still using the `deprecated-components/catalog` component. You may want to manually set the following in `JUPYTERHUB_CONFIG_OVERRIDE` since the automatic addition of the the `CATALOG_USERNAME` to the `blocked_users` set is no longer part of the default settings (since the `deprecated-components/catalog` component has been deprecated for a long time):

  ```python
  c.MagpieAuthenticator.blocked_users.add("${CATALOG_USERNAME}")
  ```

## Changes

**Non-breaking changes**
- refactoring

**Breaking changes**
- if `deprecated-components/catalog` is still used, then it needs to be manually configured (see above)
- if `JUPYTERHUB_ADMIN_USERS` is **not** set in the local environment file, the default was to automatically add the magpie admin user as a jupyterhub admin. Now the default is to add no users as an admin automatically and users must be added by being included in the `JUPYTERHUB_ADMIN_GROUP_NAME` magpie group.

## Related Issue / Discussion

- Resolves #614 

## Additional Information

- This is based on #616 only because it includes changes from that PR so that we don't have to re-do them in the new format once #616 is merged. However, we should wait until #616 is merged before merging this one.
- Once this is merged we can remove all the magpie authenticator code from https://github.com/Ouranosinc/jupyterhub since this repo will now contain the most up to date version of that code. 

## CI Operations

birdhouse_skip_ci: false


CHANGES.md
```diff
@@ -34,6 +34,60 @@
...
+      in `constants.py`. For example: `notebook_dir == constants.NOTEBOOK_DIR`. The uppercase version is 
+      preferred.
+
+  Note: if your deployment is still using the `deprecated-components/catalog` component. You may want to manually set the
+  following in `JUPYTERHUB_CONFIG_OVERRIDE` since the automatic addition of the the `CATALOG_USERNAME` to the `blocked_users`
+  set is no longer part of the default settings (since the `deprecated-components/catalog` component has been deprecated for 
+  a long time):
+
+  ```python
+  c.MagpieAuthenticator.blocked_users.add("${CATALOG_USERNAME}")
```
fmigneault: Should it be with a `os.getenv("CATALOG_USERNAME")` or jupytherhub config auto-resolves it with the templating logic? Just making sure because of the embedding of within `JUPYTERHUB_CONFIG_OVERRIDE`.

.github/workflows/precommit.yml
```diff
@@ -0,0 +1,13 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v3
+    - uses: pre-commit/action@v3.0.1
```
fmigneault: I'm worried this will cause precommit to often push extra "last minute" fix commits which will explode the number of CI runs. Since each run triggers a >1h server setup/test, I would rather avoid it. 

Maybe following https://github.com/Ouranosinc/jenkins-master/pull/5, we can revisit and enable it by applying relevant plugins to attempt early-abort/retry concurrent logic, but currently it will cause bottlenecks and spam the PR with even more notebook results comments.

birdhouse/components/jupyterhub/jupyterhub_custom/jupyterhub_custom/custom_dockerspawner.py
```diff
@@ -0,0 +1,335 @@
...
+
+    @default("default_url")
+    def _default_default_url(self) -> int:
+        """Set the URL the single-user server should start in."""
+        return "/lab"
+
+    @default("debug")
+    def _default_debug(self) -> int:
+        """Debug log output."""
+        return True
```
fmigneault: int/true return

birdhouse/components/jupyterhub/jupyterhub_custom/jupyterhub_custom/custom_dockerspawner.py
```diff
@@ -0,0 +1,335 @@
...
+        return images
+
+    # The following values override defaults in the Spawner base class.
+    # See the Spawner documentation for more details:
+    # https://jupyterhub.readthedocs.io/en/latest/reference/api/spawner.html#jupyterhub.spawner.Spawner
+
+    @default("notebook_dir")
+    def _default_notebook_dir(self) -> int:
+        """Return notebook directory path."""
+        return constants.NOTEBOOK_DIR
```
fmigneault: int/str return

birdhouse/components/jupyterhub/jupyterhub_custom/jupyterhub_custom/custom_dockerspawner.py
```diff
@@ -0,0 +1,335 @@
...
+                [
+                    "--MappingKernelManager.cull_connected=True",
+                    "--TerminalManager.cull_connected=True",
+                ]
+            )
+
+        return args
+
+    @default("extra_host_config")
+    def _default_extra_host_config(self) -> dict:
```
fmigneault: `dict[str, Any]` or a more restricted type if applicable?

birdhouse/components/jupyterhub/jupyterhub_custom/jupyterhub_custom/custom_dockerspawner.py
```diff
@@ -0,0 +1,335 @@
...
+            # https://docs.bokeh.org/en/latest/docs/user_guide/jupyter.html#jupyterhub
+            # Issue https://github.com/bokeh/bokeh/issues/12090
+            # Post on Panel forum:
+            # https://discourse.holoviz.org/t/how-to-customize-the-display-url-from-panel-serve-for-use-behind-jupyterhub-with-jupyter-server-proxy/3571
+            # Issue about Panel Preview: https://github.com/holoviz/panel/issues/3440
+            "BIRDHOUSE_HOST_URL": constants.BIRDHOUSE_HOST_URL,
+            # https://docs.dask.org/en/stable/configuration.html
+            # https://jupyterhub-on-hadoop.readthedocs.io/en/latest/dask.html
+            "DASK_DISTRIBUTED__DASHBOARD__LINK": constants.BIRDHOUSE_HOST_URL
+            + "{JUPYTERHUB_SERVICE_PREFIX}proxy/{port}/status",
```
fmigneault: Easier to read if on same line or using `()` to wrap otherwise.

birdhouse/components/jupyterhub/jupyterhub_custom/jupyterhub_custom/magpie_authenticator.py
```diff
@@ -0,0 +1,181 @@
...
+        This allows us to manage users on Magpie without having to also update the
+        JupyterHub database manually to match Magpie permissions.
+
+        Note that users on the blocked_users list will still be denied even though
+        they can authenticate through Magpie.
+        """
+        return True
+
+    @default("blocked_users")
+    def _default_blocked_users(self) -> set:
```
fmigneault: `set[str]`

birdhouse/components/jupyterhub/default.env
```diff
@@ -101,13 +108,6 @@ OPTIONAL_VARS="
   \$JUPYTERHUB_VERSION
   \$JUPYTERHUB_IMAGE
   \$JUPYTERHUB_IMAGE_URI
-  \$JUPYTERHUB_AUTHENTICATOR_AUTHORIZATION_URL
-  \$JUPYTERHUB_AUTHENTICATOR_REFRESH_AGE
-  \$JUPYTER_IDLE_SERVER_CULL_TIMEOUT
-  \$JUPYTER_IDLE_KERNEL_CULL_TIMEOUT
-  \$JUPYTER_IDLE_KERNEL_CULL_INTERVAL
-  \$JUPYTERHUB_USER_DATA_DIR
-  \$JUPYTERHUB_RESOURCE_LIMITS
```
fmigneault: Why are they removed? They are still referenced in [jupyterhub/docker-compose-extra.yml](https://github.com/bird-house/birdhouse-deploy/pull/625/changes#diff-98a0b4dfd0e9257c5232c024301313c248622690b0840ad93337cd703c5bb83b)

birdhouse/components/jupyterhub/jupyterhub_config.py.template
```diff
@@ -1,337 +1,83 @@
...
-## Note that JUPYTERHUB_DOCKER_EXTRA_HOSTS may be set by default in the local-dev-test component
-c.DockerSpawner.extra_host_config = {
-    # start init pid 1 process to reap defunct processes
-    'init': True,
-    'extra_hosts': dict(host_mapping.split(":") for host_mapping in os.getenv("JUPYTERHUB_DOCKER_EXTRA_HOSTS", "").split())
+c.JupyterHub.init_spawners_timeout = 20  # Timeout (in seconds) to wait for spawners to initialize (default 10)
+c.JupyterHub.load_roles = [
+    {
+        "name": "jupyterhub-admin",
+        "description": "Give full admin (super-user) access",
```
fmigneault: Corresponding description could be provided in the Magpie definition, using the group pre-definition as shown in https://pavics-magpie.readthedocs.io/en/latest/configuration.html#combined-configuration-file

.github/workflows/precommit.yml
mishaschwartz: This workflow won't actually make any changes to the code. It just checks whether the ruff linter/formatter passes and will fail the test if not. It will not push any changed code if it fails and it doesn't try to fix the errors (that's up to the user).

birdhouse/components/jupyterhub/default.env
mishaschwartz: They aren't template variables anymore.

.github/workflows/precommit.yml
fmigneault: OK. Thanks for validating.

birdhouse/components/jupyterhub/default.env
mishaschwartz: Wait... it already is in VARS. I guess it accidentally got added to both VARS and OPTIONAL_VARS at some point.

birdhouse/components/jupyterhub/jupyterhub_custom/jupyterhub_custom/magpie_authenticator.py
```diff
@@ -0,0 +1,182 @@
+import json
```
tlvu: This is an exact copy of the same file from https://github.com/Ouranosinc/jupyterhub/blob/b3a786a5eb22b09d41f3093b619f02f7b33348a2/jupyterhub_magpie_authenticator/jupyterhub_magpie_authenticator.py or is there any changes?

birdhouse/components/jupyterhub/docker-compose-extra.yml
```diff
@@ -28,7 +28,19 @@ services:
       USER_WORKSPACE_GID: ${USER_WORKSPACE_GID}
       JUPYTERHUB_CRYPT_KEY: ${JUPYTERHUB_CRYPT_KEY}
       JUPYTERHUB_DOCKER_EXTRA_HOSTS: ${JUPYTERHUB_DOCKER_EXTRA_HOSTS:-}
+      JUPYTERHUB_AUTHENTICATOR_AUTHORIZATION_URL: ${JUPYTERHUB_AUTHENTICATOR_AUTHORIZATION_URL:-}
```
tlvu: Some of the new env vars here redefine a default value different than in `default.env`, is there a reason?  And some do not redefine default value here, is there a reason to break the consistency?

tlvu (commented):
Wow that's some massive refactoring !  Took me 2 passes before I grasp the extend of it.

I'll have to deploy this on my test server to see if the shuffling of the existing templating vars do not break our current setup.

One quick observation is that most of the previous configs being set directly in `jupyterhub_config.py.template` migrate to `custom_dockerspawner.py`.   But some are left behind, so I wonder the ones left behind are because they can not be moved simply because they do not belongs to the docker spawner or because of other reasons?

birdhouse/components/jupyterhub/config/magpie/config.yml.template
```diff
@@ -8,3 +8,15 @@ providers:
     c4i: false
     type: api
     sync_type: api
+
```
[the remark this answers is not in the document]
fmigneault: The magpie doc used to have `permissions.cfg` and `providers.cfg` separately (still allowed and used by birdhouse-deploy), but later introduced `config.yml` that allows defining everything together since some users/groups permissions are often relevant along the specific services (providers) that get defined.

YAML was used at that point to reflect the actual data structure. 

You can actually use many extensions interchangeably. Magpie doesn't care as long as it loads as JSON-like. https://pavics-magpie.readthedocs.io/en/latest/configuration.html#configuration-file-formats

birdhouse/components/jupyterhub/jupyterhub_custom/jupyterhub_custom/magpie_authenticator.py
mishaschwartz: It is the same with some additions. Once this PR has been approved we will remove the duplicate code from that repository.

birdhouse/components/jupyterhub/docker-compose-extra.yml
tlvu: > The default is still what is in `default.env` but docker compose will log a warning if an environment variable is missing in a docker-compose.yml file.

But how can this var be empty if there is a default in `default.env`?

> One way to tell docker compose that this variable is allowed to be empty is to use the `:-` syntax. That way it will not log the warning if this variable happens to be empty.

I would still keep the warnings to indicate possible error, unless the var can really be overridden to be empty.

I am guessing you got the warnings because you override the existing default value and explicitly set the new value to be empty?  Is that just for testing or the var can be empty in "normal workflow" circumstances?

If it can be empty in  "normal workflow" circumstances, then please add a comment saying that, otherwise this looks like a default value redefinition that is really confusing and not consistent with the rest of the code !

If it can not be empty in  "normal workflow" circumstances, please keep the warnings to indicate possible error.

birdhouse/components/jupyterhub/jupyterhub_config.py.template
```diff
@@ -1,337 +1,83 @@
...
+    exec("c.Authenticator.admin_users = ${JUPYTERHUB_ADMIN_USERS}")
 
 # ------------------------------------------------------------------------------
 # Configuration overrides
 # ------------------------------------------------------------------------------
+
+# This import is required to support code inserted by JUPYTERHUB_CONFIG_OVERRIDE
+# that may use the legacy (lowercase) version of the constants. This import can
+# be removed once we decide to drop support for these legacy constants.
+from jupyterhub_custom.constants import *  # noqa: F403
```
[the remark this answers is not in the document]
tlvu: Before you ask why `JUPYTERHUB_ENABLE_MULTI_NOTEBOOKS` is used this way, we have many additional customizations so I use `JUPYTERHUB_ENABLE_MULTI_NOTEBOOKS` the same way as `JUPYTERHUB_CONFIG_OVERRIDE`.

I treat them both as generic code injection points, not specific to any particular task.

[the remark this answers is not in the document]
mishaschwartz: @tlvu Ok good to know. This is what we were wondering about in https://github.com/bird-house/birdhouse-deploy/pull/625#discussion_r2699931874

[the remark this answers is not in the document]
tlvu: > > In my ${JUPYTERHUB_ENABLE_MULTI_NOTEBOOKS} I have:
> 
> Just out of curiosity, why is this here instead of in `JUPYTERHUB_CONFIG_OVERRIDE`?

Organic grow !  

CRIM added the initial `JUPYTERHUB_ENABLE_MULTI_NOTEBOOKS` because they have the EO and NLP images flavors.

I needed to inject Ouranos specific config so I was hijacking that var but eventually it feels wrong so I added `JUPYTERHUB_CONFIG_OVERRIDE` so it is clear to all future users they can inject any config into the jupyter config.

I did not remove `JUPYTERHUB_ENABLE_MULTI_NOTEBOOKS`  to maintain backward compat with CRIM.

fmigneault (changes requested):
Something seems to have broken between the magpie-authenticator / docker-spawner configuration.

birdhouse/env.local.example
```diff
@@ -462,10 +462,10 @@ export GEOSERVER_ADMIN_PASSWORD="${__DEFAULT__GEOSERVER_ADMIN_PASSWORD}"
 ## /notebook_dir/public-share/
 #public_read_in_container = join(notebook_dir, 'public-share')
 #
-#c.DockerSpawner.volumes[public_read_on_disk] = {
+#c.DockerSpawner.volumes.update({public_read_on_disk: {
```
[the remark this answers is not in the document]
tlvu: Confirm the back-compat fix is working for my existing config.

fmigneault:
@mishaschwartz 
Thanks for the fixes. The latest run has the following error about server creation. I'm not quite sure of the cause of failure.

```python
22:54:24  ______________ notebooks-auth/test_cowbird_jupyter.ipynb::Cell 8 _______________
22:54:24  Notebook cell execution failed
22:54:24  Cell 8: Cell outputs differ
22:54:24  
22:54:24  Input:
22:54:24  if VERIFY_SSL:
22:54:24      !jhubctl run --hub {JUPYTER_URL} --auth-type token -u {TEST_USER} -n resources/user_test_cowbird_jupyter.ipynb \
22:54:24        --output-filename notebook_results.txt --stop-server --validate
22:54:24  else:
22:54:24      !jhubctl run --hub {JUPYTER_URL} --auth-type token -u {TEST_USER} -n resources/user_test_cowbird_jupyter.ipynb \
22:54:24        --no-verify-ssl --output-filename notebook_results.txt --stop-server --validate
22:54:24  
22:54:24  Traceback:
22:54:24   mismatch 'stdout'
22:54:24  
22:54:24   assert reference_output == test_output failed:
22:54:24  
22:54:24    'INFO:jhub_cl...ME_USERNAME\n' == 'INFO:jhub_cl...ut=60 [s]\r\n'
22:54:24    
22:54:24    Skipping 88 identical leading characters in diff, use -v to show
22:54:24    + lient.api:created kernel_spec=python3 kernel=JUPYTERHUB_KERNEL_UUID for jupyter
22:54:24    + INFO:jhub_client.api:deleted kernel=JUPYTERHUB_KERNEL_UUID for jupyter
22:54:24    + INFO:jhub_client.api:deleted server for username=TEST_USERNAME_USERNAME
22:54:24    - lient.api:pending spawn polling for seconds=5 [s]
22:54:24    - INFO:jhub_client.api:pending spawn polling for seconds=10 [s]
22:54:24    - INFO:jhub_client.api:pending spawn polling for seconds=15 [s]
22:54:24    - INFO:jhub_client.api:pending spawn polling for seconds=20 [s]
22:54:24    - INFO:jhub_client.api:pending spawn polling for seconds=25 [s]
22:54:24    - INFO:jhub_client.api:pending spawn polling for seconds=30 [s]
22:54:24    - INFO:jhub_client.api:pending spawn polling for seconds=35 [s]
22:54:24    - INFO:jhub_client.api:pending spawn polling for seconds=40 [s]
22:54:24    - INFO:jhub_client.api:pending spawn polling for seconds=45 [s]
22:54:24    - INFO:jhub_client.api:pending spawn polling for seconds=50 [s]
22:54:24    - INFO:jhub_client.api:pending spawn polling for seconds=55 [s]
22:54:24    - ERROR:jhub_client.api:jupyterhub server creation timeout=60 [s]
22:54:24    - Traceback (most recent call last):
22:54:24    -   File "/opt/conda/envs/birdy/bin/jhubctl", line 8, in <module>
22:54:24    -     sys.exit(main())
22:54:24    -              ^^^^^^
22:54:24    -   File "PYTHON_SITE_PACKAGES/jhub_client/__main__.py", line 5, in main
22:54:24    -     cli()
22:54:24    -   File "PYTHON_SITE_PACKAGES/jhub_client/cli.py", line 27, in cli
22:54:24    -     args.func(args)
22:54:24    -   File "PYTHON_SITE_PACKAGES/jhub_client/cli.py", line 175, in handle_run
22:54:24    -     cell_results = loop.run_until_complete(execute_notebook(**kwargs))
22:54:24    -                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
22:54:24    -   File "/opt/conda/envs/birdy/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
22:54:24    -     return future.result()
22:54:24    -            ^^^^^^^^^^^^^^^
22:54:24    -   File "PYTHON_SITE_PACKAGES/jhub_client/execute.py", line 153, in execute_notebook
22:54:24    -     return await execute_code(hub_url, cells, **kwargs)
22:54:24    -            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
22:54:24    -   File "PYTHON_SITE_PACKAGES/jhub_client/execute.py", line 79, in execute_code
22:54:24    -     jupyter = await hub.ensure_server(
22:54:24    -               ^^^^^^^^^^^^^^^^^^^^^^^^
22:54:24    -   File "PYTHON_SITE_PACKAGES/jhub_client/api.py", line 122, in ensure_server
22:54:24    -     raise TimeoutError(
22:54:24    - TimeoutError: jupyterhub server creation timeout=60 [s]
```

PR #628, opened 2026-01-05
Fix bugs in `get-components-json.include.sh` and simplify `get-services-json.include.sh`

mishaschwartz:
## Overview

`get-components-json.include.sh` was unable to handle the case where a component was commented out in the `BIRDHOUSE_EXTRA_CONF_DIRS` variable. Also, it did not display custom components (i.e. those not in one of the `birhouse/*components/` directories).

This also simplifies the code in both files to make it easier to maintain:
  
  - removed unnecessary variable declarations
  - simplify component discovery `get-components-json.include.sh`

## Changes

**Non-breaking changes**
- bug fix

**Breaking changes**
- None

## Additional Information

Note that the changes here are fully backwards compatible. 
Even though we no longer source `read-configs.include.sh` the `read_configs` and/or `read_basic_configs` functions were never called so even if you had another script that called these scripts without first reading in the relevant configuration variables, it would already be failing anyway (since neither script would have loaded the correct variables).

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/birdhouse/scripts/get-components-json.include.sh
+++ b/birdhouse/scripts/get-components-json.include.sh
@@ -24,7 +24,7 @@ for adir in ${ALL_CONF_DIRS}; do
     # component is in one of the *components directories in COMPOSE_DIR (not custom)
     # custom components are not included in the /components endpoint
     component="bird-house/birdhouse-deploy:${real_adir#${COMPOSE_DIR%/}/}"
+    BIRDHOUSE_DEPLOY_COMPONENTS_LIST="${BIRDHOUSE_DEPLOY_COMPONENTS_LIST}\"${component}\","
   fi
-  BIRDHOUSE_DEPLOY_COMPONENTS_LIST="${BIRDHOUSE_DEPLOY_COMPONENTS_LIST}\"${component}\","
 done
 export BIRDHOUSE_DEPLOY_COMPONENTS_JSON="{\"components\": [${BIRDHOUSE_DEPLOY_COMPONENTS_LIST%,}]}"


CHANGES.md
```diff
@@ -15,7 +15,18 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Fixes
+
+- Fix bugs in `get-components-json.include.sh` and simplify `get-services-json.include.sh`
+
+  `get-components-json.include.sh` was unable to handle the case where a component was commented out in
+  the `BIRDHOUSE_EXTRA_CONF_DIRS` variable. Also, it did not display custom components (i.e. those not
```
fmigneault: Just to be sure, does that mean that the following would now behave by omitting magpie?
```env
export BIRDHOUSE_EXTRA_CONF_DIRS="
  # ./components/magpie
  ./components/twitcher
"

File: birdhouse/scripts/get-components-json.include.sh [hunk 2]
```diff
@@ -12,64 +12,22 @@
...
-  | sed '/^\n*$/d' \
-)
-BIRDHOUSE_DEPLOY_COMPONENTS_LIST="${BIRDHOUSE_DEPLOY_COMPONENTS_LIST%?}"  # remove last comma
-export BIRDHOUSE_DEPLOY_COMPONENTS_JSON="{\"components\": [${BIRDHOUSE_DEPLOY_COMPONENTS_LIST}]}"
+for adir in ${ALL_CONF_DIRS}; do
+  [ -d "${adir}" ] || continue
+  real_adir="$(readlink -f "${adir}" || realpath "${adir}")"
+  if [ "${real_adir}" = "${real_adir#${COMPOSE_DIR%/}/*components/}" ]; then
+    # component is not in one of the *components directories in COMPOSE_DIR (not custom)
+    component="custom:$(basename ${adir})"
```
fmigneault: Those were purposely omitted to avoid leaking information. One principal use of custom definitions is to provide per-instance security/credentials overrides. I do not want to expose such mappings publicly.

birdhouse/scripts/get-components-json.include.sh
```diff
@@ -12,64 +12,22 @@
...
-  echo "${BIRDHOUSE_DEPLOY_COMPONENTS_LIST_ACTIVE}" \
-  | grep ${BIRDHOUSE_DEPLOY_COMPONENTS_LIST_KNOWN} \
-  | sed -E 's|^\s*([A-Za-z0-9./_-]+)\s*$|"\1",|g' \
-  | sed -E "s|^\"\./(.*)\"|\"${BIRDHOUSE_DEPLOY_COMPONENTS_BASE}\\1\"|g" \
-  | sed '/^\n*$/d' \
-)
-BIRDHOUSE_DEPLOY_COMPONENTS_LIST="${BIRDHOUSE_DEPLOY_COMPONENTS_LIST%?}"  # remove last comma
-export BIRDHOUSE_DEPLOY_COMPONENTS_JSON="{\"components\": [${BIRDHOUSE_DEPLOY_COMPONENTS_LIST}]}"
+for adir in ${ALL_CONF_DIRS}; do
+  [ -d "${adir}" ] || continue
```
fmigneault: One important advantage of the previous method was to warn immediately of incorrectly specified components (eg: typo), and warn about unresolved ones (eg: when we switched `configs/...` to `components/...`) to help the migration process / highlight backward compatibility definitions.

Now, the missing dirs are simply ignored, which could make it hard to detect whether certain configs were applied or not, notably when involving custom overrides outside this repository.

I think the warnings should remain.

fmigneault (changes requested):

birdhouse/scripts/get-components-json.include.sh
mishaschwartz: The comment before said:

> other component locations are considered 'custom' and marked as such to provide contextual information

which makes it seem like they are meant to be included and marked as custom. Maybe that comment was left in after some changes to remove them?

I don't care either way and I can remove them for now.

PR #631, opened 2026-01-12
Remove `proxy` component's dependency on `scheduler` and `scheduler-job-logrotate-nginx`

### LINKED ISSUE #618: :bulb: [Feature] Implement Nginx log parsing from STDOUT stream for CanarieAPI and prometheus-log-parser
## Description

This should be considered part 2 of https://github.com/bird-house/birdhouse-deploy/issues/593.

A quick and least disruptive solution was implemented to get production out of the water.  This issue is so we can implement a better fix.

This fix is required so we can switch back to the regular official Nginx docker image, instead of our custom build with `cron` and `logrotate`.  See the quick fix for production https://github.com/bird-house/birdhouse-deploy/pull/617.

## Concerned Organizations

@fmigneault  @mishaschwartz

mishaschwartz:
## Overview

Creates new settings in `optional-components/proxy-log-volume` that create the `proxy-logs` docker volume as well as instructing Nginx to write access logs to an additional log file (specified by `PROXY_LOG_PATH`). These settings are included as a `COMPONENT_DEPENDENCY` in components that require access to the the `proxy` access logs as a regular file. If no components require access to these logs as a regular file then the `proxy` component will only write access logs to the stdout stream for that container.

Right now, the only components that require access to logs in this way are `components/canarie-api` and `optional-components/prometheus-log-parser`. Both of these now include `optional-components/proxy-log-volume` as a `COMPONENT_DEPENDENCY`.

Note: this means that if no optional components require `optional-components/proxy-log-volume` as a dependency then logs from the `proxy` container will only be written to stdout/stderr. This means that there is no need for any additional custom log rotation handling since the logs are handled directly by docker. This means that the `proxy` service itself no longer need to be dependant on the `scheduler` and `scheduler-job-logrotate-nginx` components.

Note: a previous discussion suggested that logs could be parsed directly from the stdout stream of the `proxy` container. However, there is no way to do so that doesn't require very hacky workarounds. Possible solutions that were explored and rejected include:

- Mounting the log file from the `proxy` container from the host to the relevant containers.
  Rejected because this practice is highly discouraged by docker as the actual storage location of log files is not standardized and may be changed in future versions.
- Writing logs to a named pipe or socket within the `proxy` container.
  Rejected because this is very difficult to set up and is untested when then mounted to other containers. 
  Also, a different named pipe would be required for each consumer which is currently very difficult to set up using birdhouse's deployment tools. 

**Breaking Change**: if a custom component (not included in this repository) uses the `proxy-logs` named volume. It must now include `optional-components/proxy-log-volume` as a `COMPONENT_DEPENDENCY` for that custom component.

**Breaking Change**: if `SCHEDULER_JOB_BACKUP_ARGS` specifies `-l proxy` explicitly (not `-l '*'`) then this should be changed to `-l proxy-log-volume` since the backup script has been moved. Note that it is not necessary to specify `-l proxy-log-volume` if `--birdhouse-logs` is also specified because the log data is identical in both.

## Changes

**Non-breaking changes**
- create new volume component

**Breaking changes**
- see above 

## Related Issue / Discussion

- Resolves #618

## CI Operations

birdhouse_skip_ci: false


birdhouse/optional-components/README.rst
```diff
@@ -692,3 +692,41 @@ Note that you do *not* need an SSL certificate set up to deploy the stack in thi
...
+Proxy Log Volume
+----------------
+
+This optional setting creates a named docker volume `proxy-logs` that contains the logs directory for the `proxy` component.
+
+It also creates an Nginx configuration that instructs the proxy service to write access logs to a regular file in that directory.
+
+.. note::
+
+    By default, access logs are only written to the stdout stream of the `proxy` docker container.
```
tlvu: Oh,  you did have the intention to log the `access_log` to stdout by default but the code is not doing this !

birdhouse/optional-components/proxy-log-volume/default.env
```diff
@@ -0,0 +1,24 @@
+# PROXY_LOG_DIR is set to a different directory than the other nginx logs (/var/log/nginx) because the other
+# nginx logs are written to files that are actually symlinks to the stdout and stderr streams. We do not want
+# to share these symlinks in a volume because they will necessarily not point to the correct target once they are
+# mounted in a different container. To avoid potential confusion if trying to read from these symlinked log files
+# in a different container, it is better to just not make them shareable (as a volume) in the first place.
+export PROXY_LOG_DIR="/logs/"
+export PROXY_LOG_FILE="access_file.log"
```
tlvu: By the way, `optional-components/scheduler-job-logrotate-nginx` uses the vars `PROXY_LOG_DIR` and `PROXY_LOG_FILE`.  Should it depend on this `proxy-log-volume`?  This will clearly create a circular dependency cycle.  If our component parsing do not handle circular dependencies, forget this.

birdhouse/optional-components/proxy-log-volume/config/proxy/conf.extra-directives.d/proxy-log.conf.template
```diff
@@ -0,0 +1 @@
+    access_log  ${PROXY_LOG_PATH}  main;
```
tlvu: Should we also add `access_log  STDOUT main` so that the access logs still go to STDOUT even when we activate this log to file.  Not sure about the syntax `access_log  STDOUT main`, I just assume it would be something like that.

Putting myself in the shoes of the a new users not aware of all this problems, he just want more stats so he activates `prometheus-log-parser` and suddenly he loses his `proxy` access logs !  We will get support call for sure !  Let's avoid this support call and keep the `proxy` container logs unchanged.

tlvu:
On the topic of productizing `scheduler-job-logrotate-nginx` I'd like to also tweak the default values to be more friendly

https://github.com/bird-house/birdhouse-deploy/blob/6e16c8dfb80efbf6956a7b14fff6963fc774be87/birdhouse/optional-components/scheduler-job-logrotate-nginx/default.env#L1-L3

Proposed:
```sh
 export SCHEDULER_JOB_LOGROTATE_NGINX_JOB_SCHEDULE="*/20 * * * *"  # UTC 
 export SCHEDULER_JOB_LOGROTATE_NGINX_ROTATE=50 
 export SCHEDULER_JOB_LOGROTATE_NGINX_SIZE=20M 
```

But I wonder if increasing the file size, would the `canarie-api` and the `prometheus-log-parser` be able to handle the much larger file size on restart?

We should also not hardcode the docker image in https://github.com/bird-house/birdhouse-deploy/blob/6e16c8dfb80efbf6956a7b14fff6963fc774be87/birdhouse/optional-components/scheduler-job-logrotate-nginx/config.yml.template#L12

birdhouse/optional-components/proxy-log-volume/config/proxy/conf.extra-directives.d/proxy-log.conf.template
mishaschwartz: good idea!

mishaschwartz:
@tlvu  Are you suggesting we run it every 20 minutes or once per hour at minute 20?
If you're suggesting every 20 minutes then I think we want `0,20,40 * * * *`

> But I wonder if increasing the file size, would the canarie-api and the prometheus-log-parser be able to handle the much larger file size on restart?

It would take longer for canarie-api but I think that 20 minutes worth of logs isn't that bad. I also have a proposed change to canarie-api (keep an eye out for that later) that would mean that this would only be an issue on restart instead of requiring canarie-api to read through the entire file every minute.

This is not a problem at all for prometheus-log-parser.

birdhouse/components/proxy/nginx.conf.template
```diff
@@ -16,13 +16,11 @@ http {
     default_type  application/octet-stream;
 
     # NOTE: if this log_format ever changes, make sure to update the relevant code in the
-    #       prometheus-log-parser component as well to match the change.
+    #       components that assume this log format.
     log_format  main  '$remote_addr - $remote_user [$time_iso8601] "$request" '
                       '$status $body_bytes_sent "$http_referer" '
                       '"$http_user_agent" "$http_x_forwarded_for"';
 
-    access_log  ${PROXY_LOG_PATH}  main;
```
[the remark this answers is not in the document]
fmigneault: Yes, configurable is good. 
Our server does not have as much traffic as PAVICS, nor acts like a prod, so it is not that critical if some logs are lost or not backed up. I prefer to have sufficient cleanup to avoid disk space problems by default.

birdhouse/optional-components/proxy-log-volume/config/proxy/conf.extra-directives.d/proxy-log.conf.template
```diff
@@ -1 +1,2 @@
     access_log  ${PROXY_LOG_PATH}  main;
+    access_log  /var/log/nginx/access.log  main;
```
tlvu: Huh but this will log to the file again inside the container and we will again have log rotation problem? Can we use `/dev/stdout` (if this works)?

tlvu: Oh !  I forgot that !

birdhouse/optional-components/scheduler-job-logrotate-nginx/config.yml.template
```diff
@@ -1,11 +1,12 @@
 - name: logrotate-nginx
   comment: Rotate log files for nginx
   schedule: '${SCHEDULER_JOB_LOGROTATE_NGINX_JOB_SCHEDULE}'
-  command: bash -c 'cp /etc/logrotate.conf.orig /etc/logrotate.conf && chown root:root /etc/logrotate.conf && chmod 644 /etc/logrotate.conf && /usr/sbin/logrotate -v /etc/logrotate.conf'
+  command: bash -c 'cp /etc/logrotate.conf.orig /etc/logrotate.conf && chown root:root /etc/logrotate.conf && chmod 644 /etc/logrotate.conf && /usr/sbin/logrotate -v /etc/logrotate.conf 2>&1 | tee -a /logs/logrotate-nginx.log /dev/stderr > /dev/null'
```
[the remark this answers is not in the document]
tlvu: Oh you meant the `scheduler` container logs !  I was mistaken with the `logrotate-nginx` container logs !

I still do not see how redirecting the `logrotate` command logs inside the  `logrotate-nginx` container would impact the logging to `scheduler` container.

Also you wrote:

> * `tee` also redirects the logs to `stdout` but we don't want that because then the docker logs will have duplicate info on stdout and stderr so...
> 
> * `> /dev/null` silences the logs to stdout from `tee`

So you want to silence stdout but for me `/dev/stderr > /dev/null` means silence stderr?

PR #635, opened 2026-01-14
Add script that automatically updates postgres databases to a later version

mishaschwartz:
## Overview

In anticipation of upgrading postgres databases in the future, this introduces a script that automatically upgrades postgres databases using the backup/restore process.

This includes magpie and all WPS birds that use the postgres component. This does not include test component like `optional-components/generic_bird` and will not update custom components (ones not from this repository).

Test components are not assumed to have persistent data that needs to be updated and we cannot guarantee that other postgres databases used by components outside this repository do not require additional steps (data migrations) in order to comply with a different version of postgres. 

It will update postgres databases to the version specified by the `POSTGRES_VERSION_UPDATE` environment variable.
All of the old database files will be copied to a temporary directory in case you want to inspect them or revert this operation later on. To specify which directory to write these backups to set the `DATA_BACKUP_DIR` variable (default: `${TMPDIR:-/tmp}/birdhouse-postgres-migrate-backup/`)

Note that backups in the form of database dumps will also be written to the named volume or directory specified by the `BIRDHOUSE_BACKUP_VOLUME` variable.

For example, to update the current postgres databases to version 18.1 and write backups to `/tmp/test/`

```sh
$ POSTGRES_VERSION_UPDATE=18.1 DATA_BACKUP_DIR=/tmp/test/ birdhouse/scripts/update-postgresh.sh
```

In a future update we can update the postgres versions and tell users to run this script first in order to safely migrate data from one version to the next.

## Changes

**Non-breaking changes**
- adds new script

**Breaking changes**
- None

## Additional Information

I have tested this by upgrading postgres databases from the current version (9.6) to the latest version (18.1) and all data in the magpie and WPS bird databases are not affected.

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/birdhouse/scripts/update-postgres.sh
+++ b/birdhouse/scripts/update-postgres.sh
@@ -23,6 +23,7 @@
 #
 # $ birdhouse backup create --no-restic -a generic_bird
 # $ birdhouse compose down
+# $ ${THIS_DIR}/backup-datavolume.sh birdhouse_postgres_generic_bird "${BIRDHOUSE_BACKUP_DATA_DIR}"
 # $ docker volume rm birdhouse_postgres_generic_bird
 # $ GENERIC_BIRD_POSTGRES_IMAGE=postgres:18.1 birdhouse compose up -d
 # $ birdhouse backup restore --no-restic -a generic_bird


birdhouse/scripts/update-postgres.sh
```diff
@@ -0,0 +1,67 @@
...
+BIRDHOUSE_EXE="${THIS_DIR}/../../bin/birdhouse"
+
+eval $(${BIRDHOUSE_EXE} configs --print-log-command)
+
+: ${POSTGRES_VERSION_UPDATE:?"$(log ERROR "POSTGRES_VERSION_UPDATE must be set")"}
+
+DATA_BACKUP_DIR="${DATA_BACKUP_DIR:-"${TMPDIR:-/tmp}"/birdhouse-postgres-migrate-backup/}"
+POSTGRES_COMPONENTS="-a magpie -a $(birdhouse -q configs -c 'echo $POSTGRES_DATABASES_TO_CREATE' | sed 's/ / -a /g')"
+
+log INFO "Migrating postgres databases to version ${POSTGRES_VERSION_UPDATE}"
```
fmigneault: Maybe log these two variables immediately before doing the `birdhouse compose up -d` to allow early abort if the user notice they are wrong?

birdhouse/scripts/update-postgres.sh
```diff
@@ -0,0 +1,67 @@
...
+
+log INFO "Starting the birdhouse stack in order to backup existing databases."
+${BIRDHOUSE_EXE} compose up -d
+
+log INFO "Backing up postgres databases from the following components: $(echo "${POSTGRES_COMPONENTS}" | sed 's/[[:space:]]*-a[[:space:]]*/ /g')"
+${BIRDHOUSE_EXE} backup create --no-restic ${POSTGRES_COMPONENTS}
```
fmigneault: Wasn't `BIRDHOUSE_BACKUP_VOLUME` used for backup? What is the difference that needs introducing `DATA_BACKUP_DIR`?

birdhouse/scripts/update-postgres.sh
```diff
@@ -0,0 +1,67 @@
...
+# environment variable.
+# All of the old database files will be copied to a temporary directory in case you want to inspect
+# them or revert this operation later on. To specify which directory to write these backups to 
+# set the DATA_BACKUP_DIR variable (default: ${TMPDIR:-/tmp}/birdhouse-postgres-migrate-backup/)
+# Note that backups in the form of database dumps will also be written to the named volume or directory
+# specified by the BIRDHOUSE_BACKUP_VOLUME variable.
+# 
+# For example, to update the current postgres databases to version 18.1 and write backups to /tmp/test/
+# 
+# $ POSTGRES_VERSION_UPDATE=18.1 DATA_BACKUP_DIR=/tmp/test/ ./update-postgresh.sh
```
fmigneault: Some additional note to provide might be regarding the "types" of supported "postgres".
For example, `postgis`-based one for `stac-db` will not work because of the extra plugins and definitions. Their migration must be done separately.

birdhouse/scripts/update-postgres.sh
```diff
@@ -0,0 +1,69 @@
...
+${BIRDHOUSE_EXE} compose down
+
+log INFO "Backing up all postgres data to ${DATA_BACKUP_DIR}."
+eval "$(${BIRDHOUSE_EXE} -q configs -c 'echo "MAGPIE_PERSIST_DIR=$MAGPIE_PERSIST_DIR; POSTGRES_DATA_DIR=$POSTGRES_DATA_DIR"')"
+
+mkdir -p ${DATA_BACKUP_DIR}
+mv "${MAGPIE_PERSIST_DIR}" "${DATA_BACKUP_DIR}"
+mv "${POSTGRES_DATA_DIR}" "${DATA_BACKUP_DIR}"
+
+MAGPIE_POSTGRES_VERSION=${POSTGRES_VERSION_UPDATE} POSTGRES_VERSION=${POSTGRES_VERSION_UPDATE} ${BIRDHOUSE_EXE} compose up -d
```
tlvu: Humm will need `GENERIC_BIRD_POSTGRES_IMAGE=${POSTGRES_VERSION_UPDATE}` here as well.

tlvu (approved):
Neat little script.  Will come in handy one day for sure.

Usage question: the intended workflow is to run this script, then **immediately** checkout a newer version of `birdhouse-deploy` that has the newer postgres version? Otherwise the temporary setting of `MAGPIE_POSTGRES_VERSION=${POSTGRES_VERSION_UPDATE} POSTGRES_VERSION=${POSTGRES_VERSION_UPDATE}` will not last.  Should this intended workflow be documented?

Also I tried to see if it can also support `generic_bird` so it seems pretty easy, except the part about `generic_bird` using a data-volume instead of a volume-mount from disk.  Did I forget other details to support `generic_bird`.  We actually use `generic_bird` in production to provide a 2nd instance of Finch (`finchasync` https://pavics.ouranos.ca/twitcher/ows/proxy/finchasync?service=WPS&version=1.0.0&request=GetCapabilities).  

If I miss any steps, please let me know so I'll come back to this PR the day need to migrate `generic_bird`.

Other minor suggestions, none blocking.

birdhouse/scripts/update-postgres.sh
```diff
@@ -0,0 +1,69 @@
...
+# 
+# $ POSTGRES_VERSION_UPDATE=18.1 DATA_BACKUP_DIR=/tmp/test/ ./update-postgresh.sh
+
+THIS_FILE="$(readlink -f "$0" || realpath "$0")"
+THIS_DIR="$(dirname "${THIS_FILE}")"
+BIRDHOUSE_EXE="${THIS_DIR}/../../bin/birdhouse"
+
+eval $(${BIRDHOUSE_EXE} configs --print-log-command)
+
+: ${POSTGRES_VERSION_UPDATE:?"$(log ERROR "POSTGRES_VERSION_UPDATE must be set")"}
```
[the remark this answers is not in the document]
mishaschwartz: It's almost the same except it also exits if the variable is not set.

birdhouse/scripts/update-postgres.sh
```diff
@@ -0,0 +1,69 @@
...
+
+DATA_BACKUP_DIR="${DATA_BACKUP_DIR:-"${TMPDIR:-/tmp}"/birdhouse-postgres-migrate-backup/}"
+POSTGRES_COMPONENTS="-a magpie -a $(birdhouse -q configs -c 'echo $POSTGRES_DATABASES_TO_CREATE' | sed 's/ / -a /g')"
```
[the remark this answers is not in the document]
mishaschwartz: Yes but there's probably a better way to run this on other databases. Let me think about that a bit.

PR #641, opened 2026-02-03
Bug fixes for healthchecks and update-postgres.sh

mishaschwartz:
## Overview

  - Fix healthchecks for `finch`, `raven`, and `postgres-magpie` containers
  - update `scripts/update-postgres.sh` to ensure that the user calling the script has permissions to move
    directories that may be owned by root.

## Changes

**Non-breaking changes**
- big fixes

**Breaking changes**
- None

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/finch/docker-compose-extra.yml
```diff
@@ -15,8 +15,8 @@ services:
       test:
         [
           "CMD",
-          "curl",
-          "--fail",
+          "wget",
+          "-qO-",
```
tlvu: Just curious, this pattern `curl --fail URL` is used everywhere (ex: proxy, geoserver, jupyterhub, ...), only these have problems?  What is the problem with only these?

PR #642, opened 2026-02-06
Prometheus send alertmanager API requests to correct path

mishaschwartz:
## Overview

Prometheus communicates with alertmanager through its API which is accessible on docker's bridge network at `http://altermanager:9093/alertmanager/api/v2`. But, prometheus was not configured to include the `/alertmanager` subpath so it was actually sending alerts to `http://alertmanager:9093/api/v2` which returned a 404 error.

This fix updates the prometheus configuration so that it sends alerts to the correct URL.

## Changes

**Non-breaking changes**
- bug fix

**Breaking changes**
- None

## CI Operations

birdhouse_skip_ci: false


tlvu:
> But, prometheus was not configured to include the `/alertmanager` subpath so it was actually sending alerts to `http://alertmanager:9093/api/v2` which returned a 404 error.

Woo, nice catch but this is odd because alertmanager used to work before (there was alert on the alertmanager page and I had to silence it) !  Is this a fall out because instead of using BIRDHOUSE_FQDN we are using the container name?  Vague memory we did this change in the past, can't remember when or how.

fmigneault (approved):

PR #644, opened 2026-02-06
Allow `env.local` to use delayed eval vars immediately, if needed

tlvu:
## Overview

- Allow `env.local` to use delayed eval var immediately

  Before, `env.local` can not immediately use any of the delayed eval var because their real values are only available after `env.local` is fully read.

  So something like this in `env.local` do not work because `JUPYTERHUB_USER_DATA_DIR` is a delayed eval var:
  ```sh
  export JUPYTERHUB_README_FR="$JUPYTERHUB_USER_DATA_DIR/jupyter-readme/LISMOI.ipynb"

  if [ -f "$JUPYTERHUB_README_FR" ]; then

    export JUPYTERHUB_CONFIG_OVERRIDE="$JUPYTERHUB_CONFIG_OVERRIDE
  # /data/jupyterhub_user_data/jupyter-readme/LISMOI.ipynb
  lismoi_on_disk = join(jupyterhub_data_dir, 'jupyter-readme', 'LISMOI.ipynb')
  # /notebook_dir/LISMOI.ipynb
  lismoi_in_container = join(notebook_dir, 'LISMOI.ipynb')
  c.DockerSpawner.volumes.update({lismoi_on_disk: {
      'bind': lismoi_in_container,
      'mode': 'ro',
  }})
  "
  fi  # end if [ -f "$JUPYTERHUB_README_FR" ]
  ```

  `env.local` could simply append `JUPYTERHUB_README_FR` to `DELAYED_EVAL` list but since we need to use its value immediately in `env.local`, we need to eval it immediately.

  With the new `eval_delayed_var`, we can do the following, because `JUPYTERHUB_USER_DATA_DIR` is a delayed eval var, any vars that depend on it should also be delayed eval'ed.

  ```sh
  export JUPYTERHUB_README_FR="$JUPYTERHUB_USER_DATA_DIR/jupyter-readme/LISMOI.ipynb"
  eval_delayed_var JUPYTERHUB_README_FR

  if [ -f "$JUPYTERHUB_README_FR" ]; then
  (...)
  ```

## Changes

**Non-breaking changes**
- Adds new function  `eval_delayed_var` that `env.local` can use

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/birdhouse/README.rst
+++ b/birdhouse/README.rst
@@ -109,6 +109,14 @@ See `env.local.example <env.local.example>`_ (:download:`download </birdhouse/en
 Most variables that can be set in the local environment file (``env.local`` by default) can also be specified as environment variables when running ``bin/birdhouse`` 
 commands. Environment variables will take precedence over those specified in the ``env.local`` file.
 
+In `env.local`, if you wish to immediately use any of the delayed eval variable value, you can use `eval_delayed_var DELAYED_VAR_NAME`
+to compute their value. If not, their real values are only available after `env.local` is read.  If that delayed eval variable depends
+on another variable that you override in `env.local`, then you should define the override before calling `eval_delayed_var`.  Ex:
+`JUPYTERHUB_USER_DATA_DIR` is a delayed eval var because it depends on the value of `BIRDHOUSE_DATA_PERSIST_ROOT` which can be
+overridden in `env.local`.  So if you need the value of `JUPYTERHUB_USER_DATA_DIR` immediately in `env.local`, you need to
+`eval_delayed_var JUPYTERHUB_USER_DATA_DIR` and if you need to override `BIRDHOUSE_DATA_PERSIST_ROOT`, then you need to override it
+before calling `eval_delayed_var JUPYTERHUB_USER_DATA_DIR`.
+
 If the file `env.local` is somewhere else, symlink it here, next to `docker-compose.yml <docker-compose.yml>`_ (:download:`download </birdhouse/docker-compose.yml>`) because many scripts assume this location.
 If autodeploy scheduler job is enabled, the folder containing the `env.local` file needs to be added to `BIRDHOUSE_AUTODEPLOY_EXTRA_REPOS`.
 


fmigneault (commented):
Might be worth having this quickly mentioned around https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse#further-explanations to provide the utility function to users that might need it. 

OK for me to have the option. Users are in change to deal with the resolution order of things they set in `env.local`.

tlvu:
> I think you could actually simplify the example a lot more to make it clearer what the goal of this is.

How about just 

```sh
export JUPYTERHUB_README_FR="$JUPYTERHUB_USER_DATA_DIR/jupyter-readme/LISMOI.ipynb"

# Error, file not found because JUPYTERHUB_README_FR resolve to "${BIRDHOUSE_DATA_PERSIST_ROOT}/jupyterhub_user_data/jupyter-readme/LISMOI.ipynb" because JUPYTERHUB_USER_DATA_DIR is a delayed eval var.
if [ -f "$JUPYTERHUB_README_FR" ]; then
(...)
```
and

```sh
export JUPYTERHUB_README_FR="$JUPYTERHUB_USER_DATA_DIR/jupyter-readme/LISMOI.ipynb"
eval_delayed_var JUPYTERHUB_README_FR

# Now this works because JUPYTERHUB_README_FR resolve properly to  "/data/jupyterhub_user_data/jupyter-readme/LISMOI.ipynb"
if [ -f "$JUPYTERHUB_README_FR" ]; then
(...)
```

>  The example doesn't actually need to show the update to `JUPYTERHUB_CONFIG_OVERRIDE`

Agreed !  Sorry I didn't meant to use a complex example, that was actual production code that we use.  I think the simplification above with additional comments will make it clearer.

> Also please add at least one test for this function in `tests/unit/test_read_configs_include.py`

There is no need because that new function `eval_delayed_var` is used in `process_delayed_eval` which has a bunch of tests already.  I do not see the reason to duplicate the tests for `process_delayed_eval`.

birdhouse/README.rst
```diff
@@ -109,6 +109,14 @@ See `env.local.example <env.local.example>`_ (:download:`download </birdhouse/en
 Most variables that can be set in the local environment file (``env.local`` by default) can also be specified as environment variables when running ``bin/birdhouse`` 
 commands. Environment variables will take precedence over those specified in the ``env.local`` file.
 
+In `env.local`, if you wish to immediately use any of the delayed eval variable value, you can use `eval_delayed_var DELAYED_VAR_NAME`
+to compute their value. If not, their real values are only available after `env.local` is read.  If that delayed eval variable depends
+on another variable that you override in `env.local`, then you should define the override before calling `eval_delayed_var`.  Ex:
+`JUPYTERHUB_USER_DATA_DIR` is a delayed eval var because it depends on the value of `BIRDHOUSE_DATA_PERSIST_ROOT` which can be
+overridden in `env.local`.  So if you need the value of `JUPYTERHUB_USER_DATA_DIR` immediately in `env.local`, you need to
+`eval_delayed_var JUPYTERHUB_USER_DATA_DIR` and if you need to override `BIRDHOUSE_DATA_PERSIST_ROOT`, then you need to override it
+before calling `eval_delayed_var JUPYTERHUB_USER_DATA_DIR`.
```
fmigneault: Can the example be as follows?
Seems quicker to interpret the feature and makes sure the order is explicit. 

```bash
# in your env.local
export BIRDHOUSE_DATA_PERSIST_ROOT=/my/data  # define dependency needed by delayed var
echo "${JUPYTERHUB_USER_DATA_DIR}"           # delayed '${BIRDHOUSE_DATA_PERSIST_ROOT}/jupyterhub_user_data'
eval_delayed_var JUPYTERHUB_USER_DATA_DIR    # evaluate the desired variable
echo "${JUPYTERHUB_USER_DATA_DIR}"           # evaluated '/my/data/jupyterhub_user_data'
```

PR #647, opened 2026-02-10
Add Nvidia MPS component for managing Nvidia GPU resources

mishaschwartz:
## Overview

This creates a container running Nvidia's Multi Process Service ([MPS](https://docs.nvidia.com/deploy/mps/index.html)) which helps manage multi-user GPU access. 
It runs an alternative CUDA interface which manages resource allocation when multiple processes are running simultaneously on the same GPU.
It also allows the node admin to set additional per-user limits through the `JUPYTERHUB_RESOURCE_LIMITS` variable which configures Jupyterlab containers:

- `"gpu_device_mem_limit"`: sets the `CUDA_MPS_PINNED_DEVICE_MEM_LIMIT` environment variable
- `"gpu_active_thread_percentage"`: sets the `CUDA_MPS_ACTIVE_THREAD_PERCENTAGE` environment variable

For example, the following will give all users in the group named `"users"` access to three GPUs in their Jupyterlab container. On the first one (id = 0) only 1GB of memory is available, on the second (id = 1) only 5GB, and on the third (id = 2) only 10GB. Additionally, the container will be able to use 10% of available threads on the GPUs.

  ```shell
  export JUPYTERHUB_RESOURCE_LIMITS='
  [{
        "type": "group", 
        "name": "users", 
        "limits": {
          "gpu_ids": ["0", "1", "2"], 
          "gpu_count": 3, 
          "gpu_device_mem_limit": "0=1G,1=5G,2=10G", 
          "gpu_active_thread_percentage": "10"
        }
  }]
  '
  ```

  Note that leaving any of these limits unset will default to allowing the user full access to the given resource.

Update `CustomDockerSpawner` to make pre spawn hooks and resource limits more configurable

  Introduce `pre_spawn_hooks` and `resource_limit_callbacks` attributes to the `CustomDockerSpawner` class which 
  can be used to further customize the `CustomDockerSpawner` from optional components. This gives us a way to 
  add additional functionality without having to directly modify existing functions which may be overwritten by the
  user when they configure the spawner in `JUPYTERHUB_CONFIG_OVERRIDE`.

  This also introduces the `JUPYTERHUB_CONFIG_OVERRIDE_INTERNAL` variable which is identical to the 
  `JUPYTERHUB_CONFIG_OVERRIDE` variable except that it is intended to only be set by other components (not be the
  user in the local environment file). This allows components to customize Jupyterhub deployments without interfering
  with custom settings created by the user. Note that `JUPYTERHUB_CONFIG_OVERRIDE` has precedence over 
  `JUPYTERHUB_CONFIG_OVERRIDE_INTERNAL`.

Fixes some examples that showed that `gpu_ids` could be given as integers if they were meant to be indexes. However, due to limitation of docker they must be strings. This modifies examples so that it is clear that strings must be used and also updates the code to ensure that string values are only ever passed to docker when spawning a new jupyterlab server.

**Non-breaking changes**
- Adds new additional nvidia-multi-process-service component
- documentation fixes
- make CustomDockerSpawner more configurable
- make jupyterhub configuration more configurable by other components in a way that doesn't interfere with user-defined configurations

**Breaking changes**
- None

## CI Operations

birdhouse_skip_ci: false


birdhouse/optional-components/nvidia-multi-process-service/docker-compose-extra.yml
```diff
@@ -0,0 +1,24 @@
...
+      - nvidia_mps:/tmp/nvidia-mps
+    init: true
+    command: ["nvidia-cuda-mps-control", "-f"]
+    deploy:
+      resources:
+        reservations:
+          devices:
+            - driver: nvidia
+              count: all
+              capabilities: [gpu]
```
fmigneault: Should this be configurable as well? For example, some GPUs reserved for Jupyter and others reserved for other operations (eg: Weaver Workers) ?

Is it better to have all GPU-enabled operations connected to this MPS regardless of the way they are used, and have limited `gpu_ids` defined for jupyter vs others, or instead, have multiple sub-GPU partitions where what this MPS "sees" corresponds to 100% of a given sub-partition and its jupyter-dockerspawner resources do not need `gpu_ids` (or they do, but relative to that %sub-GPU partition) ?

birdhouse/optional-components/nvidia-multi-process-service/02-readonly-cuda-vars.sh
```diff
@@ -0,0 +1,2 @@
+readonly CUDA_MPS_PINNED_DEVICE_MEM_LIMIT
+readonly CUDA_MPS_ACTIVE_THREAD_PERCENTAGE
```
fmigneault: Maybe a comment about the justification.

birdhouse/optional-components/nvidia-multi-process-service/default.env
```diff
@@ -0,0 +1,50 @@
+export NVIDIA_MULTIPROCESS_SERVICE_DOCKER=debian
+export NVIDIA_MULTIPROCESS_SERVICE_VERSION=bookworm-slim
```
tlvu: I am surprised, in the standard debian image, nvidia tools are already installed?

PR #651, opened 2026-02-19
Fix missing homepage

tlvu:
## Overview

Unable to set custom homepage due to a missing back-compat config var mapping

Only when back-compat mode is used.  Regression since `2.4.0`.

## Changes

**Non-breaking changes**
- Add missing back-compat mapping `PROXY_ROOT_LOCATION=BIRDHOUSE_PROXY_ROOT_LOCATION`

## CI Operations

birdhouse_skip_ci: false


mishaschwartz (approved):
Looks good, pending @fmigneault's comment

PR #655, opened 2026-03-04
Improve prometheus-longterm-metrics scraping

### LINKED ISSUE #638: :bug: [BUG]: prometheus-longterm-metrics does not seem to collect any data
## Summary

Enabled
```
  ./optional-components/prometheus-log-parser/
  ./optional-components/prometheus-longterm-metrics/
  ./optional-components/prometheus-longterm-rules/
```
On both of our staging instances before the Christmas holidays, but the https://HOST/prometheus-longterm-metrics/ does not have any of the new hourly and daily metrics.

Same issues as previously described here https://github.com/bird-house/birdhouse-deploy/pull/542#issuecomment-2982134475

On the CRIM's instance https://hirondelle.crim.ca/prometheus-longterm-metrics/, I saw it has the daily metrics names but missing the hourly metrics.

When trying to graph the daily metrics, the graph is empty, as if the metrics exist but the data is not collected.

Note the data and metrics exists in the main (/prometheus) instance.  Everything seem fine on the main instance, it's the longterm instance that is missing metrics.

### To Reproduce

1. Enable the 3 optional-components.
2. Let it runs for a few days so both hourly and daily metrics are calculated.
3.

## Environment

| Server/Platform URL   | 
| Version Tag/Commit    | Our staging are on 2.20.2.
| Related issues/PR     | https://github.com/bird-house/birdhouse-deploy/pull/542
| Related components    | 
| Custom configuration  |

mishaschwartz:
## Overview

The prometheus-longterm-metrics service scrapes longterm data rules from the prometheus service. However, the scrape interval was set too long (1 hour) which meant that some data was probably lost in between scrapes. Also, the prometheus documentation recommends setting unique `external_labels` when exporting data to a different federated prometheus instance.

Note that I don't know how long is too long for a scrape interval before data starts being lost but with some experimentation it seems that 5 minutes is at least below that threshold and it's not so frequent that it seems like a waste to be polling so often. 

## Changes

**Non-breaking changes**
- bug fix

**Breaking changes**
- None

## Related Issue / Discussion

- Resolves #638 

## CI Operations

birdhouse_skip_ci: false


huard (approved):
I must confess that the way promotheus works still is hazy for me, and if this works, then I'm happy. 

If for some reason this solution has some downside, other options could  be:

 - Add `last_over_time(metric[1h])` to each hourly long-tern metric. By default, metrics are only returned if there is a value in the last 5 minutes so this would extend the lookup period. 
 - Synchronize the scraping time with the actual record timestamp so it's within the 5 minutes inverval. 

Again, happy with this solution as it feels the most straightforward.

tlvu (approved):
LGTM, deployed on my test instance, I'll report result tomorrow.

PR #656, opened 2026-03-10
Create S3 service

mishaschwartz:
## Overview
  Introduces a new service that allows users to download data using an S3 interface. This S3 interface is read-only
  for users and it's intended purpose is to serve data (like Thredds, Geoserver, or the secure data proxy).

  Administrators can create buckets on S3 using the `make-s3-bucket.sh` script. This will create a bucket and 
  create a Magpie resource for that bucket. By default, only admin users will be able to interact with a new bucket
  but additional read permissions can be added to users and groups through Magpie.

  For example, to create a bucket named "birdhouse":

  ```sh
  ./birdhouse/scripts/make-s3-bucket.sh birdhouse
  ```

  Administrators can interact with the S3 interface with admin permissions through the `s3-cli` service.
  This uses the [AWS CLI](https://aws.amazon.com/cli/) tool and supports all subcommands.

  For example, to list all buckets:

  ```sh
  ./bin/birdhouse compose run --rm s3-cli s3 ls
  ```

  or to upload some files to the "birdhouse" bucket

  ```sh
  ./bin/birdhouse compose run --rm -v ./files-to-upload:/data:ro s3-cli s3 cp /data s3://birdhouse
  ```

  Administrators can also add files to the birdhouse bucket by adding them directly to the directory
  specified by the `S3_DATA_STORE` config variable. For example, if `S3_DATA_STORE=/data/s3-data` and
  I want to add the same files in the example above to the "birdhouse" bucket:

  ```sh
  cp -r ./files-to-upload /data/s3-data/birdhouse/
  ```

  Note that when setting permission in Magpie that write permissions are ignored and read permissions
  only apply for buckets and individual files. For example, I can set read permissions on the "birdhouse"
  bucket and a file in that bucket at "birdhouse/subdir/file.nc" and that will effect whether a user can
  see the files in "birdhouse" and whether they can read that one file. 
  However, if I set a read permission on "birdhouse/subdir/" this will have no effect on whether the user 
  can or cannot read or browse files in the "subdir/" folder.

## Changes

**Non-breaking changes**
- Adds new service

**Breaking changes**
- None

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/README.rst
```diff
@@ -1105,3 +1105,66 @@ precedence for logging options are as follows:
...
+
+This S3 interface is read-only for users and it's intended purpose is to serve data 
+(like Thredds, Geoserver, or the secure data proxy).
+
+Administrators can create buckets on S3 using the ``make-s3-bucket.sh`` script. 
+This will create a bucket and create a Magpie resource for that bucket. By default, 
+only admin users will be able to interact with a new bucket but additional read 
+permissions can be added to users and groups through Magpie.
+
+For example, to create a bucket named "birdhouse":
```
fmigneault: place in code+double quotes `"birdhouse"`

birdhouse/components/README.rst
```diff
@@ -1105,3 +1105,66 @@ precedence for logging options are as follows:
...
+
+  ./bin/birdhouse compose run --rm -v ./files-to-upload:/data:ro s3-cli s3 cp /data s3://birdhouse
+
+Administrators can also add files to the birdhouse bucket by adding them directly to the directory
+specified by the ``S3_DATA_STORE`` config variable. For example, if ``S3_DATA_STORE=/data/s3-data`` 
+and we want to add the same files in the example above to the "birdhouse" bucket:
+
+.. code-block:: shell
+  
+  cp -r ./files-to-upload /data/s3-data/birdhouse/
```
fmigneault: "upload" somewhat misleading here, more like "make accessible via s3"

birdhouse/components/README.rst
```diff
@@ -1105,3 +1105,66 @@ precedence for logging options are as follows:
...
+
+.. note::
+
+  When setting permission in Magpie, write permissions are ignored and read permissions
+  only apply for buckets and individual files. 
+  
+  For example, I can set read permissions on the "birdhouse" bucket and a file in that bucket 
+  at "birdhouse/subdir/file.nc" and that will effect whether a user can see the files in "birdhouse" 
+  and whether they can read that one file.
```
fmigneault: I am guessing based on Magpie knowledge of the service being configured as `api`.

If the referenced resource is `birdhouse/subdir/`, it will work like THREDDS `directory` type with `browse` permission, while the `file.nc` would work like `file`+`read`? I am guessing the S3 object listing behaviour is directly from the API's response on `birdhouse/subdir/`, but setting `read` (i.e.: recursive) on `birdhouse` would effectively allow `read` on `birdhouse/subdir/file.nc` as well. 

Therefore, the permission on the file itself is redundant in that specific case. If the user is not fully aware of the Magpie `read`/`read-match` distinction, I am worried they might misinterpret this sentence as "if I remove `read` on the file, it won't be accessible anymore", which is not the case.

birdhouse/scripts/make-s3-bucket.sh
```diff
@@ -0,0 +1,50 @@
...
+fi
+if [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
+    echo "$HELP"
+    exit
+fi
+
+${BIRDHOUSE_EXE} compose run -e S3_BUCKET_NAME="$1" --rm --name "create-s3-bucket-${1}" --entrypoint sh s3-cli -c ' \
+    aws s3api head-bucket --bucket thredds || aws s3api create-bucket --bucket "${S3_BUCKET_NAME}" --object-ownership BucketOwnerPreferred; \
+    aws s3api put-bucket-ownership-controls --bucket "${S3_BUCKET_NAME}" --ownership-controls "Rules=[{ObjectOwnership=ObjectWriter}]"; \
+    aws s3api put-bucket-acl --bucket "${S3_BUCKET_NAME}" --acl public-read' || exit 1
```
fmigneault: If it fails here, it is clear enough with `aws-cli`'s error output or it could need more context?

birdhouse/components/s3/config/proxy/conf.extra-service.d/s3.conf.template
```diff
@@ -0,0 +1,21 @@
+    location /s3/ {
```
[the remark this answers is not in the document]
mishaschwartz: We could but we don't do that for any other similar service in the entire stack (thredds, geoserver, etc.). Why is this different?

fmigneault (commented):
Looking forward to the follow-up example PR.

PR #661, opened 2026-03-23
Mount THREDDS data to S3

mishaschwartz:
## Overview

  Automatically mount data in THREDDS to the S3 services as well when the `optional-components/mount-thredds-to-s3` component is enabled.

  This creates a new bucket named `thredds` by default (can be changed by setting the `THREDDS_S3_BUCKET_NAME` variable) which contains a symlink to the thredds data which is mounted separately to the `s3` container.

  A symlink is used so that the THREDDS data itself does not get added as a subdirectory of S3 data which is itself mounted to the S3 component from a bind mount on the host machine.

  A user accessing this data through S3 has the same permissions as if they were accessing the file through THREDDS. Users are currently not permitted to list the files in the `thredds` bucket since there is no good way to check whether the user has permission to list specific files according to the Magpie resource permissions for THREDDS.

## Changes

**Non-breaking changes**
- Creates a new optional component

**Breaking changes**
- None

## CI Operations

birdhouse_skip_ci: false


CHANGES.md
```diff
@@ -15,7 +15,23 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Changes
+
+- Mount THREDDS data to S3
+
+  Automatically mount data in THREDDS to the S3 services as well when the `optional-components/mount-thredds-to-s3`
+  component is enabled.
```
fmigneault: Maybe I'm being pedantic about it, but it is not clear here and in the component description whether enabling `optional-components/mount-thredds-to-s3` will auto-include THREDDS and S3 respectively, or whether you explicitly need all 3 simultaneously.

The error logs would tell the user quickly about it, but it could be indicated immediately here and in the component's description.

birdhouse/optional-components/mount-thredds-to-s3/post-docker-compose-up
```diff
@@ -0,0 +1,20 @@
...
+log INFO "Wait for S3 service to be available"
+for i in $(seq 5); do
+    if [ "$(docker inspect -f '{{ .State.Health.Status }}' s3)" = "healthy" ]; then
+        ln -s /thredds-data "${S3_DATA_STORE}/data/${THREDDS_S3_BUCKET_NAME}"
+        ${COMPOSE_DIR}/scripts/make-s3-bucket.sh ${THREDDS_S3_BUCKET_NAME}
+        log INFO -p "OK"
+        exit
+    fi
+    log INFO -p -n "."
+    sleep 2
```
fmigneault: Provide sleep/retry variables? In case the system running it has more things to do that just so happens to take more time/attempts.

birdhouse/optional-components/mount-thredds-to-s3/config/proxy/conf.extra-service.d/mount-thredds-to-s3.conf.template
```diff
@@ -0,0 +1,37 @@
...
+    # With the following location block, nginx will instead handle it internally.
+    location = /s3/${THREDDS_S3_BUCKET_NAME} {
+        rewrite ^/s3/${THREDDS_S3_BUCKET_NAME}$ /s3/${THREDDS_S3_BUCKET_NAME}/ last;
+    }
+
+    # Deny access to listing (browsing) the catalog because we don't have a good way to check
+    # whether they have browsing permissions for this resource.
+    # This means that users must specifically request an exact path to the thredds file they want to access.
+    # TODO: find a solution that allows us to check for browse permissions for a given prefix.
+    location = /s3/thredds/ {
```
fmigneault: Shouldn't it be `${THREDDS_S3_BUCKET_NAME}` ?

birdhouse/optional-components/README.rst
```diff
@@ -785,3 +785,21 @@ Note that leaving any of these limits unset will default to allowing the user fu
...
+Automatically mount data in THREDDS to the S3 services.
+
+This creates a new bucket named ``thredds`` by default (can be changed by setting the 
+``THREDDS_S3_BUCKET_NAME`` variable) which contains a symlink to the thredds data which is mounted separately to 
+the `s3` container.
+
+A symlink is used so that the THREDDS data itself does not get added as a subdirectory of S3 data which is itself 
+mounted to the S3 component from a bind mount on the host machine.
+
+A user accessing this data through S3 has the same permissions as if they were accessing the file through THREDDS.
```
fmigneault: Reading this at first, I thought permissions were synched between services, but you have a cleaver trick to reuse the actual THREDDS permission.

I think that needs more clarification here. Notably, if an admin wants to control who can access the file in S3, they actually have to set it on the corresponding THREDDS service's sub-resource. 

Something I'm not sure of, if a permission was set on the S3 "API" Magpie service, with a subresource under `THREDDS_S3_BUCKET_NAME`, as if it was another other "normal" S3 file, what would happen? If there are any caveats there, it should be detailed.

birdhouse/optional-components/mount-thredds-to-s3/post-docker-compose-up
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env sh
```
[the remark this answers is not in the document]
mishaschwartz: Yeah. If there's a different `sh` executable in the `PATH` then it will use that one.

Basically `#!/usr/bin/env sh` says "use the sh executable that you have chosen" and `#!/bin/sh` says "I don't care what you've chosen, use this one anyway"

`/usr/bin/env` is almost always preferred, especially if you're not 100% sure of how the user running the script has their system set up.

fmigneault:
run tests

PR #663, opened 2026-03-25
Add optional component to mount a robots.txt file

mishaschwartz:
## Overview

This adds a robots.txt file to the birdhouse stack which can be used to ask bots and web crawlers to not scrape this website.

By default it uses a file that disallows crawling by most major AI crawler bots. To specify a custom robots.txt file set the absolute path to the file as the ``ROBOTS_TXT_FILE`` configuration variable.

## Changes

**Non-breaking changes**
- Adds new optional component

**Breaking changes**
- None

## CI Operations

birdhouse_skip_ci: false


birdhouse/optional-components/robots/default.env
```diff
@@ -0,0 +1,3 @@
+# Location of the robots.txt file, default is the one in this directory
+# This should be an absolute path or a relative path from the birdhouse/ directory
+ROBOTS_TXT_FILE="./optional-components/robots/robots.txt"
```
tlvu, line 3: @mishaschwartz you're missing `export`, so in `robots/config/proxy/docker-compose-extra.yml` that var `ROBOTS_TXT_FILE` is empty, resulting in the error `services must be a mapping` at `./birdhouse-compose.sh up -d` !

Also no need for `version: "3.4"` in `robots/config/proxy/docker-compose-extra.yml`, we have removed the version string in all those files a while ago already.

tlvu: I have an upcoming change, will fix this together with my change, no need to open a new PR for this.

mishaschwartz: Thanks for catching that

Issue #665, opened 2026-03-26
:question: [Question]: Weird thredds download metrics


huard:
## Questions

I've been looking at the results from the thredds download metrics and the results look suspicious. 

https://pavics.ouranos.ca/prometheus/graph?g0.expr=sum%20(increase(thredds_transfer_size_bytes_total%5B1h%5D)%20%3E%200)&g0.tab=0&g0.display_mode=lines&g0.show_exemplars=0&g0.range_input=2h

If I sum over all labels, I get download rates of around 1400 Gb / h. This seems unrealistically high, and inconsistent with the `node_network_transmit_bytes_total`. So I suspect there's something wrong, or someone in Toronto is downloading our entire ERA5 archive, or that I misunderstand what we're measuring. 

In my mind, the node_network_transmit_bytes_total should be higher than the thredds downloads. 

https://pavics.ouranos.ca/prometheus/graph?g0.expr=sum%20by%20(remote_addr)%20(increase(thredds_transfer_size_bytes_total%5B1h%5D)%20%3E%200)&g0.tab=1&g0.display_mode=lines&g0.show_exemplars=0&g0.range_input=1h

I also note that since the thredds download is a counter with labels for the dataset, variables, ip address, etc., making queries returns so many values that requests time out, the browser crashes, etc. 

One option would be to save counter values above 0 only, or sum over some of the labels. Maybe the storage is very efficient and this is not an issue, but I thought it may be worth flagging this. 

## Environment

| Information           | Value
| --------------------- | --------------------------------------------------------------
| Server/Platform URL   | 
| Version Tag/Commit    |  
| Related issues/PR     | 
| Related components    |

fmigneault:
Maybe because of this: https://trac.nginx.org/nginx/ticket/1891

> The $bytes_sent variable returns the number of bytes sent to the client. The value is the same for the main request and all subrequests because client connection is the same for all of them. Subrequest response is sent directly to the client. There's no fake client in a subrequest with a separate byte counter.
> As for the $body_bytes_sent variable, this value currently does not make sense for a subrequest. It returns the value of $bytes_sent minus header size for the active request. For the main request it's the actual response header size, but it is zero for a subrequest.

PR #676, opened 2026-04-10
Thredds: enable Dap4 and CdmRemote and add customizations

tlvu:
## Changes

- Thredds: enable Dap4 and CdmRemote data access protocols and add extra customization options

  Existing dodsC is Dap2 (old, getting deprecated) and cdmremote appears
  to have some performance gain, according to
  https://deepwiki.com/Unidata/netcdf-java/6.3-cdm-remote-protocol

  We currently do not have any example use of these 2 new protocols but
  have to enable them so we can start experimenting with them.

  Previously all the env var customizations for Thredds and TestThredds are
  hardcoded.  Now they are all configurable via `env.local`, like other configs.

  Added a new config var `THREDDS_CONFIG_XML_EXTRA` for customizing the
  threddsConfig.xml file and a sample usage is provided in `env.local.example` to
  increase the OpenDap binary response max size to avoid having to chunk small datasets.

## Changes

**Non-breaking changes**
- All changes are non-breaking, all default values are existing previously hardcoded default values

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/thredds/default.env
```diff
@@ -47,6 +47,46 @@ export THREDDS_DATASET_DATASETSCAN_BODY='
       </filter>
 '
 
+# The minimum and maximum Java heap space memory to be allocated to the TDS
+export THREDDS_XMX_SIZE=20G
+export THREDDS_XMS_SIZE=1G
+
+# TDM related environment variables
+export THREDDS_TDM_PW=CHANGEME!
+export THREDDS_TDS_HOST=http://thredds.yourhost.net/
```
mishaschwartz, line 56: Should this be set to BIRDHOUSE_FQDN_PUBLIC?
What is this used for?

birdhouse/components/thredds/default.env
```diff
@@ -47,6 +47,46 @@ export THREDDS_DATASET_DATASETSCAN_BODY='
       </filter>
 '
 
+# The minimum and maximum Java heap space memory to be allocated to the TDS
+export THREDDS_XMX_SIZE=20G
+export THREDDS_XMS_SIZE=1G
+
+# TDM related environment variables
+export THREDDS_TDM_PW=CHANGEME!
```
mishaschwartz, line 55: If this should be changed then we should set a `__DEFAULT_THREDDS_TDM_PW` variable so and add it to `VARS` so that `read-conf-include.sh` will show a warning if it is not changed from the default.

CHANGES.md
```diff
@@ -15,7 +15,33 @@
 [Unreleased](https://github.com/bird-house/birdhouse-deploy/tree/master) (latest)
 ------------------------------------------------------------------------------------------------------------------
 
-[//]: # (list changes here, using '-' for each new entry, remove this when items are added)
+## Fixes
```
mishaschwartz: Can we please make this a separate PR. It's unrelated to everything else and now we have to wait on the discussion about the thredds part of this PR in order to fix this issue.

I know it's my PR that broke it so I'm sorry about that. I'll just make a separate PR right now and that way we can remove this. (#679)

mishaschwartz (changes requested):
Overall looks great. Just a few comments, see below:

birdhouse/components/thredds/default.env
```diff
@@ -47,6 +47,46 @@ export THREDDS_DATASET_DATASETSCAN_BODY='
       </filter>
 '
 
+# The minimum and maximum Java heap space memory to be allocated to the TDS
+export THREDDS_XMX_SIZE=20G
```
fmigneault, line 51: Seems somewhat high for a "default", assuming the stack runs many other services. Doesn't THREDDS already have 32+GB (as per [*Production Deployment Hardware Recommendations*](https://github.com/bird-house/birdhouse-deploy/tree/master/birdhouse#production-deployment-hardware-recommendations))? Is this within the same 32GB or another one on top?

birdhouse/components/thredds/default.env
fmigneault: The default (or a comment) should also give some guidance about what format of variables are actually expected (`KEY1=VAL1,KEY2=VAL2`,  `KEY1,KEY2`, `KEY1 KEY2`, or ...?).

birdhouse/components/thredds/default.env
```diff
@@ -47,6 +47,46 @@ export THREDDS_DATASET_DATASETSCAN_BODY='
...
+export THREDDS_XMS_SIZE=1G
+
+# TDM related environment variables
+export THREDDS_TDM_PW=CHANGEME!
+export THREDDS_TDS_HOST=http://thredds.yourhost.net/
+# The minimum and maximum Java heap space memory to be allocated to the TDM
+export THREDDS_TDM_XMX_SIZE=12G
+export THREDDS_TDM_XMS_SIZE=1G
```
fmigneault, line 59: Should it use the other heap variables by default?
What is the difference between these and the others?

birdhouse/components/thredds/default.env
```diff
@@ -47,6 +47,46 @@ export THREDDS_DATASET_DATASETSCAN_BODY='
...
+#  <!-- binLimit: maximum size of a binary data response, in Megabytes. Default is 500 Mbytes. -->
+#  <binLimit>2000</binLimit>
+#  </Opendap>
+#"
+export THREDDS_CONFIG_XML_EXTRA=""
+
+
 # add any new variables not already in 'VARS' or 'OPTIONAL_VARS' that must be replaced in templates here
 VARS="
   $VARS
```
fmigneault, line 92: All above are not added either to `VARS` or `OPTIONAL_VARS`?

birdhouse/optional-components/prometheus-log-parser/config/thredds/prometheus-log-exporter.py
```diff
@@ -58,6 +58,7 @@ def parse_line(line):
         # Note that `variable` is not extracted by the regex, but by the logic below.
         labels = {k: groups.get(k, "") for k in LABEL_KEYS}
         # Tweaks
+        # TODO: if needed, add tweaks for 'dap4' and 'cdmremote' once we have working example.
```
fmigneault, line 61: Seems to be important to add now. I assume that this is to better support/fix the problem Trevor faced in https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/161 ? 

If the client switches to dap4 (which is fine in itself), metrics will look like there would be a drop in access once it goes live, and @huard was already trying to demystify these metrics in https://github.com/bird-house/birdhouse-deploy/issues/665

fmigneault (commented):
Another option could have been to simply replace `thredds.env` by `thredds.env.template` and let it work the usual way, without need to "remap" the variables manually in the compose. 

I'm fine with the current approach as well, if explicit listing is prefered. Just more manual variables to maintain.

birdhouse/components/thredds/default.env
mishaschwartz: If they're only used in `docker-compose-extra.yml` then they don't need to be. The only reason to add some of them is so that the mechanism that checks if you're using default values runs properly and displays a nice warning.

birdhouse/components/thredds/default.env
tlvu: Not sure, this was the old hardcoded value that I simply carry-forward to keep back-compat.

birdhouse/components/thredds/default.env
tlvu: Again this was the old hardcoded value that I simply carry-forward to keep back-compat.  Happy to change this to something else if no problem with back-compat.

CHANGES.md
tlvu: Thanks for splitting this out.  I needed to get my PR to actually deploy so I kept it in my PR but having it in a separate PR so it can be merged quickly is the good thing.

birdhouse/components/thredds/default.env
tlvu: 20G (max TDS) + 12G (max TDM) = 32G recommendation (same recommendation, not another 32G on top).

birdhouse/components/thredds/default.env
tlvu: I honestly don't know.  Before this refactoring, these env var was taken from here.

https://github.com/bird-house/birdhouse-deploy/blob/7c41e4545125f040c2ce7ac7a8fca0fc7d6916ab/birdhouse/components/thredds/thredds.env#L1-L2

birdhouse/components/thredds/default.env
tlvu: Exact, they don't need to be used in template expansion so that's why no need to add to `VARS` or `OPTIONAL_VARS`.

birdhouse/optional-components/prometheus-log-parser/config/thredds/prometheus-log-exporter.py
tlvu: Trevor has fixed the problem for dap2 deprecation so we are not yet switching our sample notebooks to dap4. 

There are no sample notebooks using dap4 right now.   We simply want to enable this so we can test the transition.

Not sure how hard it is to generate traffic on `dap4` and `cdmremote` end points.  If easy, @huard do you think you can update the log parser?

Francis worry about missing stats once `dap4` and `cdmremote` are enabled.  Do we have a top-level "thredds" traffic counter so even if we will not have traffic counter for specific protocol, we can still have overall stats for all protocols?

tlvu:
> I'm fine with the current approach as well, if explicit listing is prefered. Just more manual variables to maintain.

I decided to not keep the `.env` file so it looks like all other components.  I think this thredds the only component using `.env` file instead of providing them directly in the compose file.

birdhouse/components/thredds/default.env
fmigneault: I'm wondering what it is actually doing and how to set a value for it, if any is actually needed? I didn't tweak this on my end since nothing ever raised an error about it. But since they would now be documented in the usual component `default.env` way, I think relevant docs should be added.

birdhouse/components/thredds/default.env
fmigneault: OK thanks.

birdhouse/components/thredds/default.env
fmigneault: OK. A quick google search showed that it differs for OPeNDAP/WMS/HTTP vs the indexings (eg: for GRIB). So they should definitely be separate/dedicated values, and it makes sense TDS is more beefed up in our case. Can leave them as is.

birdhouse/optional-components/prometheus-log-parser/config/thredds/prometheus-log-exporter.py
Zeitsperre: For future reference, the fix appears to have been downgrading a few libraries that are non-essential for the rest of the stack within the image to run properly. Most of the problems may have come from newer `pydap` but I suspect that libraries stemming from `HDF5` may also be culpable. It doesn't need to be fixed (_yet_).

CHANGES.md
tlvu: Removed this `optional-components/robots` fix from this PR and merge with `master` to get that fix.

PR #682, opened 2026-04-14
prometheus-log-parser: try to fix failed to poll target

tlvu:
## Overview

Try to fix poll error: "context deadline exceeded".

Screenshot from 2026-04-14 17-47-58

When it works fine with increased  `scrape_timeout`:

Screenshot from 2026-04-14 18-26-33

But even 55s `scrape_timeout`, sometime, it is not even enough

Screenshot from 2026-04-14 19-28-05

But can not have `scrape_timeout` higher than `scrape_interval`.

It would be useful to have some mechanism to diagnose why the prometheus-log-parser takes so much time to respond, see https://github.com/DACCS-Climate/log-parser/issues/6

We do have `export PROMETHEUS_LOG_PARSER_LOG_LEVEL=DEBUG` in `env.local` but this is all we see in `docker logs prometheus-log-parser`, basically nothing useful.

```
2026-04-14 21:36:27,775 - log_parser.log_parser - INFO - file '/var/log/proxy/access_file.log' has been replaced. Tracking will resume from the beginning of the file.
2026-04-14 21:40:46,854 - log_parser.log_parser - INFO - file '/var/log/proxy/access_file.log' has been replaced. Tracking will resume from the beginning of the file.
2026-04-14 21:42:03,771 - log_parser.log_parser - INFO - file '/var/log/proxy/access_file.log' has been replaced. Tracking will resume from the beginning of the file.
2026-04-14 21:45:02,613 - log_parser.log_parser - INFO - file '/var/log/proxy/access_file.log' has been replaced. Tracking will resume from the beginning of the file.
2026-04-14 21:46:00,297 - log_parser.log_parser - INFO - file '/var/log/proxy/access_file.log' has been replaced. Tracking will resume from the beginning of the file.
2026-04-14 21:52:01,214 - log_parser.log_parser - INFO - file '/var/log/proxy/access_file.log' has been replaced. Tracking will resume from the beginning of the file.
```

## CI Operations

birdhouse_skip_ci: false

### DIFF
--- a/birdhouse/optional-components/prometheus-log-parser/config/monitoring/scrape_configs.yml.template
+++ b/birdhouse/optional-components/prometheus-log-parser/config/monitoring/scrape_configs.yml.template
@@ -1,6 +1,11 @@
 scrape_configs:
   - job_name: log_parser
     honor_labels: true
+    # Default scrape_interval 60s in components/monitoring/prometheus.yml.template
+    # Default scrape_timeout 10s in components/monitoring/prometheus.yml.template
+    # Need long scrape_timeout because prometheus-log-parser takes a long time
+    # to respond to GET http://prometheus-log-parser:8000/metrics
+    scrape_timeout: 55s
     static_configs:
       - targets:
         - prometheus-log-parser:${PROMETHEUS_LOG_PARSER_CLIENT_PORT}


fmigneault:
Could it be that the log files are too big, and it takes to long to parse/scrape? If the interval is 60s, there shouldn't be a reason to keep too massive ones. Maybe a smaller size could be set to force early rotation of the older contents that don't need re-parsing?

fmigneault (approved):
Workaround is OK.
Would be nice to solve at the source because that might be another reason why there are missing points in the long-term result.

mishaschwartz:
I don't think it's a problem with the log files themselves (from nginx) because these files are not read when you poll the /metrics endpoint, they're read asynchronously and the metrics are stored in memory.

I think the issue is that we have so many individual metrics stored that it's taking a really long time to read them from the /metrics endpoint. The prometheus-log-parser creates one metric per rule, per file, per access method (fileserver, dodsC, etc.) which can be a huge amount of data if you have a huge amount of thredds files that are being accessed.

A few things you can check to confirm whether this is the case:

- run `docker container stats prometheus-log-parser` and look at the `MEM USAGE` column. If its using more than a GB of memory then that might be an issue
- run `time docker exec prometheus wget -qO- http://prometheus-log-parser:8000/metrics` and see what the total time it takes to check that endpoint is

So if this is the case what can we do about it?

- we can increase the scrape time like we're doing here
- we can change the parser to group some of the data together (so we collect metrics per catalog instead of per file for example)
- we can remove the `_created` metrics which may make up a significant number of the metrics, especially if each metric is rarely updated

mishaschwartz (commented):
I don't think it's a problem with the log files themselves (from nginx) because these files are not read when you poll the /metrics endpoint, they're read asynchronously and the metrics are stored in memory.

I think the issue is that we have so many individual metrics stored that it's taking a really long time to read them from the /metrics endpoint. The prometheus-log-parser creates one metric per rule, per file, per access method (fileserver, dodsC, etc.) which can be a huge amount of data if you have a huge amount of thredds files that are being accessed.

A few things you can check to confirm whether this is the case:

- run `docker container stats prometheus-log-parser` and look at the `MEM USAGE` column. If its using more than a GB of memory then that might be an issue
- run `time docker exec prometheus wget -qO- http://prometheus-log-parser:8000/metrics` and see what the total time it takes to check that endpoint is

So if this is the case what can we do about it?

- we can increase the scrape time like we're doing here
- we can change the parser to group some of the data together (so we collect metrics per catalog instead of per file for example)
- we can remove the `_created` metrics which may make up a significant number of the metrics, especially if each metric is rarely updated

tlvu:
> * run `docker container stats prometheus-log-parser` and look at the `MEM USAGE` column. If its using more than a GB of memory then that might be an issue

```
$ docker container stats prometheus-log-parser --no-stream
CONTAINER ID   NAME                    CPU %     MEM USAGE / LIMIT   MEM %     NET I/O         BLOCK I/O     PIDS
ed2f5ec8f6d0   prometheus-log-parser   0.30%     39.54MiB / 754GiB   0.01%     1.7MB / 127MB   38.3MB / 0B   3

$ docker ps |grep log-parser
ed2f5ec8f6d0   marbleclimate/log-parser:v1.1.1     "log-parser-promethe…"   2 months ago   Up 17 hours (healthy)                                                                                prometheus-log-parser
```
> * run `time docker exec prometheus wget -qO- http://prometheus-log-parser:8000/metrics` and see what the total time it takes to check that endpoint is


```
$ time docker exec prometheus wget http://prometheus-log-parser:8000/metrics
Connecting to prometheus-log-parser:8000 (172.18.0.24:8000)
saving to 'metrics'
metrics              100% |********************************| 5277k  0:00:00 ETA
'metrics' saved

real    0m0.262s
user    0m0.016s
sys     0m0.017s
```

It's really quick this time, I do not understand why.  I suppose it's because the full stack has been restarted yesterday due to the new Jupyter env go-live.

I had to do this fix because David discovered last week we have been missing stats for 12 days in production.

So it might mean the longer prometheus-log-parser runs the slower it gets?!  It might needs a restart from time to time to clear its memory?

tlvu:
While we take our time to find and fix the root cause, could we accept this work-around for production to not lose stats?

mishaschwartz:
> While we take our time to find and fix the root cause, could we accept this work-around for production to not lose stats?

Yes for sure

> It's really quick this time, I do not understand why. I suppose it's because the full stack has been restarted yesterday due to the new Jupyter env go-live.

If you recently restarted then that makes sense

> So it might mean the longer prometheus-log-parser runs the slower it gets?! It might needs a restart from time to time to clear its memory?

Yes but it will be limited by the number of files you have in thredds. So it won't increase forever but it will increase until it has a metric for every file in thredds.

Restarting every once in a while would solve the issue but it would also ensure that we're losing some metrics. I'd recommend we implement one of the solutions discussed here as well (https://github.com/bird-house/birdhouse-deploy/pull/682#pullrequestreview-4123440626) and I could even write a custom collector that sets a TTL on stale metrics if needed.

mishaschwartz (approved):

PR #684, opened 2026-04-15
update long-term metrics with rate (fix #665)

### LINKED ISSUE #665: :question: [Question]: Weird thredds download metrics

huard:
## Overview

Update long-term metrics based on comments in #665

## Changes

**Breaking changes**
- Changed 3 long-term metrics names and definitions to store rate instead of increase. 

## Related Issue / Discussion

- Resolves #665

## Additional Information

- [ ] Keep an eye on metric volume. I'm still concerned the THREDDS download will cause problems due to the large number of labels that are created (dataset, ip, variable, service). 

## CI Operations

birdhouse_skip_ci: false


fmigneault (approved):

tlvu:
I'll have to try this on my VM.  Not sure how Prometheus will react to existing metrics changing datatype.

mishaschwartz (approved):

huard:
Note that I changed the modified metrics names, so there should be no clash between old and new.

tlvu:
> Note that I changed the modified metrics names, so there should be no clash between old and new.

Oh true !  Did not notice that.  Was about to snapshot my VM but no need now :)

tlvu (approved):

PR #686, opened 2026-04-15
reduce verbosity of test bootstrap script 

fmigneault:
## Overview

reduce verbosity of test bootstrap script 

## Changes

**Non-breaking changes**
- Avoid excessive command logging from `birdhouse/scripts/bootstrap-instance-for-testsuite` and related scripts.

  Because `set -x` was used before invoking `read_configs` and all its nested calls which prints large list of variables
  and component definitions to configure them, produced logs would reach excessive lengths (tens of thousands of lines)
  before even reaching the relevant portion of the script to be debugged. The options are moved after the call and each
  nested script reset its state to better manage their verbosity independently.

**Breaking changes**
- n/a

## Related Issue / Discussion

- Fixes #673

birdhouse_skip_ci: false


birdhouse/scripts/bootstrap-instance-for-testsuite
```diff
@@ -14,7 +14,6 @@ THIS_DIR="$(dirname "${THIS_FILE}")"
 COMPOSE_DIR="${COMPOSE_DIR:-$(dirname "${THIS_DIR}")}"
 SCRIPTS_DIR="${COMPOSE_DIR}/scripts"
 
-set -x
```
tlvu, line 17: Why remove this one here?  This script do not call `read_configs` so should not be generating excessive logs?

birdhouse/scripts/bootstrap-testdata
```diff
@@ -21,6 +21,8 @@ if [ -f "${COMPOSE_DIR}/read-configs.include.sh" ]; then
     read_configs
 fi
 
+BOOTSTRAP_TESTDATA_XTRACE_STATE=$(set +o | grep xtrace)
```
tlvu: why `grep xtrace`?  Only restore if `xtrace` is present?

birdhouse/scripts/bootstrap-testdata
```diff
@@ -58,3 +60,7 @@ for afile in ${FILE_LIST}; do
         set +e
     fi
 done
+
+# reset caller state
+set +x
+eval "${BOOTSTRAP_TESTDATA_XTRACE_STATE}"
```
tlvu: By the way, there is no need to restore at the end since the script is over !  And this trace option do not propagate to child scripts like env var so it is pointless to restore at end.

For a child script to start with trace (even if no `-x` in the hashbang line of the child), the parent must call child script explicitly with `sh -x child`.

tlvu (changes requested):
Good idea, just I see the saving and restoring pattern is not required for all the changes here since there are no propagation to child scripts.

birdhouse/scripts/bootstrap-instance-for-testsuite
fmigneault: Before it was trickling down into `bootstrap-testdata` because it had `!/bin/sh -x`, which itself caused the same problem of over-logging `read_configs`. However, it is indeed not a problem anymore since I replaced the shebang and the scripts are not sourced. I'll revert in this case.

birdhouse/scripts/bootstrap-testdata
fmigneault: `-x` option is equivalent (shorthand) to `set -o xtrace`. Using either sets `xtrace` option.

birdhouse/scripts/bootstrap-testdata
fmigneault: In this specific case, indeed, but only because they are not sourced (exec'd instead by `bootstrap-instance-for-testsuite`). If they were sourced, the `-x` would make its way into the sub-scripts. 

Do you prefer we remove them? I went the "just in case" route, but I don't mind removing otherwise.

tlvu: Please remove them.  "source" case is for the `read_configs` only, else we do not "source" that often.

fmigneault: done

tlvu (approved):
LGTM

mishaschwartz (approved):
@fmigneault 

What are you using to fix the newlines in CHANGES.md?

I like that you're cleaning up the file but it makes it really confusing to see a 500+ line file change that is unrelated to this PR.

If this cleanup is a useful thing to have it should be in it's own PR and if it's being done by some other tool then maybe we should add it to the pre-commit rules so that it gets applied by everyone.

fmigneault:
@fmigneault 
I just opened the file in my IDE and it decided to do it itself.
You can tick the hide whitespaces option to avoid this noise. 
I would agree normally to keep features separate in different PRs, but it is more work than its worth to revert and separate in this case.

PR #688, opened 2026-04-20
Reduce number of metrics exported by prometheus log parser

mishaschwartz:
## Overview

The `optional-components/prometheus-log-parser` was creating way too many metrics and it was taking a really long time for prometheus to scrape the metrics from the log parser. This reduces the number of metrics created by: 

- turning off the `_created` metrics by default (can be turned on by setting `PROMETHEUS_LOG_PARSER_PROMETHEUS_DISABLE_CREATED_SERIES=False`)
- moving the `tds_service` label to its own counter named `thredds_transfer_size_by_service_bytes`

This means that the `thredds_transfer_size_bytes` counter no longer contains information about which TDS service was used to download the data. Instead the total amount of data downloaded per-service is calculated in the new counter.

## Changes

**Non-breaking changes**
- Adds new counter to prometheus metrics

**Breaking changes**
- Removes label from `thredds_transfer_size_bytes` metric. Tools that rely on the `tds_service` label of this metric will no longer work.

## Related Issue / Discussion

- Related to #682

## CI Operations

birdhouse_skip_ci: false


CHANGES.md
```diff
@@ -18,6 +18,20 @@
...
+  long time for prometheus to scrape the metrics from the log parser. This reduces the number of metrics created
+  by: 
+
+  - turning off the `_created` metrics by default (can be turned on by setting 
+    `PROMETHEUS_LOG_PARSER_PROMETHEUS_DISABLE_CREATED_SERIES=False`)
+  - moving the `tds_service` label to its own counter named `thredds_transfer_size_by_service_bytes`
+
+  This means that the `thredds_transfer_size_bytes` counter no longer contains information about which TDS service
+  was used to download the data. Instead the total amount of data downloaded per-service is calculated in the
+  new counter.
```
fmigneault, line 33: Wasn't there another nuance? Like, it is not using the combination of TDS-service + every other label, and it is this that greatly reduces since the transfer label is created separately?

fmigneault (approved):

CHANGES.md
mishaschwartz: It's taking a metric that has data in 4 dimensions `"remote_addr", "dataset", "variable", "tds_service"` and reducing it down to 3 dimensions by pulling tds_service out into it's own counter.

fmigneault: Yes, that was my understanding from the previous discussions we had. I just wondered if that needed to be mentioned here, but not blocking either way.

tlvu:
@huard Is this change compatible with the way we want to produce pour Thredds utilization report?  Your review is needed.

mishaschwartz:
@tlvu @huard We discussed this at a Friday meeting and at that time we decided this made sense. I'm going to merge this and if needed we can adjust later.

PR #690, opened 2026-04-21
Documentation refs updates + Clarify secure data proxy

### LINKED ISSUE #435: :books: [Documentation]: optional-components/README.rst still refers to ports
## Description

As mentioned by @fmigneault in https://github.com/bird-house/birdhouse-deploy/pull/428#discussion_r1513335159

> A lot of items seem to refer to accessible services via endpoints with ports, but it shouldn't be the case since https://github.com/bird-house/birdhouse-deploy/issues/222. We should adjust the doc accordingly while we are at it.

For example, the description for the emu component description states:

> Emu service will be available at http://BIRDHOUSE_FQDN:EMU_PORT/wps or https://BIRDHOUSE_FQDN_PUBLIC/TWITCHER_PROTECTED_PATH/EMU_NAME

but the URL that contains EMU_PORT is not longer available and should be removed from this documentation.

## References

Related file: [birdhouse/optional-components/README.rst](https://github.com/bird-house/birdhouse-deploy/blob/8218166d5c8c7163293a656930ff85762eff4b60/birdhouse/optional-components/README.rst)

| Server/Platform URL   | 
| Related issues/PR     | 
| Related documentation |

### LINKED ISSUE #436: :books: [Documentation]: Are we missing files in `html_extra_path` when building docs?
## Description

The `html_extra_path` variable (in docs/source/conf.py) currently contains the following files:

https://github.com/bird-house/birdhouse-deploy/blob/8218166d5c8c7163293a656930ff85762eff4b60/docs/source/conf.py#L175-L180

I'm pretty sure there are other files that should be included here as well. 

If this variable is missing files, then if there is a reference link or download link to these files in the docs, it will result in a 404 error.

We need to do a review of the docs to make sure that all code files that are referenced in the docs are included here.

## References

One example of this is:

https://github.com/bird-house/birdhouse-deploy/blob/8218166d5c8c7163293a656930ff85762eff4b60/birdhouse/README.rst?plain=1#L65

where the `optional-components/emu/docker-compose-extra.yml` file is referenced but is not present in `html_extra_path`.

| Server/Platform URL   | 
| Related issues/PR     | 
| Related documentation |

### LINKED ISSUE #685: :question: [Question]: secure-data-proxy: sample magpie permission should not go-live for real, unless they are actually useful?
## Questions

I have a usecase for `secure-data-proxy` and I notice this sample magpie permission will actually be effective if we enable `secure-data-proxy` for real.

https://github.com/bird-house/birdhouse-deploy/blob/4b66d57058e91d3c2da80d8e236cf4a271928511/birdhouse/optional-components/secure-data-proxy/config/magpie/config.yml.template#L12-L29

But since these are just "sample" should they all be commented out instead?  On the contrary, if they are actually useful, should they stay with their respective component (`weaver` and `wps_outputs-volume`) instead?

## Environment

| Server/Platform URL   | 
| Version Tag/Commit    |  
| Related issues/PR     | 
| Related components    |

### LINKED ISSUE #689: :question: [Question]: Scheduler jobs component dependency
## Questions

Shouldn't all the `optional-components/scheduler-jobs-...` have a dependency like follows
```env
COMPONENT_DEPENDENCIES="
  ./components/scheduler
"

### PR DESCRIPTION
**Pull request by fmigneault**
## Overview

This started as a cleanup of docs cross-references between components to clarify some details related to #685. While managing inter-file references and testing their generation, I came across some errors, which spiralled into a full overall of docs.

See the docs on https://github.com/bird-house/birdhouse-deploy/tree/clarify-secure-data-proxy/birdhouse and other sub-dirs. 

Let me know if you find one that I missed.

## Changes

**Non-breaking changes**
- Update documentation only, no "stack" feature
  - Update doc links and cross-references to work both on GitHub (direct sections links) and rendered Sphinx docs (things resolve where expected). 
  - Remove the `:download:` and `:ref:` for a unified `|<ref>|_` structure. Since all files are provided as-is for quick access, there is no need for explicit download. Because of relative references, the URIs are rendered correctly on GitHub as well, rather than plain `:ref:` that do not resolve when rendered by GitHub.
  - Update some code blocks to use expected lexers, which identified some syntax errors embedded within their code
  - Add Makefile `docs-...` targets for convenience
  - Reorder certain components to make reading more consistent (WPS/birds/data services, followed by auth, followed by proxies/data/volumes) rather than mixed all over the place
  - Move scheduler jobs under optional-components README to align with their actual location
  - Fix multiple links. For example `bin/birdhouse` not resolving.
  - Fix inconsistent single/double code-quotes for italic/code fragments.
  - Fix some typos found along the way

**Breaking changes**
- n/a

## Related Issue / Discussion

- Resolves #689 
- Resolves #685
- Resolves #435
- Resolves #436 (this PR offers a robust method without explicit paths)
- Closes #261 (this PR is more reliable and up-to-date)

## CI Operations

birdhouse_skip_ci: true

### CODE REVIEW

**Review by mishaschwartz — APPROVED** [review]
This looks great. Thanks for the cleanup.
I mostly reviewed this by looking at the built documentation and it looks much better than before.

**Comment by fmigneault** [comment]
@tlvu 
Waiting for your review to validate if the results look all right for you.

**Comment by tlvu** [comment]
> @tlvu Waiting for your review to validate if the results look all right for you.

Yes sorry, back from a week off, please give me a few days.

**Comment by fmigneault** [comment]
@tlvu  No worries. Thanks

File: birdhouse/components/README.rst [hunk 1]
```diff
@@ -1,83 +1,38 @@
...
+.. |birdhouse-compose.sh| replace:: ``birdhouse-compose.sh``
+.. _birdhouse-compose.sh: ../birdhouse-compose.sh
+.. |env.local.example| replace:: ``env.local.example``
+.. _env.local.example: ../env.local.example
 
 .. contents::
 
 
+.. |components-scheduler| replace:: ``./components/scheduler``
+.. _components-scheduler:
```
tlvu, line 23: This `.. _components-scheduler:` do not resolve to anything?

birdhouse/components/README.rst
```diff
@@ -167,55 +122,79 @@ Given the unattended nature, there is no UI.  Logs are used to keep trace.
 How to Enable the Component
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-- Edit ``env.local`` (a copy of env.local.example_ (:download:`download <../env.local.example>`))
```
tlvu, line 170: I believe those ":download:`download <../env.local.example>`" are for ReadTheDocs (so clicking on that link works).

The doc has to both work fine on Github and ReadTheDocs.  All relative links works fine on Github but not on ReadTheDocs if my memory is correct.

birdhouse/components/README.rst
```diff
@@ -167,55 +122,79 @@ Given the unattended nature, there is no UI.  Logs are used to keep trace.
...
+- Edit ``env.local`` (a copy of |env.local.example|_)
 
-  - Add "./components/scheduler" to ``BIRDHOUSE_EXTRA_CONF_DIRS``.
+  - Add |components-scheduler|_ to ``BIRDHOUSE_EXTRA_CONF_DIRS``.
   - Set ``BIRDHOUSE_AUTODEPLOY_EXTRA_REPOS``, ``BIRDHOUSE_AUTODEPLOY_DEPLOY_KEY_ROOT_DIR``,
     ``BIRDHOUSE_AUTODEPLOY_PLATFORM_FREQUENCY``, ``BIRDHOUSE_AUTODEPLOY_NOTEBOOK_FREQUENCY`` as desired,
     full documentation in `env.local.example`_.
   - Run once fix-write-perm_ (:download:`download <../deployment/fix-write-perm>`), see doc in script.
```
tlvu: So you do not remove all ":download:`download <../path/here>`" but some only?

What are the selection criterias?

birdhouse/components/README.rst
```diff
@@ -253,140 +231,149 @@ There are 2 tests that need to be performed:
...
+    # optionally edit env.local to remove the scheduler component from the stack
+    # then remove the running scheduler container
+    birdhouse compose up -d --remove-orphans
 
-  # optionally edit env.local to remove the scheduler component from the stack
-  # then remove the running scheduler container
-  birdhouse compose up -d --remove-orphans
 
+.. |components-monitoring| replace:: ``./components/monitoring``
+.. _components-monitoring:
```
tlvu, line 378: This `.. _components-monitoring:` points nowhere?

birdhouse/components/README.rst
```diff
@@ -890,15 +915,17 @@ Usage
...
 ---------------------------
 
-- Edit ``env.local`` (a copy of `env.local.example`_)
-- Add ``./components/geoserver`` to ``BIRDHOUSE_EXTRA_CONF_DIRS``.
+- Edit ``env.local`` (a copy from |env.local.example|_)
+- Add |components-geoserver|_ to ``BIRDHOUSE_EXTRA_CONF_DIRS``.
+
+
+.. |components-hummingbird| replace:: ``./components/hummingbird``
+.. _components-hummingbird:
```
tlvu, line 930: same here

birdhouse/components/README.rst
```diff
@@ -913,8 +940,12 @@ The service is available at ``${BIRDHOUSE_PROXY_SCHEME}://${BIRDHOUSE_FQDN_PUBLI
...
 ---------------------------
 
-- Edit ``env.local`` (a copy of `env.local.example`_)
-- Add ``./components/hummingbird`` to ``BIRDHOUSE_EXTRA_CONF_DIRS``.
+- Edit ``env.local`` (a copy from |env.local.example|_)
+- Add |components-hummingbird|_ to ``BIRDHOUSE_EXTRA_CONF_DIRS``.
+
+
+.. |components-jupyterhub| replace:: ``./components/jupyterhub``
+.. _components-jupyterhub:
```
tlvu, line 950: same here

docs/source/conf.py
```diff
@@ -52,6 +55,103 @@
...
+
+            # Only write if content changed
+            if content != original_content:
+                html_file.write_text(content, encoding='utf-8')
+        except Exception as e:
+            app.warn(f'Error processing {html_file}: {e}')
+
+
+copy_directories = [
+    "birdhouse",
```
tlvu, line 92: Recursive copy of `birdhouse` dir?

tlvu (changes requested):
I approve the spirit of this PR.

Sometime I notice a link would point nowhere, is that expected?

And please ensure all links also work on ReadTheDocs.

I will make a 2nd pass for the clarification of the Secure Proxy feature.  It was a surprise to me that the same component provide 2 distinct features and I have no way to only select one of the 2.  I only wanted a folder to expose static files, I did not want to secure WPS_output folders !

fmigneault:
@tlvu 

> Sometime I notice a link would point nowhere, is that expected?

If there are some links that I missed (and not already flagged by review), let me know. I waited for your review because I want to catch as many broken links or weird formatting as possible (some were already broken before the PR as well). 

If you meant the ones like `.. _components-jupyterhub:`, they are anchor definitions for the following section header. RST will auto-generate an anchor, but it will fluctuate based on the text of the title. The explicit `.. _components-jupyterhub:` ensures that reference is always available as `#components-jupyterhub` regardless of the section header text.

birdhouse/components/README.rst
fmigneault: see https://github.com/bird-house/birdhouse-deploy/pull/690#issuecomment-4384903955

birdhouse/components/README.rst
fmigneault: see https://github.com/bird-house/birdhouse-deploy/pull/690#issuecomment-4384903955

birdhouse/components/README.rst
fmigneault: see https://github.com/bird-house/birdhouse-deploy/pull/690#issuecomment-4384903955

birdhouse/components/README.rst
fmigneault: see https://github.com/bird-house/birdhouse-deploy/pull/690#issuecomment-4384903955

birdhouse/components/README.rst
fmigneault: I missed that one. Will adjust it as well.

birdhouse/components/README.rst
fmigneault: The reason `:download:` works is that it basically does what the `html_extra_path = [...]` setting does by injecting the reference after the fact at runtime. With the patch added that crawls the potential files to link, they should all be pre-inserted, so the reference works directly like `:download:` did. 

The only thing different is that `:download:` works relatively to the build dir (or docs dir, I don't remember exactly), but the embedded `.. _xxx:` links work relatively to the specific file they are included in. That's why there is a mixture of `../` or `./` depending on what and where is referenced to/from. 

I checked a few times with a local Sphinx build, plus directly on RTD PR-build at some point, but I can double-check again.

docs/source/conf.py
fmigneault: I'm not sure to see where the recursion happens, but the reason a copy was needed is that the symlink works for RST build working on local files, but the rendered HTML does not resolve them. So even if a file path did a valid `../` path that should land on the https://github.com/bird-house/birdhouse-deploy/blob/master/docs/source/birdhouse symlink, it would not resolve the rest of the file (just like GitHub just shows the `../../birdhouse` value rather than redirecting to it). 

That is actually most probably why the `:download:` was needed to insert the "resolved path" from the source to the output html-build dir.

birdhouse/README.rst
```diff
@@ -101,26 +123,32 @@ Quick-start
...
 
-For an example of possible override, see how the `emu service <optional-components/emu/docker-compose-extra.yml>`_ (:download:`download </birdhouse/optional-components/emu/docker-compose-extra.yml>`)
-(`README <optional-components/README.rst#emu-wps-service-for-testing>`_) can be optionally added to the deployment via the `override mechanism <https://docs.docker.com/compose/extends/>`_.
-Ouranos specific override can be found in this `birdhouse-deploy-ouranos <https://github.com/bird-house/birdhouse-deploy-ouranos>`_ repo.
+For an example of possible override, see how the |emu-service-compose|_ (details: |optional-components-emu|_)
+can be optionally added to the deployment via the `override mechanism <https://docs.docker.com/compose/extends/>`_.
+Ouranos specific override can be found in this `birdhouse-deploy-ouranos <https://github.com/bird-house/birdhouse-deploy-ouranos>`_ repository.
+
+.. |emu-service-compose| replace:: ``emu`` service configuration
+.. _emu-service-compose: birdhouse/optional-components/emu/docker-compose-extra.yml
```
tlvu:
```suggestion
.. _emu-service-compose: ./optional-components/emu/docker-compose-extra.yml
```

birdhouse/README.rst
```diff
@@ -140,28 +168,38 @@ Suggested deployment layout:
...
    ./bin/birdhouse compose up -d
 
-If you get a ``'No applicable error code, please check error log'`` error from the WPS processes, please make sure that the WPS databases exists in the
-postgres instance. See `create-wps-pgsql-databases.sh <scripts/create-wps-pgsql-databases.sh>`_ (:download:`download </birdhouse/scripts/create-wps-pgsql-databases.sh>`).
+If you get a ``'No applicable error code, please check error log'`` error from the WPS processes,
+please make sure that the WPS databases exists in the
+postgres instance. See |create-wps-pgsql-databases.sh|_.
+
+.. |create-wps-pgsql-databases.sh| replace:: ``birdhouse/scripts/create-wps-pgsql-databases.sh``
+.. _create-wps-pgsql-databases.sh: birdhouse/scripts/create-wps-pgsql-databases.sh
```
tlvu:
```suggestion
.. _create-wps-pgsql-databases.sh: ./scripts/create-wps-pgsql-databases.sh
```

birdhouse/README.rst
```diff
@@ -140,28 +168,38 @@ Suggested deployment layout:
...
+
 
 Production deployment hardware recommendations
 ----------------------------------------------
 
-RAM: at least 128 GB, Thredds 32+ GB, Geoserver 8+ GB, leaving spaces for other components and all the various Jupyter users
+RAM: at least 128 GB, THREDDS 32+ GB, GeoServer 8+ GB, leaving spaces for other components and all the various Jupyter users
```
tlvu, line 217: Ha !  I did not realized THEDDS is "officially" all caps !

birdhouse/README.rst
```diff
@@ -173,28 +211,33 @@ Note about WPS request timeout
...
-  Nginx proxy, see `nginx.conf`_ (:download:`download <birdhouse/components/proxy/nginx.conf>`).
+  Nginx proxy, see |components-proxy-nginx.conf|_.
   Any WPS requests that will take longer should use the async mode.
 
-  Default value ``PROXY_READ_TIMEOUT_VALUE`` in `default.env`_ (:download:`download <birdhouse/default.env>`).
+  Default value ``PROXY_READ_TIMEOUT_VALUE`` in |birdhouse-default.env|_.
 
   Overrideable in ``env.local`` file, as usual for all values in ``default.env`` file.
 
+.. |components-proxy-nginx.conf| replace:: ``components/proxy/nginx.conf``
```
tlvu, line 236:
```suggestion
.. |components-proxy-nginx.conf| replace:: ``components/proxy/nginx.conf.template``
```

birdhouse/README.rst
```diff
@@ -173,28 +211,33 @@ Note about WPS request timeout
...
+.. _components-proxy-nginx.conf: ./components/proxy/nginx.conf
```
tlvu, line 237:
```suggestion
.. _components-proxy-nginx.conf: ./components/proxy/nginx.conf.template
```

birdhouse/README.rst
```diff
@@ -241,6 +282,14 @@ The canarie monitoring link
 instance is ready to run the automated end-to-end test suite.  That link should
 return the HTTP response code ``200``.
 
+.. |bootstrap-instance-for-testsuite| replace:: ``scripts/bootstrap-instance-for-testsuite``
+.. _bootstrap-instance-for-testsuite: scripts/bootstrap-instance-for-testsuite
+.. |optional-components-all-public-access| replace:: ``./optional-components/all-public-access``
+.. _optional-components-all-public-access: ./optional-components/README.rst#all-public-access
```
tlvu, line 304:
```suggestion
.. _optional-components-all-public-access: ./optional-components/README.rst#give-public-access-to-all-resources-for-testing-purposes
```

birdhouse/README.rst
```diff
@@ -241,6 +282,14 @@ The canarie monitoring link
...
+.. |optional-components-secure-thredds| replace:: ``./optional-components/secure-thredds``
+.. _optional-components-secure-thredds: ./optional-components/README.rst#control-secured-access-to-resources-example
```
tlvu, line 306:
```suggestion
.. _optional-components-secure-thredds: ./optional-components/README.rst#control-secured-access-to-thredds-resources-example
```

birdhouse/README.rst
```diff
@@ -241,6 +282,14 @@ The canarie monitoring link
...
+
+
 Vagrant instructions
```
tlvu: In Vagrant instructions, ([:download:`download </vagrant_variables.yml.example>`](https://github.com/bird-house/birdhouse-deploy/blob/clarify-secure-data-proxy/birdhouse/README.rst#id1)) and ([:download:`download </birdhouse/vagrant-utils/disk-resize>`](https://github.com/bird-house/birdhouse-deploy/blob/clarify-secure-data-proxy/birdhouse/README.rst#id3)) to remove

birdhouse/README.rst
```diff
@@ -754,7 +807,6 @@ the backup and restore jobs.
   * Warning! Using this option may overwrite other docker options that are required for restic to run properly.
     Make sure you are familiar with restic commands and know what you are doing before using this feature.
 
-.. _nginx.conf: ./components/proxy/nginx.conf
 .. _default.env: ./default.env
 .. _`.bumpversion.toml`: ../.bumpversion.toml
 .. _CHANGES.md: ../CHANGES.md
```
tlvu: `CHANGES.md` and `.bumpversion.toml` not found on ReadTheDocs https://birdhouse-deploy--690.org.readthedocs.build/en/690/birdhouse/README.html

tlvu (changes requested):
Few more issues I found.  But overral the docs contain much less link errors than before, both on Github and ReadTheDocs.

birdhouse/components/README.rst
tlvu: If keeping this pointing to nothing, on GitHub the generated link is `components/README.rst#components-scheduler` but the real link is `components/README.rst#scheduler` so it does not work.  

But something un-expected is the wrong anchor works on ReadTheDocs !!?? I have no plausible explanation why !

So all empty links probably suffer the same problem.

birdhouse/components/README.rst
```diff
@@ -821,14 +842,22 @@ collection of DGGS tools and its associated data manipulation ecosystem (using `
...
+  Alternatively, employ sample DGGS dataset and configuration by enabling |optional-components-dggs-data-sample|_.
   Enabling this optional component will set ``DGGS_CONFIG_PATH`` with a predefined configuration for this sample data.
   See the `PyDGGS-API implementation <https://github.com/LandscapeGeoinformatics/pydggsapi>`_ for more details.
-- Optionally, configure variables in ``./components/dggs/default.env`` via ``env.local`` to customize the service.
+- Optionally, configure variables in |components-dggs-default.env|_ via ``env.local`` to customize the service.
+
+.. |components-dggs-default.env| replace:: ``./components/dggs/default.env``
+.. _components-dggs-default.env: ./components/dggs/default.env
+.. |optional-components-dggs-data-sample| replace:: ``./optional-components/dggs-data-sample``
+.. _optional-components-dggs-data-sample: ../optional-components/README.rst#dggs-data-sample
```
tlvu, line 858:
```suggestion
.. _optional-components-dggs-data-sample: ../optional-components/README.rst#use-the-dggs-sample-data-and-configuration
```

birdhouse/components/README.rst
```diff
@@ -1066,45 +1139,43 @@ All outputs from these processes will become available at
 the ``${BIRDHOUSE_PROXY_SCHEME}://${BIRDHOUSE_FQDN_PUBLIC}/wpsoutputs`` endpoint.
 
 By default, this endpoint is not protected. To secure access to this endpoint it is highly recommended to enable the
-``./optional-components/secure-data-proxy`` component as well.
+|optional-components-secure-data-proxy|_ component as well. Not that this component can also be further customized
+with fined-grained access permissions, while retaining the usual public access by combining the the
+|optional-components-secure-data-proxy|_ component with the |optional-components-all-public-access|_ component.
+
+.. |optional-components-all-public-access| replace:: ``./optional-components/all-public-access``
+.. _optional-components-all-public-access: ../optional-components/README.rst#all-public-access
```
tlvu, line 1149:
```suggestion
.. _optional-components-all-public-access: ../optional-components/README.rst#give-public-access-to-all-resources-for-testing-purposes
```

birdhouse/components/README.rst
```diff
@@ -1066,45 +1139,43 @@ All outputs from these processes will become available at
...
+.. |optional-components-secure-data-proxy| replace:: ``./optional-components/secure-data-proxy``
+.. _optional-components-secure-data-proxy: ../optional-components/README.rst#secure-data-proxy
```
tlvu, line 1151:
```suggestion
.. _optional-components-secure-data-proxy: ../optional-components/README.rst#control-secured-access-to-generic-data
```

Humm I think the anchor name should be static, instead of depending on the anchor text.

tlvu (changes requested):
The empty links do not work, unless the anchor name should be matching and static, instead of depending on the anchor text.

birdhouse/optional-components/README.rst
```diff
@@ -1,113 +1,323 @@
...
 
+.. shared references by multiple components definitions
+.. the short form refers to the "header" of the components section, for reference of a service by name
+.. the long form can be used to refer to the component itself, such as when to include it in a configuration
+.. links to section anchors use HTML form purposely to work both on GitHub and Sphinx rendered documentation
+.. see 'docs/source/conf.py:convert_rst_links_to_html'
+.. |magpie| replace:: Magpie
+.. _magpie: ../components/README.rst#magpie
+.. |components-magpie| replace:: ``./components/magpie``
+.. _components-magpie: ../components/README.rst#magpie
```
tlvu, line 12:
```suggestion
.. _components-magpie: ../components/magpie
```

I assume "magpie" will be point to the README but "components-magpie" is the path to the component, to match the `.. |components-magpie| replace:: ./components/magpie` ?

Same wrong pattern for all other components below.

birdhouse/optional-components/README.rst
```diff
@@ -1,113 +1,323 @@
...
+How to enable this config in ``env.local`` (a copy from |env.local.example|_):
+
+* Add |optional-components-canarie-monitoring|_ to ``BIRDHOUSE_EXTRA_CONF_DIRS``.
+
+
+.. _optional-components-scheduler-jobs:
+.. _scheduler-jobs:
+
+Scheduler Jobs
+---------------------------
```
tlvu, line 65: Same length as the header text?

birdhouse/optional-components/README.rst
```diff
@@ -786,29 +1085,36 @@ Note that leaving any of these limits unset will default to allowing the user fu
...
 ignored.
 
-Add ``./optional-components/mount-thredds-to-s3`` to ``BIRDHOUSE_EXTRA_CONF_DIRS`` to enable this component.
+Add |optional-components-mount-thredds-to-s3|_ to ``BIRDHOUSE_EXTRA_CONF_DIRS`` to enable this component.
 
 
+.. |optional-components-robots| replace:: ``./optional-components/robots``
+.. _optional-components-robots:
+
 Robots
```
tlvu: This existing section name is completely meaningless.  Something better would be "Add robots.txt file for crawlers".

tlvu (changes requested):

tlvu:
I do not see anything related to "Clarify secure data proxy" but given the size of this PR, I would rather have this clarification in a separate PR.  Keep this PR for "fixing links and remove duplicate download links" only !

I also want understand you were able to remove all the download links by adding all references to `html_extra_path` or the new
```
copy_directories = [
    "birdhouse",
    "bin",
]
```
?

birdhouse/components/README.rst
fmigneault: Replaced. 
- Works on RTD, see under https://birdhouse-deploy--690.org.readthedocs.build/en/690/birdhouse/components/README.html#how-to-enable-the-component
- Works on GitHub, see under https://github.com/bird-house/birdhouse-deploy/tree/clarify-secure-data-proxy/birdhouse/components#how-to-enable-the-component

birdhouse/optional-components/README.rst
fmigneault: It is not needed. Only requirement is "at least equal".

birdhouse/optional-components/README.rst
fmigneault: That one is special. Same thing applies to any direct `.rst` references. I tried to describe it in the `docs/source/conf.py`. 

The links are defined such as to work on GitHub first, because it does not know how to perform the `.rst` to `.html` conversion, so the link must resolve to something indexed. There are two options for this, either explicitly `../components/README.rst#magpie` or `../components/#magpie` because GitHub happens to render `README[.rst|.md]` directly whether on the file or the directory containing it, and it matches (auto-index) the kebab name of the headers. Technically, [`#components-magpie`](https://github.com/bird-house/birdhouse-deploy/tree/clarify-secure-data-proxy/birdhouse/components#components-magpie) also works, but I want to avoid creating `components-magpie => components-magpie` confusions that Sphinx sometimes complains about.

On the Sphinx side, `../components/#magpie` is "not found" because it does not understand/infer the `README.[rst|html]` from the directory path directly. It takes it literally as directory path. Normally, `../components/README.rst` would not work either because the path is not directly there, but it is only for Sphinx because of the symlink. It auto-manages the transformed `README.html` portion.

Once built on RTD, the directory `../components/magpie` is technically there, but it won't display anything because it is still a web-viewer in the end. It cannot automatically show a directory, it needs and looks for an HTML page. To browse the dir, we would need an explicit `../components/magpie/index.html`. We could actually generate that `index.html` if desired, but I felt it is more useful for the user reading the docs to be redirected to its specific component's docs rather than a list of files. 

The `replace:: ./components/magpie` portion can be whatever we want. That's just the rendering that behaves the same on GitHub/Sphinx/RDT.

birdhouse/optional-components/README.rst
fmigneault: Updated.

fmigneault:
@tlvu 

> I do not see anything related to "Clarify secure data proxy" [...]

It is only the first commit (https://github.com/bird-house/birdhouse-deploy/pull/690/changes/49e37ed5c91bd7102553fd8ed682eb1787cac773?w=1). Because it had to reference many things across components/optional-components to make sense, I had to patch the links. Then, I just went on and continued fixing those links everywhere while at it. 

I didn't create another PR because I would have to manage the copy script across them. It is still only doc updates. 

> I also want understand you were able to remove all the download links by adding all references to `html_extra_path` or the new

Both.
Anything under a dir uses the copy to avoid manually adding every possible link.
The ones at the root uses the explicit links directly in `html_extra_path` because that would cause a recursive copy of the repo if I used the same copy function. Maybe the copy function could actually do a safeguard against `docs/` to avoid it. I didn't try.

birdhouse/components/README.rst
fmigneault: I'm not sure to follow the comment and other similar suggestions about this. The anchor between `|...|` is what remains static.  We can use whichever text in the header and that anchor will always remain available and consistent. 

Technically, `#control-secured-access-to-generic-data` in that case would also exist, but it is prone to change over time if we change the header's text. 

See for example in https://github.com/bird-house/birdhouse-deploy/pull/690/changes/3d6b8a3f9e20756acdc529d7ac61649a68f16333. I just replaced the header text without replacing the anchor anywhere else that uses it, and it still works. That's the strength of this approach, edits are only local there rather than everywhere referring to that section. 

See that https://birdhouse-deploy--690.org.readthedocs.build/en/690/birdhouse/optional-components/README.html#robots does not resolve anymore because of text replacement to 
https://birdhouse-deploy--690.org.readthedocs.build/en/690/birdhouse/optional-components/README.html#add-robots-txt-file-for-bots-and-web-crawlers
However, https://birdhouse-deploy--690.org.readthedocs.build/en/690/birdhouse/optional-components/README.html#optional-components-robots works. 

If we employ the pre-baked anchor names, we can refer to them anywhere (docs or externally), and they will never become invalid, unless we move/remove the section entirely obviously. I reused the same `<component-type>-<component-name>` structure as the actual components such that if any such change happens, it is because they are completely modified (e.g.: deprecated or similar), so the docs should change anyway.

fmigneault:
@tlvu 
Doing something completely unrelated, I came across this visual I had somewhere in my files. 
~Don't know why it wasn't already in the docs, but I'll add it to further clarify.~

<details>
<summary>Old Figure</summary>
Birdhouse Secured Data Access
</details>

I realized why it wasn't included. It is not up-to-date with the actual implementation, as we employed `auth_request` rather than a proxy to make requests efficient. 

A better figure reflecting the actual implementation is added in [`0d54bec` (this PR)](https://github.com/bird-house/birdhouse-deploy/pull/690/commits/0d54bec547d652338eb671731099f5dd74721fa0)

tlvu (approved):
LGTM now.  If there are more changes needed, we can fix in another PR.

PR #693, opened 2026-04-24
update Magpie/Cowbird/Twitcher with Magpie 5.1.0 release

fmigneault:
## Overview

Update Magpie/Cowbird/Twticher with Magpie 5.1.0 release.

## Changes

**Non-breaking changes**

- Magpie / Twitcher / Cowbird: Address error with permission synchronization under a special user.

  When a special user with reserved keyword such as the main Magpie "administrator" was applied a permission, the registration process of the permissions could, in certain cases, yield an error causing the procedure to fail. Notably, if the admin user already exists, which is typically the case since it is required and the service auto-generates it if missing on startup, the permission registration would cause an error since it itself tries to create the user, which is forbidden by the API. This is normally automatically handled by 409 conflict by the registration script, but another 400 error code is used for the special user case, which was not caught. 

  In most deployments, this error is not observed since the administrator *group* is typically employed to apply permissions on multiple "admins" simultaneously, and it does not have the same error handling conditions.

  Because the registration procedure is leveraged by Cowbird to perform a batch multi-permission/resource update, which is the same operation employed by the loaded Magpie configurations, it needs to align with the Magpie API. Cowbird happened to have been using a slightly outdated (although compatible) Magpie 4.3.0 version. Because Magpie and Twitcher have synchronized releases, it is also updated. Both Cowbird, Twitcher and Magpie itself are now aligned with version 5.1.0. These updates also include various dependency security fixes that they share.

**Breaking changes**
- n/a

## Related Issue / Discussion

  - Relates to https://github.com/Ouranosinc/Magpie/pull/650.
  - Relates to https://github.com/Ouranosinc/Magpie/pull/651.
  - Relates to https://github.com/Ouranosinc/cowbird/pull/104.

## CI Operations

birdhouse_skip_ci: false


fmigneault:
run tests

mishaschwartz (approved):
Looks good!

mishaschwartz:
What is this Test (Python3.11) check from? Is this blocking the merge?

image

fmigneault:
Looks like a branch protection rule.
@Zeitsperre did you add it? 

It won't run until https://github.com/bird-house/birdhouse-deploy/pull/694/changes#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957f, and it effectively blocks the PRs until then (unless some override/bypass is allowed).

mishaschwartz:
Oh, that makes sense. Yeah we shouldn't add that as a branch protection rule until #694 has been merged and the action is on the master branch. @Zeitsperre can you remove that rule for now?

Zeitsperre:
@mishaschwartz Sorry about that; I migrated the repo from the obsolete branch protections to the new rules-based approach. It'll be safe to fully enable after #694 is in. The build check is disabled for now.

fmigneault:
@mishaschwartz @Zeitsperre 
How do you want to sync the few pending PRs currently?
Should I wait after the CI auto-tag or I proceed with this? 
Same question about the Jupyter and docs-related adjustments.

Zeitsperre:
@fmigneault This PR could be a low-risk test to see if the new auto-tag system is working as intended. There _might_ be some small adjustments to make.

One thing worth mentioning is that there will likely be a handful of automated PRs opened once #694 is merged; When changes are made to the Dependabot file, it runs and opens PRs if required.

mishaschwartz:
Yes let's use this as a test but at this point do we want to wait until Monday in case something breaks and we don't want to leave it sitting over the weekend?

Zeitsperre:
Things are bad enough with CopyFail and now DirtyFrag this week. Let's wait until Monday ;)

tlvu:
@fmigneault any Magpie DB upgrade here?  If yes I'd want to try an upgrade using our production DB data first.

fmigneault:
@tlvu 
> any Magpie DB upgrade here? If yes I'd want to try an upgrade using our production DB data first.

No DB update. The minor version change is because of an API-side change of HTTP error code logic, so technically a change of behaviour (but one that is expected/desired).

tlvu:
> > any Magpie DB upgrade here? If yes I'd want to try an upgrade using our production DB data first.
> 
> No DB update. The minor version change is because of an API-side change of HTTP error code logic, so technically a change of behaviour (but one that is expected/desired).

Great !  Approving.

tlvu (approved):

PR #700, opened 2026-05-15
[skip jenkins] Add gpg key and adjust workflow label

Zeitsperre:
## Overview

This PR includes two very small adjustments required for normal GitHub Workflow operations.

A GPG Key specific for Bird-house projects has now been added (with access to this repo enabled). This should have been in my original PR, but I forgot to add it.

The other change is a label adjustment to address a small race condition.

This also adds a section to the README explaining the privileges and keys required in order to run the various GitHub Workflows.

## Changes

**Non-breaking changes**

- Adds a necessary GPG key to the `create-tag.yml` workflow.
- Adjusts the label on `auto-accept-ci-changes.yml` to the `github_actions` label.
- Adds a section to bottom of `birdhouse/README.rst` that explains the settings required for GitHub Workflows.

## Additional Information

The `label.yml` workflow that categorized PRs as "ci/operations" would only be completed after the `auto-accept-ci-changes.yml` already has started. This meant that the only time that this workflow would properly work would be after a change was pushed to the branch after being opened.


.github/workflows/create-tag.yml
```diff
@@ -44,6 +44,15 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Import GPG Key
+        uses: crazy-max/ghaction-import-gpg@2dc316deee8e90f13e1a351ab510b4d5bc0c82cd # v7.0.0
+        with:
+          gpg_private_key: ${{ secrets.BIRDHOUSE_HELPER_BOT_GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.BIRDHOUSE_HELPER_BOT_GPG_PRIVATE_KEY_PASSWORD }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
```
mishaschwartz: Since we're using this to sign a tag, not a commit, shouldn't this be [`git_tag_gpgsign`](https://github.com/crazy-max/ghaction-import-gpg#inputs) instead?

Zeitsperre: Good call!

.github/workflows/create-tag.yml
```diff
@@ -44,6 +44,15 @@ jobs:
...
           python-version: ${{ matrix.python-version }}
 
+      - name: Import GPG Key
+        uses: crazy-max/ghaction-import-gpg@2dc316deee8e90f13e1a351ab510b4d5bc0c82cd # v7.0.0
+        with:
+          gpg_private_key: ${{ secrets.BIRDHOUSE_HELPER_BOT_GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.BIRDHOUSE_HELPER_BOT_GPG_PRIVATE_KEY_PASSWORD }}
+          git_user_signingkey: true
+          git_tag_gpgsign: true
+          trust_level: 5
```
fmigneault, line 54: If the git-user (github action running the workflow in this case?), wouldn't that make it a level 4, since it is not "birdhouse-helper-bot" itself?

Not blocking, but just out of curiosity. I'm not an expert nor know what could be the impact of lvl4 vs lvl5 in this case.

Zeitsperre: In all honesty, I don't really know. This was an example that I reused from elsewhere. If you have opinions on this, I more than welcome them.

mishaschwartz: I don't think it really matters in this case because Github doesn't actually use the web-of-trust (WOT) model to verify commits using a decentralized network, it just uses the public key that you've uploaded to verify (just a regular PKI model).

Even if it was using WOT to verify commits then you'd still want to use trust level 5 because we control the private key as well, so we have to trust the signer because we are the ones using the private key to do the signing.

tlvu:
Just curious, last time the create-tag workflow failed (https://github.com/bird-house/birdhouse-deploy/actions/runs/25896036654/job/76109218802) was because the git username was not set.  So adding the GPG key will solve that?

Is it possible to reference in that `create-tag.yml` the documentation how to set it up properly so other can also maintain it?  Same doc for the new `release.yml`.

.github/workflows/create-tag.yml
fmigneault: I wouldn't consider myself knowledgeable about it either further than what was already said. All good as is.

fmigneault (approved):

Zeitsperre:
> Just curious, last time the create-tag workflow failed (https://github.com/bird-house/birdhouse-deploy/actions/runs/25896036654/job/76109218802) was because the git username was not set. So adding the GPG key will solve that?

Precisely. The GPG key has a username and email associated with it. `git` knows to get its author information from there.

> Is it possible to reference in that `create-tag.yml` the documentation how to set it up properly so other can also maintain it? Same doc for the new `release.yml`.

I suppose I can add this somewhere, sure.

tlvu:
> Precisely. The GPG key has a username and email associated with it. `git` knows to get its author information from there.

Ah, thanks for the explanation.  I was trying to connect the dots!  This GPG key is the bot's one, not your personal GPG right?

Zeitsperre:
Indeed, the GPG key is the bot one, but it's tied to my account as we don't have a bot account that we could use.

tlvu:
> Indeed, the GPG key is the bot one, but it's tied to my account as we don't have a bot account that we could use.

So the commit author is still you?  We currently have this bot user https://github.com/ouranos-deploy, would that work?  The credential to this bot user already shared via our credential sharing mechanism if you need it.

Zeitsperre:
I could definitely migrate the key, or I could create a new one with that user. We're using this key in a few places already, so it would make sense to do this at the org level.

Would it be cleaner to make a `bird-house-bot` user though?

tlvu:
> Would it be cleaner to make a `bird-house-bot` user though?

Ah oops forgot this is bird-house org.  Yes that would be cleaner.   So we need one bot user per org?  If we have a generic user like `automation` or `deploy` or `commit-bot` can it be shared across orgs or it has to be one bot per org?  

I wouldn't consider this blocking.  We can merge this PR now and switch the user in a different one.  I am guessing a name debate for the new bot user would take some time.

tlvu (approved):
Please just add some docs for future maintainer.

Zeitsperre:
I'm hesitant to have a shared user bot across multiple orgs, simply for security reasons. Sharing access/responsibility according to our respective orgs seems safer/easier.

`ouranosinc-bot` and `crim/luqia-bot` and `bird-house-bot` seem descriptive enough as far as names go. Doesn't have to be fancy.

tlvu:
> `ouranosinc-bot` and `crim/luqia-bot` and `bird-house-bot` seem descriptive enough as far as names go. Doesn't have to be fancy.

Name debate possible for shared bot only.  If we're going the separate bot per org then yes it's simpler for the naming.  Up to you guys.

Zeitsperre:
This is good to go, AFAIC!

mishaschwartz (approved):
Yup! looks good

tlvu:
Good for me, thanks.

Zeitsperre:
@mishaschwartz This PR will fix the tagging workflow!

tlvu:
> @mishaschwartz This PR will fix the tagging workflow!

@Zeitsperre FYI we have all 3 approved this PR.

PR #702, opened 2026-05-19
Group dependabot updates for better handling by other workflows

mishaschwartz:
## Overview
Only group patch and minor updates together for updates to github actions. This will allow the `dependabot-auto-approve` job to approve and merge these updates without having to wait for approval if the group also includes a major update.

## Changes

**Non-breaking changes**
- updating grouping rules for dependabot

**Breaking changes**
- None

## Additional Information

## CI Operations

birdhouse_skip_ci: false


.github/dependabot.yml
```diff
@@ -15,9 +15,10 @@ updates:
         patterns:
           - "*"
         update-types:
+          # only group patch and minor updates which will allow the dependabot-auto-approve job to approve and
+          # merge these updates without having to wait for approval if the group also includes a major update.
           - patch
           - minor
-          - major
```
Zeitsperre, line 20: If we remove this here, we won't receive major updates for GitHub Actions at all. This is fine, but it just means that when workflows start emitting deprecation warnings, we'll need to manually intervene.

mishaschwartz: No, this just means that only patch and minor updates will be grouped together. 
Major updates will be put in their own PR

mishaschwartz: From [the documentation](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#groups--)

> When groups is used to define rules:
>    All updates for dependencies that match a rule are combined in a single pull request.
>    If a dependency matches more than one rule, it's included in the first group that it matches.
>    Any outdated dependencies that do not match a rule are updated in individual pull requests.

So since major updates are not included in the grouping rule they will be put in their own PR (see the last line above).

Zeitsperre: Oh, alright. I was sure it was the opposite. My mistake!

Zeitsperre (approved):
LGTM

.github/dependabot.yml
```diff
@@ -15,9 +15,10 @@ updates:
         patterns:
           - "*"
         update-types:
+          # only group patch and minor updates which will allow the dependabot-auto-approve job to approve and
+          # merge these updates without having to wait for approval if the group also includes a major update.
```
tlvu: Do you actually mean " if the group do NOT include a major update"?

tlvu (approved):
LGTM but it looks like the comment needs clarification or additional punctuation to help reading properly?

.github/dependabot.yml
mishaschwartz:
```suggestion
          # merge these updates without having to wait for human approval if the group also includes a major update.
```

@tlvu ?

tlvu: So my understanding is you are only grouping minor and patch and these will be auto approve and merge.  But if there is a major update, that will need manual human approval.

```suggestion
          # merge these updates without having to wait for approval if the group only includes minor or patch update.  Major update will need human manual approval.
```

tlvu:
Update the changelog too, same text there.

mishaschwartz:
@tlvu 

Please see updated language and let me know if that makes sense.

tlvu (approved):
LGTM now, thanks.

tlvu:
OH LOL, the PR description needs to match too :)

fmigneault (approved):

PR #703, opened 2026-05-19
Fix non-posix shell syntax in the `optional-components/emu` component

mishaschwartz:
## Overview

The `default.env` file in the `emu` component was using `==` to do string comparison which is not a bash-ism.
When loading this component in a non-bash shell (bourne shell `sh` is used by default) this will cause a syntax error and the relevant `if` block will always fail. This may cause the `wps_volume` to not be properly loaded causing the component to be misconfigured.

## Changes

**Non-breaking changes**
- bug fix

**Breaking changes**

## Additional Information

## CI Operations

birdhouse_skip_ci: false


tlvu (approved):
Good catch !  You used emu?  Any similar pattern in other default.env?

mishaschwartz:
> Good catch ! 

Thanks!

> You used emu? 

Not really, I was testing something out that required me to add every component and it broke when I added emu so that's why I was investigating that.

> Any similar pattern in other default.env?

Nope! Everything else looks good as far as this bug is concerned.

fmigneault (approved):

PR #705, opened 2026-05-19
Apply logging arguments when running configuration "print" commands through `bin/birdhouse`

mishaschwartz:
## Overview

The logging flags available through the `bin/birdhouse` interface are not applied to the current shell if the output of the `configs --print-config-command` and `configs --print-log-command` are `eval`ed in order to load configuration settings or log functions in the current shell environment.

For example, if you run:

```sh
eval $(bin/birdhouse --quiet configs --print-config-command)
```

the log outputs will *not* be silenced as expected. Similarly, the following:

```sh
eval $(bin/birdhouse --log-file example.log configs --print-log-command)
```

will *not* write logs to `example.log`.

This change ensures that the expected logging variables are set properly so that setting these values will have the expected effect.

## Changes

**Non-breaking changes**
- bugfix

**Breaking changes**
- None

## CI Operations

birdhouse_skip_ci: false


tlvu (approved):
I suppose it looks good.  Add a test or 2 in case future changes by someone else break this?

Also I forgot why the config has to be unset at the end?   I think it's due to this unset requirement that this problem arise?

So this would happen on the nested call only (when proper unset did not occur between 2 calls)?  But your example do not require nested call so I am a bit lost as why this situation would happen.

mishaschwartz:
@tlvu 

Tests are a good idea. I'll add some

> Also I forgot why the config has to be unset at the end?

It's mostly for courtesy so that it doesn't overwrite settings that had previously been set. We could easily not revert them back, it just depends on what we think the user will expect.

> So this would happen on the nested call only (when proper unset did not occur between 2 calls)?

For the resetting/unsetting the variables then it really only matters when you're making nested calls.
What this PR is trying to fix though doesn't have anything to do with resetting/unsetting the variables though (we were doing that already). 
This PR fixes the issue that the environment variables that configure the way that logging is set up didn't get set *at all*.

tlvu:
> This PR fixes the issue that the environment variables that configure the way that logging is set up didn't get set _at all_.

Oh sorry, looking at the changes again I see it now !  

Basically all the variables set during cli args parsing are not preserved.

Should there be a list of var names, like `VARS` or `OPTIONAL VARS` list that is collected directly during cli args parsing and the set and unset function just iterate over that list so future new variables (if you add more options to the cli parsing later) will transparently be set and preserve?

Not blocking.  Can merge this PR now and do this improvement later or just log an issue for later.  I don't want to hold back this PR.  Just a suggestion.

fmigneault (approved):

mishaschwartz:
> Should there be a list of var names, like VARS or OPTIONAL VARS list ... 

yeah I was thinking about that but when I started to actually implement it, the code got pretty obscure and there was quite a few exceptions because of handling resetting vs. unsetting and it just seemed a lot clearer to write it like this. We can revisit it later when I have time.

tlvu:
> We can revisit it later when I have time.

Absolutely

tests/unit/test_cli.py
```diff
@@ -180,6 +184,83 @@ def test_configs_set_env_file(cli_path, run, local_env_file, tmp_path, flag):
...
+               ("--log-file", "test2.log", "BIRDHOUSE_LOG_FILE", Path.cwd() / "test2.log"),
+               ("--log-file", "INFO test2.log", "BIRDHOUSE_LOG_DEST_OVERRIDE", f":DEBUG:fd:1:INFO:file:{Path.cwd() / 'test2.log'}", f":INFO:file:{Path.cwd() / 'test2.log'}"),
+               ("--log-level", "INFO", "BIRDHOUSE_LOG_LEVEL", "INFO"),
+               ("--backwards-compatible", "", "BIRDHOUSE_BACKWARD_COMPATIBLE_ALLOWED", "True"),
+               ("--env-file", Path.cwd() / "example2.env", "BIRDHOUSE_LOCAL_ENV", Path.cwd() / "example2.env"))
+
+
+@pytest.mark.parametrize("env", ("full", "empty"))
+@pytest.mark.parametrize("params", CONFIG_PARAMS, ids=[f"{c[0]}{' <arg>' if c[1] else ''}" for c in CONFIG_PARAMS])
+def test_configs_print_config_command(cli_path, run, config_command_env, root_dir, params, env):
```
tlvu, line 214: Could you add some docs what this test is trying to do?  To help the future maintainer.

For me it looks way too complicated to understand.

tests/unit/test_cli.py
```diff
@@ -180,6 +184,83 @@ def test_configs_set_env_file(cli_path, run, local_env_file, tmp_path, flag):
...
+            assert f"{env_var}=''" in suffix
+
+
+@pytest.mark.parametrize("env", ("full", "empty"))
+@pytest.mark.parametrize("params", CONFIG_PARAMS[:-1], ids=[f"{c[0]}{' <arg>' if c[1] else ''}" for c in CONFIG_PARAMS[:-1]])
+def test_configs_print_log_command(cli_path, run, config_command_env, root_dir, params, env):
+    flag, flag_value, env_var, env_value, *empty_values = params
+    unset_env = [] if env == "full" else ["BIRDHOUSE_BACKWARD_COMPATIBLE_ALLOWED", "BIRDHOUSE_LOCAL_ENV"]
+    env = config_command_env if env == "full" else {}
+    proc = run(f"{cli_path} {flag} {flag_value} configs --print-log-command", unset_env=unset_env, env=env)
```
tlvu, line 257: This one is almost identical to the other one?  Can this be folded together?  Or both calls common functions.

mishaschwartz: They call different functions. `--print-config-command` is different from `--print-log-command` and they should be tested separately.

PR #706, opened 2026-05-19
[skip jenkins] Add spellchecker to pre-commit hooks

Zeitsperre:
## Overview

This PR adds a spellchecking tool to the `pre-commit` hooks and addresses several typos throughout the code base.

The dictionary of ignored words can either be passed via a newline-separated list of words or directly via command-line argument. The ignored-words-list argument can be found in the `pre-commit` hook.

Misspelled words will be caught by the linting tools and will need to be addressed before other checks are performed.

## Changes

**Non-breaking changes**
- Adds a `pre-commit` check for common spelling mistakes.
- Spelling mistakes found across all files within the project have been corrected, including older entries in the `CHANGES.md`.

## Additional Information

https://github.com/codespell-project/codespell

## CI Operations

birdhouse_skip_ci: false


tlvu (approved):
LGTM but I just want to ensure the spell-checker will not touch the code (variable name) but only comments?  And what if we have a "look-a-like" product, ex: if ever Ngix actually exists, will it still auto-correct to Nginx?

Zeitsperre:
> LGTM but I just want to ensure the spell-checker will not touch the code (variable name) but only comments?

This is solely a spellchecker. It doesn't perform any corrections. All corrections (or exceptions) need to be done manually.

> And what if we have a "look-a-like" product, ex: if ever Ngix actually exists, will it still auto-correct to Nginx?

We would then add 'ngix' to the `ignored-words-list` in the `pre-commit-config.yml`.

tlvu:
> This is solely a spellchecker. It doesn't perform any corrections. All corrections (or exceptions) need to be done manually.

OH !!! I thought all the corrections in here are made by the new spell-checker !  I've seen you added other pre-commit hook in other repos that directly clean some metadata in notebooks when they are pushed to a PR so I though this one does the same.

mishaschwartz (approved):

.pre-commit-config.yaml
```diff
@@ -11,3 +11,8 @@ repos:
       - id: ruff-format
         files: birdhouse/components/jupyterhub/jupyterhub_custom/jupyterhub_custom/
         args: [ --config=birdhouse/components/jupyterhub/jupyterhub_custom/ruff.toml ]
+  - repo: https://github.com/codespell-project/codespell
+    rev: v2.4.2
+    hooks:
+      - id: codespell
+        args: [ '--ignore-words-list=afile,bu,te,tre,vas' ]
```
fmigneault: Is there a way to have a more manageable file?

.pre-commit-config.yaml
```diff
@@ -11,3 +11,8 @@ repos:
       - id: ruff-format
         files: birdhouse/components/jupyterhub/jupyterhub_custom/jupyterhub_custom/
         args: [ --config=birdhouse/components/jupyterhub/jupyterhub_custom/ruff.toml ]
+  - repo: https://github.com/codespell-project/codespell
+    rev: v2.4.2
+    hooks:
+      - id: codespell
```
fmigneault, line 17: Does it have a locale preference? E.g.: Canadian "favour" over US "favor".

fmigneault (approved):

.pre-commit-config.yaml
Zeitsperre: I don't think it favours one spelling over another. Instead, it catches clearly misspelled words (one or two letters off) and provides suggestions for the closest possible word(s) it could be.

Their dictionaries include US and GB locales, so Canadian English supported enough.

.pre-commit-config.yaml
Zeitsperre: You can use a text file with a newline-separated list of ignored words, if desired. This list can also live in a `pyproject.toml`, but we don't rely on one here.

We might be able to define our own dictionaries, but those are only really useful if we're consistently misspelling words.

.pre-commit-config.yaml
Zeitsperre: For reference, we use a mix of Canadian English and American English in `xclim`, for documentation and code, respectively, and `codespell` doesn't give us any false positives there.

.pre-commit-config.yaml
fmigneault: If you have the doc nearby, I think its best to set it up right away. The file can be along other CI configs under `.github`.

PR #708, opened 2026-05-30
fix `DELAYED_EVAL` pre-export management

fmigneault:
## Overview
Fix compose env resolution to preserve internal `DELAYED_EVAL` list when re-applying caller environment.

  Calling `read_configs` re-applies exported variables from the caller process using `reset_process_env`.
  If the caller shell had already exported `DELAYED_EVAL` preemptively (by mistake or by sourcing a file exporting it),
  that step could overwrite the internally assembled delayed-eval list, causing following invocations to rely on a stale
  list of variables. Many derived paths (e.g.: `${BIRDHOUSE_DATA_PERSIST_ROOT}/...`) would then remain unresolved.
  Notably, this could fail compose commands (not files using `.template`, but directly including delayed-eval variables)
  with errors such as `undefined volume ${BIRDHOUSE_DATA_PERSIST_ROOT}/...` due to invalid portions of volume mounts
  (e.g.: `${COWBIRD_MONGODB_DATA_DIR}:/data/db` becomes `${BIRDHOUSE_DATA_PERSIST_ROOT}/mongodb_cowbird_persist:/data/db`)
  because the environment variables are used directly in the compose files rather than by shell/templating resolution.

  The fix filters `DELAYED_EVAL` from the staged caller environment before running `reset_process_env`,
  ensuring the internally computed list (if any) is not overwritten by caller values breaking following references.

## Changes

**Non-breaking changes**
- add missing `export` for `COWBIRD_MONGODB_DATA_DIR`
- moved Cowbird's `EXTRA_VARS=` definition closer to `DELAYED_EVAL` for clarity
- fix handling of `DELAYED_EVAL` in cases of stale/broken env

**Breaking changes**
- n/a

## Additional Information

Example test.
Do something like the following (any var using a delayed eval):
```shell
export DELAYED_EVAL=''
bin/birdhouse -L DEBUG configs -c  'echo "COWBIRD_MONGODB_DATA_DIR=${COWBIRD_MONGODB_DATA_DIR}"'
```
Output result **before** the PR:
```shell
COWBIRD_MONGODB_DATA_DIR=
```

Still on the **before** PR, simply add the missing `export`
```diff
- COWBIRD_MONGODB_DATA_DIR='${BIRDHOUSE_DATA_PERSIST_ROOT}/mongodb_cowbird_persist'
+ export COWBIRD_MONGODB_DATA_DIR='${BIRDHOUSE_DATA_PERSIST_ROOT}/mongodb_cowbird_persist'
```

Now we get half the solution of the PR, but still a problem is observed.
The output is (not evaled) because of the `export DELAYED_EVAL=''` that already exists: 
```
COWBIRD_MONGODB_DATA_DIR=${BIRDHOUSE_DATA_PERSIST_ROOT}/mongodb_cowbird_persist
```

With **after** PR:

Output result (as expected because early `DELAYED_EVAL` is dismissed, so it gets populated properly):
```shell
COWBIRD_MONGODB_DATA_DIR=/data/mongodb_cowbird_persist
```

## CI Operations

birdhouse_skip_ci: false


CHANGES.md
```diff
@@ -52,9 +68,9 @@
 
   The `optional-components/prometheus-log-parser` was creating way too many metrics and it was taking a really
   long time for prometheus to scrape the metrics from the log parser. This reduces the number of metrics created
-  by:
```
mishaschwartz, line 55: I don't love that these style fixes always make their way into unrelated PRs so I made #711 so that we can actually enforce these.

mishaschwartz (approved):

CHANGES.md
tlvu: Agreed, I am also annoyed by unrelated changes in PR.  I prefer focused and smaller PR, easier to review.

tests/unit/test_read_configs_include.py
```diff
@@ -327,6 +327,118 @@ def test_delayed_eval_custom_value(self, read_config_include_file, exit_on_error
...
+        }
+        proc = self.run_func(
+            read_config_include_file,
+            extra,
+            (
+                'echo "$COWBIRD_MONGODB_DATA_DIR"; '
+                'echo "PRECOND_HAS_COWBIRD_DELAYED=$(echo \" $DELAYED_EVAL \" | tr "\\n" " " '
+                '| grep -q "[[:space:]]COWBIRD_MONGODB_DATA_DIR[[:space:]]" && echo 1 || echo 0)"; '
+                'echo "HAS_COWBIRD_DELAYED=$(echo \" $DELAYED_EVAL \" | tr "\\n" " " '
+                '| grep -q "[[:space:]]COWBIRD_MONGODB_DATA_DIR[[:space:]]" && echo 1 || echo 0)"; '
```
tlvu: PRECOND_HAS_COWBIRD_DELAYED and HAS_COWBIRD_DELAYED have identical check?  If yes, why 2 different vars?

tlvu (changes requested):
I like the fix to block `DELAYED_EVAL` from being "saved" accidentally by someone doing an `export DELAYED_EVAL=something` in the shell before bringing up the stack.  

Any other "internal" variables falling on the same case?  `VARS`, `OPTIONAL_VARS`?  All the append kind of vars I guess since the "existing env var take predecence over config file" feature.  Not blocking, can be in a different PR.  Let's get this one in first.

The test seems overly complicated.  We can easily search for some specific string using python, why need to use `grep`?

tests/unit/test_read_configs_include.py
fmigneault: I've simplified it.

fmigneault:
@tlvu 

> Any other "internal" variables falling on the same case? VARS, OPTIONAL_VARS?

Possibly, but I have not investigated further. 
This PR is based only on the minimal case I encountered. 

>  We can easily search for some specific string using python, why need to use grep?

I massively simplified the test now. Should be easier to interpret.
I left them as grep commands because I want them to run within the eval env context (as if they were actually ran that way).

tlvu (approved):
Indeed, the new test rewrite with all variants folded together in one test is much easier to understand now, even with the `grep` mixed in, thanks !

PR #709, opened 2026-06-01
JSON endpoints were being set before template variable expansion

mishaschwartz:
## Overview

Because `service`, `components`, and `version` endpoints were being written before `.template` files were being processed, the `service` JSON endpoint could contain information that was out of date.

This fixes the issue by moving all of the logic that determines the content and creates the files for these endpoints to the `proxy/pre-docker-compose-up.include` file which will always run after the `.template files are processed.

## Changes

**Non-breaking changes**
- Bug fix

**Breaking changes**

## CI Operations

birdhouse_skip_ci: false


tlvu (approved):
LGTM.  Nice race condition catch !

CHANGES.md
```diff
@@ -15,7 +15,17 @@
...
+# Fixes
+
+- JSON endpoints were being set before template variable expansion
+
+  Because `service`, `components`, and `version` endpoints were being written before 
+  `.template` files were being processed, the `service` JSON endpoint could contain information that was out of date.
+
+  This fixes the issue by moving all of the logic that determines the content and 
+  creates the files for these endpoints to the `proxy/pre-docker-compose-up.include`
+  file which will always run after the `.template files are processed.
```
fmigneault: missing ` after ".template"

fmigneault (commented):
Thanks for the fix. Just minor typo in changelog.

fmigneault (approved):

PR #711, opened 2026-06-01
Add additional pre-commit hooks to enforce file format styles

mishaschwartz:
## Overview

  These hooks do the following:

  - fix trailing whitespace at the end of lines
  - format JSON files
  - detect whether a private key is being committed
  - ensures a single newline at the end of each file

  This is to enforce styles for this project and to also keep other PRs uncluttered
  since some developers use local tools that automatically enforce similar style
  rules on files and others don't.

## Changes

**Non-breaking changes**
- pre-commit hooks

**Breaking changes**
- None

## CI Operations

birdhouse_skip_ci: false


.pre-commit-config.yaml
```diff
@@ -11,3 +11,11 @@ repos:
       - id: ruff-format
         files: birdhouse/components/jupyterhub/jupyterhub_custom/jupyterhub_custom/
         args: [ --config=birdhouse/components/jupyterhub/jupyterhub_custom/ruff.toml ]
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v6.0.0  # Use the ref you want to point at
+    hooks:
+      - id: trailing-whitespace
+      - id: pretty-format-json
+        args: [ --autofix ]
```
Zeitsperre: This hook sees Jupyter Notebooks as JSON files:
```suggestion
      - id: pretty-format-json
        args: [ --autofix ]
        exclude: .ipynb
```

tlvu (approved):
LGTM.  I see 2 .svg changes but the image looks the same.  I guess it's safe.

This brings a question could there be "text" file that should not be auto-formatted?  Does the `trailing-whitespace` hook also suport the `excludes` option like the `pretty-format-json` one?

There could also have here-document syntax in regular file (.sh, .py, ...) that actually wants to preserve the trailing whitepace, how can we tell the tool to exclude that block?

Zeitsperre:
> This brings a question could there be "text" file that should not be auto-formatted? Does the `trailing-whitespace` hook also suport the `excludes` option like the `pretty-format-json` one?

The `excludes` option is used by `pre-commit`/`prek`, not unique to the hook, so yes. All hooks can be configured using it.

mishaschwartz:
> I see 2 .svg changes but the image looks the same. I guess it's safe.

svg files are technically still text files since they're XML so I think the tool just removed some whitespace

> This brings a question could there be "text" file that should not be auto-formatted?

I guess there could be... do we have a file like that in the repo right now or are you just thinking about the future?

> There could also have here-document syntax in regular file (.sh, .py, ...) that actually wants to preserve the trailing whitepace, how can we tell the tool to exclude that block?

No the tool only allows us to exclude entire files, not sections of files. What would the use-case be where we have a heredoc with trailing whitespace that we need to keep?

tlvu:
> > This brings a question could there be "text" file that should not be auto-formatted?
> 
> I guess there could be... do we have a file like that in the repo right now or are you just thinking about the future?

Future.


> > There could also have here-document syntax in regular file (.sh, .py, ...) that actually wants to preserve the trailing whitepace, how can we tell the tool to exclude that block?
> 
> No the tool only allows us to exclude entire files, not sections of files. What would the use-case be where we have a heredoc with trailing whitespace that we need to keep?

Also for the future, to avoid surprise, but no usecase right now.

I already approve by the way.

.pre-commit-config.yaml
```diff
@@ -11,3 +11,12 @@ repos:
...
```
fmigneault: Please add `--no-sort-keys` and revert the corresponding changes.
I would rather not change JSON formats structures.

For example, I follow the DGGS format specifically with the source repository so it makes diffs easier to investigate. The auto-sort causes more trouble in such case.

fmigneault (changes requested):

.pre-commit-config.yaml
Zeitsperre: The default args I use for this hook are:
```yaml
- id: pretty-format-json
  args: [ '--autofix', '--no-ensure-ascii', '--no-sort-keys' ]
```

fmigneault: Good point about the `--no-ensure-ascii` as well. 
It might make our `é`, `à`, etc. unicodes ugly in service metadata.

Zeitsperre (approved):
Mostly whitespace changes. Adding prek to conda config makes sense as well. Should we say somewhere that contributors should run `prek install` at some point to ge the hooks active?

Otherwise, LGTM!

fmigneault (approved):
Thanks! All good :shipit:

PR #712, opened 2026-06-01
Add auth cache in nginx for s3 endpoint

mishaschwartz:
## Overview

Adds a cache for the `s3` auth endpoint (that calls Twitcher's verify endpoint) so that repeated requests to the S3 store don't overwhelm twitcher and cause client connection issues if twitcher or nginx fails to properly authorize access to a resource.

This issue may arise if a client is processing a large S3 object in parallel by accessing different byte-ranges simultaneously. This results in many simultaneous requests to the same resource, each of which hits twitcher's verify endpoint. With this change, only the first request each minute will hit twitcher and the rest will only hit the cache.

Some implementation details to note:

  - responses are only cached for 1 minute to ensure that if a user's permissions are changed (on Magpie) their previous
    permissions expire quickly
  - the value of Magpie's cookie as well as the auth header is used as a cache key which allows us to cache cookie based
    and token-based authentication methods

## Changes

**Non-breaking changes**
- Adds cache

**Breaking changes**
- None

## Additional Information

Note that this could be implemented for other endpoints as well (i.e. thredds).

## CI Operations

birdhouse_skip_ci: false


birdhouse/components/s3/config/proxy/conf.extra-service.d/s3.conf.template
```diff
@@ -12,6 +12,13 @@
         internal;
         proxy_pass ${BIRDHOUSE_PROXY_SCHEME}://${BIRDHOUSE_FQDN_PUBLIC}${TWITCHER_VERIFY_PATH}$request_uri;
         proxy_pass_request_body off;
+
+        # cache auth requests so we don't overwhelm twitcher when a client is accessing multiple chunks
+        # of the same file in parallel (for example).
+        proxy_cache "s3_auth_cache";
+        proxy_cache_key "$request_uri:uri:$http_authorization:cache:$cookie_${MAGPIE_COOKIE_NAME}";
+        proxy_cache_valid 200 401 403 1m;
```
tlvu: Any reasons someone would want a longer or shorter cache timeout than 1m?  Worth to make it configurable?

tlvu (approved):
LGTM

birdhouse/components/magpie/default.env
```diff
@@ -61,6 +61,9 @@ export MAGPIE_NETWORK_CREATE_MISSING_PEM_FILE=true
 # translate MAGPIE_NETWORK_PEM_FILES to the location of the files on the magpie container
 export MAGPIE_NETWORK_PEM_FILES_ON_CONTAINER='$(echo "/magpie-pem/${MAGPIE_NETWORK_PEM_FILES#:}" | sed "s|:|:/magpie-pem/|g" )'
 
+# explicitly declare the magpie cookie name so that it can be referred to by other components
+export MAGPIE_COOKIE_NAME=auth_tkt
```
fmigneault, line 65: Defined here but not injected in the container? If modified, the Cookie and API won't work.

birdhouse/components/s3/config/proxy/conf.extra-service.d/s3.conf.template
```diff
@@ -12,6 +12,13 @@
...
+        proxy_ignore_headers Cache-Control Expires Set-Cookie;
```
fmigneault: If the cookie needs to be set of the first API call (when cache is not yet set), won't `Set-Cookie` be missing?

birdhouse/components/s3/config/proxy/conf.extra-service.d/s3.conf.template
```diff
@@ -12,6 +12,13 @@
         internal;
         proxy_pass ${BIRDHOUSE_PROXY_SCHEME}://${BIRDHOUSE_FQDN_PUBLIC}${TWITCHER_VERIFY_PATH}$request_uri;
         proxy_pass_request_body off;
+
+        # cache auth requests so we don't overwhelm twitcher when a client is accessing multiple chunks
+        # of the same file in parallel (for example).
+        proxy_cache "s3_auth_cache";
+        proxy_cache_key "$request_uri:uri:$http_authorization:cache:$cookie_${MAGPIE_COOKIE_NAME}";
```
fmigneault, line 19: Is the specific token name actually necessary?

If it was `...:$cookie_magpie_twitcher_s3`, wouldn't it already be distinct enough for users and service combinations?

birdhouse/components/s3/config/proxy/conf.extra-service.d/s3.conf.template
mishaschwartz: I think that 1m is a reasonable default but it's always worth making things configurable

birdhouse/components/magpie/default.env
mishaschwartz: good catch. thanks

birdhouse/components/s3/config/proxy/conf.extra-service.d/s3.conf.template
mishaschwartz: `proxy_ignore_headers` tells nginx to ignore response headers from the upstream response. Nginx won't cache a response if it contains a cookie so this must be disabled in order for caching to work at all (twitcher's verify endpoint sets a cookie).

However, the cookie is still visible to nginx as a variable and can be used as part of the cache key. Think about it this way:

1. nginx receives first request (contains cookie)
2. nginx gets a cache miss and sends it to the twitcher/ows/verify
3. twitcher/ows/verify returns a response that contains a cookie
4. nginx ignores the cookie and caches the response (based on the cookie from the request)
5. nginx receives a second request (contains cookie)
6. nginx gets a cache hit and returns the cached response

If nginx didn't ignore the cookie in step 4 then it would never cache the response. The cache is based on the cookie sent in the request, not the one sent in the response.

If the magpie cookie expires then the cache will keep the resource available to the user for one additional minute (by default) but that's the tradeoff we have to deal with when we're implementing caches.

birdhouse/components/s3/config/proxy/conf.extra-service.d/s3.conf.template
mishaschwartz: `$cookie_magpie_twitcher_s3` would read a cookie from the request named "magpie_twitcher_s3" which is not the name of the cookie that is set by magpie/twitcher.

fmigneault: OK. I misunderstood the way the command worked.

birdhouse/components/s3/config/proxy/conf.extra-service.d/s3.conf.template
fmigneault: OK thanks for the explanation.

fmigneault (approved):
All good :shipit:

PR #714, opened 2026-06-05
fix invalid `JUPYTER_VERSION` in CanarieAPI config 

### LINKED ISSUE #713: :bug: [BUG]: Use JUPYTERHUB_VERSION in Jupyterhub canarie-api template
## Summary

The JupyterHub canarie API template uses an undefined `${JUPYTER_VERSION}` variable, causing the service info endpoint to return the literal string `${JUPYTER_VERSION}`.

## Details
The template currently has:
https://github.com/bird-house/birdhouse-deploy/blob/22dae698d197cf59ea81dfc188f6c70ddd6e70ff/birdhouse/components/jupyterhub/config/canarie-api/canarie_api_monitoring.py.template#L5-L6

Since releaseTime uses ${JUPYTERHUB_VERSION}, I think version should use the same variable.

### To Reproduce

1.Run the Jupyterhub component with canarie.
2. Visit [canarie/Jupyter/service/info](https://garryoak.pacificclimate.org/canarie/Jupyter/service/info)
3. The version field is ${JUPYTER_VERSION} instead of an actual version.

## Environment

| Server/Platform URL   | https://garryoak.pacificclimate.org
| Version Tag/Commit    | 2.28.1
| Related issues/PR     | NA
| Related components    | Jupyterhub, canarie-api
| Custom configuration  | NA

## Concerned Organizations

All organizations.

fmigneault:
## Overview
`JUPYTER_VERSION` was used instead of the available `JUPYTERHUB_VERSION` variable.

>[!NOTE]
> There are actually quite a few other misnamed `JUPYTER_` variables (not matching the `jupyterhub` naming convention. However, I left them as it to avoid introducing breaking changes / need to apply backward compat mappings since they are also referenced  by `jupyter_...` names in Python scripts.
>
> If this is desired (probably should for consistency/avoid confusion, but not breaking), it can be done in a more substential follow-up PR. 

## Changes

**Non-breaking changes**
- JupyterHub: Fix invalid `JUPYTER_VERSION` instead of `JUPYTERHUB_VERSION` in CanarieAPI configuration.

  

**Breaking changes**
- n/a

## Related Issue / Discussion

- Fixes https://github.com/bird-house/birdhouse-deploy/issues/713

## CI Operations

birdhouse_skip_ci: false


tlvu (approved):
LGTM

mishaschwartz (approved):

PR #717, opened 2026-06-11
restart `canarie-api` on `proxy` restart

fmigneault:
## Overview

Create a `canarie-api -> proxy` dependency to ensure config reload on service restart.

## Changes

**Non-breaking changes**
- CanarieAPI: Fix reflected metadata from server udpates.

  Add a dependency on `canarie-api -> proxy` such that `canarie-api` is always restarted (since `proxy` always restarts)
  to make sure that server metadata updates in `docker_configuration.py` are considered by its internal API process to
  reload them. Without this, updating the server (e.g.: pulling a new version and doing `bin/birdhouse compose up -d`)
  would not reflect the updated server/node version changes reported by CanarieAPI endpoints, since the actual service
  definition of `canarie-api` would not have changed (volume-mount/service config unchanged, only file contents change).

**Breaking changes**
- n/a

## Additional Information

I used the `restart` dependency because it is the less intrusive and easiest to maintain IMO. The drawback is that it restarts each time, even when the config is unchanged, but I assume this would occur in most cases when "important enough" changes happen that would need a restart anyway.

### Alternatives considered
1. use `develop.watch`: 
   - that would work by detecting the contents changed, but requires doing `docker compose watch` or `docker compose up --watch` to ensure it is actually running, since not enabled by default
   - might be unreliable if the watch stops, since it is a separate process, not within docker/config checks itself, even if the fields are present within the compose config
   - technically intended for local development, therefore no guarantee "prod" side
 
2. manual check of some sort + `docker compose restart canarie-api` call
   - custom logic to manage in the scripts, therefore more stuff to maintain
   - probably some edge cases we wouldn't have tought of 

3. use some hash of the contents in the file name
   - auto-reload/detect since the volume mount would change
   - seems way more hacky and needs cleanup since files will accumulate with different hashes
   - custom script logic again

## CI Operations

birdhouse_skip_ci: false


tlvu (approved):
LGTM

mishaschwartz (approved):

PR #724, opened 2026-06-16
Adjust triggers and labels for dependabot auto-approve

### LINKED ISSUE #723: :question: [Question]: PR dependabot auto-approved not processed
## Questions

In cases where the "auto-merge" feature was *NOT* enabled before an approval is submitted, it looks like dependabot sometimes hangs to process it. At the very least, it seems to wait until another PR is merged, which triggers a `synchronize` (from dependabot rebasing its own PR), which then matches the below workflow conditions.

I am wondering if the following event triggers are probably missing something like https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_review to flag the "on approved" case?

https://github.com/bird-house/birdhouse-deploy/blob/a80c746bf81ba5f1aa3cc53e2556dab2514ae390/.github/workflows/auto-accept-ci-changes.yml#L3-L11

Examples pending (at least until it syncs at some point):
- https://github.com/bird-house/birdhouse-deploy/pull/721
- https://github.com/bird-house/birdhouse-deploy/pull/722

@Zeitsperre Thoughts?

## References

- #700 
- #694

## Environment

CI/CD

Zeitsperre:
## Overview

Adjusts the label and triggering mechanisms to ensure that both GitHub Actions and Python CI/Test dependencies updates are automatically approved.

## Changes

**Non-breaking changes**
- Added two new labels (`ci/actions` and `ci/python`)
- Added a trigger for dependabot auto-approve on PR review

## Related Issue / Discussion

- Resolves #723 

## CI Operations

birdhouse_skip_ci: false


fmigneault (approved):

.github/workflows/auto-accept-ci-changes.yml
mishaschwartz, line 8: Add `tests/requirements.*` to this as well or else this won't run for test dependencies

.github/workflows/auto-accept-ci-changes.yml
```diff
@@ -9,6 +9,9 @@ on:
     types:
       - opened
       - synchronize
+  pull_request_review:
```
mishaschwartz: I don't think I understand the use case for this... 

We want this to run when someone submits a review and it's approved? But if it gets auto-merged when the PR is opened then there will never be a chance to approve it because it will be merged right away.

I thought the point of this PR was to avoid the approval process. If an admin approves the PR manually then they can also just merge it manually too.

.github/dependabot.yml
mishaschwartz, line 12: should we be adding this prefix to the other one too?

.github/workflows/auto-accept-ci-changes.yml
```diff
@@ -20,7 +23,9 @@ jobs:
     environment: automation
     if: >
       github.event.pull_request.user.login == 'dependabot[bot]' &&
-      contains(github.event.pull_request.labels.*.name, 'github_actions')
+      ((contains(github.event.pull_request.labels.*.name, 'ci/actions') ||
```
mishaschwartz: So just so I understand...

- the PR has to be submitted by dependabot
- the PR can have either label OR it can be an approved review

When would it be an approved review and not have one of those labels?
When would it be submitted by dependabot an not have one of those labels?

Can we just simplify this to:

```
github.event.pull_request.user.login == 'dependabot[bot]'
```

Since currently all dependabot PRs should trigger this action

mishaschwartz (changes requested):
Unless I'm mistaken I don't think that there is anything preventing this from merging a PR even if the tests fail. Previously that didn't matter because we were only auto-merging changes for the CI dependencies, not the test dependencies but it matters now.

We either need to make this dependent on the tests workflow or we need to make the test workflow required by adding a branch protection rule for it (I would recommend this one).

.github/workflows/auto-accept-ci-changes.yml
mishaschwartz: We should change this name as well since it is no longer just for "GitHub Actions"

mishaschwartz (changes requested):

.github/workflows/auto-accept-ci-changes.yml
mishaschwartz: Also note that this means that every time anyone submits a review to *any* PR this workflow will run (even if the action is immediately skipped).

tlvu (approved):
Trevor is handling all github workflows for Ouranos so LGTM, pending Misha comments.

Also I just want to be sure, are we auto-merging patch and minor version updates only or we are also auto-merge major version changes?

.github/workflows/auto-accept-ci-changes.yml
Zeitsperre: > Also note that this means that every time anyone submits a review to _any_ PR this workflow will run (even if the action is immediately skipped).

This is not the case. I believe it counts only for the PR when the review is done within the scope of its context. We use this trigger all over the place and that's how it works.

I'll admit that the docs do a bad job of explaining this: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_review

Zeitsperre: I think the challenge here is that my original intention was to automatically accept certain kinds of PRs from Dependabot, but the existing rules specify only the CODEOWNERS can give ultimate approval before something is merged.

The workflow here can accept and set auto-merge, but it will always need CODEOWNER approval (unless I create a rule exception).

.github/dependabot.yml
Zeitsperre: Don't we want to run changes to Python deps on Jenkins that could affect the tests?

.github/workflows/auto-accept-ci-changes.yml
Zeitsperre: Good point. The conditionals were getting a bit unwileldly as well.

.github/dependabot.yml
Zeitsperre: I'll set it for both. GitHub tests are enough AFAIK

.github/workflows/auto-accept-ci-changes.yml
mishaschwartz: > This is not the case.

If you list multiple conditions in the "on:" block then _any_ of those conditions will trigger a workflow to run. This means that we're saying "trigger this when there is a PR to master targeting these files OR if there is a pull request review submitted". 

The "when there is a PR to master targeting these files" part only limits the `pull_request`, not the `pull_request_review`.

Look... I made an example repo here (https://github.com/mishaschwartz/workflow-example) that has a simple workflow with the same "on:" condition that we have here. I created 3 PRs to demonstrate:

- https://github.com/mishaschwartz/workflow-example/pull/4 does not trigger the workflow because I updated a file not in the "paths:" list
- https://github.com/mishaschwartz/workflow-example/pull/6 does trigger the workflow because I did update a file in the "paths:" list
- https://github.com/mishaschwartz/workflow-example/pull/7 did not initially trigger a workflow because it makes the same change as https://github.com/mishaschwartz/workflow-example/pull/4 but then I submitted a review and a workflow was triggered

mishaschwartz: > unless I create a rule exception

Yes we definitely need a rule exception otherwise I don't think this workflow makes sense to have. I think you can do this by creating a bypass list for the codeowner branch rule and add the dependabot user to the bypass list

Zeitsperre: Sorry to get back to this so late. Thanks for all the testing you did here @mishaschwartz. Will remove this trigger and adjust the rules a bit.

Zeitsperre:
> Are we auto-merging patch and minor version updates only or we are also auto-merge major version changes?

We are only auto-merging patch and minor updates. Major updates will need to pass by the CODEOWNERS for approval to be pushed to `main`.

Zeitsperre:
@mishaschwartz The rules for PRs are set to allow Dependabot to bypass them. There should be no more issue, hopefully.

Zeitsperre:
I've removed the `environment` from the auto-approve workflow. This was essentially removing all other "secrets" that the GitHub Action for signing commits required. I can see it being very useful in cases where we don't want any risk of tokens being leaked (e.g. Trusted Publishing), but it's overkill for Dependabot.

This will fix the failures in #733 and #729.

mishaschwartz (approved):
@Zeitsperre Thanks for your work on this!

PR #730, opened 2026-07-06
update default crim-ca/stac-app:2.4.0

fmigneault:
## Overview

STAC API dependency updates.

## Changes

- STAC: Update default API version to `crim-ca/stac-app:2.4.0` for dependency updates and security fixes.

  - Update Docker with Python 3.13.14 for security fixes.
  - Update `stac-fastapi.api==6.3.2` and `stac-fastapi.pgstac==6.3.1` with extension breaking change from `core` module.

**Breaking changes**
- none if using the component as-is, but the component itself has a breaking change if using custom overrides

## CI Operations

birdhouse_skip_ci: false


tlvu (approved):
LGTM as we do not use any override.  But just for the sake of completeness, can you update the migration doc for those that actually use the override and will encounter incompatible problems?

tlvu:
> update the migration doc

https://github.com/bird-house/birdhouse-deploy/blob/master/docs/source/migration_guide.rst and reference that migration guide in the Changelog.

fmigneault:
@tlvu 
I'm not sure anyone does, or if they do, how even it is done.
I added this warning as a indicator that "*if you do something custom*" like manual overrides of source code to add extensions, it could break, but there is nothing currently specifically in the repo to perform such overrides.

Maybe @mishaschwartz with some scripts to update the collection schema/queryables? But I don't think there is an impact regarding loaded extenions.

tlvu:
> manual overrides of source code to add extensions

Oh that level of override.  I though only overriding the config files volume-mounted by the stack.

I am fine then as long as you think this is unlikely for anyone to have upgrade failure.

mishaschwartz (approved):
Yeah that's fine. I don't do anything custom on my end either. 

In general, I think that we don't need to warn people "this may break things if you're overwriting the source code". If they're doing that then they already know they're taking that risk.

PR #731, opened 2026-07-10
Jupyterhub: bump version to 5.5.0

mishaschwartz:
## Overview

This version includes some bug fixes and changes to an authenticator not used by birdhouse.

See [the changelog](https://jupyterhub.readthedocs.io/en/stable/reference/changelog.html#id3) for specifics.

## Changes

**Non-breaking changes**
- jupyterhub version: 5.5.0-20260710

**Breaking changes**
- None

## CI Operations

birdhouse_daccs_configs_branch: master
birdhouse_skip_ci: false

PR #740, opened 2026-07-29
Thredds: authenticate with twitcher verify instead of going through twitcher's proxy

mishaschwartz:
## Overview

The Thredds endpoints go through Twitcher's proxy (`twitcher/ows/proxy` by default) which adds some overhead to the requests. In order to avoid this and increase efficiency, Twitcher endpoints now authenticate with the Twitcher verify endpoint (`twitcher/ows/verify` by default) and then access the thredds service directly.

**Deprecations**

Note that some older scripts and notebooks were prone to accessing Thredds directly through Twitcher itself (e.g. `twitcher/ows/proxy/thredds/...`). In order to continue supporting these requests this change adds an Nginx rewrite rule to permanently redirect to the equivalent `thredds/...` endpoint. However, accessing Thredds through the Twitcher proxy directly is now considered deprecated and support for this will be dropped in a later version.

## Changes

**Non-breaking changes**
- change internal authentication path for thredds

**Breaking changes**
- None

## Related Issue / Discussion

- Closes #341

## Additional Information

This is a redo of #341 that is out of date with a bunch of the code base.

## CI Operations

birdhouse_skip_ci: false


CHANGES.md
```diff
@@ -15,7 +15,21 @@
...
+  The Thredds endpoints go through Twitcher's proxy (`twitcher/ows/proxy` by default) which adds some overhead
+  to the requests. In order to avoid this and increase efficiency, Twitcher endpoints now authenticate with the
+  Twitcher verify endpoint (`twitcher/ows/verify` by default) and then access the thredds service directly.
+
+  **Deprecations**
+
+  Note that some older scripts and notebooks were prone to accessing Thredds directly through Twitcher itself
+  (e.g. `twitcher/ows/proxy/thredds/...`). In order to continue supporting these requests this change adds an
+  Nginx rewrite rule to permanently redirect to the equivalent `thredds/...` endpoint. However, accessing Thredds
+  through the Twitcher proxy directly is now considered deprecated and support for this will be dropped in a later
```
tlvu:
```suggestion
  through the Twitcher proxy directly is allowed to avoid breaking older projects using the older link (ensure reproducible science)
```

Can not be dropped.  We have a tons of notebooks hardcoding the old path.

Also these notebooks could have been copied by our numerous users as a starting point for their older projects so we absolutely do not want to break their older finished project to not go against "reproducible science".

mishaschwartz:
@tlvu 

> Can you remind me why verify link is faster than the proxy link? 

It's faster for services that send a lot of data because it doesn't have to pass through twitcher which does it's own buffering.

> Can verify link do everything the proxy link can? 

No, it doesn't support request and response hooks

> TestThredds is a mirror for Thredds, can you check if it is also behind the proxy link and make the similar adjustment? 

It does not use twitcher

> As for the older twitcher/ows/proxy/thredds/ it has to be preserved forever. Please remove the mention about deprecation of that link.

:cry:

> There should not have any performance penalty if user go through the old link right? It should perform a permanent redirect once and then subsequent requests will hit the new link.

There should be no penalty

> Oh wait, did you forget the permanent redirect from the old link to the new link?

No

CHANGES.md
```diff
@@ -23,13 +23,10 @@
   to the requests. In order to avoid this and increase efficiency, Twitcher endpoints now authenticate with the
   Twitcher verify endpoint (`twitcher/ows/verify` by default) and then access the thredds service directly.
 
-  **Deprecations**
-
   Note that some older scripts and notebooks were prone to accessing Thredds directly through Twitcher itself
   (e.g. `twitcher/ows/proxy/thredds/...`). In order to continue supporting these requests this change adds an
   Nginx rewrite rule to permanently redirect to the equivalent `thredds/...` endpoint. However, accessing Thredds
-  through the Twitcher proxy directly is allowed to avoid breaking older projects using the older link (ensure reproducible science)
```
tlvu, line 31: I'd like to keep the mention of "reproducible science"?  That's the real reason we want to keep the redirect.  Why remove this part?  If you don't want this here, leave that justification with the proxy conf?

birdhouse/components/thredds/config/proxy/conf.extra-service.d/thredds.conf.template
```diff
@@ -1,10 +1,26 @@
...
+        proxy_set_header Host $host;
+        proxy_set_header Content-Length "";
+        proxy_set_header X-Original-URI $request_uri;
+        proxy_set_header X-Forwarded-Proto $real_scheme;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Host $host:$server_port;
+    }
+
+    # Some old scripts and notebooks accessed threads via the twitcher proxy directly.
+    # In order to support this older syntax this redirects to the correct location.
```
tlvu, line 24:
```suggestion
    # In order to support this older syntax this redirects to the correct location.
    # Older finished and submitted projects are using the old link to access Thredds data
    # and we want to avoid breaking them to ensure reproducible science.
```

mishaschwartz:
Honestly, it's because I disagree with the policy that we cannot even contemplate eventually deprecating something in this repo because of an old design decision.

I am perfectly happy to take a long time to deprecate and eventually remove part of the API but keeping it forever is a bad choice. 

In this case I would like to do the following:

- permanent redirect (301) to the right endpoint and add a comment that this is deprecated and **may** eventually be removed
- after a long time change it to Gone (410) with an explanation of where to find the correct URL in the response message.
- after an even longer time remove the route entirely

All of the above will be documented in the Changelog and the migration guide. Any script that is being used even somewhat regularly will encounter the 301 or the 410 and be able to be updated. This should be our policy for deprecating and removing any endpoints.

If PAVICS specifically wants to keep the 301 forever we can create a separate optional component when the route is eventually removed and you can enable that component for PAVICS.

tlvu:
Good idea for the optional-component so PAVICS can keep it forever but we won't force that onto other org.

Agreed with the "301 -> 401 -> remove for real" to gradually phase out old endpoints with optional-components for orgs stuck with backward compatibility support.

Can you create the optional components that does the 301 now and make thredds component depend on it for now.  

At the time of 401, we'll just remove that dependency.  Anyone that wants to preserve the 301 can simply enable the optional component before hand and the 401 transition and future removal is "transparent".  If both 301 and 401 exists at the same time, how to make 301 take precedence?

I'll approve this PR for now because I am going on vacation so I won't block your merge but please recreate the current state that 301 is the default behavior by creating the optional-component and make thredds component depends on it.

tlvu (approved):
LGTM, excellent handling of both 301 and 410.

mishaschwartz:
appologies for the force push... the unittest github action simply won't run and I was trying to force it to re-run. It is obviously triggered since it shows up in the pending checks list but it doesn't actually run and the actions tab does not show anything for the unittest job.

Could someone with more permissions on this repo check if we've hit some resource limit or something?

@fmigneault @Zeitsperre 

I cannot merge this until that action finishes since it's a required check

Zeitsperre:
@mishaschwartz The commit message includes `[skip ci]` which prevents the CI from running on GitHub. If we wanted to skip only for jenkins, it would be `[skip jenkins]`.

mishaschwartz:
I see.... 

That's a problem then because the last commit on any PR that's going to bump the version is going to include [skip ci] since that is automatically added to the bumpversion commit.

I see that I actually ran into this problem the last time I made a release by making an empty commit at the end (see: https://github.com/bird-house/birdhouse-deploy/pull/731). 

I'm going to do the same thing again but we should probably remove the [skip ci] from the bumpversion commit message or else we're going to run into this issue every time.

